PR #38, opened 2024-11-15
Updating Mistral README w/ Warning 

mvanniasingheTT:


tstescoTT (approved):
Verified this warning message is the same as in tt-metal for Mistral 7B at https://github.com/tenstorrent/tt-metal/blob/main/models/demos/wormhole/mistral7b/README.md

We should keep this warning here until the issue is resolved in tt-metal.

PR #39, opened 2024-11-18
Add Logging Utils

mvanniasingheTT:
# Adding Logging 

This PR adds logging functionality to the online and offline mock models as well as to the vllm api server. Note there was a previous PR with the same changes but this PR is rebased off main and has a cleaner commit history.

### Note:
There is some repeated code from `logging_utils.py` in `mock_vllm_model.py` since usage of these functions and classes in the online and offline mock model inference scripts and requires refactoring of the tt-inference-server directory naming. Repeated code can be removed following the refactoring.


vllm-tt-metal-llama3-70b/src/run_vllm_api_server.py
```diff
@@ -25,6 +30,7 @@ def get_encoded_api_key(jwt_secret):
     return encoded_jwt
 
 
+@patch.object(LLMEngine, "__init__", new=init_wrapper)
```
tstescoTT, line 33: the init_wrapper should be named to indicate it is for logging only, e.g. logging_init_wrapper.

tests/mock_vllm_model.py
```diff
@@ -2,11 +2,12 @@
...
+from datetime import datetime
+import json
 from dataclasses import dataclass
 from typing import List
 
 import torch
-from loguru import logger
 
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
 from tt_metal.models.demos.t3000.llama2_70b.tt.llama_common import (
```
tstescoTT, line 13: should the import path be `models.demos.t3000.llama2_70b.tt.llama_common` for these tt-metal modules?

tests/mock_vllm_model.py
```diff
@@ -318,3 +335,105 @@ def forward(
                 kv_cache=kv_cache,
                 prompt_lens=prompt_lens,
             )
+
+
+class RawStatLogger(StatLoggerBase):
```
milank94, line 340: Is this Class different from the one in `vllm-tt-metal-llama3-70b/src/logging_utils.py`? Can you explain the difference?

tests/mock_vllm_api_server.py
```diff
@@ -13,8 +13,14 @@
...
 from vllm.worker.tt_worker import TTWorker, TTCacheEngine
-from mock_vllm_model import new_init_cache_enginer, new_allocate_kv_cache, MockModel
+from mock_vllm_model import (
+    new_init_cache_enginer,
+    new_allocate_kv_cache,
+    MockModel,
+    init_wrapper,
+)
 from vllm.engine.multiprocessing.engine import run_mp_engine
+from vllm.engine.llm_engine import LLMEngine
```
milank94, line 23: Any reason why it's `LLMEngine` and not `AsyncLLMEngine` for the online server?

mvanniasingheTT: I believe `AsyncLLMEngine` is a wrapper for `LLMEngine`. For the online server, the logging functionality still gets implemented in an `LLMEngine` . But for this case, [line 127 always returns false](https://github.com/tenstorrent/vllm/blob/dev/vllm/entrypoints/openai/api_server.py#L127C27-L127C48)  (since `pipeline_parellel_size = 1` so  [`is_unsupported_config()`] (https://github.com/tenstorrent/vllm/blob/main/vllm/engine/multiprocessing/client.py#L120) returns false. This means `MQLLMEngine` is created. The `LLMEngine` object is created in [`MQLLMEngine` ](https://github.com/tenstorrent/vllm/blob/dev/vllm/engine/multiprocessing/engine.py#L78).  Since the relevant attributes are created in `LLMEngine` that is why it is used as opposed to `MQLLMEngine`

tests/mock_vllm_model.py
mvanniasingheTT: It is the same. It is duplicated since due to the current directory naming structure with the hyphens it is not possible to import `logging_utils`. Is there a better way to go about this?

vllm-tt-metal-llama3-70b/src/logging_utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class RawStatLogger(StatLoggerBase):
+    def __init__(self, num_scheduler_steps, batch_size) -> None:
+        self.time_to_first_token = []
+        self.time_per_output_token = []
+        self.num_scheduler_steps = num_scheduler_steps
+        self.batch_size = batch_size
+        timestamp = datetime.now().strftime("%Y-%m-%d_%H%M%S")
+        self.filepath = f"/home/user/tests/statistics_{timestamp}.jsonl"
```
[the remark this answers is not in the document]
mvanniasingheTT: The `/home/user` is the path in the docker container. If a user is following the instructions for running the mock model, an instruction in the README is given to mount the test directory to this location. Should I do this differently?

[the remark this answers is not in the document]
milank94: No that's perfect -- thanks for the explanation.

tests/mock_vllm_model.py
milank94: Ok makes sense. Can you add a header to the file then that you think should be removed after the refactor?

Maybe something with a `# TODO:` identifier so that we can keep track of it?

tests/mock_vllm_api_server.py
milank94: Got it -- thank you for the explanation.

vllm-tt-metal-llama3-70b/src/run_vllm_api_server.py
mvanniasingheTT: got it, it is changed now

tests/mock_vllm_model.py
mvanniasingheTT: yes, you are right. We no longer use the symlink so you are correct.  It is updated now.

vllm-tt-metal-llama3-70b/src/logging_utils.py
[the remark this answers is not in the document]
mvanniasingheTT: Actually, I made a change to store it in `cache_root` since this path is the persistent storage rather than have it go under the `tests` directory. I think this helps it be more clean and is more general so if someone runs the actual model, the logging is not saved to the `test` directory.

tstescoTT (approved):

PR #41, opened 2024-11-27
Prompt generation CLI and utils

tstescoTT:
# change log
- adding prompt_client_cli.py for testing vllm open ai inference server with prompts sent over requests with prompts sent over requests
- adding example using llama 3.1 instruct jinja template
- add utils/README.md documentation for prompt_client_cli.py usage

### DIFF
--- a/utils/prompt_generation.py
+++ b/utils/prompt_generation.py
@@ -94,7 +94,7 @@ def generate_random_prompts(
         )
     else:
         raise ValueError(
-            f"Invalid distribution method: '{distribution}'. Use 'uniform' or 'max_length'."
+            f"Invalid distribution method: '{distribution}'. Must be 'fixed', 'uniform', or 'normal'."
         )
 
     # Generate random tokens for all prompts


utils/README.md
```diff
@@ -0,0 +1,150 @@
...
+- `--batch_size`: Number of concurrent requests
+- `--max_prompt_length`: Maximum length for generated prompts
+- `--output_seq_len`: Maximum length for completions
+- `--num_full_iterations`: Number of times to repeat the full prompt set
+- `--vary-batch-size`: Randomize batch sizes using normal distribution
+- `--input_seq_len`: Fixed length for input sequences (-1 for variable)
+- `--inter_batch_delay`: Delay between batches in seconds
+- `--no-stream`: Disable streaming responses
+- `--dataset`: Source dataset (random, alpaca_eval)
+- `--distribution`: Prompt length distribution (fixed, uniform, normal)
```
milank94, line 38: I'm not seeing `--distribution` under `prompt_client_cli.py`. 

Does `--distribution` need to be used with `--input_seq_len`? If `--input_seq_len` is set to a length such as 128, does that mean that `--distribution` must be set to `fixed`?

utils/prompt_generation.py
```diff
@@ -0,0 +1,424 @@
...
+    parser.add_argument(
+        "--distribution",
+        type=str,
+        default="fixed",
+        choices=[
+            "fixed",
+            "uniform",
+            "normal",
+        ],
+        help="Distribution method for selecting random prompt lengths ('uniform' or 'max_length').",
```
milank94, line 400:
```suggestion
        help="Distribution method for selecting random prompt lengths ('fixed', 'uniform', or 'max_length').",
```

utils/prompt_generation.py
```diff
@@ -0,0 +1,424 @@
...
+            args.tokenizer_model,
+        )
+    elif args.dataset is not None:
+        assert (
+            args.max_prompt_length > -1
+        ), "max_length must be set for datasets prompts."
+        logger.info(f"Generating prompts from the '{args.dataset}' dataset...")
+        if args.dataset == "alpaca_eval":
+            prompts = load_alpaca_eval_dataset_samples(args.num_prompts)
+        elif args.task is not None:
```
milank94, line 331: What should the user set `--dataset` to if they want to generate from a `--task`?

If they set `--dataset random` they will generate the random set and if they set `--dataset alpaca_eval` they will get the AlpacaEval dataset irrespective of what `--task` is set to. But, if they want a `--tast` set, what should `--dataset` be?

utils/README.md
milank94: Looks from here: https://github.com/tenstorrent/tt-inference-server/pull/41/files#diff-b27d6427aa7798c85e8aa50b17a6fd49d0a133be880921927eeb11a50232684fR306 that if a user sets `--input_seq_len -1` then it will just automatically get set to `--max_prompt_length`

utils/prompt_generation.py
```diff
@@ -0,0 +1,424 @@
...
+# Define a function to generate random prompts using a model's vocabulary
+def generate_random_prompts(
+    num_prompts: int,
+    max_length: int,
+    input_seq_len: int,
+    distribution: str,
+    tokenizer_model=None,
+):
+    """
+    Generate random prompts using the model's vocabulary.
```
milank94, line 64: Love docstrings. We should align on a standard for consistency. Should they be used for all Classes, Methods, and Functions.

utils/prompt_generation.py
```diff
@@ -0,0 +1,424 @@
...
+    if distribution == "fixed":
+        prompt_lengths = [input_seq_len] * num_prompts
+    elif distribution == "uniform":
+        prompt_lengths = torch.randint(1, input_seq_len, (num_prompts,)).tolist()
+    elif distribution == "normal":
+        prompt_lengths = (
+            torch.normal(mean=input_seq_len, std=input_seq_len / 4, size=(num_prompts,))
+            .clamp(1, max_length)
+            .round()
+            .to(torch.int32)
```
milank94, line 92: Any reason why casting "normal" to `torch.int32` but not "uniform"?

utils/prompt_generation.py
```diff
@@ -0,0 +1,424 @@
...
+            .tolist()
+        )
+    else:
+        raise ValueError(
+            f"Invalid distribution method: '{distribution}'. Use 'uniform' or 'max_length'."
```
milank94, line 97:
```suggestion
            f"Invalid distribution method: '{distribution}'. Use 'fixed', 'uniform', or 'max_length'."
```

utils/prompt_generation.py
tstescoTT: yes, uniform uses `torch.randint` so its already an int. Both need to be ints because these are the prompt context lengths.

utils/prompt_generation.py
tstescoTT: Agreed, we can use a prompt template for generation draft doc strings that meet that standard too!

utils/prompt_generation.py
tstescoTT: I added a commit to remove the --task support, I think it's overly complicated to have LM eval prompt templating in this CLI. I didnt find it very useful, can add it back if there's a real use case for testing.

utils/README.md
tstescoTT: `--distribution` is only used in generate_random_prompts() which is called if `--dataset` is set to `"random"`. `--input_seq_len` in that case is the parameter for the distribution, i.e.`uniform: max`, `normal: mean, std`, `fixed -> all same length`.

tstescoTT: `max_prompt_length` is more important for when `--dataset = alpaca_eval`.

utils/prompt_client_cli.py
```diff
@@ -0,0 +1,430 @@
...
+
+def call_inference_api(
+    prompt,
+    response_idx,
+    prompt_len,
+    stream,
+    headers,
+    api_url,
+    max_tokens,
+    vll_model,
```
mvanniasingheTT, line 72: should this be vllm_model ?

tstescoTT: yes, that's the model lookup vLLM is hosting it on.

utils/README.md
milank94: Should we have checks for that? I don't want to complicate the code base, but it sounds like some combinations of configs wont be compatible together.

tstescoTT: They are compatible or disabled for the different usecases, e.g. for `--distribution` if the variable isnt needed it wont be used so the default is fine. The example usage shows the commands for expected usecases.

milank94 (approved):

mvanniasingheTT (approved):

PR #42, opened 2024-11-27
Logging utils and config

tstescoTT:
# change log
- move logging_utils.py to utils/logging_utils.py
- add set_vllm_logging_config() to add logging to timestamped logfile as well as to stdout using vLLM logging config

### DIFF
--- a/vllm-tt-metal-llama3-70b/src/run_vllm_api_server.py
+++ b/vllm-tt-metal-llama3-70b/src/run_vllm_api_server.py
@@ -10,10 +10,7 @@ import json
 import jwt
 from vllm import ModelRegistry
 
-# importing logging utils
-from logging_utils import logging_init_wrapper
-from vllm.engine.llm_engine import LLMEngine
-from unittest.mock import patch
+from utils.logging_utils import set_vllm_logging_config
 
 # importing from tt-metal install path
 from models.demos.t3000.llama2_70b.tt.llama_generation import TtLlamaModelForGeneration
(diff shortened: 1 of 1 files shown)


vllm-tt-metal-llama3-70b/src/run_vllm_api_server.py
```diff
@@ -30,11 +27,22 @@ def get_encoded_api_key(jwt_secret):
...
+    print(f"setting vllm logging config at: {config_path}")
+    print(f"setting vllm logging file at: {log_path}")
+    # note: the vLLM logging environment variables do not cause the configuration
+    # to be loaded in all cases, so it is loaded manually in set_vllm_logging_config
+    os.environ["VLLM_CONFIGURE_LOGGING"] = "1"
+    os.environ["VLLM_LOGGING_CONFIG"] = str(config_path)
+    # stop timeout during long sequential prefill batches
+    # e.g. 32x 2048 token prefills taking longer than default 30s timeout
+    # timeout is 3x VLLM_RPC_TIMEOUT
+    os.environ["VLLM_RPC_TIMEOUT"] = "200000"  # 200000ms = 200s
```
milank94, line 42: Does it make sense to put this in the `.env` file that gets passed during the launch? Along with the other variables such as `WH_ARCH_YAML=wormhole_b0_80_arch_eth_dispatch.yaml`

tstescoTT: linking to conversation at https://github.com/tenstorrent/tt-inference-server/pull/45#discussion_r1862878924

I'd recommend keeping this here for now, and refactor out into a separate config file that we can keep in version control.

milank94 (approved):

vllm-tt-metal-llama3-70b/src/run_vllm_api_server.py
```diff
@@ -30,11 +27,22 @@ def get_encoded_api_key(jwt_secret):
     return encoded_jwt
 
 
-@patch.object(LLMEngine, "__init__", new=logging_init_wrapper)
```
mvanniasingheTT, line 33: I may be missing something, but without using the wrapper, how is the raw logging class for the api server being initialized?

tstescoTT: We will use it for benchmarking only. We're not enabling the stats logging by default in the main run script.

PR #43, opened 2024-11-27
Tests with mock model vllm open ai inference server

tstescoTT:
# change log

- adding Dockerfile for mock model vllm open ai server
- fix mock model from sending stop token early
- adding tests for prompt gen scripts

### DIFF
--- a/tests/mock_vllm_model.py
+++ b/tests/mock_vllm_model.py
@@ -1,13 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
 import copy
-import os
 import time
-from datetime import datetime
-import json
 from dataclasses import dataclass
 from typing import List
-from pathlib import Path
 
 import torch
+
+from vllm.engine.metrics import logger
 
 from models.demos.t3000.llama2_70b.tt.llama_common import (
     setup_llama_env,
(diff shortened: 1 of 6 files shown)


tests/mock_vllm_model.py
```diff
@@ -286,24 +272,24 @@ def decode_forward(
...
-                )
-            else:  # if start_pos is a tensor ie. is different across batch, now in decode mode
-                # if start position is greater than index to send EOT
-                cache_idxs = start_pos.to(dtype=torch.int64)
-                send_token_mask = cache_idxs > send_index
-                # find positions where start pos passes send_index (ie. done decoding) + make 1D
-                batch_indices = torch.nonzero(send_token_mask).squeeze()
-                # assign a high logit at at the send _token index so model will select it and generate the EOT so that generation stops
-                logits[batch_indices, 0, send_token] = 100.0
+        # EOT_ID = 128009
```
milank94, line 275: Should we leave this commented out?

tests/utils/test_prompt_generation.py
milank94, line 1: Is this related to work from #41 ?

tstescoTT: ya I just put all the tests related changes in this PR. If you think its worth the effort I can move them onto #41.

milank94: When you first merge in #41 , this will just come ontop?

tstescoTT: Yupe that works.

tests/mock_vllm_model.py
```diff
@@ -286,24 +272,24 @@ def decode_forward(
...
+        # # EOS_ID = 128001
+        # send_index = 200
```
mvanniasingheTT, line 277: nice catch! but, can we have a variable fixed EOR send index based on the max number of tokens allowed when the mock model get initialized? Or does it not matter that there is an EOT at the end of the output sequence for the mock model?

tstescoTT: It doesn't matter for the mock model, vLLM will cut off the generation as expected. I'm leaving the code block in case we need it to test a specific case where a token gets sent at specific time/output length.

tests/mock_vllm_model.py
tstescoTT: I'm leaving the code block in case we need it to test a specific case where a token gets sent at specific time/output length. If thats unattractive I can break it out into a function that is turned off.

tests/mock_vllm_model.py
mvanniasingheTT: That makes sense. Thanks for clarifying!

mvanniasingheTT (approved):

milank94 (approved):

tests/mock_vllm_model.py
tstescoTT: added in https://github.com/tenstorrent/tt-inference-server/pull/43/commits/1e82b4201e0300560181067ccc1b389f07a96c40

PR #50, opened 2024-11-29
Add Locust testing scripts and configurations

milank94:
Adding scripts to run [Locust](https://locust.io/) server load testing. The purpose of the Locust tests is to provide detailed server metrics such as Requests Per Second (RPS).


locust/locustfile.py
```diff
@@ -0,0 +1,67 @@
...
+def load_custom_data(**kwargs):
+    global data_iter
+    data_iter = DataReader()
+
+
+class ServeUser(FastHttpUser):
+    # Set test parameters
+    network_timeout = NETWORK_TIMEOUT
+    connection_timeout = CONNECTION_TIMEOUT
+    headers = {"Authorization": get_authorization()}
```
tstescoTT, line 51:
```suggestion
    headers = {"Authorization": f"Bearer {get_authorization()}"}
```

tstescoTT: With adding "Bearer " to the Auth header locust tests ran fine with JWT auth turned on for me.

tstescoTT (approved):
After adding the Auth header change the locust tests are running as expected. Tested with `tests/mock_vllm_api_server.py`. Approved on condition we add that minor fix.

locust/locustfile.py
milank94: I kept getting "unauthorized" with Bearer. I need to check which I was using, jwt or auth. But, can you test out both and recommend which method we should stick to?

PR #55, opened 2024-12-12
adding online benchmarking scripts

tstescoTT:
# change log

- add utils/prompt_client.p::PromptClient as vLLM client handling authentication and health checks
- addres trace capture: https://github.com/tenstorrent/tt-inference-server/issues/56
- improve prompt generation and handling with utils/prompt_configs.py and utils/batch_processor.py
- remove explictly setting stop token in prompt client, this causes issues with instruct models correctly configured with instruct tokenizer
- add trace capturing ahead of performance measurement in benchmarking scripts
- add online benchmarking script using vllm/benchmarks/benchmark_serving.py
- add vllm benchmarking patch at benchmarking/benchmark_serving.patch handling best_of which is unsupported in current Tenstorrent vllm fork
- add benchmarking/prompt_client_online_benchmark.py to measure performance with different batch handling
- update benchmarking docs
- update prompt CLI and util docs
- update mock model to be faster and not send stop tokens unexpected
- add `benchmarking`, `evals`, and `tests` to Docker image vllm-tt-metal-llama3-70b/vllm.llama3.src.Dockerfile


benchmarking/README.md
```diff
@@ -36,3 +36,78 @@ python examples/offline_inference_tt.py --measure_perf --max_seqs_in_batch 32 --
...
+    --dataset random
+```
+
+#### using vllm/benchmarking/benchmark_serving.py
+Within the Docker container, use the benchmark_serving.patch file:
+```
+cd ~/app/src
+python run_vllm_api_server.py
+```
+This simply stops the benchmarking script from sending the `best_of` arg which is not supported and causes issues.
```
milank94, line 61: Is there an issue open for this in vLLM? Should it be something we open and get a permanent fix for?

benchmarking/README.md
```diff
@@ -36,3 +36,78 @@ python examples/offline_inference_tt.py --measure_perf --max_seqs_in_batch 32 --
 - `--max_seqs_in_batch` (default: `32`):
   - **Maximum batch size** for inference, determining the number of prompts processed in parallel.
 
+### Online Benchmarking
+
+#### single user
+
+```bash
+python utils/prompt_client_cli.py \
```
milank94, line 44: Does this give any performance metrics back for the single user?

benchmarking/README.md
```diff
@@ -36,3 +36,78 @@ python examples/offline_inference_tt.py --measure_perf --max_seqs_in_batch 32 --
...
+Output token throughput (tok/s):         166.07
+Total Token throughput (tok/s):          10794.77
+--------------------------------------------------
+               Time to First Token                  
+--------------------------------------------------
+Mean TTFT (ms):                          358.26
+Median TTFT (ms):                        358.45
+P99 TTFT (ms):                           361.67
+--------------------------------------------------
+     Time per Output Token (excl. 1st token)
```
milank94, line 93: From their reporting, what is different between TPOT and ITL?

benchmarking/vllm_online_benchmark.py
```diff
@@ -0,0 +1,128 @@
...
+    timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    result_dir = cache_dir / "vllm_online_benchmark_results" / f"results_{timestamp}"
+    result_dir.mkdir(parents=True, exist_ok=True)
+
+    prompt_client = PromptClient(env_config)
+    # note: there isnt a better way to pass an api key to the vllm benchmarking script
+    os.environ["OPENAI_API_KEY"] = prompt_client._get_authorization()
+
+    # Define benchmarking context length (isl, osl) pairs
+    context_lens = [
```
milank94, line 79: Does this need to be coordinated with what is set in `prompt_client_online_benchmark.py`?

utils/batch_processor.py
milank94, line 1: Can you provide a description of what this Class is trying to achieve?

benchmarking/vllm_online_benchmark.py
tstescoTT: No, only the vLLM benchmark script https://github.com/tenstorrent/vllm/blob/dev/benchmarks/benchmark_serving.py uses this authorization method.

benchmarking/README.md
tstescoTT: It prints them to stdout, I think this will be simpler to add to the `run_vllm_api_server.py` script so it covers that case as well. This could be done by setting the `--request-rate` for that benchmark to a number higher than e2e request latency.

benchmarking/README.md
tstescoTT: It's an issue with how we support parameters in our vLLM fork. Just filed https://github.com/tenstorrent/vllm/issues/44 to support this.

benchmarking/vllm_online_benchmark.py
milank94: Sorry, I meant the `context_lens`

tstescoTT: They dont need to be but if for example we want to run the same context ISL/OSL pairs for max batch (using vllm benchmarking script) and single user with prompt_client_online_benchmark.py then they would need to be the same.

benchmarking/README.md
tstescoTT: TPOT is the average ITL over the request. Only TTFT, ITL, and latency (request latency) are raw measurements made. TPOT is calculated in https://github.com/tenstorrent/vllm/blob/dev/benchmarks/benchmark_serving.py#L311

utils/batch_processor.py
tstescoTT: BatchProcessor runs multiple concurrent requests to the backend inference server (vLLM in this case). This adds some functionality for sending requests with a specific max number of requests allowed that is independent with the backend batch_size. Mostly this is for testing continous batching and seq lens, but can be used as an alternative method for benchmarking as in `benchmarking/prompt_client_online_benchmark.py` measuring TTFT as experienced by users by not exceeding the backend concurrent user capacity and having requests queued on the backend server before processing starts by the model.

milank94: A suggestion is to add that description to the file / Class. That's a good explanation for the user.

milank94 (approved):
Looks great. Pending one suggestion to add a description under `BatchProcessor`.

PR #60, opened 2024-12-19
Fix early exiting behaviour in setup_weights and add setup of cache d…

### LINKED ISSUE #57: vLLM llama3.1-70b setup script not creating tt_metal_cache/cache_repacked-llama-3.1-70b-instruct directory
In testing of deploying this server for another application, I believe I have encountered a problem in running the setup script for the vLLM llama3.1-70b model, `tt-inference-server/vllm-tt-metal-llama3-70b/setup.sh`

The problem is that the `tt_metal_cache/cache_repacked-llama-3.1-70b-instruct` directory is not being created, causing the server to fail to start. As far as I know, the `TtLlamaModelForGeneration` from tt-metal is responsible for constructing the directory, using the `LLAMA3_CACHE_PATH` environment variable (which is set by the setup.sh script). This behaviour needs to be investigated to understand why this directory is not being created in the persistent volume.

bgoelTT:
This PR closes #57. I have:
- changed the `setup.sh` script to not exit if the weight directory has been setup already (needed to run a new step)
- added a new `setup_tt_metal_cache()` step to the `setup.sh` script to create the TT Metal cache directory if it does not already exist

### DIFF
--- a/vllm-tt-metal-llama3-70b/setup.sh
+++ b/vllm-tt-metal-llama3-70b/setup.sh
@@ -277,7 +277,7 @@ setup_weights() {
         echo
         echo "If directory does not have correct weigths, to re-download or copy the model weights delete the directory."
         echo "✅ Model weights setup is already complete, check if directory contents are correct."
-        exit 1
+        return 0
     fi
 
     # TODO: support HF_TOKEN for downloading models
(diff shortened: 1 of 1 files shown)


tstescoTT (approved):
LGTM

Issue #64, opened 2024-12-24
Add API authorization to YOLOv4 server


bgoelTT:
Currently the YOLOv4 server lacks API key authorization for it's main inference request endpoint. This needs to be added and should use a similar system to the other servers (see TT-Metal Llama 3.1 70B, and TT-Metal Mistral 7B). This system uses a JWT (JSON Web Token) to encode information in the "Authorization" header.

See https://github.com/tenstorrent/tt-inference-server/blob/dddad09bb7bbca9d2a86595fc8adcea60ef30111/tt-metal-llama3-70b/src/inference_api_server.py#L474 for how API authentication works on the server and https://github.com/tenstorrent/tt-inference-server/blob/main/tt-metal-llama3-70b/README.md#jwt_token-authorization for how to send requests with the corresponding API key

PR #65, opened 2024-12-30
Add JWT API key authentication to YOLOv4

bgoelTT:
This PR addresses #64 by adding JWT API key authentication to the main inference endpoint. Authentication is performed similarly to the other inference servers, where the `Authentication` header is sent with a Bearer token.

# Tests
Tests were added to `tt-metal-yolov4/tests/test_inference_api.py` to demonstrate valid and invalid API requests.

### DIFF
--- a/tt-metal-yolov4/server/scripts/jwt_util.py
+++ b/tt-metal-yolov4/server/scripts/jwt_util.py
@@ -1,0 +1,52 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+#!/usr/bin/env python3
+import argparse
+import json
+import os
+
+import sys
+
+import jwt
+
+
+def main():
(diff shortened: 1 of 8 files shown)


tt-metal-yolov4/server/scripts/jwt_util.py
milank94, line 1: I'm assuming this will be part of a larger restructuring effort later on, but we should think about moving some of these utils in a common folder path (no directly in the model-specific implementation folder).

milank94 (approved):
Updates look good. Some reorganization could be done to consolidate some of the common utils such as `jwt_util.py` with the other models.

@sbennettTT please ensure that these updates are propagated to AI Playground

tt-metal-yolov4/server/scripts/jwt_util.py
bgoelTT: Yes we should refactor the repo in general and specifically, I have filed an issue here #66 to capture the common utilities refactoring

sbennettTT (approved):
nothing crazy, good authentication implementation

Issue #37, opened 2024-11-15
Initial vLLM setup fails due to missing HuggingFace permissions


milank94:
When following initial setup steps from: https://github.com/tenstorrent/tt-inference-server/tree/main/vllm-tt-metal-llama3-70b#vllm-tt-metalium-llama-31-70b-inference-api, fails due to missing HF token permissions to download [config](https://huggingface.co/meta-llama/Meta-Llama-3.1-70B/resolve/main/config.json) for `Meta-Llama-3.1-70B`.

```bash
docker run \
>   --rm \
>   -it \
>   --env-file .env \
>   --cap-add ALL \
>   --device /dev/tenstorrent:/dev/tenstorrent \
>   --volume /dev/hugepages-1G:/dev/hugepages-1G:rw \
>   --volume ${PERSISTENT_VOLUME?ERROR env var PERSISTENT_VOLUME must be set}:/home/user/cache_root:rw \
>   --shm-size 32G \
>   --publish 7000:7000 \
>   ghcr.io/tenstorrent/tt-inference-server/tt-metal-llama3-70b-src-base-vllm:v0.0.1-tt-metal-385904186f81-384f1790c3be
2024-11-15 09:07:01.845 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.pearson_correlation_coefficient be migrated to C++?
2024-11-15 09:07:01.847 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.Conv1d be migrated to C++?
2024-11-15 09:07:01.848 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.conv2d be migrated to C++?
2024-11-15 09:07:01.851 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.unsqueeze_to_4D be migrated to C++?
2024-11-15 09:07:01.852 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.from_torch be migrated to C++?
2024-11-15 09:07:01.852 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.to_torch be migrated to C++?
2024-11-15 09:07:01.852 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.to_device be migrated to C++?
2024-11-15 09:07:01.852 | WARNING  | ttnn.decorators:operation_decorator:801 - Should ttnn.from_device be migrated to C++?
[...]

Issue #23, opened 2024-10-24
Docker run support for HF_TOKEN authentication using env var pass in


tstescoTT:
To keep convention with vLLM Docker containers described in https://github.com/vllm-project/vllm/blob/main/docs/source/serving/deploying_with_docker.rst, e.g.:
```
docker run --runtime nvidia --gpus all \
    -v ~/.cache/huggingface:/root/.cache/huggingface \
    --env "HUGGING_FACE_HUB_TOKEN=<secret>" \
    -p 8000:8000 \
    --ipc=host \
    vllm/vllm-openai:latest \
    --model mistralai/Mistral-7B-v0.1
```

`HUGGING_FACE_HUB_TOKEN` is being deprecated in favor of `HF_TOKEN` as defined in https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#deprecated-environment-variables, so it's recommended to use `HF_TOKEN` here.

This could be used to allow users to download models without having to enter credentials manually where possible.

tstescoTT:
Addressed in https://github.com/tenstorrent/tt-inference-server/pull/67

Issue #14, opened 2024-09-26
llama model install script support for llama CLI and huggingface hub


tstescoTT:
Latest Llama model releases deprecate the download.sh scripts in favor of `llama` CLI: https://github.com/meta-llama/llama-models?tab=readme-ov-file#download

Similar support using huggingface can be added: https://github.com/meta-llama/llama-models?tab=readme-ov-file#access-to-hugging-face

tstescoTT:
Addressed in https://github.com/tenstorrent/tt-inference-server/pull/67

PR #67, opened 2025-01-14
Llama 3.x model support, setup.sh script multiple model support using HF download

tstescoTT:
# change log:
- add multiple model support using persistent_volume/model_envs/*.env
- setup using Hugging Face huggingface-cli to download models: llama model install script support for llama CLI and huggingface hub #14
- add model setup for llama 3.x
- address Initial vLLM setup fails due to missing HuggingFace permissions #37
- address Docker run support for HF_TOKEN authentication using env var pass in #23
- renamed vllm-tt-metal-llama3-70 to vllm-tt-metal-llama3 for all llama 3.x models
- updated documentation for v0 drop
- add Docker Ubuntu 22.04 option for vLLM llama 3.x


vllm-tt-metal-llama3/vllm.llama3.src.ubuntu-22.04-amd64.Dockerfile
milank94, line 1: Is there any other difference other than `FROM local/tt-metal/tt-metalium/ubuntu-22.04-amd64:$TT_METAL_DOCKERFILE_VERSION` between this Dockerfile an the one for 20.04?

Wondering if that could be parameterized and this combined into just a single file?

milank94 (approved):

vllm-tt-metal-llama3/vllm.llama3.src.ubuntu-22.04-amd64.Dockerfile
tstescoTT: I tried doing this with a multistage build, but it doesnt work well. Multistage builds are best when the fork happens later in the process and all combinations are desired to be built. Unfortunately, because tt-metal isnt publishing release images for Ubuntu 22.04 we have to build them manually here, and that makes avoiding the Ubuntu 22.04 build easier as the default so there are less steps. 

One way to do this is to put as much of the RUN commands as possible into a setup.sh script and copy then run that, but this wont cover everything (e.g. CMD, COPY, etc.) and would require extra testing.

The best way I found to do this is to use TT_METAL_DOCKERFILE_URL instead of TT_METAL_DOCKERFILE_VERSION so we can pass in a different base image entirely. This support locally built Ubuntu 22.04 images, and will support GHCR Ubuntu 22.04 tt-metal images once those are published.

tstescoTT: added this in https://github.com/tenstorrent/tt-inference-server/pull/67/commits/11141de96438fc2d61823a6bbf5ace4a0cc955d9

PR #80, opened 2025-01-27
YOLOv4 improvements

### LINKED ISSUE #63: Add health check to YOLOv4 server
The YOLOv4 server currently does not implement a `/health` endpoint to perform health checks. It would be useful to propose a concrete definition of what a "healthy" server is, so here are my and other's thoughts on what criteria a server should satisfy to be considered healthy:
- Ready to respond to inference requests: This means the server has successfully loaded models, initialized configurations, and prepared any necessary resources (e.g., caching, tracing) to serve requests
  - a binary status that should be evaluated set once
- Performing as expected under real-world conditions: The server responds to inference requests within a predefined latency threshold and with expected output integrity
  - a binary status that should be evaluated and set continuously

The YOLOv4 server's new `/health` endpoint should implement functionality to evaluate the above criteria.

bgoelTT:
Adds healthcheck endpoint and now performs resizing on server-side

Closes #63

### DIFF
--- a/tt-metal-yolov4/server/fast_api_yolov4.py
+++ b/tt-metal-yolov4/server/fast_api_yolov4.py
@@ -4,6 +4,7 @@
 import os
 import logging
 from fastapi import FastAPI, File, HTTPException, Request, status, UploadFile
+from fastapi.responses import JSONResponse
 from functools import wraps
 from io import BytesIO
 import jwt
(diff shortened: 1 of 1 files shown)


anirudTT (approved):
Tested on TT-Studio GUI.
LGTM!
Screenshot 2025-01-28 at 10 13 24 AM

PR #83, opened 2025-01-31
Fixes for Llama 3.x support, adding image input tooling, tt-studio environment variable handling

tstescoTT:
## change log

* Add benchmark summary support handling for vllm benchmark script, add documentation example
* Rename client side scripts batch_size options to max_concurrent to indicate client side concurrent request limits
* Stop-gap handling of MESH_DEVICE for Llama 3.x models
* Add image input support for image-text-to-text models in client scripts and tools
  - Added support for image input in trace capturing
  - Added new parameters for image width and height
  - Implemented handling of both text-only and image+text trace captures
* Modified setup script improvements:
  - Improved environment variable handling and persistence storage integration
  - Added IMPL_ID field (set to "tt-metal" for all current models)
  - Introduced MODEL_VERSION and MODEL_ID variables for better versioning
  - Enhanced HF token validation with character count check
* Fixed the vLLM model registration logic:
  - Added missing ModelRegistry.register_model call for TTLlamaForCausalLM for legacy implementation models
* Updated benchmark path handling to use $HOME environment variable instead of hardcoded /home/user path
* Added support for a new model "DeepSeek-R1-Distill-Llama-70B" in the model setup configurations

### DIFF
--- a/benchmarking/vllm_online_benchmark.py
+++ b/benchmarking/vllm_online_benchmark.py
@@ -126,8 +126,9 @@ def main():
             / f"vllm_online_benchmark_{run_timestamp}_{mesh_device}_isl-{isl}_osl-{osl}_maxcon-{max_concurrent}_n-{num_prompts}.json"
         )
         logger.info(f"\nRunning benchmark {i}/{len(combinations)}")
+        user_home = os.environ.get("HOME")
         run_benchmark(
-            benchmark_script="/home/user/vllm/benchmarks/benchmark_serving.py",
+            benchmark_script=f"{user_home}/vllm/benchmarks/benchmark_serving.py",
             params=params,
             model=env_config.vllm_model,
             port=env_config.service_port,
--- a/setup.sh
+++ b/setup.sh
@@ -121,7 +121,7 @@ setup_model_environment() {
         META_DIR_FILTER=""
         REPACKED=1
         ;;
-        "llama-3.3-70b-instruct")
+        "Llama-3.3-70B-Instruct")
         IMPL_ID="tt-metal"
         MODEL_NAME="Llama-3.3-70B-Instruct"
         HF_MODEL_REPO_ID="meta-llama/Llama-3.3-70B-Instruct"
@@ -129,7 +129,7 @@ setup_model_environment() {
         META_DIR_FILTER=""
         REPACKED=1
         ;;
-        "llama-3.2-11b-vision-instruct")
+        "Llama-3.2-11B-Vision-Instruct")
         IMPL_ID="tt-metal"
         MODEL_NAME="Llama-3.2-11B-Vision-Instruct"
         HF_MODEL_REPO_ID="meta-llama/Llama-3.2-11B-Vision-Instruct"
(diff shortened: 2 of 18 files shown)


benchmarking/vllm_online_benchmark.py
```diff
@@ -126,8 +126,9 @@ def main():
             / f"vllm_online_benchmark_{run_timestamp}_{mesh_device}_isl-{isl}_osl-{osl}_maxcon-{max_concurrent}_n-{num_prompts}.json"
         )
         logger.info(f"\nRunning benchmark {i}/{len(combinations)}")
+        user_home = os.environ.get("HOME")
         run_benchmark(
-            benchmark_script="/home/user/vllm/benchmarks/benchmark_serving.py",
+            benchmark_script=f"{user_home}/vllm/benchmarks/benchmark_serving.py",
```
ppetrovicTT, line 131: can we use "vllm_dir" instead?

we set "vllm_dir" in the docker, and in vllm/tt-metal doc and tt-inference-server's development.doc it is mentioned as a mandatory variable, but i don't see references to it in code?

setup.sh
```diff
@@ -9,18 +9,19 @@ set -euo pipefail  # Exit on error, print commands, unset variables treated as e
...
-    echo "  llama-3.2-1b-instruct"
-    echo "  llama-3.1-70b-instruct"
-    echo "  llama-3.1-70b"
-    echo "  llama-3.1-8b-instruct"
-    echo "  llama-3.1-8b"
-    echo "  llama-3-70b-instruct"
-    echo "  llama-3-70b"
-    echo "  llama-3-8b-instruct"
-    echo "  llama-3-8b"
+    echo "  DeepSeek-R1-Distill-Llama-70B"
```
ppetrovicTT, line 12: i'm up for this, as it will align id with metal's expectations! (model_config.py, L137)

but we need we need to change the names in setup_model_environment as well then.
i've done that locally, will push soon. you can the usage chunk if you wish?

setup.sh
```diff
@@ -74,6 +75,7 @@ get_hf_env_vars() {
         echo "HF_TOKEN environment variable is not set. Please set it before running the script."
         read -r -s -p "Enter your HF_TOKEN: " input_hf_token
         echo
+        echo "entered HF_TOKEN contains: ${#input_hf_token} characters, expected 37."
```
ppetrovicTT, line 78: we can ping HF API to check if the token is valid
i have that in my change, will follow up!

ppetrovicTT (approved):

setup.sh
tstescoTT: Nice, can you open a separate PR on that or share the code you're using? I had a TODO to also check repo access for the target model and message to user on 404.

benchmarking/vllm_online_benchmark.py
tstescoTT: yupe, vllm_dir is better.

tstescoTT:
Moved this PR to https://github.com/tenstorrent/tt-inference-server/pull/88 to use the proposed RC git workflow.

PR #84, opened 2025-01-31
Dev flag in Docker; Split .env into host and container; Auto-patch VLLM

ppetrovicTT:
- Align casing in model names with tt-metal expectations
- Provide weights from a local folder
- Workaround non-working apt-get keyring
- Development flag in docker to allow sudo and debug symbols
- Patch VLLM for best-of arg while building docker
- Separate host and container .env files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ python_env
 
 # persistent storage volume
 persistent_volume
+model_envs
 
 # Output Files
 *netlist.yaml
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -16,6 +16,7 @@ python_env
 
 # persistent storage volume
 persistent_volume
+model_envs
```
tstescoTT, line 19: the `model_envs` dir should be inside the `persistent_volume` so we shouldn't need to add a .gitignore

ppetrovicTT: i've had an issue where peristent_volume was created inside tt-inference-server folder even though i've provided a different one. so i thought that a different folder should be better to have! i see this issue is now fixed on dev, so i'm not going to create another folder :)

ppetrovicTT:
Closing this PR, as I will be rolling out smaller ones, and based on dev branch.

PR #91, opened 2025-02-03
Provide a model from your local storage

ppetrovicTT:
- Third option to provide a model, next to downloading from HF and Meta.
- Fix when copying links in HF download
- Folder suggestions when choosing local folders

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -89,7 +89,7 @@ get_hf_env_vars() {
     # get HF_HOME
     if [ -z "${HF_HOME:-}" ]; then
         echo "HF_HOME environment variable is not set. Please set it before running the script."
-        read -r -p "Enter your HF_HOME [default: $HOME/.cache/huggingface]:" input_hf_home
+        read -e -r -p "Enter your HF_HOME [default: $HOME/.cache/huggingface]:" input_hf_home
         echo
         input_hf_home=${input_hf_home:-"$HOME/.cache/huggingface"}
         if [ ! -d "$input_hf_home" ]; then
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -229,7 +229,7 @@ setup_model_environment() {
     # Safely handle potentially unset environment variables using default values
     PERSISTENT_VOLUME_ROOT=${PERSISTENT_VOLUME_ROOT:-$DEFAULT_PERSISTENT_VOLUME_ROOT}
     # Prompt user for PERSISTENT_VOLUME_ROOT if not already set or use default
-    read -r -p "Enter your PERSISTENT_VOLUME_ROOT [default: ${DEFAULT_PERSISTENT_VOLUME_ROOT}]: " INPUT_PERSISTENT_VOLUME_ROOT
+    read -e -r -p "Enter your PERSISTENT_VOLUME_ROOT [default: ${DEFAULT_PERSISTENT_VOLUME_ROOT}]: " INPUT_PERSISTENT_VOLUME_ROOT
```
tstescoTT, line 232: adding readline editing is good idea.

tstescoTT (approved):
I like adding support for local model weights. Which models did you test the local download flow with?

ppetrovicTT:
> I like adding support for local model weights. Which models did you test the local download flow with?

For tt-metal demos, there are Llama weights on NAS - [link](https://github.com/tenstorrent/tt-metal/blob/27d4416e11ee44c37a4cd047d74936ba98d101c4/models/demos/llama3/lt#L1031)

I initially thought that we will be using these, and then figured it might be the case already for customers (and devs) already!

For my specific testing I was using Llama-3.2-1B-Instruct model.

tstescoTT:
> > I like adding support for local model weights. Which models did you test the local download flow with?
> 
> For tt-metal demos, there are Llama weights on NAS - [link](https://github.com/tenstorrent/tt-metal/blob/27d4416e11ee44c37a4cd047d74936ba98d101c4/models/demos/llama3/lt#L1031)
> 
> I initially thought that we will be using these, and then figured it might be the case already for customers (and devs) already!
> 
> For my specific testing I was using Llama-3.2-1B-Instruct model.

Great, thanks for mentioning the weights locations and model. Looks good to me!

PR #95, opened 2025-02-04
Separate host and container envs

ppetrovicTT:
I had a clashing LLAMA_DIR variable, which pushed me to separate environments for host and container.
Then I found it cleaner not to pollute container environment with host's variables.

Instead of keeping everything in persistent/{MODEL_NAME}.env, I made a folder "persistent/{MODEL_NAME}" and put a pair of files: host.env and container.env.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -29,25 +29,25 @@ usage() {
 # globals
 readonly REPO_ROOT=$(dirname "$(realpath "$0")")
 
-check_and_prompt_env_file() {
+check_and_prompt_host_env_file() {
     local MODEL_NAME_KEY="MODEL_NAME"
     local MODEL_NAME=""
     # Check if .env file exists
-    if [[ -f "${ENV_FILE}" ]]; then
+    if [[ -f "${HOST_ENV_FILE}" ]]; then
         # Extract the MODEL_NAME value from .env
-        echo "found ENV_FILE: ${ENV_FILE}"
-        FOUND_MODEL_NAME=$(grep "^$MODEL_NAME_KEY=" "$ENV_FILE" | cut -d '=' -f2) || FOUND_MODEL_NAME=""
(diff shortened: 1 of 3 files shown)


tstescoTT:
I'm not against this approach but I'll flag this will require changes in https://github.com/tenstorrent/tt-studio for compatibility and require users to run the setup.sh again to get the new env var file format.

How did the clash occur? I've found it confusing to have host and container paths but am not sure breaking them into separate files is the best approach. Ideally we could remove/deprecate some environment variables that arent needed anymore first.

ppetrovicTT:
When I was running other vLLM tests locally, I had to set up LLAMA_DIR on my host.
And the Docker image needed LLAMA_DIR pointing to a different location (inside the Docker).

I glanced at tt-studio, and I wonder would it be enough to just provide the container.env?

Actually, I have a third scenario:
I was curious how can we run tt-inference-server locally, without Docker, so that we have tt-metal and vLLM exposed for easier development and model debugging. And I would need the same variables , but pointing to the local folders, instead of docker ([Confluence](https://tenstorrent.atlassian.net/wiki/spaces/PSE/pages/874348699/Benchmarking+Llama+with+vLLM#Build-Packages), Step 3). So I thought of adding a third file - host-dev.env! I don't see this as a big problem, as they're all generated, but I might be missing something.

Please let me know what you think!

Thank you!

tstescoTT:
I finally got around to refactoring the setup_host.py code in https://github.com/tenstorrent/tt-inference-server/pull/223

TLDR; we no longer need the per model environment variable files! 

We're tracking the run locally without Docker feature in https://github.com/tenstorrent/tt-inference-server/issues/123

tstescoTT:
@ppetrovicTT do we still need this PR merged or have these changes been superseded and we can close this PR?

ppetrovicTT:
Let's just close it! Thanks!

Issue #76, opened 2025-01-22
setup.sh to check host for running out of disk space when setting up new model


tstescoTT:
When downloading all weights, repacking, tt-metal caches - it's not uncommon for the user to need to manage their storage capacity. Having the `setup.sh` script fail during download, repack, for the model crash during tt-metal cache generation because the host runs out of disk space is a nuisance to debug.

Possible solution:
The `setup.sh` script can check for a minimum amount of required disk space for each desired model.

tstescoTT:
Addressed in https://github.com/tenstorrent/tt-inference-server/pull/103

PR #103, opened 2025-02-07
 Release Candidate v0.0.3: add model implementation list, mark preview models

tstescoTT:
# change log
* update documentation to show all supported models and give correct links, mark experimental preview models
* setup.sh supports all base and Instruct models
* add host disk and RAM check in setup.sh addressing https://github.com/tenstorrent/tt-inference-server/issues/76 with #106 
* update setup documentation in vllm-tt-metal-llama3/README.md to describe other models and put setup and installation first to avoid confusion

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -9,13 +9,19 @@ set -euo pipefail  # Exit on error, print commands, unset variables treated as e
 usage() {
     echo "Usage: $0 <model_type>"
     echo "Available model types:"
-    echo "  Qwen2.5-72B-Instruct"
-    echo "  Qwen2.5-7B-Instruct"
-    echo "  DeepSeek-R1-Distill-Llama-70B"
+    echo "  Qwen2.5-72B-Instruct (preview)"
+    echo "  Qwen2.5-72B (preview)"
+    echo "  Qwen2.5-7B-Instruct (preview)"
+    echo "  Qwen2.5-7B (preview)"
+    echo "  DeepSeek-R1-Distill-Llama-70B (preview)"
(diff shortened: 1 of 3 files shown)


setup.sh
```diff
@@ -9,13 +9,19 @@ set -euo pipefail  # Exit on error, print commands, unset variables treated as e
...
+    echo "  Qwen2.5-7B (preview)"
+    echo "  DeepSeek-R1-Distill-Llama-70B (preview)"
     echo "  Llama-3.3-70B-Instruct"
-    echo "  Llama-3.2-11B-Vision-Instruct"
-    echo "  Llama-3.2-3B-Instruct"
-    echo "  Llama-3.2-1B-Instruct"
+    echo "  Llama-3.3-70B"
+    echo "  Llama-3.2-11B-Vision-Instruct (preview)"
+    echo "  Llama-3.2-11B-Vision (preview)"
+    echo "  Llama-3.2-3B-Instruct (preview)"
```
milank94, line 21: Will this cause confusion if the user enters "Llama-3.2-3B-Instruct (preview)", but the conditional is looking for "Llama-3.2-3B-Instruct"?

I would remove the "(preview)" from `setup.sh` as it can cause more confusion and points that need to be updated. Keeping this in the README is good.

tstescoTT: Agreed. I'll remove the preview marker from the setup.sh.

tstescoTT: fixed in #106

milank94 (approved):

PR #113, opened 2025-02-24
Support Stable Diffusion 1.4 & 3.5-medium

bgoelTT:
This PR introduces support for two new stable diffusion inference servers: Stable-Diffusion-1.4 and Stable-Diffusion-3.5-medium

The server implementations compose the corresponding models implemented in `tt-metal`. They are wrapped in a Flask API that is served via a Gunicorn WSGI application. The servers also make use of the persistent volume mechanism, so support has been added to `setup.sh`.


anirudTT:
Was able to test the `./setup.sh `to set up persistent volume mechanism for s.d 1.4 and 3.5 and successfully deploy both models via `tt-studio`.

setup.sh
```diff
@@ -199,15 +201,25 @@ setup_model_environment() {
     # MIN_DISK: safe lower bound on available disk (based on 2 bytes per parameter and 2.5 copies: HF cache, model weights, tt-metal cache)
     # MIN_RAM: safe lower bound on RAM needed (based on repacking 70B models)
     case "$1" in
-        "Qwen2.5-72B"|"Qwen2.5-72B-Instruct")
```
tstescoTT, line 202: It looks like this is deleting `"Qwen2.5-72B"|"Qwen2.5-72B-Instruct"` models definition in `setup.sh`.

setup.sh
```diff
@@ -9,6 +9,8 @@ set -euo pipefail  # Exit on error, print commands, unset variables treated as e
 usage() {
     echo "Usage: $0 <model_type>"
     echo "Available model types:"
+    echo "  Stable-Diffusion-1.4 (preview)"
```
tstescoTT, line 12: For other preview models we decided to not include the (preview) string here to avoid confusion that users need to type it to setup the model.

utils/authentication.py
tstescoTT, line 1: Could this file be re-named to e.g. `flask_authentication.py` so it doesnt confuse developers with it having different requirements and use than everything else in top level utils directory?

tstescoTT:
@bgoelTT can we close this PR in favor of new SDXL inference server implementation?

tstescoTT:
Thank you, lets keep the branch archived https://github.com/tenstorrent/tt-inference-server/tree/ben/sd3.5

Issue #117, opened 2025-02-24
Benchmark performance run script


tstescoTT:
Improve on https://github.com/tenstorrent/tt-inference-server/blob/main/benchmarking/vllm_online_benchmark.py and add a script that handles full setup, e.g. as described in https://gist.github.com/milank94/1b7c31556f5f6a13c56553e2cccc5823

## Requirements

- single bash command to setup and run benchmarks (vLLM server + client side requests)
- can be run either in tt-inference-server Docker image, or from an existing tt-metal development environment
- output is stored so that each run can be found and data extracted easily
- can parameterize over MESH_DEVICE={N150, N300, T3K} as supported by tt-transformers

PR #128, opened 2025-02-26
WIP: tstesco/benchmark-run-script

tstescoTT:
adding benchmarking/run_benchmarks.py to startup vllm server, run benchmarking client script, then tear down vllm server

Addresses https://github.com/tenstorrent/tt-inference-server/issues/117

### DIFF
--- a/benchmarking/run_benchmarks.py
+++ b/benchmarking/run_benchmarks.py
@@ -15,10 +15,13 @@ def main():
     vllm_log_file_path = (
         Path(os.getenv("CACHE_ROOT", ".")) / "logs" / f"run_vllm_{log_timestamp}.log"
     )
-    vllm_log = open(vllm_log_file_path, "w")
+    vllm_log_file_path.parent.mkdir(parents=True, exist_ok=True)
+    # note: line buffering of log file to reduce disk IO overhead on vLLM run process
+    # run_vllm_api_server.py uses runpy so vLLM process inherits sys.stdour and std.stderr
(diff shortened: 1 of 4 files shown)


benchmarking/run_benchmarks.py
```diff
@@ -0,0 +1,65 @@
...
+    vllm_process = subprocess.Popen(
+        ["python", "-u", "/home/container_app_user/app/src/run_vllm_api_server.py"],
+        stdout=vllm_log,
+        stderr=vllm_log,
+        text=True,
+        env=env_vars,
+    )
+
+    # set benchmarking env vars
+    # TODO: figure out the correct MESH_DEVICE, use same logic as run vllm script
```
stisiTT, line 29: Lifted this from [vllm_online_benchmark.py](https://github.com/tenstorrent/tt-inference-server/blob/main/benchmarking/vllm_online_benchmark.py) and repurposed it for here - could this work as a solution for the TODO?

from utils.prompt_configs import EnvironmentConfig
env_config = EnvironmentConfig()
env_vars["MESH_DEVICE"] = env_config.mesh_device

tstescoTT: That pulls the MESH_DEVICE from environment variables https://github.com/tenstorrent/tt-inference-server/blob/dev/utils/prompt_configs.py#L42

For 1st pass I was thinking the default mapping in https://github.com/tenstorrent/tt-inference-server/blob/dev/vllm-tt-metal-llama3/src/run_vllm_api_server.py#L68

Ideal solution determines this based on the physical hardware is being targeted.

PR #158, opened 2025-03-20
Fixing timeout issue for 3.2 11B Vision-Instruct, only

stisiTT:


evals/eval_config.py
```diff
@@ -128,6 +128,7 @@ class EvalConfig:
                 apply_chat_template=False,
                 use_chat_api=True,
                 batch_size=16,
+                gen_kwargs={**{"stream": "False"}, "timeout": "600"},
```
tstescoTT, line 131: is this needed in gew_kwargs AND the top level config for lm-eval==0.4.3?

PR #167, opened 2025-03-27
tstesco/user-docs

tstescoTT:
# change log
* add docs/workflows_user_guide.md
* add link to model code in main README.md LLM table
* add Llama 3.2 11B Vision tt_metal_commit=v0.56.0-rc47
* deprecate warning for manual evals and benchmarking documentation


workflows/model_config.py
```diff
@@ -254,13 +273,15 @@ def infer_param_count(hf_model_repo: str) -> int:
         tt_metal_commit="v0.56.0-rc47",
         vllm_commit="e2e0002ac7dc",
         status="supported",
+        code_link="https://github.com/tenstorrent/tt-metal/tree/v0.56.0-rc47/models/demos/llama3",
     ),
     ModelConfig(
         device_configurations={DeviceTypes.N150, DeviceTypes.N300, DeviceTypes.T3K},
         hf_model_repo="meta-llama/Llama-3.1-8B",
```
stisiTT, line 280: No Instruct? I am noticing no Instruct models are listed at all on tt-metal. Do we not technically support them, or is it implied they're supported?

docs/workflows_user_guide.md
```diff
@@ -0,0 +1,305 @@
...
+
+and stopped with:
+```bash
+# container ID was 
+docker stop <container-id>
+```
+
+#### Model setup
+
+On first run, if the model weights are not already downloaded the `workflows/setup_host.py` script will be called.
```
stisiTT, line 124: suggested grammar change "if [the model weights are not already downloaded], then..."

docs/workflows_user_guide.md
```diff
@@ -0,0 +1,305 @@
...
+
+The `benchmarks` workflow
+
+```bash
+python3 run.py --model Llama-3.2-1B-Instruct --device n300 --workflow benchmarks
+```
+
+### Benchmarking Steps
+
+Lets dissect the runtime logs (stdout and stderr + streamed line-by-line to `workflow_logs/run_logs`):
```
stisiTT, line 153: I don't know how much we care about getting specific about Active vs. Passive writing in our documentation. But if we do, then this line could be made more active. "Here, we dissect the runtime logs..." might be closer. FWIW, I've been told technical documentation is typically more active.

stisiTT (approved):

docs/workflows_user_guide.md
tstescoTT: accepted proposed language 👍

docs/workflows_user_guide.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Server workflow
+
+The `server` workflow runs the vLLM inference server for the model in detached process using either `--docker-server` or `--local-server` and exits.
+
+Once the server is running, multiple runs of the client side workloads, i.e. `benchmarks` and `evals` can be run without tearing the model inference server down. This saves time during development.
+
+#### Docker server
+
+To run the inference server with docker use `--docker-server` flag:
```
stisiTT, line 103: Do we state somewhere that an already-running server (that meetings the reqs) is an option? The options above don't seem to make that clear (although I could see it being implied), and I could fathom us being wary of saying that considering the version control issues with builds. 

Just wondering if we can also find a way to save ourselves from the question of whether or not users require the server workflow.

tstescoTT: worth mentioning, I'll add that.

PR #228, opened 2025-05-09
Add --override-docker-image

acvejicTT:
As a part of this issue https://github.com/tenstorrent/tt-shield/issues/10 working on building a Docker image as part of the CI/CD process in the `tt-shield` repo.
To enable this, I had to add the `--override-docker-image` argument in the code.
I started from @vmilosevic 's branch, so the PR includes additional changes beyond mine.


ppetrovicTT:
Overall, I'd keep the changes relevant only for the docker image override and ignore everything else we're doing to stabilize the CI. It would be cleaner for the commit, I think.

acvejicTT:
It didn’t work when I used main with my changes, which is why I started from Vladimir’s branch — the one we knew was working.
I’ll create a branch from dev and test it. It’s quite possible that the original issue wasn’t related to the branch but something else entirely.

ppetrovicTT:
> It didn’t work when I used main with my changes, which is why I started from Vladimir’s branch — the one we knew was working. I’ll create a branch from dev and test it. It’s quite possible that the original issue wasn’t related to the branch but something else entirely.

sure, i don't expect it to work end-to-end.. but let's go step-by-step! this should just open the possibility to override the docker image - if that happens, that's already enough :)

PR #229, opened 2025-05-09
Server Side Tokenization

stisiTT:
These codes modify PromptClient for server-side tokenization, provides: 

1. brute force exploration to test lossiness of tokenizers and their ability to be stabilized through truncation
2. function to generate a stable random prompt


tstescoTT (changes requested):
How do you recommend we use this code in this repo? There is the analysis code with separate dependencies from the main repo code. Maybe we should have an `investigations` directory for stuff like that.

Can you add a module or example for how to integrate with the benchmarking script at https://github.com/tenstorrent/vllm/blob/tstesco/benchmark-uplift/benchmarks/benchmark_serving.py?

stisiTT:
When I finish the update for tests (name of that workflow to also be changed for clarity), where I'll be building a new bare-bones benchmarking-like script I'll use instead of vLLM's. From there, it should become clearer how and where we should use it.

tstescoTT:
Could we still use vLLM's benchmarking script if possible and just patch it to use your prompt generation?

tstescoTT:
can we close this in favour of https://github.com/tenstorrent/tt-inference-server/pull/272?

PR #245, opened 2025-05-16
Llama3 on Blackhole

bgoelTT:
This PR adds a new `ModelConfig` for Blackhole p100 and p150 devices on Llama-3.1-8B(-Instruct).

This PR adds a couple new features:
* ability to have multiple `ModelConfig`s for the same weight-modelImpl combination for different devices
* model_id is now composed of
  * model_name
  * implementation_id
  * device_type


tstescoTT (approved):
LGTM

PR #257, opened 2025-05-21
Add device-id argument and use persistent volume env variable if provided

acvejicTT:
- In our CI setup, we have a machine with two devices. To select which one to use, we need to provide a `device-id`.
- The `PERSISTENT_VOLUME_ROOT` environment variable was not used during the setup.

I ran 3 tests using this branch to validate the behavior with device-id:
- `llmbox` without `device-id` arg

- `N150` - `device-id=1`
https://github.com/tenstorrent/tt-shield/actions/runs/15163158332/job/42634094551
- `N300` - `device-id=0`
https://github.com/tenstorrent/tt-shield/actions/runs/15163984147/job/42636933194

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -79,6 +79,10 @@ def run_docker_server(args, setup_config):
     mesh_device_str = DeviceTypes.to_mesh_device_str(device)
     container_name = f"tt-inference-server-{short_uuid()}"
 
+    device_path = "/dev/tenstorrent"
+    if hasattr(args, 'device_id') and args.device_id is not None:
+        device_path = f"{device_path}/{args.device_id}"
+
     if args.dev_mode:
         # use dev image
         docker_image = docker_image.replace("-release-", "-dev-")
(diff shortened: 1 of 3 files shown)


workflows/run_docker_server.py
```diff
@@ -114,7 +118,7 @@ def run_docker_server(args, setup_config):
         "--name", container_name,
         "--env-file", str(default_dotenv_path),
         "--cap-add", "ALL",
-        "--device", "/dev/tenstorrent:/dev/tenstorrent",
+        "--device", f"{device_path}:{device_path}",
```
tstescoTT, line 121: Do we want to map devices equivalently in the container? Previously we needed everything in the container to be device 0. i.e `/dev/tenstorrent/1:/dev/tenstorrent/0`

acvejicTT: I tested it and it seems that it works.
Here is 1:1 mapping:
https://github.com/tenstorrent/tt-shield/actions/runs/15163158332/job/42634094551#step:8:88

and here it is 0:0:
https://github.com/tenstorrent/tt-shield/actions/runs/15163984147/job/42636933194#step:8:88

I can test 1:0 mapping, but I'm expecting it will work too.

acvejicTT: Another solution would could be to keep mapping `/dev/tenstorrent:/dev/tenstorrent` but pass an argument to container which device-id to use.

acvejicTT: @tstescoTT 
I got an info from the team that they are working on `TT_METAL_VISIBLE_DEVICES` option which will enable tt-metal to run on specified devices, even when multiple are mapped.
Until then I think this implementation is valid.

Issue #240, opened 2025-05-14
HF cache downloads incomplete models


tstescoTT:
on commit: [8902008](https://github.com/tenstorrent/tt-inference-server/commit/89020084f29bdde434717bfd315ab4bdffd78c71)

`setup_host.py` downloads of HF models are missing main model files, e.g. for Qwen2.5-7B-Instruct in https://huggingface.co/Qwen/Qwen2.5-7B-Instruct/tree/main

```
$ ll huggingface/hub/models--Qwen--Qwen2.5-7B-Instruct/snapshots/a09a35458c702b33eeacc393d103063234e8bc28/
merges.txt             tokenizer_config.json  tokenizer.json         vocab.json 
```

The error message shown was unhelpful:
```
2025-05-14 22:52:06.453 | INFO     | models.tt_transformers.tt.model_config:__init__:397 - Cache directory: /home/container_app_user/cache_root/tt_metal_cache/cache_Qwen2.5-7B-Instruct/N300
Process SpawnProcess-1:
Traceback (most recent call last):
  File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/container_app_user/vllm/vllm/engine/multiprocessing/engine.py", line 391, in run_mp_engine
    engine = MQLLMEngine.from_engine_args(engine_args=engine_args,
  File "/home/container_app_user/vllm/vllm/engine/multiprocessing/engine.py", line 139, in from_engine_args
    return cls(
  File "/home/container_app_user/vllm/vllm/engine/multiprocessing/engine.py", line 78, in __init__
    self.engine = LLMEngine(*args, **kwargs)
  File "/home/container_app_user/vllm/vllm/engine/llm_engine.py", line 335, in __init__
    self.model_executor = executor_class(
  File "/home/container_app_user/vllm/vllm/executor/executor_base.py", line 47, in __init__
    self._init_executor()
  File "/home/container_app_user/vllm/vllm/executor/tt_executor.py", line 28, in _init_executor
    self.driver_worker.load_model()
  File "/home/container_app_user/vllm/vllm/worker/tt_worker.py", line 209, in load_model
    self.model_runner.load_model()
  File "/home/container_app_user/vllm/vllm/worker/tt_model_runner.py", line 145, in load_model
[...]

Issue #262, opened 2025-05-22
initial setup fails on 2nd run when setup flow exits unexpectedly


tstescoTT:
Some examples of bugs under this umbrella that must be manual resolved by removing the directories:
- If the setup_hosts.py model download fails the partially downloaded files may show as fully downloaded causing errors, or the repo may go through model setup flow every single time as in https://github.com/tenstorrent/tt-inference-server/issues/240
- If run_vllm_api_server.py was run initially using cloud image the sym link dir in persistent_volume may be wrong permissions as in https://github.com/tenstorrent/tt-inference-server/issues/255

tstescoTT:
Need to test if this is fixed and can be closed.

PR #265, opened 2025-05-22
fix HF download issue

tstescoTT:
# changelog

* fix https://github.com/tenstorrent/tt-inference-server/issues/240 which seems to have been caused by making models re-download after failed detailed check for files present only to then gate re-download by check for dir existence.
* set HF_HOME from HOST_HF_HOME passed to run.py so its used correctly.
* fix related issues in https://github.com/tenstorrent/tt-inference-server/issues/262
* add default logging of the tt-inference-server git sha to run_8.log

### DIFF
--- a/run.py
+++ b/run.py
@@ -7,6 +7,7 @@ import os
 import argparse
 import getpass
 import logging
+import subprocess
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


bgoelTT:
One thing that has happened to me, is if execute `workflows/build_release_docker_images.py` but cancel it half-way through because I want to change something (i.e. tt-metal commit SHA).

These directories still exist https://github.com/tenstorrent/tt-inference-server/blob/a64353a547a9d08c9c604b1546adfa692d707815/workflows/build_docker.sh#L173-L176

Then if I execute it again, it complains that `tt-metal` has already been cloned and it exits.

tstescoTT:
> One thing that has happened to me, is if execute `workflows/build_release_docker_images.py` but cancel it half-way through because I want to change something (i.e. tt-metal commit SHA).
> 
> These directories still exist
> 
> https://github.com/tenstorrent/tt-inference-server/blob/a64353a547a9d08c9c604b1546adfa692d707815/workflows/build_docker.sh#L173-L176
> 
> Then if I execute it again, it complains that `tt-metal` has already been cloned and it exits.

Tracking this in another issue 👍 https://github.com/tenstorrent/tt-inference-server/issues/266

bgoelTT (approved):

PR #286, opened 2025-05-28
Failures handling

ppetrovicTT:
- When vLLM server fails abort benchmarks/evals
- Reduce timeouts for server start
- Use sys.exit to propagate return codes between subprocesses

### DIFF
--- a/benchmarking/run_benchmarks.py
+++ b/benchmarking/run_benchmarks.py
@@ -197,8 +197,13 @@ def main():
     env_config.jwt_secret = args.jwt_secret
     env_config.service_port = args.service_port
     env_config.vllm_model = model_config.hf_model_repo
+
     prompt_client = PromptClient(env_config)
-    prompt_client.wait_for_healthy(timeout=7200.0)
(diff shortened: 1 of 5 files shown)


benchmarking/run_benchmarks.py
```diff
@@ -197,8 +197,13 @@ def main():
     env_config.jwt_secret = args.jwt_secret
     env_config.service_port = args.service_port
     env_config.vllm_model = model_config.hf_model_repo
+
     prompt_client = PromptClient(env_config)
-    prompt_client.wait_for_healthy(timeout=7200.0)
+    if not prompt_client.wait_for_healthy(timeout=30 * 60.0):
```
tstescoTT, line 202: Because the time the model takes on 1st run is specific to the model this should be a model_config.py parameter. Also possible to set an option not wait for the max 1st run time (including tt-metal cache generation).

Until this can be addressed I'm happy with setting to 30 mins.

tstescoTT: Added https://github.com/tenstorrent/tt-inference-server/issues/288 to track adding that.

PR #290, opened 2025-05-29
Make sure to install torch for cpu

ppetrovicTT:
After v1.13 the CPU version of Torch is available only through wheel, not through PyPI.


tstescoTT (commented):
LGTM

PR #300, opened 2025-06-01
Update register_tt_models to use string paths instead of imports to support newer vLLM versions

skhorasganiTT:
- In newer versions of vLLM, registered model classes that are imported must be PyTorch classes. Modify model registration to use string paths instead of imports to use soft registration to bypass that requirement for TT models. Also with this change, only the model being evaluated will be imported and not other TT models.
- Fix renamed vLLM logging import
- Note: The changes from this PR are required when https://github.com/tenstorrent/vllm/tree/dev has been updated to a more recent upstream version (April 2025).

Shield CI for Llama1B-T3K (dispatched with https://github.com/tenstorrent/tt-metal/pull/22720, rebased vLLM branch, and the tt-inference-server commit from this PR): https://github.com/tenstorrent/tt-shield/actions/runs/15450428085

### DIFF
--- a/utils/logging_utils.py
+++ b/utils/logging_utils.py
@@ -29,12 +29,17 @@ LOG_PATH = Path(os.getenv("CACHE_ROOT", ".")) / "logs" / f"vllm_{LOG_TIMESTAMP}.
 
 
 def get_logging_dict(log_path, level="DEBUG"):
+    # TODO: remove this once all vLLM versions have been updated to use the new logging formatter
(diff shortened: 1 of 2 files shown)


tstescoTT (approved):
CI appears to be passing, if any final issues I can fix post-integrating some other incoming changes. Thanks @skhorasganiTT !

tstescoTT:
Tested backwards compatibility works on commits before the vLLM upstream rebase.

PR #317, opened 2025-06-09
quick try

acvejicTT:


acvejicTT:
Hey @teijo I agree it could bring some value.
I ran into a blocker with HF and JWT tokens, so I decided to wait for you to return.
The tokens need to be available, and they can be added as repo secrets by an admin.
Since this is a public repo, I wanted to check with you first, can we proceed with this approach, or do we need to take a different route?

Issue #311, opened 2025-06-05
Add Eval implementation: LiveCodeBench


tstescoTT:
https://github.com/LiveCodeBench/LiveCodeBench

stisiTT:
Daily - Qwen2.5 7B issues with host-memory. Possibly cache-holding/weights issue or lack-of-releasing of memory. @arobergeTT to provide logs so team can support identifying issue and solution.

Able to run with Llama 3.2, but accuracy calculation seems to be broken/wrong.

arobergeTT:
Results for Llama3.2-3B-Instruct model: 

```
|    Tasks    |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|-------------|------:|------|-----:|------|---|-----:|---|-----:|
|livecodebench|      1|none  |     0|acc   |↑  |0.0711|±  |0.0079|
```

```
### Accuracy Evaluations for Llama-3.2-3B-Instruct on n300

| model                 | device | task_name     | accuracy_check | score | ratio_to_reference | gpu_reference_score | ratio_to_published | published_score                         |
|-----------------------|--------|---------------|----------------|-------|--------------------|---------------------|--------------------|-----------------------------------------|
| Llama-3.2-3B-Instruct | n300   | livecodebench | PASS ✅        | 7.11  | 1.42               | [5](TBD)            | 1.32               | [5.40](https://bigcode-bench.github.io) |

```

Reference score is not correct. Needs to be fixed.

arobergeTT:
Things to check:

1. test dataset release date vs. model knowledge cutoff date 
2. Performance depending on test difficulty (Easy, Medium, Hard) 
3. Performance vs GPU (@gtobarTT) 
4. "system prompt" 
5. Check perf using EvalScope and compare.

arobergeTT:
Example log of a Passed test case:

```
🔧 DOC_TO_TARGET DEBUG - Question ID: abc387_f
📝 Found public_test_cases field
📝 Public test cases type: <class 'str'>
📝 Public test cases (first 100 chars): [{"input": "3 3\n2 1 1", "output": "6", "testtype": "stdin"}, {"input": "4 9\n1 1 1 1", "output": "2
✅ Parsed public test cases from JSON string
📝 Number of public test cases: 3
📝 Processing public test case 1: {'input': '3 3\n2 1 1', 'output': '6', 'testtype': 'stdin'}
✅ Added public test case 1
[...]

PR #318, opened 2025-06-09
Qwen3-32B integration

bgoelTT:
This PR adds Qwen3-32B into the model config. It also:
* defines GPU reference scores for accuracy evaluations
  * AIME24
  * Math500
  * GPQA-Diamond
* sets the top performance targets in the benchmark reference


workflows/workflow_venvs.py
```diff
@@ -80,6 +80,18 @@ def setup_evals(
     )
     return True
 
+def setup_evals_livecodebench(
```
ppetrovicTT, line 83: are these evals Qwen3-only or will they be used for all models?

workflows/workflow_venvs.py
```diff
@@ -80,6 +80,18 @@ def setup_evals(
...
+    venv_config: VenvConfig,
+    model_config: "ModelConfig",  # noqa: F821
+    uv_exec: Path,
+) -> bool:
+    logger.warning("this might take 5 to 15+ minutes to install on first run ...")
```
ppetrovicTT, line 88: Is this something that is mandatory for all evaluations?
On CI, every run is the first run.

workflows/model_config.py
```diff
@@ -221,7 +221,7 @@ def _infer_data(self):
             # Note: default to release image, use --dev-mode at runtime to use dev images
             # TODO: Use ubuntu version to interpolate this string
             _default_docker_repo = "ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64"
-            _max_tag_len = 12
+            _max_tag_len = 128
```
ppetrovicTT, line 224: shorter shas are more readable.. is there a reason why we need the full sha?

vllm-tt-metal-llama3/src/run_vllm_api_server.py
```diff
@@ -219,6 +221,9 @@ def runtime_settings(hf_model_id):
             {
                 "meta-llama/Llama-3.1-70B-Instruct": {},
                 "meta-llama/Llama-3.3-70B-Instruct": {},
+                "Qwen/Qwen3-32B": {
+                    "VLLM_ALLOW_LONG_MAX_MODEL_LEN": 1,
```
ppetrovicTT, line 225: if we go for longer seqlens than the model is trained for, won't this have a negative effect on accuracy?

vllm-tt-metal-llama3/src/run_vllm_api_server.py
```diff
@@ -301,7 +306,7 @@ def model_setup(hf_model_id):
         logger.warning(
             "DEPRECATION WARNING: ENABLE_AUTO_TOOL_CHOICE will be removed, use VLLM_OVERRIDE_ARGS env var directly or via --vllm-override-args in run.py CLI"
         )
-        args["enable-auto-tool-choice"] = None
+        args["enable-auto-tool-choice"] = True
```
ppetrovicTT, line 309: should we set the parameter through vllm-override-args instead?

evals/eval_config.py
```diff
@@ -71,6 +71,172 @@ class EvalConfig:
...
+                    "temperature": 0.6,
+                    "top_k": 20,
+                    "top_p": 0.95,
+                },
+                seed=42,
+                num_fewshot=0,
+                batch_size=32,
+                log_samples=True,
+            ),
+            # TODO: uncomment after https://github.com/tenstorrent/tt-inference-server/issues/311 is closed
```
ppetrovicTT, line 197: i prefer that this is kept for another PR

workflows/workflow_venvs.py
bgoelTT: They will be used for other models which require livecodebench in the future. If we ever want to test coding-finetuned models we will use this accuracy evaluation. However, it will just be used by Qwen3-* for now.

workflows/workflow_venvs.py
bgoelTT: Are you referring to the `logger.warning` we emit or the fact that it takes a decent amount of time to install these dependencies into the workflow virtual environment?

For 1: no we can disable the warning if you want, 2: this will only be installed for models that require running LiveCodeBench, which will just be Qwen3 for now.

workflows/model_config.py
bgoelTT: No reason, this was a debug change that I will revert

vllm-tt-metal-llama3/src/run_vllm_api_server.py
bgoelTT: Yes I much prefer this.

Should I just remove this environment variable handling entirely? Or perhaps move this from a `logger.warning` to a `DeprecationError`?

workflows/workflow_venvs.py
ppetrovicTT: i was curious, as i would expect that adding a new evaluation would be a separate PR.
does this also need https://github.com/tenstorrent/tt-inference-server/issues/311 closed first?

workflows/workflow_venvs.py
ppetrovicTT: I was referring to the install time.

Also, it crossed my mind that if this requires longer install time, it should be pushed into docker instead of doing it here?

workflows/workflow_venvs.py
bgoelTT: No, we can add LiveCodeBench as a part of Qwen3-32B's eval_config later. The purpose of this PR is just to get Qwen3 integrated into tt-inference-server to use in Shield CI.

ppetrovicTT (approved):

Issue #127, opened 2025-02-25
Qwen2.5 72B Model Readiness (FKA: eval runs)


tstescoTT:
~~Running on customer lab machine blocked by https://github.com/tenstorrent/vllm/issues/61~~

Published results from Qwen 2.5 technical report: https://arxiv.org/abs/2412.15115

<img width="686" alt="Image" src="https://github.com/user-attachments/assets/00cb3510-62ad-4e01-8bed-3c342808536b" />

- GPQA (5-shot): 49.0
- MATH (4-shot): 83.1
- IFEval (?): 84.1
- MMLU-Pro (5-shot): 71.1

For IFEval, paper only references origianl IFEval paper (Zhou et al., 2023) does not state if reported value is mean of 4 IFEval metrics or one of them. `prompt_level_strict_acc` is closest to actual values.

Third party repro results: https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard#/?search=qwen2.5-7B&official=true

- IFEval (leaderboard_ifeval): 86.38
- GPQA (leaderboard_gpqa, 0-shot): 16.67 -> multiple choice (requires logprobs)
- MMLU_pro (leaderboard_mmlu_pro, 5-shot): 51.40 -> multiple choice (requires logprobs)

tstescoTT:
Can attempt running eval with batch 1 (num_concurrent=1)

tstescoTT:
tt_metal_commit: v0.56.0-rc33
vllm_commit: e2e0002ac7dc
device: T3K


| Model                        | Eval Task | Score |  Published |
|----------------------| ------| ------------------| -------- |
| Qwen2.5-72B-Instruct | leaderboard_ifeval | 83.7 | 84.1 |
| Qwen2.5-72B-Instruct | gpqa_diamond_generative_n_shot | 41.41 | 49.0 |
| Qwen2.5-72B-Instruct | mmlu_pro | 42.2 | 71.1 |

tstescoTT:
Adding math eval support:

expected value 59.82: https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard#/?official=true&search=Qwen2.5-72B-Instruct


tt_metal_commit: v0.56.0-rc33
vllm_commit: e2e0002ac7dc
device: T3K

| Model                        | Eval Task | Score |  Published |
|----------------------| ------| ------------------| -------- |
[...]

PR #320, opened 2025-06-10
Fix TTQwen2ForCausalLM registration error with graceful fallback

stisiTT:
**Problem:**
The TTQwen2ForCausalLM model registration was failing because it referenced a non-existent `Qwen2ForCausalLM` class in the TT-metal module. The actual available class is `QwenForCausalLM`.
**Solution:**

- Added try/catch around `TTQwen2ForCausalLM` registration
- Try to register using the correct TT-specific `QwenForCausalLM` class
- Fall back to standard `transformers:Qwen2ForCausalLM` if TT implementation fails
- Added warning logging for troubleshooting

**Impact:**

- Fixes runtime errors when loading Qwen2.5 models
- Maintains compatibility with both TT-specific and standard implementations
- Improves error handling and debugging visibility

**Testing:**

- Resolves the `AttributeError: module 'models.tt_transformers.tt.generator_vllm' has no attribute 'Qwen2ForCausalLM'` error when running Qwen2.5 models.

### DIFF
--- a/benchmarking/benchmark_config.py
+++ b/benchmarking/benchmark_config.py
@@ -50,6 +50,12 @@ MAX_CONCURRENCY_BENCHMARK_COMMON_ISL_OSL_PAIRS = [
     (16000, 64),
 ]
 
+# Image resolution pairs for multimodal benchmarks
+# Format here is isl, osl, image_height, image_width, images_per_prompt
+ISL_OSL_IMAGE_RESOLUTION_PAIRS = [
+    (128, 128, 512, 512, 1),   # Base resolution
+]
+
 
 def get_num_prompts(input_len, output_len, max_concurrency):
     # Large sequences (slowest) -> fewest prompts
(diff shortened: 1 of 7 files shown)


stisiTT:
Should unblock resolving solutions in #127

vllm-tt-metal-llama3/src/run_vllm_api_server.py
```diff
@@ -59,9 +59,16 @@ def register_tt_models():
...
         path_llama_text = f"{path_ttt_generators}:LlamaForCausalLM"
 
-        ModelRegistry.register_model(
-            "TTQwen2ForCausalLM", f"{path_ttt_generators}:Qwen2ForCausalLM"
-        )
+        try:
+            ModelRegistry.register_model(
+                "TTQwen2ForCausalLM", f"{path_ttt_generators}:QwenForCausalLM"
+            )
+        except (ImportError, AttributeError) as e:
```
bgoelTT, line 66: Do we need to catch both exception types? It looked like we only needed to catch the `AttributeError`

vllm-tt-metal-llama3/src/run_vllm_api_server.py
```diff
@@ -59,9 +59,16 @@ def register_tt_models():
...
+            logger.warning(f"Failed to register TTQwen2ForCausalLM: {e}")
```
bgoelTT, line 67:
```suggestion
            logger.warning(f"Failed to register TTQwen2ForCausalLM: {e}, attempting to register older model signature")
```

bgoelTT (changes requested):
We need to make it clear that we retry the import if it first fails.

Also, the model registry path needs to be updated to point to the correct path

Issue #316, opened 2025-06-09
Llama-3.3-70B benchmark on TG breaches trace region size


ppetrovicTT:
The trace region size is breached during the benchmark of `Llama-3.3-70B` on TG.
https://github.com/tenstorrent/tt-shield/actions/runs/15522977342/job/43699215506#step:9:715

Problematic configuration:
```
Capturing traces for input_seq_len=16000, output_seq_len=64
```

Haven't tested:
```
Capturing traces for input_seq_len=16384, output_seq_len=128
Capturing traces for input_seq_len=32000, output_seq_len=128
```

<details>
<summary>Server log:</summary>

```log
ERROR 06-08 22:58:15 [engine.py:163]   File "/home/container_app_user/vllm/vllm/utils.py", line 2428, in run_method
ERROR 06-08 22:58:15 [engine.py:163]     return func(*args, **kwargs)
ERROR 06-08 22:58:15 [engine.py:163]   File "/home/container_app_user/vllm/vllm/worker/tt_worker.py", line 350, in execute_model
ERROR 06-08 22:58:15 [engine.py:163]     output = self.model_runner.execute_model(
ERROR 06-08 22:58:15 [engine.py:163]   File "/home/container_app_user/tt-metal/python_env/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
ERROR 06-08 22:58:15 [engine.py:163]     return func(*args, **kwargs)
ERROR 06-08 22:58:15 [engine.py:163]   File "/home/container_app_user/vllm/vllm/worker/tt_model_runner.py", line 418, in execute_model
ERROR 06-08 22:58:15 [engine.py:163]     next_token_ids = self._execute_model_single_step(
ERROR 06-08 22:58:15 [engine.py:163]   File "/home/container_app_user/vllm/vllm/worker/tt_model_runner.py", line 548, in _execute_model_single_step
ERROR 06-08 22:58:15 [engine.py:163]     outputs = self.model.prefill_forward(**execute_model_kwargs)
ERROR 06-08 22:58:15 [engine.py:163]   File "/home/container_app_user/tt-metal/models/demos/llama3_subdevices/tt/generator_vllm.py", line 121, in prefill_forward
ERROR 06-08 22:58:15 [engine.py:163]     return super().prefill_forward_text(*args, **kwargs)
[...]

PR #321, opened 2025-06-10
Extend trace region size for Llama TG

ppetrovicTT:
### Related issue
https://github.com/tenstorrent/tt-inference-server/issues/316

### DIFF
--- a/workflows/model_config.py
+++ b/workflows/model_config.py
@@ -446,7 +446,7 @@ config_templates = [
                     "sample_on_device_mode": "all",
                     "fabric_config": "FABRIC_1D",
                     "worker_l1_size": 1344544,
-                    "trace_region_size": 95693824,
+                    "trace_region_size": 103772160,
                 },
             ),
         },


ssanjayTT (approved):

workflows/model_config.py
```diff
@@ -446,7 +446,7 @@ def expand_to_configs(self) -> List["ModelConfig"]:
                     "sample_on_device_mode": "all",
                     "fabric_config": "FABRIC_1D",
                     "worker_l1_size": 1344544,
-                    "trace_region_size": 95693824,
+                    "trace_region_size": 103772160,
```
bgoelTT, line 449: I see that this number is coming from the server log error here https://github.com/tenstorrent/tt-inference-server/issues/316

Does increasing the trace region allow us to capture traces for:
```
Capturing traces for input_seq_len=16384, output_seq_len=128
Capturing traces for input_seq_len=32000, output_seq_len=128
```

ppetrovicTT: I asked the models team today how can we calculate this size in advance and I was told to _try it_.
So this was my fist try :)

But I'll run some Shield CI tests first and see if longer seqlens require higher numbers and then use that one instead!

bgoelTT: Ok, do you need this to be merged to `dev` to do your tests?

ppetrovicTT:
We don't need this anymore.

After https://github.com/tenstorrent/tt-metal/commit/309e2b3785d3ce4d6686804536939cd789ea5a46 we don't trace anymore seqlens longer than 8k, as the performance is not impacted for such long seqlens. Therefore, the trace region size doesn't grow over 95000000.

We can even reduce it, if you wish? Or keep it as is.

PR #334, opened 2025-06-20
Enable configuration of vLLM args in model_config

bgoelTT:
This PR:
* enables configuration of vLLM args in model_config
* sets `num_scheduler_steps=1` for TG on Llama-3.3-70B(-Instruct)
* uses a new tt-metal commit which fixes the 32k OOM issue (https://github.com/tenstorrent/tt-metal/pull/23786)

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -115,6 +115,7 @@ def run_docker_server(args, setup_config):
         "VLLM_MAX_NUM_SEQS": model_config.device_model_spec.max_concurrency,
         "VLLM_MAX_MODEL_LEN": model_config.device_model_spec.max_context,
         "VLLM_MAX_NUM_BATCHED_TOKENS": model_config.device_model_spec.max_context,
+        "TT_METAL_ENABLE_ERISC_IRAM": 1,
     }
 
     # Pass model config override_tt_config if it exists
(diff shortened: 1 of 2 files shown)


workflows/run_docker_server.py
```diff
@@ -115,6 +115,7 @@ def run_docker_server(args, setup_config):
         "VLLM_MAX_NUM_SEQS": model_config.device_model_spec.max_concurrency,
         "VLLM_MAX_MODEL_LEN": model_config.device_model_spec.max_context,
         "VLLM_MAX_NUM_BATCHED_TOKENS": model_config.device_model_spec.max_context,
+        "TT_METAL_ENABLE_ERISC_IRAM": 1,
```
ppetrovicTT, line 118: Perhaps `model_config.device_type == DeviceTypes.Galaxy` instead of `1`?

ppetrovicTT (changes requested):
Don't we need a change in `run_vllm_api_server.py` as well?

`override_tt_config` is passed all the way to vLLM, so it's fine to have it defined as json.
but `VLLM_NUM_SCHEDULER_STEPS` is an env var, and i don't see it matching?

workflows/run_docker_server.py
bgoelTT: I removed this change because I'm waiting on confirmation if this is required

ppetrovicTT (approved):

PR #345, opened 2025-06-25
docker image needs triton to run normally

ryanzhang-hlit:


tstescoTT:
Hi @scorbunny-cinderace, thanks for contributing to tt-inference-server!

I'm not familiar with how you're running into this issue requiring triton as a python dependency. Could you please open an issue to go through your setup and where you're hitting the error?

As first step, if you're not already I'd recommend using the Docker images (you may need to build them) and working off of the `dev` branch if you're not finding what you need on `main`.

scorbunny-cinderace:
I'm running Qwen2.5-7B on n300 with logs below and it seems to be fine.



But the docker log shows the it needs triton and the outputs is stopped. It seems the container running iss stucked.



So I installed Triton and it ran fine after the installation.

tstescoTT:
Okay, glad it worked! The INFO message for triton is from vLLM I believe, and you can safely disregard it as we're not using GPUs and don't need the triton dependency.

scorbunny-cinderace:
You’re right. It worked by removeing triton. I've been misled by the terrible internet environment here.

scorbunny-cinderace:
Thank you for your support. I'm closing this PR.

PR #355, opened 2025-06-30
Change default max concurrency to include data-parallel

rdraskicTT:
Each data-parallel instance can have a batch of up to 32 prompts. We need to change the default to enable bigger batches for data-parallel configurations.
Switching to DP=4 since it's currently in the best shape.

- [x] [shield ci run](https://github.com/tenstorrent/tt-shield/actions/runs/16147158518)

### DIFF
--- a/workflows/model_config.py
+++ b/workflows/model_config.py
@@ -150,7 +150,12 @@ class DeviceModelSpec:
         # need to use __setattr__ because instance is frozen
         # Set default concurrency and context if not provided
         if not self.max_concurrency:
-            _default_max_concurrent = 32
+            data_parallel = (
+                self.override_tt_config["data_parallel"]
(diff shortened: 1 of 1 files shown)


tstescoTT:
Thanks for raising this, @rdraskicTT. Is the scale purely linear in num concurrent? e.g. if I'm understanding this correctly, for DP=16 it would be 16*32 = 512 concurrent users for Llama 3.1 8B.

I prefer the strategy you just added of changing the specific model-device config so we can label the default and if scaling is not linear we can set it per model-device.

rdraskicTT:
@tstescoTT My hope is that it will be close to linear, I'm investigating it now. I still have some other perf issues that I'm investigating as well so we probably won't see the best performance even after adding this.

tstescoTT (approved):

PR #363, opened 2025-07-01
Acvejic/generate empty report

acvejicTT:
Add model and device to report even when it fails.
We would like to force creation of report even when one or all steps fail.
1) It will pick up benchmarks if evals fail and vice-versa.
2) If we have model and device we can more easily connect the data in Superset and more easily visualize it.


workflows/run_workflows.py
```diff
@@ -209,5 +209,7 @@ def run_workflows(args):
         return return_codes
     else:
         return_codes.append(run_single_workflow(args))
+        args.workflow = WorkflowType.REPORTS.name
+        return_codes.append(run_single_workflow(args))
```
tstescoTT, line 213: For --reports workflow this will run twice. We may get rid of the reports workflow but I've found it useful in the past having it decoupled.

acvejicTT: @tstescoTT @ppetrovicTT 
how about this:
```
    else:
        return_codes.append(run_single_workflow(args))
        if WorkflowType.from_string(args.workflow) != WorkflowType.REPORTS:
            args.workflow = WorkflowType.REPORTS.name
            return_codes.append(run_single_workflow(args))
```

tstescoTT: Ya that should work, thanks!

PR #367, opened 2025-07-03
P100 benchmark targets

acvejicTT:
Add targets in order to generate report that could be parsed and pushed to superset


acvejicTT:
@tstescoTT I've put the values same as for n150, makes sense?

PR #369, opened 2025-07-03
Update Llama-3.2-11B model commits

ssanjayTT:
Just updating the tt-metal and vllm commits for better perf on the vision benchmarks.


benchmarking/benchmark_config.py
```diff
@@ -153,8 +153,8 @@ def get_num_prompts(input_len, output_len, max_concurrency):
                     BenchmarkTaskParams(
                         isl=isl,
                         osl=osl,
-                        max_concurrency=_max_concurrency // 2,
```
tstescoTT, line 156: I'm wondering why this was set to // 2 previously. Any thoughts?

ssanjayTT: I had set it to // 2 because Llama-3.2-11B used to hang on batch size 16.  So I wanted to test with batch-size 8 instead.

tstescoTT (approved):

PR #386, opened 2025-07-14
Add ability to pass multiple device-id arguments as comma-separated list

bgoelTT:
To support running multiple `--workflow server --docker-server`'s, we need to be able to map more than just one `--device-id` into the docker container.

This PR:
* adds this argument parsing to `run.py`
* adds unit tests
  * and fixes broken ones
* adds device string creation to be passed down to `docker run ...`

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -87,9 +87,17 @@ def run_docker_server(args, setup_config):
         else mesh_device_str
     )
 
+    # create device mapping string to pass to docker run
     device_path = "/dev/tenstorrent"
-    if hasattr(args, "device_id") and args.device_id is not None:
-        device_path = f"{device_path}/{args.device_id}"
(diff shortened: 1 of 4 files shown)


workflows/run_docker_server.py
```diff
@@ -87,9 +87,17 @@ def run_docker_server(args, setup_config):
         else mesh_device_str
     )
 
+    # create device mapping string to pass to docker run
     device_path = "/dev/tenstorrent"
-    if hasattr(args, "device_id") and args.device_id is not None:
-        device_path = f"{device_path}/{args.device_id}"
+    device_map_strs = (
+        ["--device", f"{device_path}:{device_path}"]
+        if not getattr(args, "device_id", None)
```
tstescoTT, line 94: nit: conditional inside of definition is hard to follow with list comprehension, prefer it being a classic if / else block.

PR #395, opened 2025-07-19
add --reset-venvs

tstescoTT:
# change log

* add `--reset-venvs` CLI arg to handle blowing away uv installed .workflows_venvs which may get into dependency issues if changes occur and previous versions are on the system installed. This solves many upgrading issues simply.

### DIFF
--- a/run.py
+++ b/run.py
@@ -9,6 +9,7 @@ import argparse
 import getpass
 import logging
 import subprocess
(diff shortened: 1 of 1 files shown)


bgoelTT (approved):
This is really useful, I find myself manually `rm -rf .workflow_venvs` a decent amount.

PR #402, opened 2025-07-22
Add sampling on device to llama 8b config

rdraskicTT:
Issue https://github.com/tenstorrent/tt-metal/issues/24326
Once https://github.com/tenstorrent/tt-metal/pull/25510 gets merged we can activate this for perf improvement

### DIFF
--- a/workflows/model_config.py
+++ b/workflows/model_config.py
@@ -711,6 +711,7 @@ config_templates = [
                 default_impl=True,
                 override_tt_config={
                     "data_parallel": 4,
+                    "sample_on_device_mode": "decode_only",
                 },
             ),
         },


workflows/model_config.py
```diff
@@ -711,6 +711,7 @@ def expand_to_configs(self) -> List["ModelConfig"]:
                 default_impl=True,
                 override_tt_config={
                     "data_parallel": 4,
+                    "sample_on_device_mode": "decode_only",
```
ppetrovicTT, line 714: why not "all" like llama70b? (line 490 in this file)

rdraskicTT: `tt_transformers` prefill doesn't accept `sampling_params`. I'm not even sure what does this option do in prefill for llama70b, looks like it's just not being used.

ppetrovicTT: ok, then maybe it can be added later.
i remember this was a big deal once it was activated on llama70b.
@skhorasganiTT @sraizada-tt, your input is appreciated as always!

ppetrovicTT (approved):

workflows/model_config.py
ppetrovicTT: one more question, should we activate this by default on all models?

skhorasganiTT: Yeah currently TT-Transformers doesn't have sampling on device for prefill. Also for decode it currently only supports argmax/greedy-decoding (temperature = 0), so you should be careful when turning this on to not pass in requests with top-p/top-k sampling params (it will assert out in the generator if it receives temp != 0 and crash the server). Once TT-Transformers supports greedy + top-pk on device (ideally for both prefill and decode), similar to llama70b-6U, then we can turn it on by default for all models.

Issue #416, opened 2025-07-26
Broken Llama-70b model configs due to missing repack weights script


bgoelTT:
With https://github.com/tenstorrent/tt-metal/pull/25373 landing, this is raised when attempting to run the Llama-70B family on `main`:
`2025-07-26 01:01:01,409 - setup_host.py:467 - ERROR: ⛔ Failed to download repack_weights.py`

We need to either make a new release or create a quick patch to main

PR #417, opened 2025-07-26
Pin repack URL to specific commit

bgoelTT:
Hotfix for #416


tstescoTT (approved):
Thanks Ben, can you please add the same change to `dev` as well?

PR #422, opened 2025-07-29
SD 3.5 Large support & mock mode & threads instead of queues

idjuricTT:
Changes

* SD3.5 large basic support
* Mock mode for device runner
* Using sub mashes to max possible size
* Threads instead of processes (due to sharing of the objects)


idjuricTT:
Since it's only minor changes and doesn't affect rest of the tt-inference-server merging

Issue #429, opened 2025-08-05
ValueError: mutable default for field device_id when using --device-id argument


acvejicTT:
When running run.py with device-id argument we are getting the error
```
ValueError: mutable default <class 'list'> for field device_id is not allowed: use default_factory
```

#### Steps to Reproduce:
Run any command with the --device-id argument:

```
python3 run.py --model Llama-3.2-1B-Instruct --device n150 --workflow release --docker-server --device-id 0
```

Full traceback:
```
Traceback (most recent call last):
  File "run.py", line 413, in <module>
    sys.exit(main())
  File "run.py", line 333, in main
    model_spec = get_runtime_model_spec(args)
  File "/home/ubuntu/actions-runner/_work/tt-shield/tt-shield/tt-inference-server/workflows/model_spec.py", line 1115, in get_runtime_model_spec
    model_spec.apply_runtime_args(args)
  File "/home/ubuntu/actions-runner/_work/tt-shield/tt-shield/tt-inference-server/workflows/model_spec.py", line 542, in apply_runtime_args
    CliArgsClass = make_dataclass("cli_args", fields)
  File "/usr/lib/python3.8/dataclasses.py", line 1230, in make_dataclass
    return dataclass(cls, init=init, repr=repr, eq=eq, order=order,
  File "/usr/lib/python3.8/dataclasses.py", line 1019, in dataclass
    return wrap(cls)
  File "/usr/lib/python3.8/dataclasses.py", line 1011, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
  File "/usr/lib/python3.8/dataclasses.py", line 861, in _process_class
    cls_fields = [_get_field(cls, name, type)
  File "/usr/lib/python3.8/dataclasses.py", line 861, in <listcomp>
    cls_fields = [_get_field(cls, name, type)
  File "/usr/lib/python3.8/dataclasses.py", line 745, in _get_field
    raise ValueError(f'mutable default {type(f.default)} for field '
ValueError: mutable default <class 'list'> for field device_id is not allowed: use default_factory
```

bgoelTT:
This has been address in #434

PR #434, opened 2025-08-05
Allow parsing of dict and list from CLI for workflow args

bgoelTT:
This PR creates a fix for #429 and restores the functionality for passing the `--device-id` workflow argument


bgoelTT:
To prevent this from happening in the future, I will try to address #435 ASAP

acvejicTT (approved):
LGTM

PR #438, opened 2025-08-05
Fix Docker container permissions error by moving JSON config outside CACHE_ROOT

stisiTT:
## Problem

Docker server workflow fails during startup with:

```
chmod: changing permissions of '/home/container_app_user/cache_root/tt_model_spec_.json': Read-only file system
```

The model spec JSON file was mounted read-only inside `CACHE_ROOT`, but `docker-entrypoint.sh` runs `chmod -R 2775` on the entire `CACHE_ROOT` directory, which fails on read-only files.

## Solution

Mount the JSON file to `/home/container_app_user/model_spec/` instead of inside `CACHE_ROOT` to avoid the recursive chmod operation. Added the model spec directory path to `SetupConfig` for better maintainability.

## Changes

**`workflows/setup_host.py`**:
- Added `container_model_spec_dir: Path = containter_user_home / "model_spec"` to SetupConfig class

**`workflows/run_docker_server.py`**:
- Changed `docker_json_fpath` to use `setup_config.container_model_spec_dir / json_fpath.name`

**`tests/test_run_arguments.py`**:
- Updated mock_setup_config fixture to include container_model_spec_dir

## Testing

- Existing test `test_docker_server_mounts_model_spec_json` passes
- Verified JSON file is mounted outside CACHE_ROOT at `/home/container_app_user/model_spec/`
- No breaking changes

**Type**: Bug Fix  
**Priority**: High

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -97,7 +97,7 @@ def run_docker_server(model_spec, setup_config, json_fpath):
         model_spec.docker_image
     ), f"Docker image: {model_spec.docker_image} not found on GHCR or locally."
 
-    docker_json_fpath = setup_config.container_config_dir / json_fpath.name
+    docker_json_fpath = setup_config.container_model_spec_dir / json_fpath.name
     # CACHE_ROOT needed for the docker container entrypoint
     # TT_CACHE_PATH has host path
     # TT_MODEL_SPEC_JSON_PATH has dynamic path
--- a/tests/test_run_arguments.py
+++ b/tests/test_run_arguments.py
@@ -92,7 +92,7 @@ def mock_setup_config():
     """Mock setup configuration for docker server."""
     mock_config = MagicMock()
     mock_config.cache_root = Path("/tmp/cache")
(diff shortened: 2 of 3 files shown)


workflows/run_docker_server.py
```diff
@@ -97,7 +98,7 @@ def run_docker_server(model_spec, setup_config, json_fpath):
         model_spec.docker_image
     ), f"Docker image: {model_spec.docker_image} not found on GHCR or locally."
 
-    docker_json_fpath = setup_config.cache_root / json_fpath.name
+    docker_json_fpath = Path("/home/container_app_user/config") / json_fpath.name
```
bgoelTT, line 101: Can we add this path to the `SetupConfig` in `workflows/setup_host.py`?

bgoelTT (changes requested):

workflows/run_docker_server.py
bgoelTT: Actually, can the directory be called `model_specification`? We want to move away from `config`

stisiTT: I believe this is now resolved. Thank you.

Issue #445, opened 2025-08-06
Qwen2.5 72B Buffer Crash


stisiTT:
When running benchmarking or evaluations on a Qwen2.5 model, the vLLM server crashes with the following error
```
2025-08-06 02:12:02.957 | critical |          Always | Creating trace buffers of size 25914368B on MeshDevice 0, but only 25000000B is allocated for trace region. (assert.hpp:107)
```

PR #446, opened 2025-08-06
Fix: Increase trace_region_size for Qwen2.5 models to prevent vLLM crashes

stisiTT:
## Problem

When running benchmarking or evaluations on Qwen2.5 models, the vLLM server crashes with the following error:
```
2025-08-06 02:12:02.957 | critical | Always | Creating trace buffers of size 25914368B on MeshDevice 0, but only 25000000B is allocated for trace region. (assert.hpp:107)
```

**Root Cause**: 
- vLLM sets a default `trace_region_size` of 25MB (25,000,000 bytes) for trace buffer allocation
- Qwen2.5 models require ~25.9MB for trace buffers, exceeding the default limit
- This causes a fatal assertion failure during device initialization

## Solution

Added `trace_region_size: 30000000` (30MB) to the `override_tt_config` for all Qwen2.5 model specifications in `workflows/model_spec.py`.

### Models Updated

1. **Qwen2.5-72B with llama3_impl** (T3K device)
2. **Qwen2.5-72B with tt_transformers_impl** (T3K device)  
3. **Qwen2.5-7B with tt_transformers_impl** (N300 device)
4. **Qwen2.5-7B with tt_transformers_impl** (T3K device)

### Technical Details

- **Previous**: Used vLLM default of 25MB trace region size
- **Updated**: Explicitly set 30MB trace region size via `override_tt_config`
- **Scope**: Changes only affect Qwen2.5 models experiencing the issue

## Related Issues
Issue #445

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -754,10 +754,7 @@ spec_templates = [
                 device=DeviceTypes.T3K,
                 max_concurrency=32,
                 max_context=128 * 1024,
-                default_impl=True,
-                override_tt_config={
-                    "trace_region_size": 30000000,
-                },
+                default_impl=False,
             ),
         ],
         status=ModelStatusTypes.EXPERIMENTAL,
@@ -777,7 +774,7 @@ spec_templates = [
                 max_context=128 * 1024,
                 default_impl=True,
                 override_tt_config={
-                    "trace_region_size": 30000000,
+                    "trace_region_size": 26000000,
                 },
             ),
         ],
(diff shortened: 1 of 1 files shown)


workflows/model_spec.py
```diff
@@ -755,6 +755,9 @@ def expand_to_specs(self) -> List["ModelSpec"]:
                 max_concurrency=32,
                 max_context=128 * 1024,
                 default_impl=True,
```
bgoelTT, line 757: Actually, we should only have one default implementation for Qwen2.5-72B-* on T3K, so can we make this False
```suggestion
                default_impl=False,
```

bgoelTT (changes requested):

bgoelTT (approved):

PR #449, opened 2025-08-06
Fix: Add MAX_PREFILL_CHUNK_SIZE env var to Qwen2.5 72B

stisiTT:
Also, Update model specifications for Qwen2.5: bump tt_metal and vllm commit

Running fix and release workflow [here](https://github.com/tenstorrent/tt-shield/actions/runs/16790982178)

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -760,7 +760,6 @@ spec_templates = [
         status=ModelStatusTypes.EXPERIMENTAL,
         env_vars={
             "VLLM_ALLOW_LONG_MAX_MODEL_LEN": 1,
-            "MAX_PREFILL_CHUNK_SIZE": "16",
         },
     ),
     ModelSpecTemplate(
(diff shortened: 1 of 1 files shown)


workflows/model_spec.py
```diff
@@ -747,8 +747,8 @@ def expand_to_specs(self) -> List["ModelSpec"]:
     ModelSpecTemplate(
         weights=["Qwen/Qwen2.5-72B", "Qwen/Qwen2.5-72B-Instruct"],
         impl=llama3_impl,
-        tt_metal_commit="v0.56.0-rc33",
-        vllm_commit="e2e0002ac7dc",
+        tt_metal_commit="v0.62.0-rc8",
+        vllm_commit="c348d08",
```
bgoelTT, line 751: We should not update this model spec as this is for the legacy implementation.

bgoelTT (changes requested):
Thanks Sam, this must have been mangled during the model spec rebase

bgoelTT (approved):

Issue #468, opened 2025-08-08
Add recovery for workers if they're dead


idjuricTT:
In case of worker unintentionally dying, add recovery mechanism

PR #469, opened 2025-08-08
Recover dead workers

idjuricTT:
Recover dead workers and restart them.

Related item #468

### DIFF
--- a/tt-metal-sdxl/model_services/scheduler.py
+++ b/tt-metal-sdxl/model_services/scheduler.py
@@ -173,6 +173,8 @@ class Scheduler:
                 
                 self.worker_info[worker_id]['error_count'] += 1
                 
+                self.logger.warning(f"Worker {worker_id} error cound is : {self.worker_info[worker_id]['error_count']}")
+                
                 if task_id is None:
                     self.listener_running = False
                     break
--- a/tt-metal-sdxl/model_services/scheduler.py
+++ b/tt-metal-sdxl/model_services/scheduler.py
@@ -5,6 +5,7 @@
 import asyncio
 from multiprocessing import Process, Queue as Queue  # Need multiprocessing queues
 from threading import Lock
+import time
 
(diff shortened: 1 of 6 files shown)


tt-metal-sdxl/model_services/scheduler.py
```diff
@@ -90,14 +93,57 @@ def startWorkers(self):
...
+        
+        # Clean up old process if it exists
+        if worker_id in self.worker_info:
+            old_process = self.worker_info[worker_id]['process']
+            if old_process.is_alive():
+                old_process.terminate()
+                old_process.join(timeout=5.0)
+        
+        # Start new worker
+        self._start_worker(worker_id)
```
ztorlakTT, line 140: What do you think about passing restart_count as a parameter to the _start_worker function, instead of setting it manually in the following line? That way, the function signature could look like: 
def _start_worker(self, worker_id: int, restart_count: int = 0) 
and you could simply pass restart_count directly to worker_info[worker_id].

tt-metal-sdxl/model_services/scheduler.py
```diff
@@ -272,4 +323,35 @@ def _get_max_queue_size(self, settings: Settings) -> int:
...
+                    process = info['process']
+                    if not process.is_alive():
+                        dead_workers.append(worker_id)
+                
+                # Restart dead workers
+                for worker_id in dead_workers:
+                    restart_count = self.worker_info[worker_id].get('restart_count', 0)
+                    
+                    # Optional: Limit restart attempts
+                    if restart_count < 5:  # Max 5 restarts per worker
```
ztorlakTT, line 344: I'd suggest moving this check into the _restart_worker function. Since it's a private method used only within this class, it could handle the logic of whether the worker should be restarted. This would make the code safer and cleaner. As a bonus, you'd avoid extracting restart_count twice when restarting.

PR #488, opened 2025-08-13
Add model and device to image perf result

acvejicTT:


acvejicTT:
Tested - https://github.com/tenstorrent/tt-shield/actions/runs/16910071646/job/47938983723

Issue #470, opened 2025-08-08
--service_port argument is not being forwarded to containers


bgoelTT:
Despite passing `--service_port 8001` to `run.py`, the actual vLLM server is still being hosted on the default 8000. The docker container is correctly performing the port bindings so this is a simple fix

PR #500, opened 2025-08-14
Add service port to vLLM args

bgoelTT:
This PR addresses #470 

By simply adding the service port passed by the user at runtime to the vLLM args in the device model spec, we can control the port that the vLLM server is hosted on


tstescoTT (approved):
LGTM

PR #501, opened 2025-08-14
Acvejic/fix issue pyluwen

acvejicTT:
Fix problems that came with https://github.com/tenstorrent/tt-smi/commit/2ca67178be6ea56cf33b7312856d7530b6d7f01f

### DIFF
--- a/tt-metal-sdxl/Dockerfile
+++ b/tt-metal-sdxl/Dockerfile
@@ -78,10 +78,6 @@ RUN apt-get update && apt-get install -y \
     protobuf-compiler \
     libprotobuf-dev \
     pkg-config \
-    # Rust dependencies
-    rustc \
-    cargo \
-    # user deps
     htop \
     screen \
     tmux \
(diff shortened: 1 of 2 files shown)


acvejicTT:
Test run https://github.com/tenstorrent/tt-shield/actions/runs/16969782122/job/48103340272

tstescoTT:
I would expect the rust install to automatically create these directories. Is there a permissions issue with running rust install from non root user? or is the rust install broken?

tstescoTT (approved):
Thanks for adding these changes and testing the build. We will need tt-smi and other systems tools available in the container for some new features.

I would expect that the rust install would be possible from non-sudo user, but that can be solved at a future time. Adding https://github.com/tenstorrent/tt-inference-server/issues/503 to track future work to clean up rust install, likely after changes to how pyluwen is packaged and installed.

tstescoTT:
Looking into this a bit further ... does https://github.com/tenstorrent/tt-metal/pull/26756 fix this?

acvejicTT:
Could be, same env variables. I've already meged this, no harm, but we can test it without it.

PR #505, opened 2025-08-14
uplift Llama-3.2-1B tt-metal and vllm versions

tstescoTT:


tstescoTT:
Verified local install and run works with these commits as well.

PR #556, opened 2025-08-26
Use SDXL class for inference

idjuricTT:


tt-metal-sdxl/tt_model_runners/sdxl_runner_trace.py
```diff
@@ -106,54 +88,31 @@ async def load_model(self, device)->bool:
...
-                    )
-            except Exception as e:
-                self.logger.error(f"Error in ttnn.distribute block: {e}")
-                raise
+            self.tt_sdxl = TtSDXLPipeline(
+                ttnn_device=self.ttnn_device,
+                torch_pipeline=self.pipeline,
+                pipeline_config=TtSDXLPipelineConfig(
+                    capture_trace=True,
+                    vae_on_device="device_vae",
```
ipotkonjak-tt, line 109: This is expected to be a bool value. True if we are using device VAE else False.

tt-metal-sdxl/tt_model_runners/sdxl_runner_trace.py
```diff
@@ -106,54 +88,31 @@ async def load_model(self, device)->bool:
...
-                        self.pipeline.scheduler.config.rescale_betas_zero_snr,
-                        self.pipeline.scheduler.config.final_sigmas_type,
-                    )
-            except Exception as e:
-                self.logger.error(f"Error in ttnn.distribute block: {e}")
-                raise
+            self.tt_sdxl = TtSDXLPipeline(
+                ttnn_device=self.ttnn_device,
+                torch_pipeline=self.pipeline,
+                pipeline_config=TtSDXLPipelineConfig(
```
ipotkonjak-tt, line 107: Jfyi I've added default values for `capture_trace`, `vae_on_device` and `encoders_on_device`. They are all True, so inference server doesn't need to worry about those if you want trace and encoders and vae on device. Which looks like your use case from the code.

PR #559, opened 2025-08-27
Reduce context length for llama 8B on n150

rdraskicTT:
https://github.com/tenstorrent/tt-shield/issues/109

To unblock model CI, OOM issue still tracked here https://github.com/tenstorrent/tt-metal/issues/23854

- [x] [llama 8B on n150](https://github.com/tenstorrent/tt-shield/actions/runs/17409316531)

### DIFF
--- a/benchmarking/benchmark_config.py
+++ b/benchmarking/benchmark_config.py
@@ -144,7 +144,8 @@ else:
                             num_prompts=get_num_prompts(isl, osl, 1),
                         )
                         for isl, osl in BATCH_1_BENCHMARK_COMMON_ISL_OSL_PAIRS
-                        if (isl, osl, 1) not in perf_ref_task_runs.get(_device, [])
+                        if (isl, osl, 1) not in perf_ref_task_runs.get(_device, []) 
+                        and isl <= model_spec.device_model_spec.max_context
                     ]
                     + [
                         BenchmarkTaskParams(
@@ -156,6 +157,7 @@ else:
                         for isl, osl in MAX_CONCURRENCY_BENCHMARK_COMMON_ISL_OSL_PAIRS
                         if (isl, osl, _max_concurrency)
                         not in perf_ref_task_runs.get(_device, [])
+                        and isl <= model_spec.device_model_spec.max_context
                     ]
                     + 
                     (
--- a/benchmarking/benchmark_config.py
+++ b/benchmarking/benchmark_config.py
@@ -145,7 +145,7 @@ else:
                         )
                         for isl, osl in BATCH_1_BENCHMARK_COMMON_ISL_OSL_PAIRS
                         if (isl, osl, 1) not in perf_ref_task_runs.get(_device, []) 
-                        and isl <= model_spec.device_model_spec.max_context
+                        and (isl + osl) <= model_spec.device_model_spec.max_context
                     ]
                     + [
                         BenchmarkTaskParams(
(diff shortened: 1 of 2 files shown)


bgoelTT:
In my opinion, we shouldn't sacrifice our testing to accomodate a broken model. It looks like a fix is on the way https://github.com/tenstorrent/tt-metal/issues/23854#issuecomment-3245384170.

@rdraskicTT since Models CI will error out and stop the workload, what is the motivation for this change?

bgoelTT:
That makes sense but we need to be very careful that we don't get false positives from Models CI. If we reduce the context length to 8k and everything passes then we'll get a green checkmark when we really shouldn't.

We technically do perform a 128k context len test when we perform max concurrency sweeps, we will send `max_concurrency` requests, each with 16k input seq len. This specific configuration (and the smaller ones which get run before it) will max out the full context length. If you're asking about single-request context length then yes we have a gap here.

rdraskicTT:
> That makes sense but we need to be very careful that we don't get false positives from Models CI. If we reduce the context length to 8k and everything passes then we'll get a green checkmark when we really shouldn't.

You are right, but it's on model owners to track the CI health and update these specs accordingly once improvements are added. I think this is a step in a good direction, when CI is red for a long time, nobody knows if something got worse or if it's still failing the same, this way we at least have a baseline.

> We technically do perform a 128k context len test when we perform max concurrency sweeps, we will send `max_concurrency` requests, each with 16k input seq len. This specific configuration (and the smaller ones which get run before it) will max out the full context length. If you're asking about single-request context length then yes we have a gap here.

Ah ok, thanks for the explanation.

workflows/model_spec.py
```diff
@@ -1069,7 +1069,7 @@ def expand_to_specs(self) -> List["ModelSpec"]:
             DeviceModelSpec(
                 device=DeviceTypes.N150,
                 max_concurrency=32,
-                max_context=64 * 1024,
+                max_context=8 * 1024,
```
bgoelTT:
```suggestion
                # revert back to 64 * 1024 once https://github.com/tenstorrent/tt-metal/issues/23854#issuecomment-3245384170 is resolved
                max_context=8 * 1024,
```

benchmarking/benchmark_config.py
```diff
@@ -156,6 +157,7 @@ def get_num_prompts(input_len, output_len, max_concurrency):
                         for isl, osl in MAX_CONCURRENCY_BENCHMARK_COMMON_ISL_OSL_PAIRS
                         if (isl, osl, _max_concurrency)
                         not in perf_ref_task_runs.get(_device, [])
+                        and isl <= model_spec.device_model_spec.max_context
```
bgoelTT, line 160: Same as above

benchmarking/benchmark_config.py
```diff
@@ -144,7 +144,8 @@ def get_num_prompts(input_len, output_len, max_concurrency):
                             num_prompts=get_num_prompts(isl, osl, 1),
                         )
                         for isl, osl in BATCH_1_BENCHMARK_COMMON_ISL_OSL_PAIRS
-                        if (isl, osl, 1) not in perf_ref_task_runs.get(_device, [])
+                        if (isl, osl, 1) not in perf_ref_task_runs.get(_device, []) 
+                        and isl <= model_spec.device_model_spec.max_context
```
bgoelTT, line 148:
```suggestion
                        and (isl + osl) <= model_spec.device_model_spec.max_context
```
It should be the total `isl + osl` value that is compared to the max_context

bgoelTT (changes requested):
Looks good, just a few changes to help us in the future

bgoelTT (approved):
Thank you for making these changes!

rdraskicTT:
@bgoelTT Thanks for staying with me on this one! I just have another question, I haven't figured out if I need to change something in evals, how big are these inputs?

PR #568, opened 2025-08-28
Add matmul throttle for Llama 8B to resolve 6U hang

rdraskicTT:
Performance is not significantly affected while hang is resolved.
- [x] [llama 8B DP=4](https://github.com/tenstorrent/tt-shield/actions/runs/17307250689)

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -1148,6 +1148,9 @@ spec_templates = [
                     "data_parallel": 4,
                     "sample_on_device_mode": "decode_only",
                 },
+                env_vars={
+                    "TT_MM_THROTTLE_PERF": 5,
+                },
             ),
         ],
         status=ModelStatusTypes.FUNCTIONAL,


bgoelTT:
@rdraskicTT I've fiddled around with this variable before and have a few questions:
* do we need the most aggressive throttling? I.e, can we get away with level 1 or 2?
* this only affects WH Galaxy right? Not Llama-3.1-8B on n150/n300?

rdraskicTT:
> @rdraskicTT I've fiddled around with this variable before and have a few questions:
> 
> * do we need the most aggressive throttling? I.e, can we get away with level 1 or 2?

I'm not sure, I quickly tested all levels and I saw it working even with 1 but I didn't perform any stress tests so I didn't want to risk. I only saw a minor perf drop so this seemed the best way to unblock this hang.

> * this only affects WH Galaxy right? Not Llama-3.1-8B on n150/n300?

Right, for some reason 6U galaxy has this problem which is why this wasn't noticed before.

PR #569, opened 2025-08-29
Sadaesoye/update sd35

sadesoyeTT:
- Update SD35 to use latest code.


tt-metal-sdxl/tests/test_sd35_device_runner.py
sadesoyeTT, line 1: @idjuricTT , how do you run tests? I keep getting the following module error while trying to run this test as well as `test_runner_fabric.py` and` test_sd35_device_runner.py` .

PR #586, opened 2025-09-01
Whisper locks

idjuricTT:


tt-metal-sdxl/tests/conftest.py
ztorlakTT, line 1: Why have you deleted this file? It was used for tests

tt-metal-sdxl/tt_model_runners/whisper_runner.py
```diff
@@ -73,15 +74,17 @@ def __init__(self, device_id: str):
         # Limit threading for stability during inference
         os.environ['OMP_NUM_THREADS'] = '1'
         os.environ['MKL_NUM_THREADS'] = '1'
-                        
+        self.ttnn_lock = multiprocessing.RLock()
```
ztorlakTT, line 77: Should we add a lock in the other runners as well? In that case, we can move this to the base runner

tt-metal-sdxl/tests/conftest.py
idjuricTT: IMO we should use our own tests, not conftests - I should remove that metal test too

PR #596, opened 2025-09-02
reduce RAM overhead requirement because of failing CI machines

tstescoTT:
# change log
* reduce RAM overhead requirement because of failing CI machines, similarly reduce required disk because host hf cache can now be passed through without copies by default


acvejicTT (approved):
Lgtm thanks

PR #606, opened 2025-09-03
default evals batch_size=1, use max_concurrent to specify concurrent requests

tstescoTT:
# change log

* in evals/eval_config.py setting default batch_size=1 and use max_concurrent for all evals


bgoelTT (approved):
Okay looks good now, good catch on the `model_kwargs.max_concurrent` override

Issue #608, opened 2025-09-03
vLLM Docker image fails to build due to pyluwen (unavailable)


tstescoTT:
For example:
```
python3 scripts/build_docker_images.py --build-metal-commit v0.62.0-rc36
```

```
#9 865.7 ERROR: Could not find a version that satisfies the requirement pyluwen (unavailable) (from tt-smi) (from versions: 0.7.9, 0.7.10, 0.7.11)
#9 865.7 ERROR: No matching distribution found for pyluwen (unavailable)
```

PR #609, opened 2025-09-03
tstesco/fix-vllm-docker

tstescoTT:
# change log
* address #608 adding RUSTUP_HOME and CARGO_HOME handling so tt-smi and syseng tools install correctly


bgoelTT (approved):
Thanks for fixing this Tom

Issue #657, opened 2025-09-12
LLM first run tt-metal cache generation times out and causes run failure


tstescoTT:
Confusing to debug for first time users, the run may also fail for benchmarking and then pass for evals.

**Proposed fix:** 
Track the tt-metal cache generation on first run by device type and use an increased timeout awaiting the vLLM server to become available.

PR #658, opened 2025-09-12
tstesco/model-cache-timeout

tstescoTT:
# change log

* address #657, add utils/cache_monitor.py and support in util/prompt_client.py


bgoelTT (changes requested):
Just one requested change to support the DP param in TTT

PR #664, opened 2025-09-16
Return base64 image responses

idjuricTT:
Added returning base64 responses for image generation.

Why: When returning multiple images this is the only proper way of returning those.

Response looks:

image

Note: Tested on both SDXL, SD-3.5 and image-client for benchmarks

### DIFF
--- a/tt-metal-sdxl/model_services/image_service.py
+++ b/tt-metal-sdxl/model_services/image_service.py
@@ -12,15 +12,17 @@ class ImageService(BaseService):
     def __init__(self):
         super().__init__()
         self.image_manager = ImageManager("img")
-
+    
     def post_process(self, result):
-        return self.image_manager.convert_image_to_bytes(result)
+        """Convert PIL Image objects to base64 array"""
+        return self.image_manager.images_to_base64_list(result)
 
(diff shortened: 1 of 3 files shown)


tt-metal-sdxl/utils/image_manager.py
```diff
@@ -36,14 +36,57 @@ def delete_image(self, filename: str) -> bool:
         file_path.unlink()
         return True
 
+    def convert_image_to_base64(self, image: Image.Image, format="JPEG", quality=85):
```
ztorlakTT, line 39:
```suggestion
    def _convert_image_to_base64(self, image: Image.Image, format="JPEG", quality=85):
```

tt-metal-sdxl/model_services/image_service.py
```diff
@@ -12,15 +12,17 @@ class ImageService(BaseService):
     def __init__(self):
         super().__init__()
         self.image_manager = ImageManager("img")
-
+    
     def post_process(self, result):
-        return self.image_manager.convert_image_to_bytes(result)
```
ztorlakTT, line 17: We can remove `convert_image_to_bytes` from `ImageManager` since it is not used

ztorlakTT (approved):

PR #669, opened 2025-09-16
Add streaming support for Whisper audio transcription

ztorlakTT:
This PR implements real-time streaming capabilities for audio transcription using Whisper. 

Key changes:
• Added streaming mode to AudioService that processes audio in segments or time-based chunks
• Implemented two-level streaming architecture: service-level orchestration with model-level streaming for individual chunks
• Enhanced FastAPI endpoint to support streaming responses with proper HTTP headers
• Added intelligent audio processing that uses diarization segments when available, falling back to overlapping time-based chunks
• Maintains backward compatibility with existing non-streaming transcription requests

The streaming implementation provides faster time-to-first-token and allows clients to receive partial transcription results as they become available, improving user experience for real-time audio processing scenarios.


tt-media-server/model_services/audio_service.py
```diff
@@ -86,4 +87,60 @@ def stop_workers(self):
...
+            self.logger.error(f"Model-level streaming failed: {e}")
+            raise
+        finally:
+            self.scheduler.result_futures.pop(request._task_id, None)
+    
+    def _yield_final_streaming_result(self, result: dict):
+        if not 'final_result' in result:
+            raise Exception(f"Streaming result missing 'final_result': {result}")
+        
+        final_result_data = result['final_result']
```
idjuricTT, line 141: Could this overlap with multiple requests?

idjuricTT:
Approach makes sense IMO, it would be good to check how other inference servers are doing streaming just to compare our approach with those

PR #670, opened 2025-09-16
fix openai sdk dep, it can be managed by vllm requirements

tstescoTT:
# change log

* fix openai sdk dep, it can be managed by vllm requirements

---

This currently leads to 
```
(python_env) container_app_user@7016ef1d0885:~/app/src$ pip freeze | grep openai
openai==1.90.0                                                                                            
(python_env) container_app_user@7016ef1d0885:~/app/src$ pip freeze | grep datasets                                                                                                                                   
datasets==2.9.0
```


acvejicTT (approved):
lgtm

Issue #391, opened 2025-07-18
[Model Readiness Support]: Qwen2.5-VL-7B-Instruct


tstescoTT:
Note: this is a pure markdown version of new Model Readiness Support template, actual template will look different.

### Model Name

Name of model, typically HF repo ID without org name.

Qwen2.5-VL-7B-Instruct

### Model URL

Provide a link to the model publisher's repository, weights download, or documentation.

https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct

### Model Tenstorrent code implementation

Provide a link to the Tenstorrent model implementation demo code.

https://github.com/tenstorrent/tt-metal/tree/main/models/tt_transformers

### What accuracy evals are needed?

Corresponds to the evals/eval_config.py task_name.

From image benchmark table: https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct

- MMMU_val
- DocVQA_test
- ChartQA_test

### Model Readiness Support Preconditions

Preconditions for Model Readiness testing, non-blocking requirements must be provided before final run but do not block work. These tasks are decoupled and can be worked on by different people.

- [ ] Tenstorrent code implementation has all runtime preconditions and configurations documented.
- [ ] vLLM or other inference server integration complete.
- [ ] [Non-blocking] Performance targets (Top Perf) set in benchmarking/benchmark_targets/model_performance_reference.json.
- [ ] [Non-blocking] Required evals are implemented and integrated for model in evals/eval_config.py.
- [ ] [Non-blocking] GPU accuracy reference scores collected and set for model in evals/eval_config.py.

### Model Readiness Support Checklist

Required tasks to add support for a new model and are tightly coupled, ideally done by same person in sequence. This is independent of the measured model performance and accuracy scores and work to optimize performance.

- [ ] Inference server can serve model via run.py server workflow.
- [ ] Inference server can run release workflow. Release markdown report for target hardware pasted in comment below.
- [ ] Specific inference server docs cover how to serve model and run release workflow.
- [ ] Model is added to top level README table.
- [ ] Model is added to tt-shield Model Readiness CI nightly run producing release report data.

tstescoTT:
daily:
- @ssanjayTT to start running on Qwen2.5-VL-7B and 3B for setting up evals to run.

tstescoTT:
daily:
- blocked by model hangs.

ssanjayTT:
[...]

Issue #233, opened 2025-05-13
Model readiness support: Qwen2.5-VL-72B


tstescoTT:

tstescoTT:
daily:
- blocked by vLLM integration pending

stisiTT:
Daily - Continue to find issues; still working with branches rather than RCs. ENV issues persist. @ssanjayTT  and @tstescoTT  to reconnect and evaluate the model issues as they exist today.

tstescoTT:
daily:
- need to optimize for specific image resolutions, can pad to those. e.g. 300 DPI resolution.

tstescoTT:
daily:
- @ssanjayTT built image and ran with latest bug fix, still hitting some import errors.

tstescoTT:
daily:
- @ssanjayTT to test again after vllm branch re-based to numpy fix.

tstescoTT:
daily:
- need vLLM rebase on https://github.com/tenstorrent/vllm/tree/skhorasgani/qwenvl

tstescoTT:
daily:
- @ssanjayTT fixing vLLM environment failures. 
- Possibly for 72B model, we may only be able to support 32k context with batch-1.

tstescoTT:
daily:
- Now thinking batch-32 should be possible with large image resolutions.

bgoelTT:
daily:
- @gwangTT diagnosing > batch 1 bug in online vLLM server

tstescoTT:
daily:
- vLLM rebase merged in https://github.com/tenstorrent/vllm/commit/6c2a9ea5ff7a4212c7177359d2f859ccce5d88ec
- @ssanjayTT to start testing on LoudBox

tstescoTT:
daily:
- @ssanjayTT found hangs during trace capture, this happens not during tt-metal CI but when we do multiple trace captures for different ISL.
- @ssanjayTT to use tt-triage to debug

PR #676, opened 2025-09-16
Adding Qwen2.5-VL models

ssanjayTT:
# Related tickets
#391 #392 #233 
## Purpose
The goal of this PR is simply to add model support for the Qwen2.5-VL models of all sizes 3B, 7B, and 72B along with three multimodal evals for each with their published scores. Currently the models are not in a state for actually running these evals/benchmark suites and need further optimization/ bug fixing. The goal is simply to have them as experimental models in tt-inference-server and be able to build the docker images for these models for further testing. Tasks like adding gpu-reference scores for evals and benchmark targets shall be done later.

## Changelog
- added all three models to the model spec
- for each model we have 3 evals: docvqa, chartqa and mmmu_val with their published scores added
- modified docker build scripts for qwen2.5-vl models. Since these models have separate requirements outlined in tt-metal, we modify our docker image build script to take a new parameter called `--specific-model-image` which by default is None and leads to the regular tt-transformers requirements being installed. Otherwise depending on the model family (in this case qwen25-vl), we install the model specific requirements and publish an image for that model family.


vllm-tt-metal-llama3/vllm.tt-metal.src.cloud.Dockerfile
```diff
@@ -178,6 +179,20 @@ COPY --chown=${CONTAINER_APP_USERNAME}:${CONTAINER_APP_USERNAME} \
     "VERSION" "${APP_DIR}/VERSION"
 
 # Install additional app requirements into the copied venv
+RUN /bin/bash -c "source ${PYTHON_ENV_DIR}/bin/activate \
+    && pip install compressed-tensors"
```
tstescoTT, line 183: Is this needed for Qwen2.5-VL? I recall it was needed previously for TTT and was able to be removed.

vllm-tt-metal-llama3/vllm.tt-metal.src.cloud.Dockerfile
```diff
@@ -178,6 +179,20 @@ COPY --chown=${CONTAINER_APP_USERNAME}:${CONTAINER_APP_USERNAME} \
...
+
+ARG APP_DIR="${HOME_DIR}/app"
+WORKDIR ${APP_DIR}
```
tstescoTT, line 186: WORKDIR gets set after this step. Is this redundant?

vllm-tt-metal-llama3/vllm.tt-metal.src.cloud.Dockerfile
```diff
@@ -84,17 +87,15 @@ RUN /bin/bash -c "git clone https://github.com/tenstorrent-metal/tt-metal.git ${
...
 
 # Build vllm - clone with minimal history and clean
 RUN /bin/bash -c "git clone https://github.com/tenstorrent/vllm.git ${vllm_dir} \
     && cd ${vllm_dir} \
     && git checkout ${TT_VLLM_COMMIT_SHA_OR_TAG} \
     && source ${PYTHON_ENV_DIR}/bin/activate \
     && pip install --upgrade pip \
-    && pip install -e . --extra-index-url https://download.pytorch.org/whl/cpu \
-    && rm -rf ${vllm_dir}/.git"
+    && pip install -e . --extra-index-url https://download.pytorch.org/whl/cpu"
```
tstescoTT, line 98: What is the model size tradeoff for keeping the git history?

tstescoTT (commented):
The `specific_model_image` support addresses a key issue with Qwen2.5-VL having different python dependencies than the other vLLM integrated models.

I think this is a good solution. However I would like to see if a single image could contain python dependencies for all vLLM integrated models. This can be tested by combining both dependency files together seeing if a valid version exists for compatibility for all Python deps. Does this work?

If a single image works that would reduce complexity considerably for users needing to download an additional Docker image as well as CI and Models Release process.

We could keep this functionality for future use either way.

Changes would be required to tt-shield in order to support the new docker image build args for `specific_model_image`.

vllm-tt-metal-llama3/vllm.tt-metal.src.cloud.Dockerfile
ssanjayTT: So this is something I added for debugging because I wanted to switch to another branch. I shall be removing this soon.

tstescoTT:
@ssanjayTT can you please rebase this?

gwangTT:
> > Are you able to address this with a single image?
> 
> This depends entirely on if the dependencies in https://github.com/tenstorrent/tt-metal/blob/67884fcafb502cee219268b0c8a2f02799a97f29/models/demos/qwen25_vl/requirements.txt interfere with other dependencies of the models in tt-transformers. @gwangTT what do you think about this. Is this a feasible approach?

Sarva and I had a discussion on Slack and the conclusion was that there should be no interference now.

tstescoTT (approved):

PR #681, opened 2025-09-17
Gemma3 Update

ppetrovicTT:
- New version of Gemma3 supports multi-image prompts and works with uplifted vLLM.
- Add `MedGemma` model - same architecture as `Gemma3`, but trained on medical data.
- Add support for `gemma-3-1b-it` text model; Context length is 32k, per specification.
- Remove T3K target for `gemma-3-4b-it`, as the model cannot be run across 8 devices.

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -707,11 +707,33 @@ class ModelSpecTemplate:
 spec_templates = [
     ModelSpecTemplate(
         weights=[
-            "google/gemma-3-4b-it",
+            "google/gemma-3-1b-it",
         ],
         impl=tt_transformers_impl,
-        tt_metal_commit="87b758d",
-        vllm_commit="03cb300",
+        tt_metal_commit="dc85f59",
+        vllm_commit="87fe4a4",
+        device_model_specs=[
+            DeviceModelSpec(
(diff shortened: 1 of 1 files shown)


workflows/model_spec.py
```diff
@@ -707,24 +707,36 @@ def expand_to_specs(self) -> List["ModelSpec"]:
...
+        ],
+        impl=tt_transformers_impl,
+        tt_metal_commit="dc85f59",
+        vllm_commit="87fe4a4",
+        device_model_specs=[
             DeviceModelSpec(
-                device=DeviceTypes.N300,
+                device=DeviceTypes.N150,
                 max_concurrency=32,
                 max_context=128 * 1024,
```
bgoelTT, line 741: the 4B model on n150 supports a larger max context than the 1B model?

bgoelTT (approved):
Can we also run `google/gemma-3-12b-it`?

ppetrovicTT:
> Can we also run `google/gemma-3-12b-it`?

I've never tried it, so I hesitated to add it blindly.

bgoelTT:
We can throw it through Models CI and see what happens :)

PR #686, opened 2025-09-18
Add tt-mesh-graph-desc-path argument for T3K mesh partitioning support

### LINKED ISSUE #624: Support Galaxy Partitioning in T3K
To enable Galaxy Partitioning container per tray, two updates are required:

## 1. Update commit ids in ModelSpec
In the T3K ModelSpec, update `tt_metal_commit` and `vllm_commit` to known stable commits that support the partitioning container:
```
tt_metal_commit = "v0.62.0-rc33"
vllm_commit = "e7c329b"
```
at https://github.com/tenstorrent/tt-inference-server/blob/dev/workflows/model_spec.py#L967

@bgoelTT and I tested these commits on Galaxy6U.

## 2. Add mesh graph descriptor setting
Set the environment variable `TT_MESH_GRAPH_DESC_PATH` when running the containers as T3K on Galaxy:
```bash
TT_MESH_GRAPH_DESC_PATH=<path-to-tt-metal>/tt_metal/fabric/mesh_graph_descriptors/t3k_mesh_graph_descriptor.yaml
```

### Options for setting TT_MESH_GRAPH_DESC_PATH
- Option1 (preferred): External environment variable support
Allow passing external environment variables as arguments to `run.py`, which are then propagated into `docker_env_vars` in [run_docker_server.py#L106](https://github.com/tenstorrent/tt-inference-server/blob/dev/workflows/run_docker_server.py#L106).

- Option2: Extend ModelSpec to support the mesh graph descriptor
Add a mechanism to specify MeshGraphDescriptor within the ModelSpec. This option generalize beyond partitioning use cases, however this is higher complexity than Option1.

- Option3: Add new ModelSpec for partitioning Galaxy
Create a dedicated ModelSpec for partitioning Galaxy as T3K. This is simplest to apply directly, however it introduces much partitioning ModelSpecs, increasing maintenance burden.

I recommend *Option 1* because its straightforward and reduces duplication.
I'm not clear on the design principles for `tt-inference-server`, so I'd appreciate any comments you could provide.

CC: @bgoelTT

sott0n:
close #624 

This PR adds support for specifying a TT mesh graph description file path to enable proper T3K partitioning when running inference containers on Galaxy hardware.
    
## Changes:
- Add `--tt-mesh-graph-desc-path` CLI argument to run.py
- Pass tt_mesh_graph_desc_path parameter through setup_host function
- Set TT_MESH_GRAPH_DESC_PATH environment variable in Docker containers
    
When running partitioned containers on Galaxy hardware, tt-metal may auto-detect the hardware as Galaxy instead of T3K, causing failures. This argument allows explicit specification of the T3K mesh graph descriptor path to ensure proper hardware detection and partitioning behavior in containerized environments.
    
## Usage:
To run a partitioning container per tray1(device ids=0,1,2,3,4,5,6,7) as T3K:
```
python run.py \
  --model Llama-3.3-70B-Instruct \
  --workflow server \
  --docker-server \
  --device t3k \
  --device-id 0,1,2,3,4,5,6,7 \
  --tt-mesh-graph-desc-path /home/container_app_user/tt-metal/tt_metal/fabric/mesh_graph_descriptors/t3k_mesh_graph_descriptor.yaml
```

## Tests
I tested running Llama3.3-70B using partitioned containers, with four containers running simultaneously (one per tray, trays 1-4).
During this process, @bgoelTT and I confirmed that it is necessary to uplift the commit Ids of tt-metal and vLLM in order for the setup to run correctly.

This is my first PR for tt-inference-server. I believe we need a test to verify that environment variables are being passed correctly during `docker run` when this argument is provided, however I'm unsure where the corresponding test should be added. Could you please advise?
`test_run_arguments.py` would be the appropriate place, but I could not find a matching test class there.


run.py
```diff
@@ -150,6 +150,11 @@ def parse_arguments():
         type=str,
         help="Predefined eval dataset limit mappings: ['ci-nightly', 'ci-long', 'ci-commit', 'smoke-test']",
     )
+    parser.add_argument(
+        "--tt-mesh-graph-desc-path",
+        type=str,
+        help="Path to TT mesh graph description file (sets TT_MESH_GRAPH_DESC_PATH environment variable)",
```
bgoelTT, line 156:
```suggestion
        help="Path to TT mesh graph description file (sets TT_MESH_GRAPH_DESC_PATH environment variable in docker container)",
```

bgoelTT: Can you also add a check in `run::validate_runtime_args` that raises a `NotImplementedError` if `--tt-mesh-graph-desc-path` is set without `--docker-server` also being set?

workflows/run_docker_server.py
```diff
@@ -111,6 +109,15 @@ def run_docker_server(model_spec, setup_config, json_fpath):
         "TT_MODEL_SPEC_JSON_PATH": docker_json_fpath,
     }
 
+    # Add TT_MESH_GRAPH_DESC_PATH if provided
+    # When running on Galaxy with partitioning (per tray), we need to explicitly
+    # set the T3K mesh graph descriptor path. Without this, tt-metal will auto-detect
+    # the hardware as Galaxy instead of T3K, causing failures in partitioned containers.
```
bgoelTT, line 115:
```suggestion
    # When running on Galaxy with partitioning (per-tray or sub-tray), we need to explicitly
    # set the mesh graph descriptor path. Without this, tt-metal will auto-detect
    # the hardware as Galaxy instead of the desired topology (i.e. n150 or T3K), causing failures in partitioned containers.
```

workflows/run_docker_server.py
```diff
@@ -111,6 +109,15 @@ def run_docker_server(model_spec, setup_config, json_fpath):
...
+    if setup_config.tt_mesh_graph_desc_path:
+        docker_env_vars["TT_MESH_GRAPH_DESC_PATH"] = (
+            setup_config.tt_mesh_graph_desc_path
+        )
```
bgoelTT, line 119: As mentioned in our discussion, use the `args` local variable to parse the `tt_mesh_graph_desc_path` attribute, if set

workflows/setup_host.py
bgoelTT, line 1: This file should be unchanged (running the `pre-commit` is fine) as we should now parse the CLI args via the `ModelSpec`

bgoelTT (changes requested):
Looks good, just a few comments / suggestions

@tstescoTT I'm going to approve this once the fixes are made but I'm wondering if we should allow the arbitrary setting of environment variables in the `--docker-server` via the CLI. I will be filing an issue in vLLM that will defer the setting of this `TT_MESH_GRAPH_DESC_PATH` env var inside the `tt_worker.py` as a follow-up to this.

workflows/setup_host.py
sott0n: Removed this change (only keep changes with pre-commit)

run.py
sott0n: I added this validation under the `validate_runtime_args`.

sott0n:
@tstescoTT @bgoelTT Got it, that approach sounds good to me. I think we should close this PR and instead create a new one to add Galaxy_T3K to DeviceType. Is anyone already working on this? If not, I'll take it on.

PR #732, opened 2025-09-23
uplift tt-metal llama 70b glx, fix vllm overrides

tstescoTT:
# change log

* uplift tt-metal llama 70b glx, fix vllm overrides

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -613,6 +613,12 @@ class ModelSpec:
                 "override_tt_config",
                 json.loads(args.override_tt_config),
             )
+            # Update vllm_args to include the new override_tt_config
+            merged_vllm_args = {
+                **self.device_model_spec.vllm_args,
+                "override_tt_config": args.override_tt_config,
+            }
+            object.__setattr__(self.device_model_spec, "vllm_args", merged_vllm_args)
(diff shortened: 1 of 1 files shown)


tstescoTT:
> 1. ah, does this mean we weren't setting the override tt config in the vLLM CLI? (this could explain why @stisiTT your DP experiments weren't working)
> 2. do these uplifted commits include the new batch prefill work?

1. Could be the case! not sure how those were being run.
2. Yes it should, however I'm not sure that's being enabled fully. In particular, this is with ring distributed SDPA which gives a prefill boost to high ISL: https://github.com/tenstorrent/tt-metal/commit/4dcc88f813a3208d7f7cfdcf92bf34686d709824
The performance needs further characterization in decode and there is a patch required in https://github.com/tenstorrent/tt-metal/commit/d8073ed500c7a66a3faff9dd3045ff9456531795.

This is maybe a little too cutting-edge given the current way we are using `dev`, but I want to push forward on Llama 70B Galaxy given the other known issues. Also want to move towards the Model Release process and publish release images + model_spec.py changes more frequently.

bgoelTT:
Sounds good, I believe the batched prefill changes only kick in for certain ISL sizes at this moment. I think this happens transparently.

PR #755, opened 2025-09-26
Speaker-Aware VAD Segment Merging with Duration Limit for Whisper

ztorlakTT:
This PR introduces a new method, `_merge_vad_segments_by_speaker_and_duration`, in `AudioManager` to create Whisper input chunks that respect both speaker boundaries and a maximum duration. Chunks are merged up to a target duration (default 30s).
This approach improves Whisper performance by providing speaker-consistent, optimally sized audio segments for transcription.

### DIFF
--- a/tt-media-server/model_services/base_service.py
+++ b/tt-media-server/model_services/base_service.py
@@ -82,13 +82,8 @@ class BaseService(ABC):
             self.scheduler.result_futures[request._task_id] = future
             
         try:
-            result = await asyncio.wait_for(future, timeout=settings.default_inference_timeout_seconds)
+            result = await future
             return result
-        except asyncio.TimeoutError:
-            self.logger.error(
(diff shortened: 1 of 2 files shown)


tt-media-server/model_services/base_service.py
```diff
@@ -77,15 +77,23 @@ async def pre_process(self, request):
...
-        self.scheduler.result_futures[request._task_id] = future
+        
+        with self.scheduler.result_futures_lock:
+            self.scheduler.result_futures[request._task_id] = future
+            
         try:
-            result = await future
+            result = await asyncio.wait_for(future, timeout=settings.default_inference_timeout_seconds)
             return result
+        except asyncio.TimeoutError:
```
idjuricTT, line 87: This is redundant to the timeout we have on a device runner level

idjuricTT (approved):

PR #800, opened 2025-10-02
Increase timeout for YOLOv4 weight distribution

changh95:
# Background

When running YOLOv4 on tt-media-server, there is a timeout limit on the weight distribution stage.

This was set to 120 seconds originally.

But this was not enough when running it on N300 cloud VM.

# Action

We set the timeout value to a generous amount - 300 seconds.

# Results

Works now 👍 

image


changh95:
@tstescoTT My bad! Fixed now.

idjuricTT (approved):

ztorlakTT (approved):

PR #806, opened 2025-10-06
set TT_MM_THROTTLE_PERF for Qwen3-32B and QwQ-32B on galaxy

tstescoTT:
# change log 

* set TT_MM_THROTTLE_PERF for Qwen3-32B and QwQ-32B on galaxy


bgoelTT (approved):
This will use the maximum MM throttling, we never understand if we could set it to level 1 (highest duty cycle)

Can we try with level 3, and if it fails in CI then we adjust accordingly? I.e., to level 2 if it passes, or level 4 if it fails.

PR #807, opened 2025-10-06
Uplift versions for Qwen3-32B & add Galaxy_T3K DeviceModelSpec

bgoelTT:
Experimentally verified that `TT_MM_THROTTLE_PERF=2` was the lowest throttle level that resolved hangs on GALAXY_T3K. Going to throttle level 1 still hit hang during prefill at ISL=8k


bgoelTT:
Okay nice, using level 2 passed benchmarks and evals

PR #857, opened 2025-10-15
Add support for xla models

vmilosevic:
This PR introduces several changes to how the inference server uses Forge to run models. The goal is to switch to using the TT-XLA frontend and enable running all supported models in future,  through a uniform interface for handling model inputs and outputs.

Changes:

- Use tt-xla (pjrt plugin) instead of tt-forge-fe for running models
- Added support for models: resnet, mobilenetv2, vovnet
- Select model using env variables

```
# instead
export MODEL_RUNNER=forge

# use new model runners
export MODEL_RUNNER=tt-xla-resnet
export MODEL_RUNNER=tt-xla-vovnet
export MODEL_RUNNER=tt-xla-mobilenetv2
```

- Set cache service for faster initial model loading in our environments

```
export IRD_LF_CACHE=http://aus2-lfcache.aus2.tenstorrent.com/ 
(or http://large-file-cache.large-file-cache.svc.cluster.local/ in Civ2)
```

- Use python 3.11 environment
- Skip resizing image as it will be handled by model loader
- Updated demo code, split html into css, js and html


tt-media-server/main.py
```diff
@@ -35,3 +35,8 @@ async def lifespan(app: FastAPI):
 
 # Mount static files
 app.mount("/static", StaticFiles(directory="static"), name="static")
+
+# Launch main app for local testing
+if __name__ == "__main__":
```
idjuricTT: Do we need this change? I'm not sure how it would affect the runtime

tt-media-server/tt_model_runners/forge_runners/README.md
```diff
@@ -30,16 +30,32 @@ This module provides:
...
+Model is CNN model name to load
+   - mobilenetv2
+   - resnet
+   - vovnet
+
+Set appropriate HF_TOKEN to load weights from Huggingface.
+IRD_LF_CACHE is out large file caching service, in IRD enviroment use http://aus2-lfcache.aus2.tenstorrent.com/
+
 ```bash
+export MODEL_RUNNER=forge
```
idjuricTT: Suggestion: Would it make sense to use the existing architecture  to support models?
I mean instead of using Forge as model_runner to have different runners that will load the models

idjuricTT (approved):

idjuricTT (approved):

PR #879, opened 2025-10-16
fix media server create_env.sh

acvejicTT:


acvejicTT:
Successful run:
https://github.com/tenstorrent/tt-shield/actions/runs/18556354431/job/52894699105

PR #898, opened 2025-10-17
Quickfix Benchmarking Workflow

ljovanovicTT:
## Problem
`MODEL` was passed in the wrong format, causing release pipeline crash


acvejicTT (approved):
lgtm

PR #899, opened 2025-10-17
updated and formatted performance reference json

gtobarTT:
Updated performance json file.


benchmarking/benchmark_targets/model_performance_reference.json
```diff
@@ -473,62 +4781,6 @@
...
-                "num_prompts": 8,
-                "targets": {
-                    "theoretical": {
-                        "ttft_ms": 50,
-                        "tput_user": 80
-                    }
-                }
-            },
-            {
-                "isl": 2048,
```
tstescoTT: Are we intentionally removing the ISL=2048 target?

tstescoTT (approved):

Issue #123, opened 2025-02-25
Setup tool to run in tt-metal dev env


tstescoTT:
Setup tool - bash oneliner to be run in tt-metal developer environment to run vLLM model inference server using model developer code.

Script can additionally run:

- https://github.com/tenstorrent/tt-inference-server/issues/117
- https://github.com/tenstorrent/tt-inference-server/issues/120
- https://github.com/tenstorrent/tt-inference-server/issues/122

bgoelTT:
The final setup tool should *not* require a containerized environment. The tool cannot user docker, thus, it should be able to bind to the `tt-metal` development environment such that the environment is unaffected.

tstescoTT:
For first pass, using the defaults assumed by the tt-metal setup in https://github.com/tenstorrent/tt-metal/tree/main/models/demos/llama3 should be fine. i.e. setting the environment variables accordingly and checking the model weights exist in this place, or in the tt-inference-server format (created by running setup.sh).

tstescoTT:
Note from @rjakovljevicTT: we should be able to pass a certain commit SHA (not only the top of the main) when running the pipeline. That will enable to create a pipeline that runs the tests for my submitted change in e.g. metal/llm/tt-inference-server.

tstescoTT:
addressed in https://github.com/tenstorrent/tt-inference-server/pull/2437

PR #912, opened 2025-10-20
Run workflows against already running local server

ppetrovicTT:
For development purposes, it is neat to just run the workflows (benchmarks, evals) against the already running local server.
This way I can run the server manually and breakpoint in the code, while actually running the benchmarks required by the CI.

### DIFF
--- a/docs/workflows_user_guide.md
+++ b/docs/workflows_user_guide.md
@@ -76,10 +76,10 @@ You will need to accept the terms for any specific gated datasets or model repos
 | Option                   | Description                                                                                      |
 |--------------------------|--------------------------------------------------------------------------------------------------|
 | `--docker-server`        | Run the inference server inside a **Docker container**.                                          |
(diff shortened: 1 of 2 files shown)


tstescoTT:
Hey @ppetrovicTT , reporting this here:

I think there is some confusion and maybe I'm misunderstanding this PR. Because the `--local-server` feature was still in testing and documented. This feature is for starting the vLLM server locally, i.e. finding where tt-metal is installed with vLLM,
setting all env vars, and running the vLLM command with args to run just the client side scripts against a pre-setup vLLM server, you can run `--workflow {benchmarks,evals}` without setting `--docker-server` or `--local-server` to start vLLM.

tstescoTT:
cross ref: https://github.com/tenstorrent/tt-inference-server/issues/123

ppetrovicTT:
Sorry for the confusion, I will close this.

I believe I've had some issues starting just benchmarks, from the setup side (installing venvs, etc.) so I created this PR to skip unnecessary parts. But I've tried recently with only `--workflow evals` and it worked.

PR #925, opened 2025-10-21
used VersionMode.SUGGESTED unless explicit need

tstescoTT:
# change log

* used VersionMode.SUGGESTED unless explicit need

For `Llama-3.3-70B` on Galaxy and T3K there is not a known explicit need for firmware/KMD minimum versions but we want to include the versions as suggested from what was tested and found to be working.

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -1189,11 +1189,11 @@ spec_templates = [
         system_requirements=SystemRequirements(
             firmware=VersionRequirement(
                 specifier=">=18.6.0",
-                mode=VersionMode.SUGGESTED,
+                mode=VersionMode.STRICT,
             ),
             kmd=VersionRequirement(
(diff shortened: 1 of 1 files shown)


workflows/model_spec.py
```diff
@@ -1209,11 +1209,11 @@ def expand_to_specs(self) -> List["ModelSpec"]:
         system_requirements=SystemRequirements(
             firmware=VersionRequirement(
                 specifier=">=18.8.0",
-                mode=VersionMode.STRICT,
+                mode=VersionMode.SUGGESTED,
             ),
             kmd=VersionRequirement(
                 specifier=">=2.2.0",
-                mode=VersionMode.STRICT,
+                mode=VersionMode.SUGGESTED,
```
bgoelTT: Good with this, I just made it strict because that's what I tested, not because it's a hard requirement

PR #995, opened 2025-10-28
make logic simpler in get_trace_context_lens()

tstescoTT:
##  change log

* make logic simpler in get_trace_context_lens()

This is based on recommendation in https://github.com/tenstorrent/tt-inference-server/pull/907

### DIFF
--- a/utils/prompt_client.py
+++ b/utils/prompt_client.py
@@ -61,7 +61,7 @@ def get_trace_context_lens(
     Returns:
         List of (input_seq_len, output_seq_len) tuples
     """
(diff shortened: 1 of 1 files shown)


nidhinjose-git:
@tstescoTT I see that the logic is reverted back to its original form in this commit: https://github.com/tenstorrent/tt-inference-server/commit/337c12df17fe8ffd3a6ee57de601004980363afb

Was it intentional?

tstescoTT:
Nope, just a bad merge https://github.com/tenstorrent/tt-inference-server/pull/1014#discussion_r2476391924

Not a big issue, the logic here doesn't make a difference for trace capture. We'll put it back to the one in https://github.com/tenstorrent/tt-inference-server/pull/995

Issue #971, opened 2025-10-27
Update to benchmark workflows


milank94:
1. Add an argument for the “URL” — so we can use the same benchmark script to target different deployments without having to modify the code each time.
2. Add an argument for selecting the endpoint — the current workflow for text models still uses /v1/completions, but since that endpoint is now deprecated, we’ve moved to /v1/chat/completions.

It would be great if we could add an argument to choose which endpoint to use in the benchmark workflow.
These changes would help make the benchmark workflow more adaptable across environments and consistent with the latest API usage.

PR #996, opened 2025-10-29
adding --host and --endpoint run.py CLI args

tstescoTT:
# change log

* adding --host and --endpoint run.py CLI args addressing https://github.com/tenstorrent/tt-inference-server/issues/971

### DIFF
--- a/benchmarking/run_benchmarks.py
+++ b/benchmarking/run_benchmarks.py
@@ -227,6 +227,7 @@ def main():
     logger.info("Wait for the vLLM server to be ready ...")
     env_config = EnvironmentConfig()
     env_config.jwt_secret = jwt_secret
+    env_config.api_key = api_key
     env_config.vllm_model = model_spec.hf_model_repo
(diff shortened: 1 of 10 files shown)


bgoelTT (commented):
This looks great, just one comment

bgoelTT (commented):
Sorry for moving the goalpoasts, I didn't make this clear

With these changes we can specify host, service port, and endpoint. But what if we wanted to target a URL from the internet? We don't actuall want to include the `f":{service_port}"` part, we pretty much just want `f"{host}{endpoint}"`

Looking at `benchmark_serving.py` they use `base_url` https://github.com/tstescoTT/vllm/blob/b25243544da81a93e6c83ec0e17c4dda484d1494/benchmarks/benchmark_serving.py#L832, which if set, does what I've described above

To make you not do anymore work, maybe we can make `benchmarks` and `evals` conditionally set `--base-url` in the exec command if `args.host` has been set, and ignore `args.service_port`?

Issue #952, opened 2025-10-24
Try removing .git folder from docker image


idjuricTT:
To trim image size try removing .git folders when creating a docker images

PR #1013, opened 2025-10-29
Reduce Docker Image Size

ljovanovicTT:
## Issue
[#952 Reduce Docker Image Size](https://github.com/tenstorrent/tt-inference-server/issues/952)

## Problem
Docker Image was over 34GB in size, making it hard to distribute.
```
sdxl-inf-server   latest    ef1f9dbd23f9   4 minutes ago   34.5GB
```

With these changes, the final image size was reduced to around 20GB.
```
sdxl-inf-server-twostage  latest    3c2b66c9c0ca   48 minutes ago   20.6GB
```

## Solution
- Removed build-only directories
- Created a [multi-stage](https://docs.docker.com/get-started/docker-concepts/building-images/multi-stage-builds/) build o minimize unnecessary layers and reduce the final image size


idjuricTT (approved):
Looks good but please let's have all sucesfull runs before merging so we don't break the upcoming release

fivanovicTT (approved):

ztorlakTT (approved):

PR #1014, opened 2025-10-29
Add signal file to end of trace warmup

bgoelTT:
[A standard probing pattern in k8s Pods is to execute commands inside the container, whose completion status determines the value of the probe.](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command)

To facilitate accurate and precise probing of the vLLM warmup mechanism (not `/health` endpoint), the warmup will now write a signal file upon completion. The existence of this signal file can be used by k8s probes (specifically the `readinessProbe`) to determine if the server is ready to serve traffic.

Example probe setup for Llama-3.3-70B-Instruct on Galaxy:
```yaml
ports:
- containerPort: 8000
  name: http
# probes if the server has started
startupProbe:
  httpGet:
    path: /health
    port: http
  # tune these params based on average expected startup duration
  initialDelaySeconds: 120
# probes if the server is alive
livenessProbe:
  httpGet:
    path: /health
    port: http
  # tune these params based on desired liveness polling rate
  failureThreshold: 3
  periodSeconds: 10
# probes if the server is ready to handle traffic
readinessProbe:
  exec:
    # This command checks if the file /tmp/ready exists
    command:
    - test
    - -f
    - /tmp/ready
  # delay 10 mins after startupProbe succeeds, wait for warmup
  initialDelaySeconds: 660
  failureThreshold: 6
  periodSeconds: 10
```


tstescoTT (approved):

utils/prompt_client.py
```diff
@@ -53,15 +53,19 @@ def get_trace_context_lens(
...
-        
+
     Returns:
         List of (input_seq_len, output_seq_len) tuples
     """
-    return [(seq_len, output_len) for seq_len in PADDED_SEQ_LENS if (seq_len + output_len) <= max_context]
+    return [
+        (seq_len, output_len)
+        for seq_len in PADDED_SEQ_LENS
+        if seq_len <= (max_context + output_len)
```
tstescoTT: Looks like this wasnt rebased.

PR #1029, opened 2025-10-30
Llama-3.1-8B trace region sizes

nostojicTT:
Llama-3.1-8B supports prefill tracing from https://github.com/tenstorrent/tt-metal/pull/29827

N150 - Up to 1k sequence length, as op-to-op gaps are too small above that limit
N300, T3K, Galaxy - Up to 8k sequence length

Due to memory limitations, the trace region sizes need to be different between platforms.

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -1404,7 +1404,7 @@ spec_templates = [
     ModelSpecTemplate(
         weights=["meta-llama/Llama-3.1-8B", "meta-llama/Llama-3.1-8B-Instruct"],
         impl=tt_transformers_impl,
-        tt_metal_commit="9b67e09",
+        tt_metal_commit="20fbdc2",
         vllm_commit="a91b644",
         device_model_specs=[
             DeviceModelSpec(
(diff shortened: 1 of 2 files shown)


workflows/model_spec.py
```diff
@@ -1472,6 +1481,7 @@ def expand_to_specs(self) -> List["ModelSpec"]:
                 override_tt_config={
                     "data_parallel": 4,
                     "sample_on_device_mode": "decode_only",
+                    "trace_region_size": 33000000,
```
ppetrovicTT: this is for blackhole quietbox. did you actually run this platform?

workflows/model_spec.py
```diff
@@ -1404,26 +1404,35 @@ def expand_to_specs(self) -> List["ModelSpec"]:
...
         vllm_commit="a91b644",
         device_model_specs=[
             DeviceModelSpec(
                 device=DeviceTypes.N150,
                 max_concurrency=32,
                 max_context=64 * 1024,
                 default_impl=True,
+                override_tt_config= {
+                    "trace_region_size": 25000000,
+                },
```
bgoelTT: This is the default in vLLM so we do not need to override https://github.com/tenstorrent/vllm/blob/0edd242e0f91ca92cb95624f1acb1237fc9c2008/vllm/worker/tt_worker.py#L541

bgoelTT (changes requested):
If you have to bump the `trace_region_size` for T3K you'll also need to do it for `GALAXY_T3K`

bgoelTT (approved):

PR #1094, opened 2025-11-06
media-server release report from perf benchmarks and evals

fivanovicTT:
### Link to GitHub issue
[922](https://github.com/tenstorrent/tt-inference-server/issues/922)

### Summary
Fix the following:

1. Cannot hardcode evals passing, and the benchmarks should not have the eval results in them: https://github.com/tenstorrent/tt-inference-server/blob/dev/utils/media_clients/audio_client.py#L155
2. t/s/u uses len(total_text.split()) (word count), not model tokens. This is not the correct measurement, it should be the model tokens https://github.com/tenstorrent/tt-inference-server/blob/dev/utils/media_clients/audio_client.py#L340
3. Only the first request’s t/s/u and RTR are recorded in reports instead of averaging across all requests https://github.com/tenstorrent/tt-inference-server/blob/dev/utils/media_clients/audio_client.py#L157, should use means

### DIFF
--- a/utils/media_clients/audio_client.py
+++ b/utils/media_clients/audio_client.py
@@ -2,27 +2,33 @@
 #
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
-from pathlib import Path
-import time
-from typing import Optional
-import logging
-import requests
-import json
+# Standard library imports
 import asyncio
+import json
+import logging
+import sys
+import time
+from pathlib import Path
+from typing import Optional
+
+# Third-party imports
 import aiohttp
+import requests
+from transformers import AutoTokenizer
+
+# Local imports
 from .base_strategy_interface import BaseMediaStrategy
 from .test_status import AudioTestStatus
-import sys
-from pathlib import Path
+
 # Add project root to Python path
 project_root = Path(__file__).resolve().parent.parent.parent
 if str(project_root) not in sys.path:
     sys.path.insert(0, str(project_root))
 
 from workflows.utils import (
-    is_streaming_enabled_for_whisper,
+    get_num_calls,
     is_preprocessing_enabled_for_whisper,
-    get_num_calls
+    is_streaming_enabled_for_whisper,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


idjuricTT (approved):

utils/media_clients/audio_client.py
```diff
@@ -309,14 +319,14 @@ async def _transcribe_audio_streaming_on(self, is_preprocessing_enabled: bool) -
                             audio_duration = result.get("duration")
                             logger.info(f"Found audio duration in chunk: {audio_duration}s")
 
+                        # Calculate tokens for this chunk only
+                        chunk_tokens = len(text.split()) if text.strip() else 0
```
tstescoTT: Is this still not using the tokenizer to count the tokens?

utils/media_clients/test_status.py
```diff
@@ -45,17 +45,17 @@ def to_dict(self) -> Dict[str, Any]:
 class AudioTestStatus(BaseTestStatus):
     """Test status for audio transcription models."""
     def __init__(self, status: bool, elapsed: float, ttft: Optional[float] = None,
-                tpups: Optional[float] = None, rtr: Optional[float] = None):
+                tsu: Optional[float] = None, rtr: Optional[float] = None):
         super().__init__(status, elapsed)
         self.ttft = ttft
-        self.tpups = tpups
+        self.tsu = tsu
```
tstescoTT: thanks for fixing this as well 🙏

tstescoTT:
Tested and appears to be working, you can see the change from 7.8 (using word split counting) to 9.8 t/s/u from tokenization change below:
```
#### Audio Benchmark Sweeps for whisper-large-v3 on n150

| Num Requests | TTFT (ms) | Streaming enabled | Preprocessing enabled | Accuracy Check | T/S/U | RTR  |
|--------------|-----------|-------------------|-----------------------|----------------|-------|------|
|            2 |    1418.0 | True              | False                 |              2 |  7.68 | 2.22 |
|            2 |    1373.2 | True              | False                 |              2 |  7.8  | 2.26 |
|            2 |    1378.0 | True              | False                 |              0 |  9.82 | 2.23 |
|            2 |    5249.5 | True              | True                  |              0 |  9.24 | 1.46 |

```

tstescoTT (approved):

PR #1191, opened 2025-11-18
fix: standardize authentication to use VLLM_API_KEY for both vllm and trace capture

nidhinjose-git:
Production deployments mostly use VLLM_API_KEY rather than JWT_SECRET (atleast with flexai). When the container starts, most deployments would already expose VLLM_API_KEY.

With the changes in this PR, both trace capture and vLLM server now follow the same priority:
1. First check for VLLM_API_KEY
2. If not present, fallback to JWT_SECRET
3. If JWT_SECRET is not present, no authorization.

Previously, trace capture and vLLM server used different environment variable names for the same authentication token (VLLM_API_KEY and AUTHORIZATION). Since trace capture runs inside the same container as the vLLM server, we will have to expose the same token as both variables with the same token values. Let both components use VLLM_API_KEY instead.

### DIFF
--- a/utils/prompt_client.py
+++ b/utils/prompt_client.py
@@ -660,7 +660,6 @@ class PromptClient:
 def run_background_trace_capture(
     hf_model_repo: str,
     service_port: int,
-    jwt_secret: str = None,
     supported_modalities: List[str] = None,
     max_context: int = None,
     context_lens: List[Tuple[int, int]] = None,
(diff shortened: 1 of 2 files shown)


nidhinjose-git:
@tstescoTT I have now brought back `AUTHORIZATION` env variable. Now,

****vLLM** behavior:**

  1. Check `VLLM_API_KEY` first

  2. If not present, fall back to `JWT_SECRET`

  3. If `JWT_SECRET` is not present, no authorization is applied

****Trace capture** behavior:**

  1. Check `VLLM_API_KEY` first

  2. If not present, check `AUTHORIZATION`

  3. If not present, fall back to `JWT_SECRET`

  4. If `JWT_SECRET` is not present, no authorization header is sent
  
  
All previous behaviors remain unchanged, except that `VLLM_API_KEY` will now be used as the authentication token when present, overriding `JWT_SECRET` and `AUTHORIZATION`.

PR #1193, opened 2025-11-19
Fix missing new MGD path for Galaxy_T3K

sott0n:
For transition MGD2 on Galaxy_T3K, we need to specify a new MGD path into `TT_MESH_GRAPH_DESC_PATH ` with fabric config. If no fabric configuration is specified, the system assumes it is running on a T3K in Galaxy and therefore selects the TorusXY, which results in a fabric validation failure.

PR #1199, opened 2025-11-19
Add Gemma3-27B into Galaxy_T3K

sott0n:
Added Gemma3-27B model spec for Galaxy_T3K with MGD2 related to #1193 

@bgoelTT In https://github.com/tenstorrent/tt-inference-server/pull/1023, you removed the `galaxy_t3k` pattern — is there any particular reason for this? Without this, the report workflow should fail with an error.


sott0n:
Yes, we need this. The customer has requested to run Gemma3 as T3K in Partitioning Galaxy.

sott0n:
@tstescoTT could you review it?

tstescoTT (approved):

PR #1241, opened 2025-11-25
Support plain text response format for audio transcriptions

ztorlakTT:
Added `return_json_response` parameter (default: `true`) to audio transcription requests.
When set to false, the endpoint returns only the transcribed text instead of the full JSON response object

### DIFF
--- a/tt-media-server/open_ai_api/audio.py
+++ b/tt-media-server/open_ai_api/audio.py
@@ -83,8 +83,7 @@ async def transcribe_audio(
 
             if audio_transcription_request.return_json_response:
                 return result.to_dict()
-            else:
-                return Response(content=result.text, media_type="text/plain")
+            return Response(content=result.text, media_type="text/plain")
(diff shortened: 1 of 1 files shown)


ljovanovicTT (approved):

tt-media-server/open_ai_api/audio.py
```diff
@@ -76,7 +81,10 @@ async def transcribe_audio(
                     f"Unexpected response type: {type(result).__name__}. Expected response class with to_dict() method."
                 )
 
-            return result.to_dict()
+            if audio_transcription_request.return_json_response:
+                return result.to_dict()
+            else:
+                return Response(content=result.text, media_type="text/plain")
```
fivanovicTT:
```suggestion
            if audio_transcription_request.return_json_response:
                return result.to_dict()

            return Response(content=result.text, media_type="text/plain")
```

PR #1269, opened 2025-11-27
Generalize segment-based parallel processing pattern in BaseService

ztorlakTT:
Refactored the segment-based parallel processing pattern into `BaseService` to enable code reuse across `AudioService` and `ImageService`.

### DIFF
--- a/tt-media-server/model_services/audio_service.py
+++ b/tt-media-server/model_services/audio_service.py
@@ -4,10 +4,10 @@
 
 from config.settings import settings
 from domain.audio_processing_request import AudioProcessingRequest
-from domain.audio_text_response import combine_transcription_responses
 from model_services.base_service import BaseService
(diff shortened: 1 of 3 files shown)


tt-media-server/model_services/audio_service.py
```diff
@@ -2,12 +2,9 @@
 #
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
-import asyncio
-from typing import List
-
 from config.settings import settings
 from domain.audio_processing_request import AudioProcessingRequest
-from domain.audio_text_response import AudioTextResponse, AudioTextSegment
+from domain.audio_text_response import combine_transcription_responses
```
idjuricTT: I'm a bit unsure should this be a part of domain?

ztorlakTT: I moved this function to audio_manager

PR #1272, opened 2025-11-27
#33058 (metal issue) Limit torch threads in SDXL inference.

pavlepopovic:
SDXL has some torch code in encode_prompts path (such as concat, chunk etc).
When ran as 32 independent users, each process can spawn as many torch threads as it wants, leading to thread explosions in these cases, producing huge overheads (such as +3s for encoding prompts), if 32 requests arrive at the same time.

To avoid this, we can limit number of threads the torch is available to spawn.

### DIFF
--- a/tt-media-server/tt_model_runners/sdxl_generate_runner_trace.py
+++ b/tt-media-server/tt_model_runners/sdxl_generate_runner_trace.py
@@ -68,6 +68,14 @@ class TTSDXLGenerateRunnerTrace(BaseSDXLRunner):
         self._apply_request_settings(requests[0])
         
         self.logger.debug(f"Device {self.device_id}: Starting text encoding...")
+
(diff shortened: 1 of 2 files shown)


tt-media-server/tt_model_runners/sdxl_generate_runner_trace.py
```diff
@@ -68,6 +68,14 @@ def run_inference(self, requests: list[ImageGenerateRequest]):
         self._apply_request_settings(requests[0])
         
         self.logger.debug(f"Device {self.device_id}: Starting text encoding...")
+
+        # Limit the number of threads torch can create in order to avoid thread explosion when running multi-process scenarios (such as 32 processes on a galaxy).
+        # This way, torch can create only one thread per process, instead of predefined number of them (32).
```
idjuricTT: Would this be better to be in init of base sdxl runner?

This way we're going to run the check on every inference and instance of this class will leave through out the whole server runtime

idjuricTT (approved):

PR #1294, opened 2025-12-01
Reduce Inference Server Forge Docker Image Size

ljovanovicTT:
## Problem
It was possible to further reduce Docker image size.

## Implementation
- use lean base image for both runtime and build stages
- introduce two-stage build

## Results
```
REPOSITORY         TAG        IMAGE ID       CREATED       SIZE
inf-server-forge   original   c2fa70f9c552   1 day ago   25.9GB
```
```
REPOSITORY     TAG       IMAGE ID       CREATED         SIZE
forge-server   latest    8dd7886ef2e3   4 minutes ago   9.59GB
```


tt-media-server/Dockerfile.forge
ztorlakTT: Can we delete these lines as well?

PR #1300, opened 2025-12-02
Limit torch threads

idjuricTT:


tt-media-server/config/settings.py
```diff
@@ -73,7 +73,7 @@ class Settings(BaseSettings):
     audio_chunk_duration_seconds: Optional[int] = None
     max_audio_duration_seconds: float = 60.0
     max_audio_duration_with_preprocessing_seconds: float = (
-        300.0  # 5 minutes when preprocessing enabled
+        30000  # 5 minutes when preprocessing enabled
```
ztorlakTT: Can you please update the comment, and README file?

ztorlakTT (approved):

tt-media-server/config/settings.py
idjuricTT: I reverted this one as well 😄

PR #1301, opened 2025-12-02
Extend Multiprocessing Queue

ljovanovicTT:
## Problem
Implementing a custom queue introduced significant overhead on every read and write operation. Also, with this design switching to another multiprocessing Queue implementation in the future is simple and configurable

## Results
**:bar_chart: Performance Comparison (100,000 objects):**
   **Multiprocessing Queue:**
     Write: 0.0758s (1,319,382 obj/s)
     Read:  1.8540s (53,938 obj/s)
     
   **Custom Queue:**
     Write: 0.0802s (1,246,889 obj/s)
     Read:  1.5287s (65,417 obj/s)
     
   **Performance Ratios (higher is better for custom queue):**
     Write: 0.95x slower
     Read:  1.21x faster


idjuricTT (approved):

idjuricTT (approved):

tt-media-server/model_services/tt_queue.py
```diff
@@ -0,0 +1,67 @@
...
+    def get(self, block=True, timeout=None):
+        """Get item, checking leftover cache first if batching enabled"""
+        if self.batch_enabled:
+            try:
+                item = self._leftover_items.get_nowait()
+                return item
+            except Empty:
+                return super().get(block=block, timeout=timeout)
+        else:
+            return super().get(block=block, timeout=timeout)
```
ztorlakTT:
```suggestion
        if self.batch_enabled:
            try:
                return self._leftover_items.get_nowait()
            except Empty:
                pass
        return super().get(block=block, timeout=timeout)
```

ztorlakTT (approved):

Issue #1395, opened 2025-12-10
Img2Img not working


fivanovicTT:
Currently, there are multiple issues and solutions for Img2Img implementation:
Example of CI run: https://github.com/tenstorrent/tt-shield/actions/runs/20095317317

model name `stable-diffusion-xl-base-1.0-img-2-img` is not valid

Solution might be to extend in `model_spec.py` for SDXL and Img2Img:
```
weights=[
    "stabilityai/stable-diffusion-xl-base-1.0",
    "stabilityai/stable-diffusion-xl-base-1.0-img-2-img",
],
```

PR #1428, opened 2025-12-12
Fix Img2Img

fivanovicTT:
### Link to GitHub issue
[1395 - Img2Img not working](https://github.com/tenstorrent/tt-inference-server/issues/1395)

### Summary
Img2Img not working


workflows/model_spec.py
```diff
@@ -1925,14 +1936,19 @@ def expand_to_specs(self) -> List["ModelSpec"]:
...
+        ],
         tt_metal_commit="e95ffa5",
         impl=tt_transformers_impl,
         min_disk_gb=15,
         min_ram_gb=6,
         docker_image="ghcr.io/tenstorrent/tt-media-inference-server:0.4.0-e95ffa59adbe39237525161272141cbbb603c686",
         model_type=ModelType.IMAGE,
         inference_engine=InferenceEngine.MEDIA.value,
+        # img2img uses the same weights as base SDXL
+        hf_weights_repo="stabilityai/stable-diffusion-xl-base-1.0",
```
bgoelTT: Can you confirm this is a way to provide a layer of indirection despite using the same model weights? Also allows us to just select the tests that apply to image-gen vs image-image?

bgoelTT: Okay yeah I see now. @tstescoTT what are your thoughts on not having a 1:1 mapping between the `weights` ModelSpecTemplate field and a real HF repo?

tstescoTT: I'm confused, https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0-img-2-img is not a HF repo, why are we labeling it as hf_model_repo?

tstescoTT: I think we need a 2nd level mapping of the model modality or `tasks` as they're called in HF UI:

Image

fivanovicTT: Hey @tstescoTT, https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0-img-2-img is not a HF repo, because `base SDXL` and `Img2Img` are using the same weights => weights of `base SDXL`.

The issue I was facing is that for single weight we have 2 models basically - `base SDXL` and `Img2Img`. If we had only weight of the `base SDXL`, we wouldn't be able to differentiate between `base SDXL` and `Img2Img` in our reporting (because we use weight of the model from `model_spec.py` and use it for reporting throught the tt-inference-server). 

There were 2 options as far as I have seen:
1. Add bunch of `if` statements in the code (kinda 'hack' it) - of which I wasn't really fan of
2. Add optional field in `model_spec.py` which will be default to actual weights, but for our `base SDXL`/`Img2Img` to actually both point to weights of the `base SDXL`, but using the different name: `stable-diffusion-xl-base-1.0-img-2-img` for our reporting stuff

I chose option 2 because it was a way cleaner option and easy to maintain (added comments to document this decision).

Let me know if you want to discuss on the meeting/slack how to implement like HF do:
> I think we need a 2nd level mapping of the model modality or tasks as they're called in HF UI:

PR #1467, opened 2025-12-15
BAAI/bge-large-en-v1.5 model runner

dmadicTT:


tt-media-server/tt_model_runners/vllm_bge_large_en_runner.py
```diff
@@ -0,0 +1,48 @@
...
+        self.llm.embed(prompts)
+        self.logger.info(f"Device {self.device_id}: Model warmup completed")
+
+        return True
+
+    def run_inference(self, requests: list[TextEmbeddingRequest]):
+        self.logger.debug(
+            f"VLLMBGELargeENRunner: Running inference for {len(requests)} requests"
+        )
+        prompts = [req.input for req in requests]
```
ljovanovicTT: Do we need a verification of `max_num_batched_tokens` for now?

tt-media-server/tt_model_runners/vllm_bge_large_en_runner.py
```diff
@@ -0,0 +1,48 @@
...
+        self.logger.info(f"Device {self.device_id}: Loading model...")
+
+        prompts = [
+            "The capital of France is Paris",
+        ]
+        self.llm = LLM(
+            model="BAAI/bge-large-en-v1.5",
+            max_model_len=384,
+            max_num_seqs=8,
+            max_num_batched_tokens=3072,
```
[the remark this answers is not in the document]
dmadicTT: Unfortunately, we can't. They are specific to the model.
Should we move these values to `vllm_forge_qwen_embedding_runner.py`?

tt-media-server/tt_model_runners/vllm_bge_large_en_runner.py
ljovanovicTT: In `VLLMForgeEmbeddingQwenRunner` we use a Tokenizer to check whether the number of tokens in the requests is less than `max_num_batched_tokens` - but in this case it might make sense to leave this up to vLLM

dmadicTT: Ahaaa :)
We can leave this to vLLM. It has it's own scheduler that considers this.

PR #1505, opened 2025-12-17
CPU worker dealay in s

idjuricTT:


ljovanovicTT (approved):

tt-media-server/model_services/cpu_workload_handler.py
```diff
@@ -114,7 +114,7 @@ def _start_workers(self):
             worker.start()
             self.workers.append(worker)
             self.logger.info(f"Started {self.name} worker {i} with PID {worker.pid}")
-            time.sleep(settings.new_device_delay_seconds)
+            time.sleep(settings.new_runner_delay_seconds)
```
dmadicTT: side question, should we use `await asyncio.sleep` to not block the whole process?

idjuricTT: for this one it's on purpose since we want those to start in sequence and block each other

PR #1519, opened 2025-12-18
Add has_builtin_warmup field to model specs and auto-control trace ca…

### LINKED ISSUE #1471: Fix benchmarking trace capture handling for new internal vLLM trace capture
Enable users to run immediate benchmarks on starting server, e.g.:
```
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow benchmarks --docker-server --dev-mode
```
- this needs to wait until internal vLLM model impl finishes trace capture so that initial perf measurements do not include waiting for trace capture or program compilation to finish.

Currently users need to run:
```
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow release --docker-server --dev-mode
# or
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow server --docker-server --dev-mode
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow evals
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow benchmarks
# or 
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow server --docker-server --dev-mode
sleep 900
run.py --model Llama-3.3-70B-Instruct --device t3k --workflow benchmarks
```

stisiTT:
### Problem

Benchmark workflows could start before trace capture finished, skewing initial performance measurements. Users had to manually disable trace capture or add delays. Additionally, models with builtin warmup (which already capture traces during initialization) were redundantly capturing traces client-side.

### Solution

Add automatic trace capture control based on whether a model has builtin warmup, while respecting explicit user overrides.

### Changes

1. **Added `has_builtin_warmup` attribute** to `ModelSpecTemplate` and `ModelSpec` (defaults to `True`)
   - Models with fully implemented `warmup_model_prefill()` that capture traces: `has_builtin_warmup=True`
   - Models with stub/missing warmup: `has_builtin_warmup=False`

2. **Set `has_builtin_warmup=False` for Qwen2.5-VL models** (3B, 7B, 32B, 72B)
   - These models have stub warmup implementations that don't capture traces

3. **Automatic trace capture control in benchmarks workflow**
   - Models with `has_builtin_warmup=True` → automatically disable client-side trace capture
   - Models with `has_builtin_warmup=False` → automatically enable client-side trace capture
   - User's explicit `--disable-trace-capture` flag always takes precedence

### Behavior

- **Without `--disable-trace-capture`**: Automatic behavior based on `has_builtin_warmup`
- **With `--disable-trace-capture`**: User's choice is respected regardless of `has_builtin_warmup`

### Impact

- Eliminates redundant trace capture for models with builtin warmup
- Ensures trace capture for models without builtin warmup
- Enables immediate benchmarks: `run.py --workflow benchmarks --docker-server` works without manual delays
- Maintains backward compatibility: explicit flags are always respected

Fixes #1471

### DIFF
--- a/benchmarking/run_benchmarks.py
+++ b/benchmarking/run_benchmarks.py
@@ -174,22 +174,15 @@ def main():
     service_port = cli_args.get("service_port", os.getenv("SERVICE_PORT", "8000"))
     disable_trace_capture = cli_args.get("disable_trace_capture", False)
 
-    # Check if user explicitly set --disable-trace-capture flag
-    # For store_true flags, True means user explicitly passed the flag
-    user_explicitly_set = disable_trace_capture
-
     # Automatically control trace capture based on has_builtin_warmup
     # Only apply automatic logic if user hasn't explicitly set --disable-trace-capture
-    if not user_explicitly_set and hasattr(model_spec, "has_builtin_warmup"):
+    if not disable_trace_capture and hasattr(model_spec, "has_builtin_warmup"):
         if model_spec.has_builtin_warmup:
             # Model has builtin warmup - disable client-side trace capture
             disable_trace_capture = True
-        else:
-            # Model doesn't have builtin warmup - enable client-side trace capture
-            disable_trace_capture = False
             logger.info(
-                "Model does not have builtin warmup (has_builtin_warmup=False), "
-                "automatically enabling trace capture for benchmarks workflow"
+                "Model has builtin warmup (has_builtin_warmup=True), "
+                "automatically disabling trace capture for benchmarks workflow"
             )
 
     device = DeviceTypes.from_string(device_str)


benchmarking/run_benchmarks.py
```diff
@@ -174,6 +174,32 @@ def main():
...
+    # Automatically control trace capture based on has_builtin_warmup
+    # Only apply automatic logic if user hasn't explicitly set --disable-trace-capture
+    if not user_explicitly_set and hasattr(model_spec, "has_builtin_warmup"):
+        if model_spec.has_builtin_warmup:
+            # Model has builtin warmup - disable client-side trace capture
+            disable_trace_capture = True
+            logger.info(
+                "Model has builtin warmup (has_builtin_warmup=True), "
+                "automatically disabling trace capture for benchmarks workflow"
+            )
```
bgoelTT: Because the new preference is to have trace capture be opt-in, we shouldn't log here

workflows/model_spec.py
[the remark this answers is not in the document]
tstescoTT: yes, I recommend setting `has_builtin_warmup=False` for all except the latest released model impls, and then as we add new model impls to newer releases we flip them to `has_builtin_warmup=True` if the model does have the builtin warmup implemented in tt-metal (e.g. tt-transformers) and vLLM or media server.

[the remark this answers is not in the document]
bgoelTT: 👍 @stisiTT can we flip the default and then update these ModelSpecs:
* Llama-70B on Galaxy
* Llama-8B on Galaxy
* Qwen3-32B on Galaxy

The media server models already handle the model warmup by their respective clients

benchmarking/run_benchmarks.py
```diff
@@ -174,6 +174,24 @@ def main():
     service_port = cli_args.get("service_port", os.getenv("SERVICE_PORT", "8000"))
     disable_trace_capture = cli_args.get("disable_trace_capture", False)
 
+    # Check if user explicitly set --disable-trace-capture flag
+    # For store_true flags, True means user explicitly passed the flag
+    user_explicitly_set = disable_trace_capture
```
bgoelTT: Why not just use `disable_trace_capture` as the variable instead of creating an identity assignment?

benchmarking/run_benchmarks.py
```diff
@@ -174,6 +174,24 @@ def main():
...
+
+    # Automatically control trace capture based on has_builtin_warmup
+    # Only apply automatic logic if user hasn't explicitly set --disable-trace-capture
+    if not user_explicitly_set and hasattr(model_spec, "has_builtin_warmup"):
+        if model_spec.has_builtin_warmup:
+            # Model has builtin warmup - disable client-side trace capture
+            disable_trace_capture = True
```
bgoelTT: This is where the logging should occur now. Pretty much we should log when we override the default behaviour of `disable_trace_capture`

bgoelTT (approved):
Thanks Sam this looks a lot better now from the CLI perspective.

PR #1556, opened 2025-12-23
Set model source based on inference engine instead of env variable

mjeremicTT:
This PR introduces changes related to adding default values for model source based on inference engine value, removing MODEL_SOURCE environment variable dependency when running run.py.


mjeremicTT:
We will still rely on our CI and setting MODEL_SOURCE, so closing this for now.

PR #1602, opened 2025-12-25
TPOT easy wins

idjuricTT:
Small changes to address TPOT:

- Do not create logs during streaming
- Do not do string concatenation during streaming
- Avoid counting chunks during streaming 
- Minor changes to test runner to make it more precise

### DIFF
--- a/tt-media-server/model_services/device_worker.py
+++ b/tt-media-server/model_services/device_worker.py
@@ -163,17 +163,13 @@ def device_worker(
                             result_generator = await device_runner._run_inference_async(
                                 [inference_request]
                             )
-                            chunk_count = 0
 
+                            chunk_key = inference_request._task_id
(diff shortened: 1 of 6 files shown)


tt-media-server/model_services/device_worker.py
```diff
@@ -163,17 +163,13 @@ async def handle_streaming():
                             result_generator = await device_runner._run_inference_async(
                                 [inference_request]
                             )
-                            chunk_count = 0
 
+                            chunk_key = inference_request._task_id
```
dmadicTT: How can we afford not having the `chunk_count` in the key anymore?
Is it because we migrated away from the future map to queues? The migration from @ztorlakTT 

Did this negatively impact performance or it's more a chore?

ztorlakTT (approved):

PR #1622, opened 2025-12-26
Added new models mochi1 and wan2.2 to modelspec

mdobrosavljevicTT:


idjuricTT (approved):

workflows/model_spec.py
```diff
@@ -2460,6 +2461,68 @@ def expand_to_specs(self) -> List["ModelSpec"]:
...
+        device_model_specs=[
+            DeviceModelSpec(
+                device=DeviceTypes.P150X4,
+                max_concurrency=32,
+                max_context=128 * 1024,
+                default_impl=True,
+            ),
+            DeviceModelSpec(
+                device=DeviceTypes.GALAXY,
+                max_concurrency=32,
```
idjuricTT: Same here

idjuricTT (approved):

ztorlakTT (approved):

workflows/model_spec.py
```diff
@@ -2460,6 +2461,68 @@ def expand_to_specs(self) -> List["ModelSpec"]:
             ),
         ],
     ),
+    ModelSpecTemplate(
+        weights=["wan2.2"],
+        tt_metal_commit="d9c663b",
+        impl=tt_transformers_impl,
```
bgoelTT: I know this is the de-facto `impl` value for media server models in model_spec.py but we need to break this pattern.

Can we not make new `ImplSpec`s for each of the media server model implementations that don't fall into forge-vllm, tt-vllm-plugin, etc?

bgoelTT (changes requested):

PR #1313, opened 2025-12-03
Support LLM streaming

ztorlakTT:
Example:
https://github.com/user-attachments/assets/ab6976cd-3a62-4376-adde-b33860ef0ca5

PR #1624, opened 2025-12-26
Use VLLM Forge runner to serve tt-xla models

ljovanovicTT:


tt-media-server/open_ai_api/llm.py
dmadicTT: These changes are just making it OpenAI compatible endpoint

tt-media-server/config/vllm_settings.py
dmadicTT: Started this way not to blow up the settings file.
What do you think about migrating to this approach? https://github.com/tenstorrent/tt-inference-server/issues/1627

tt-media-server/tt_model_runners/vllm_forge_runner.py
ztorlakTT: You can use RequestOutputKind.FINAL_ONLY when stream=false, similar to what I did here: https://github.com/tenstorrent/tt-inference-server/pull/1313/changes#diff-641955ec81a9c5adc6df539fe8f2b8b56ca9d141836053d0b86275b5c630134aR73-R74.
This way you wouldn't have to iterate through chunks in the non-streaming case.

PR #1628, opened 2025-12-29
Fix Forge Issues

ljovanovicTT:
## Problem
With recent changes, Forge models started failing on CI

### Ci runs:
https://github.com/tenstorrent/tt-shield/actions/runs/20525697558

### DIFF
--- a/tt-media-server/config/constants.py
+++ b/tt-media-server/config/constants.py
@@ -131,8 +131,8 @@ MODEL_RUNNER_TO_MODEL_NAMES_MAP = {
     ModelRunners.TT_MOCHI_1: {ModelNames.MOCHI_1},
     ModelRunners.TT_WAN_2_2: {ModelNames.WAN_2_2},
     ModelRunners.TT_WHISPER: {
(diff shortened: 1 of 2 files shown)


tt-media-server/config/constants.py
```diff
@@ -131,8 +131,8 @@ class ModelServices(Enum):
     ModelRunners.TT_MOCHI_1: {ModelNames.MOCHI_1},
     ModelRunners.TT_WAN_2_2: {ModelNames.WAN_2_2},
     ModelRunners.TT_WHISPER: {
-        ModelNames.DISTIL_WHISPER_LARGE_V3,
         ModelNames.OPENAI_WHISPER_LARGE_V3,
+        ModelNames.DISTIL_WHISPER_LARGE_V3,
```
ljovanovicTT: Another thing to note is that when weights are pre-downloaded, only the first entry in the map is fetched, so since OpenAI Whisper has higher priority, I moved it to be the first item in the list

fivanovicTT (approved):

Issue #1589, opened 2025-12-24
Add input params for CNN endpoint


idjuricTT:
New input params:

- Input format base64/file - probably we can reuse some of audio file input 
- Output format verbose / json - json something like [{"object":"image"}, {"confidence_level": 65,..] - verbose would be just image,65,dog,99
- Top_K - int - i.e. top 3 objects
- Min_confidence level - filter out all below some threshold i.e. 70%
- Multiple images - tbd - if base_64 split into multiple requests

PR #1631, opened 2025-12-29
Add input params for CNN endpoint

fivanovicTT:
### Link to GitHub issue
[1589](https://github.com/tenstorrent/tt-inference-server/issues/1589)

### Summary
New input params:

- Input format base64/file - probably we can reuse some of audio file input 
- Output format verbose / json - json something like [{"object":"image"}, {"confidence_level": 65,..] - verbose would be just image,65,dog,99
- Top_K - int - i.e. top 3 objects
- Min_confidence level - filter out all below some threshold i.e. 70%
- Multiple images - tbd - if base_64 split into multiple requests

### DIFF
--- a/tt-media-server/tt_model_runners/forge_runners/forge_runner.py
+++ b/tt-media-server/tt_model_runners/forge_runners/forge_runner.py
@@ -202,7 +202,9 @@ class ForgeRunner(BaseDeviceRunner):
         raw_predictions = self.loader.output_postprocess(output, top_k=top_k)
         self.logger.info("Convert list of dicts and parse probability strings")
         predictions = []
-        for label, probability in zip(raw_predictions["labels"], raw_predictions["probabilities"]):
+        for label, probability in zip(
+            raw_predictions["labels"], raw_predictions["probabilities"]
+        ):
             prob = float(probability.rstrip("%"))
             if prob >= min_confidence:
                 predictions.append({"label": label, "probability": prob})


tt-media-server/tt_model_runners/forge_runners/forge_runner.py
```diff
@@ -150,3 +179,57 @@ def base64_to_pil_image(self, base64_string, target_mode="RGB"):
             image = image.convert(target_mode)
 
         return image
+
+    def _postprocess_with_filters(
```
dmadicTT: Just an idea: I would split the postprocessing and filtering responsibilities, then filtering is independent and can be easily unit-tested.

tt-media-server/tt_model_runners/forge_runners/forge_runner.py
```diff
@@ -113,17 +114,45 @@ def run(self, image_search_requests: List[ImageSearchRequest]):
...
+            )
+
+            # Format response based on response_format
+            formatted_result = self._format_response(
+                predictions, request.response_format
+            )
+
+            return [formatted_result]
+
+    def _get_pil_image(
```
dmadicTT: What do you think about having a separate class for this?

```python
class ImageDecoder:
    def get_pil_image_from_base64(self, base64_string: str, image_mode: str = "RGB") -> Image.Image:
        ...

    def get_pil_image_from_bytes(self, image_bytes: bytes, image_mode: str = "RGB") -> Image.Image:
        ...
```

tt-media-server/open_ai_api/cnn.py
```diff
@@ -2,24 +2,74 @@
...
+    min_confidence: Optional[float] = Form(70.0),
+) -> ImageSearchRequest:
+    """Parse the image search request from the request body."""
+    content_type = request.headers.get("content-type", "").lower()
+
+    if CONTENT_TYPE_JSON in content_type:
+        json_body = await request.json()
+        return ImageSearchRequest(**json_body)
+
+    if file is not None:
```
dmadicTT: Should we still check is it `multipart/form-data`?

tt-media-server/open_ai_api/cnn.py
```diff
@@ -2,24 +2,74 @@
...
+        file_content = await file.read()
+
+        return ImageSearchRequest(
+            prompt=file_content,
+            response_format=response_format or ImageSearchResponseFormat.JSON.value,
```
dmadicTT: As it already has default value I would not re-define `or ImageSearchResponseFormat.JSON.value`

tt-media-server/domain/image_search_request.py
```diff
@@ -12,19 +13,53 @@
...
 class ImageSearchRequest(BaseRequest):
     # Base64-encoded image
-    prompt: str
+    prompt: Union[str, bytes]
+
+    # Number of top predictions to return
+    top_k: int = 3
+
+    # Response format: "json" or "verbose"
+    response_format: str = "json"
```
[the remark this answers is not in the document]
fivanovicTT: I agree, I was following the convention we have for other models as well ([audio for example](https://github.com/tenstorrent/tt-inference-server/blob/fe00b075425620fdc35a1740e9cd2a4c7dd756a8/tt-media-server/domain/audio_processing_request.py#L19))

Maybe we should revisit this and update all of the endpoints accordingly?

tt-media-server/tt_model_runners/forge_runners/forge_runner.py
fivanovicTT: I actually renamed method name to: `_postprocess_model_output` it makes more sense actually. Basically, there is no filtering - I just pass these arguments as parameters downstream i extract data and that's it

tt-media-server/tt_model_runners/forge_runners/forge_runner.py
fivanovicTT: I agree, but not sure is it worth it to create a separate class etc just for 8 lines of code 🤔?

tt-media-server/domain/image_search_request.py
```diff
@@ -4,27 +4,63 @@
...
             base64.b64decode(v)
             return v
         except Exception as e:
             raise ValueError(f"Invalid base64 image: {e}")
 
+    @field_validator("top_k")
+    @classmethod
+    def validate_top_k(cls, v):
+        if v < 1:
+            raise ValueError("top_k must be at least 1")
```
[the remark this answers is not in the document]
idjuricTT: Sample:

Optional[int] = Field(default=20, ge=12, le=50)

tt-media-server/domain/image_search_request.py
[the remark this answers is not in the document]
fivanovicTT: I agree. Good catch, we should really refactor this part of code for all endpoint

idjuricTT (approved):

ztorlakTT (approved):

tt-media-server/tt_model_runners/forge_runners/forge_runner.py
```diff
@@ -150,3 +181,103 @@ def base64_to_pil_image(self, base64_string, target_mode="RGB"):
...
+
+        self.logger.info("Convert list of dicts and parse probability strings")
+        predictions = []
+        for label, probability in zip(labels, probabilities):
+            prob = float(probability.rstrip("%"))
+            predictions.append({"label": label, "probability": prob})
+
+        return predictions
+
+    def _legacy_postprocess(
```
vmilosevic: I wouldn't add code to handle legacy loader interface, lets keep it simple and use model loader from branch until its merged to main
```
FORGE_MODELS_SHA="copilot/add-top-k-argument-to-postprocess
```
Like in https://github.com/tenstorrent/tt-inference-server/pull/1653/changes#diff-09dc2b241af1611459f9b20d946acd841dad9243db2abf0242e6be4a5165cc1e

Issue #1561, opened 2025-12-23
Add CANCELLED (and CANCELLING) Job State


ztorlakTT:
Currently, when `cancel_job()` is called, the job is immediately deleted from memory and database. For fine-tuning jobs, we should preserve job metadata and allow users to see that a job was cancelled (not just disappeared).

PR #1638, opened 2025-12-30
Add persistence layer to store the data about jobs in an SQLite database

### LINKED ISSUE #1560: Store Job Results in Database
Currently, job results are only stored in memory and in files. We need to persist job results (model checkpoints, metrics, logs) to the database for durability and easier retrieval.

mcupacTT:
Closes #1560 and #1561 

### Description
This PR adds an option to store data from the job manager into an SQLite database.

### Changes and Motivation
- CANCELLED and CANCELLING states are integrated
    - Workflow: If job is _cancelled_ - it first goes to the _cancelling_ state, and upon successful cancellation it is marked as _cancelled_
    - when cancelled, the job is not deleted neither from the memory nor from the database - but rather marked _cancelled_
- on shutdown, the jobs are removed from the memory, but persisted in the database
    - _in-progress_ or _queued_ jobs are cancelled during shutdown, and upon successful shutdown, marked as cancelled in the database
- `cleanup_old_jobs` is the only place where jobs are deleted from the database as well as from the memory
- on job manager startup, `_restore_jobs_from_db` is run 
    - which fetches jobs data from the database to the memory 
    - in case the shutdown was not graceful - we might encounter jobs which remained as _queued_, _in-progress_, or _cancelling_ on restore
        - if we encounter jobs as _queued_ or _in-progress_ - we mark them as _failed_ on restore
        - if we encounter _cancelling_ jobs - we mark them as _cancelled_ on restore
- the `result` field of `Job` class is renamed to `result_path` to enforce that only paths to the files can be saved there
    -  the reason: to prevent database performance issues related to large file storage (images, model weights) and to provide a unified interface for all job types, including inference and fine-tuning
    - the tests are updated accordingly
    - the error is raised if non-string type is saved to `result_path` field
- the `object` field of Job class is renamed to `job_type` for clarity
- the sqlite db file location is configured in `settings.py`

### Tests
- For the tests, there were two options: to create a separate test suite with separate tests for the database scenario or to extend the test_job_manager tests
   - i decided to extend the already existing job_manager test suite to ensure that the same logic is tested at the same place - meaning that on future changes, less tests need to be changed
   - The `job_manager` fixture is parametrized for database and non-database case - meaning that all `job_manager` tests will be run for both scenarios

### Issues to be resolved/discussed in the follow-up PRs
- Database functions need to be called with `asyncio.to_thread` to ensure that the calls are non-blocking
- discuss if the connection to the database should remain open for more transactions and not only one to improve performance
- `os.remove(result_path)` in `cleanup_job` should be wrapped in `asyncio.to_thread` as well - since the communication with the disk may be slow
- discuss if the database calls should be under the job lock similar to writing to `_jobs` dictionary
- add more database-only tests and handle logic - what if db file is corrupted, schema is changed etc
- add a secret to the database connection
- update api side according to the current logic
- add tests to check if the database insertion or updates fails, if the workflow continues as desired/or raises an error
- add more tests to test race conditions


tt-media-server/config/settings.py
```diff
@@ -69,6 +69,7 @@ class Settings(BaseSettings):
     job_retention_seconds: int = 86400
     job_max_stuck_time_seconds: int = 10800
     enable_job_persistence: bool = False
+    job_database_path: str = "./tt-media-server/jobs.db"
```
idjuricTT: I would rather use something like ./jobs.db since tt-media-server folder doesn't exist in the docker image

tt-media-server/utils/job_manager.py
```diff
@@ -5,15 +5,15 @@
 import asyncio
 import os
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, field
```
idjuricTT: @ztorlakTT could you please double check this file?

tt-media-server/config/settings.py
mcupacTT: changed

tt-media-server/utils/job_manager.py
```diff
@@ -282,49 +327,80 @@ def _cleanup_job(self, job: Job):
...
                     completed_at=db_job.get("completed_at"),
-                    error=db_job.get("error"),
+                    result_path=db_job.get("result_path"),
+                    error=db_job.get("error_message"),
                 )
 
-                if db_job.get("result"):
-                    job.result = db_job["result"]
+                # If job was stuck, mark it as failed or cancelled and sync to database
+                if original_status not in ["completed", "failed", "cancelled"]:
```
```suggestion
                if not job.is_terminal():
```

tt-media-server/utils/job_manager.py
```diff
@@ -282,49 +327,80 @@ def _cleanup_job(self, job: Job):
...
+                    if original_status == "cancelling":
+                        job.mark_cancelled()
+                    else:
+                        job.mark_failed(
+                            "server_restart", "Job interrupted by system restart"
+                        )
+
+                    # override the completed_at time, since it was cancelled due to system restart
+                    # here we override it with the creation time, since we don't know the time of the system restart
```
ztorlakTT: nit:
```suggestion
                    # we override the completed_at time with the creation time, since we don't know the time of the system restart
```

tt-media-server/utils/job_manager.py
```diff
@@ -148,33 +169,39 @@ def get_job_metadata(self, job_id: str) -> Optional[dict]:
                 return job.to_public_dict()
             return None
 
-    def get_job_result(self, job_id: str) -> Optional[Any]:
-        """Get job result if completed."""
+    def get_job_result_path(self, job_id: str) -> Optional[str]:
```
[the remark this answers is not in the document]
mcupacTT: Updated, and also updated the base_service function `get_job_result` to `get_job_result_path` - since it has the same purpose

tt-media-server/utils/job_manager.py
mcupacTT: Rewritten

PR #1669, opened 2026-01-08
Fix version comparison to support prerelease versions

sott0n:
SpecifierSet excludes prereleases by default, causing versions like '2.6.0-rc1' to fail matching against '>=2.3.0'. Use contains() with `prereleases=True` to correctly handle rc, alpha, beta, and dev versions.

### DIFF
--- a/workflows/run_local_setup_validation.py
+++ b/workflows/run_local_setup_validation.py
@@ -8,7 +8,6 @@ import json
 import logging
 import os
 from packaging.specifiers import SpecifierSet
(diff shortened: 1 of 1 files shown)


sott0n:
When I fork the repository and create a PR, the Actions that are supposed to post comments on the PR fail due to insufficient permissions.

```
      message: 'Resource not accessible by integration',
      documentation_url: 'https://docs.github.com/rest/issues/comments#create-an-issue-comment',
      status: '403'
```

PR #1670, opened 2026-01-08
Add --skip-system-sw-validation hint to version check error messages 

sott0n:
UPDATED
based on https://github.com/tenstorrent/tt-inference-server/pull/1670#pullrequestreview-3642349193
For the validation cause an error by given a prerelease version, modify the error message to recommend using `--skip-system-sw-validation` flag.

---
SpecifierSet excludes prereleases by default, causing versions like '2.6.0-rc1' to fail matching against '>=2.3.0'. Use contains() with prereleases=True to correctly handle rc, alpha, beta, and dev versions.

currently failure is
```
File "/home/ttuser/.local/lib/tt-inference-server/workflows/run_local_setup_validation.py", line 287, in <module>
main()
File "/home/ttuser/.local/lib/tt-inference-server/workflows/run_local_setup_validation.py", line 273, in main
_enforce_requirement("KMD", kmd_version, kmd_requirement)
File "/home/ttuser/.local/lib/tt-inference-server/workflows/run_local_setup_validation.py", line 256, in _enforce_requirement
raise ValueError(message)
ValueError: KMD version '2.6.0-rc1' does not satisfy the STRICT requirement '>=2.3.0'.
```

### DIFF
--- a/workflows/run_local_setup_validation.py
+++ b/workflows/run_local_setup_validation.py
@@ -248,10 +248,23 @@ def main():
         ) -> None:
             version_specifier = version_requirement.specifier
             enforcement_mode = version_requirement.mode
-            meets_requirement = Version(version) in SpecifierSet(version_specifier)
+            parsed_version = Version(version)
+            meets_requirement = parsed_version in SpecifierSet(version_specifier)
             if meets_requirement:
                 return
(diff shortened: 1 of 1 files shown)


sott0n:
@bgoelTT Make sense, Thanks for explanation. I'll update to modify error message as you mentioned.

Issue #1253, opened 2025-11-26
Simplify container image to match standard vLLM patterns - Remove JSON config dependency and mandatory custom environment variables and host volume mounts


nidhinjose-git:
We deploy TT inference images using Kubernetes rather than through the `tt-inference-server run.py` script. This means we must expose host volume mounts, model spec JSON file, and TT-specific environment variable configuration through our Kubernetes orchestration layer - tasks that are otherwise automated by the `tt-inference-server` `run.py`.

Currently, we serve Nvidia and AMD inference workloads in production. For these deployments, our Kubernetes pods start without any host volume mounts and require only standard `VLLM_*` environment variables. Model weights are downloaded fresh for each inference session, and we don't persist vLLM cache to storage (though this may change in the future). 

We would like Tenstorrent container images to have fewer deployment constraints and work similarly to Nvidia/AMD images for seamless Kubernetes integration.

**Current behavior**: Requires model_spec JSON file on host + volume mounts for metal cache and model weights + TT env vars.

**Required behavior**: Remove mandatory host volume mounts, embed JSON spec inside container image and remove (or make it optional) as many TT specific env variables as possible.

#### 1. **CACHE_ROOT**  - Make it Optional  
**Current**: Required host volume mount for general cache directory
**Proposed**: Make it optional for stateless deployments where we don't use persistent general cache

#### 2. **TT_MODEL_SPEC_JSON_PATH** - Embed Model Specs in Container Image
**Current**: Mandatory - requires model spec JSON file on host and mount to container.
**Proposed**:  Make it optional. When this environment variable becomes optional, the complete model specs JSON file should be embedded within the container image at a default location. This raises the question of how to properly identify the entry within the JSON file from the running container image. A model spec is identified as `id_{impl_name}_{model_name}_{device}`. The implementation can be assumed to be `"tt-transformers"` as default. The `HF_MODEL` environment variable can contain the model name, and a new environment variable might be required to pass the device type.  `TT_MODEL_SPEC_JSON_PATH` can still be used to override the embedded specs when custom JSON files are needed.

#### 3. **MODEL_WEIGHTS_PATH** - Make it Optional  
**Current**: Required host volume mount for pre-downloaded model weights
**Proposed**: 
- Make optional for deployments that prefer vLLM to download weights on startup
- When not specified, vLLM handles model downloading from HuggingFace
- Acceptable trade-off: longer startup time vs. no persistent storage requirement

[...]

PR #1685, opened 2026-01-09
Update Qwen3 Embedding 4B Model Spec Environment Variables

ljovanovicTT:
## Problem
This line was causing issues by setting an incorrect mesh grid shape. Since this line does not give us control over what is being configured, therefore making it harder to spot possible issues, we reverted to explicitly setting only the necessary environment variables
```
**model_spec.device_model_spec.env_vars,
```
## Implementation
- Explicitly set environment variables when running Docker
- Update environment variables for `Qwen3 Embedding 4B` model
- Update environment variable name used in benchmarking script in `EmbeddignStrategy`

## Testing
Successful CI runs:
https://github.com/tenstorrent/tt-shield/actions/runs/20853567630

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -121,6 +121,8 @@ def get_embedding_docker_env_vars(model_spec, args):
         "MODEL": model_spec.model_name,
         "DEVICE": model_spec.device_type.name.lower(),
         "DEVICE_IDS": device_ids_str,
+        # TODO: Remove these VLLM explicit parameters
+        # https://github.com/tenstorrent/tt-inference-server/issues/1253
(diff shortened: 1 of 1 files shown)


workflows/run_docker_server.py
```diff
@@ -121,7 +121,15 @@ def get_embedding_docker_env_vars(model_spec, args):
         "MODEL": model_spec.model_name,
         "DEVICE": model_spec.device_type.name.lower(),
         "DEVICE_IDS": device_ids_str,
-        **model_spec.device_model_spec.env_vars,
```
bgoelTT: Please leave a TODO comment referring to https://github.com/tenstorrent/tt-inference-server/issues/1253

Issue #1503, opened 2025-12-17
Add spec tests for BAAI/bge-large-en-v1.5


dmadicTT:
Reference https://github.com/tenstorrent/tt-inference-server/pull/1390/commits/63ac9c2d34572dede5f7670e1783976aa0721f29

bgoelTT:
@dmadicTT @ljovanovicTT can you please confirm this issue can be closed?

PR #1687, opened 2026-01-09
Add BGE-Large Spec Tests

ljovanovicTT:
## Issue
[#1503 Add Spec Tests for BAAI/bge-large-en-v1.5](https://github.com/tenstorrent/tt-inference-server/issues/1503)

## Problem
We need to add spec tests `BAAI/bge-large-en-v1.5` for standardized testing
Since dimensions param is not supported on `BAAI/bge-large-en-v1.5`, we are not enabling `EmbeddingParamTest` for this model.

## Implementation
- Add `EmbeddingLoadTest` 
- Add `EmbeddingParamTest`
- Add `TestEmbeddingClientStrategy`
- Minor fix in report generation to determine target value
- Enable tests in `server_tests_config.json`

## Testing
Successful CI runs:
https://github.com/tenstorrent/tt-shield/actions/runs/20959310827

### DIFF
--- a/tests/server_tests/test_cases/embedding_load_test.py
+++ b/tests/server_tests/test_cases/embedding_load_test.py
@@ -95,4 +95,6 @@ class EmbeddingLoadTest(BaseTest):
         logger.info(
             f"\n🚀 Avg time for {batch_size} concurrent requests: {avg_duration:.2f}s"
         )
-        logger.info(f"🚀 Avg time for {batch_size} concurrent requests: {avg_duration:.2f}s")
+        logger.info(
+            f"🚀 Avg time for {batch_size} concurrent requests: {avg_duration:.2f}s"
+        )


tests/server_tests/test_cases/embedding_load_test.py
```diff
@@ -0,0 +1,98 @@
...
+                tasks = [timed_request(session, i + 1) for i in range(batch_size)]
+                results = await asyncio.gather(*tasks)
+                requests_duration = max(results)
+                total_duration = sum(results)
+                avg_duration = total_duration / batch_size
+                return requests_duration, avg_duration
+            if iteration == 0:
+                print("🔥 Warm up run done.")
+
+        print(f"\n🚀 Time taken for individual concurrent requests : {results}")
```
fivanovicTT: Nit: I would recommend to use `logger` instead of print, just to be the same across code base

ljovanovicTT: Great catch, thank you for noticing! This is fixed with the latest commit

PR #1718, opened 2026-01-13
Align text-to-speech model with the latest dev

ztorlakTT:
## Changes

   - Changed from `/tts` to `/audio/speech` to match OpenAI's `/v1/audio/speech` format
   - Updated router registration in `open_ai_api/__init__.py` to use `/audio` prefix
   - Improved naming consistency with other API modules
   - Removed function that handled both `multipart/form-data` as it wasn't needed
   - Now it uses FastAPI's automatic JSON body parsing via Pydantic model
   - Removed streaming functionality as it is not needed
   - Removed response format handling (no raw audio option which wasn't working)
   - Always returns JSON with base64-encoded audio

There are two issues that still need to be addressed: 
https://github.com/tenstorrent/tt-inference-server/issues/1719
https://github.com/tenstorrent/tt-inference-server/issues/1720


tests/server_tests/test_cases/tts_integration_test.py
```diff
@@ -102,22 +99,22 @@ def test_valid_auth_token(self):
...
-            json={"text": "Test", "stream": False},
+            json={"text": "Test"},
         )
         # Should get 200 or a processing error, not auth error
         assert response.status_code not in [401, 403]
 
 
-class TestTTSNonStreaming:
-    """Test non-streaming TTS generation"""
+class TestTTS:
```
idjuricTT: Can we make this extend base test so we can run it in the framework?

idjuricTT (approved):

PR #1731, opened 2026-01-14
Include reports from tests workflow in release output

mjeremicTT:
This PR introduces changes related to adding reports from tests workflow to release output folder in workflow_logs directory.


bgoelTT (approved):
Models CI artefacts look good!

Issue #1658, opened 2026-01-05
Add support for vLLM bench tool (server, latency, throughput options)


tstescoTT:
`vllm bench` CLI is the new default benchmarking method vLLM supports.
see docs: https://docs.vllm.ai/en/latest/benchmarking/cli/?h=benchma#online-benchmark

vLLM bench tool code from https://github.com/vllm-project/vllm/tree/main/benchmarks

tstescoTT:
addressed in https://github.com/tenstorrent/tt-inference-server/pull/1802

PR #1742, opened 2026-01-14
add tokenize detokenize endpoints

knovokmetTT:
Our current benchmarks use a forked version of vllm, that will expect tokenization/detokenization to be server sided. We currently do not have the expected endpoints in our media server
This PR creates the endpoints our vllm fork expects to be there, implementing them the same way the actual vllm does

### DIFF
--- a/tt-media-server/tests/test_tokenizer.py
+++ b/tt-media-server/tests/test_tokenizer.py
@@ -63,24 +63,49 @@ class TestTokenizeEndpoint:
     """Tests for tokenize endpoint."""
 
     @pytest.fixture
-    def mock_tokenizer(self):
-        """Create a mock tokenizer."""
+    def mock_llama_tokenizer(self):
+        """Create a mock Llama tokenizer with real token values."""
+        # Real values from Llama 3.2 3B for "Hello world"
+        # with special tokens: [128000, 9906, 1917]
+        # without special tokens: [9906, 1917]
+        # token strs: ['<|begin_of_text|>', 'Hello', 'Ġworld']
         tokenizer = MagicMock()
-        tokenizer.encode = MagicMock(return_value=[1, 2, 3, 4, 5])
-        tokenizer.model_max_length = 8192
+        tokenizer.encode = MagicMock(return_value=[128000, 9906, 1917])
+        tokenizer.model_max_length = 131072
         tokenizer.convert_ids_to_tokens = MagicMock(
-            return_value=["<s>", "Hello", "world", "</s>", "<pad>"]
+            return_value=["<|begin_of_text|>", "Hello", "Ġworld"]
         )
         return tokenizer
 
     @pytest.fixture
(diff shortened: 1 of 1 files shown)


tt-media-server/domain/tokenize_response.py
```diff
@@ -0,0 +1,7 @@
+from pydantic import BaseModel
```
idjuricTT: Missing SPDX header

tt-media-server/domain/detokenize_request.py
```diff
@@ -0,0 +1,6 @@
+from domain.base_request import BaseRequest
+
+
+class DetokenizeRequest(BaseRequest):
+    model: str | None = None
```
idjuricTT: Suggestion: Add a default one to settings and use that one if value is none, it'll save us enum check since most likely we have one model at the time

tt-media-server/resolver/tokenizer_resolver.py
```diff
@@ -0,0 +1,20 @@
+from transformers import AutoTokenizer
+from config.constants import SupportedModels
+from utils.logger import TTLogger
+import threading
+
+tokenizer_cache = {}
+tokenizer_cache_lock = threading.Lock()
```
dmadicTT: What is the motive to have tokenizer cache?

tt-media-server/open_ai_api/tokenizer.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Detokenize the tokens
+        text = tokenizer.decode(request.tokens, skip_special_tokens=False)
+
+        return DetokenizeResponse(prompt=text)
+    except HTTPException:
+        raise
+    except Exception as e:
+        logger.error(f"Error detokenizing tokens: {e}")
+        raise HTTPException(status_code=500, detail=f"Detokenization failed: {str(e)}")
```
dmadicTT: Minor for our case, but good to align:
While it's fine to log the stack trace or the exception, usually you don't return it to the user, as it exposes implementation details

tt-media-server/open_ai_api/tokenizer.py
```diff
@@ -0,0 +1,81 @@
...
+    except HTTPException:
+        raise
+    except Exception as e:
+        logger.error(f"Error tokenizing text: {e}")
+        raise HTTPException(status_code=500, detail=f"Tokenization failed: {str(e)}")
+
+
+@router.post("/detokenize")
+def detokenize(request: DetokenizeRequest) -> DetokenizeResponse:
+    """ """
```
dmadicTT: You can remove those empty comments

bgoelTT:
@knovokmetTT thanks for this PR! We actually want to remove this endpoint requirement from the LLM benchmarking workflow with #1658. If you want to expose this from media-server for other reasons that's great, but for benchmarking we want to move towards removing this requirement.

ljovanovicTT (approved):

PR #1754, opened 2026-01-15
Implement Gemma Trainer with SST2 dataset to demonstrate the functionality of training on tt-inference-server

### LINKED ISSUE #1614: Implement first fine-tuning experiment to demonstrate the functionality of the whole process
Currently, the jobs are scheduled, but no real training/fine-tuning is executed. We need to implement first fine-tuning experiment

mcupacTT:
Closes #1614. 

This pull request introduces a new, fully implemented training workflow for LoRA fine-tuning of the Gemma 1.1 2B model, along with improved dataset configuration and job management for training tasks. It replaces the previous placeholder LoRA trainer with a production-ready runner, adds dataset configuration options, and ensures training jobs are properly created and tested.

**Key changes:**

### Training Workflow Overhaul

- Added a new `TrainingGemmaLoraRunner` in `tt_model_runners/forge_training_runners/training_gemma_lora_runner.py` 
- _TrainingGemmaLoraRunner code is mostly adopted from tt-blacksmith repository_

### Dataset Configuration

- Introduced new environment variables `DATASET_LOADER` and `DATASET_MAX_LENGTH` for dataset selection, documented in `README.md`.
- Implemented SST2 dataset download from Hugging Face and preprocessing
- _SSTDataset code is mostly adopted from tt-blacksmith repository_

### Training Job Management

- Updated `TrainingRequest` to include all necessary hyperparameters - and stored other releveant attributes as private attributes into it
- Implemented logic in `TrainingService.create_job` to set dataset loader, max length, and output model path, and to create output directories as needed.
    - The dataset related data are stored in the database through `request` attribute
- For `TRAINING` jobs, the output path is set already at the beginning comparing to `VIDEO` jobs where it is set after the runner finishes. 
    - The reason: TRAINING jobs have longer duration - we should allow that user can request current state of the training model even before the job finishes _(this logic will be implemented in the follow-up PR)_

### Dependency Updates

- Updated `tt-forge` to version 0.8.0 
- Added `datasets` and `peft` as dependencies for model training in  `forge-runners/requirements.txt`.

### Loss curve of the example fine-tuning with default hyperparameters
plots


tt-media-server/model_services/training_service.py
```diff
@@ -1,10 +1,40 @@
...
+        if job_type != JobTypes.TRAINING:
+            raise ValueError(
+                "The job type must be TRAINING, since the chosen model service is TrainingService"
+            )
+        # check that dataset loader is set
+        if not self.settings.dataset_loader:
+            raise ValueError("The dataset loader must be set")
+
+        request._dataset_loader = self.settings.dataset_loader
+        request._dataset_max_length = self.settings.dataset_max_length
```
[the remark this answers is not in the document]
mcupacTT: That is now a good discussion and I will list the tradeoffs here, so you can tell me your opinion which I will implement then
- My idea of putting it here was that the dataset information is already available in the warmup function - so that we can download and preprocess the dataset before the run method 
     - meaning that run method does not have to wait on this, and that user can submit multiple requests with the same dataset settings without needing to wait on it 
- On the other hand, it makes sense that the user should submit the dataset of choice in the same api request - and that maybe they would change dataset_max_length in multiple api calls
    - and also we have less env variables, since dataset info is only required for training job types

CC @vladimirjovanovicTT to loop in for this discussion

tt-media-server/config/settings.py
```diff
@@ -51,6 +52,10 @@ class Settings(BaseSettings):
     trace_region_size: int = 34541598
     download_weights_from_service: bool = True
 
+    # Dataset settings
+    dataset_loader: str = DatasetLoaders.SST2.value
+    dataset_max_length: int = 32
```
vladimirjovanovicTT: nit: dataset_max_sequence_length might be a better name

tt-media-server/tests/test_job_manager.py
```diff
@@ -573,6 +573,32 @@ async def task_func(req):
         result_path = job_manager.get_job_result_path("job-123")
         assert result_path is None
 
+    @pytest.mark.asyncio
+    async def test_get_job_result_path_set_on_creation(self, job_manager, mock_request):
```
vladimirjovanovicTT: These tests seem to be related to video - let's modify towards training workloads.

tt-media-server/tt_model_runners/forge_training_runners/training_gemma_lora_runner.py
```diff
@@ -0,0 +1,225 @@
...
+
+        global_step = 0
+        running_loss = 0.0
+        self.model.train()
+        try:
+            for epoch in range(request.num_epochs):
+                for batch in tqdm(self.train_dataloader):
+                    self.optimizer.zero_grad()
+
+                    # batch = device_manager.prepare_batch(batch)
```
vladimirjovanovicTT: I guess we don't need this.

tt-media-server/tt_model_runners/forge_training_runners/training_gemma_lora_runner.py
```diff
@@ -0,0 +1,225 @@
...
+                    batch["labels"].view(-1),
+                )
+                total_val_loss += loss.item()
+                # predictions = shift_logits.argmax(dim=-1)
+
+                torch_xla.sync(wait=True)
+
+                num_val_batches += 1
+
+                # TODO: add option to print examples from predictions
```
vladimirjovanovicTT: Can you link an issue, if this TODO is relevant?

PR #1769, opened 2026-01-15
Add system_info data to run_spec

mdobrosavljevicTT:
Added system info (software versions and topology) to run_spec.


acvejicTT (approved):
LGTM

PR #1776, opened 2026-01-16
Text-to-Speech API should return raw WAV audio instead of base64-encoded JSON

vpetrovicTT:
### Link to GitHub issue
[1720](https://github.com/tenstorrent/tt-inference-server/issues/1720)

### Summary

Added support for direct WAV file responses from the TTS endpoint, alongside the existing JSON option.

 Changes

1. Added ResponseFormat.AUDIO enum
2. Added response_format field to TextToSpeechRequest (default: "audio")
3. Implemented post_process in TextToSpeechService with CpuWorkloadHandler for base64 → WAV decoding
4. Refactored endpoint with handle_tts_request pattern
5. Support for multiple response formats:
6. "audio" or "wav" → direct WAV file (Content-Type: audio/wav)
7. "verbose_json" or "json" → JSON with base64 audio

### DIFF
--- a/tt-media-server/tests/test_tokenizer.py
+++ b/tt-media-server/tests/test_tokenizer.py
@@ -350,7 +350,9 @@ class TestDetokenizeEndpoint:
         assert exc_info.value.detail == "Detokenization failed"
 
     @patch("open_ai_api.tokenizer.logger")
-    def test_detokenize_with_tokenizer_loading_error_raises_exception(self, mock_logger):
+    def test_detokenize_with_tokenizer_loading_error_raises_exception(
+        self, mock_logger
+    ):
         """Test that tokenizer loading errors are handled properly."""
         with patch("open_ai_api.tokenizer.AutoTokenizer") as mock_auto:
             mock_auto.from_pretrained.side_effect = Exception("Loading failed")


tt-media-server/domain/text_to_speech_request.py
```diff
@@ -30,6 +31,10 @@ def validate_text(cls, text):
     )
     speaker_id: Optional[str] = None  # ID for pre-configured speaker embeddings
 
+    # Response format and streaming
+    response_format: str = ResponseFormat.AUDIO.value  # ResponseFormat.AUDIO for WAV bytes, ResponseFormat.VERBOSE_JSON or ResponseFormat.JSON for JSON
```
ztorlakTT: If you check the OpenAI API parameters for the text-to-speech endpoint (https://platform.openai.com/docs/api-reference/audio/createSpeech), it says that `mp3`, `opus`, `aac`, `flac`, `wav`, and `pcm` formats are supported. I think we should stick with that and for now it's fine to only support the wav audio format.

tt-media-server/open_ai_api/text_to_speech.py
```diff
@@ -4,19 +4,52 @@
...
+        If response_format is "audio" or "wav", returns WAV bytes directly.
+        Otherwise, returns JSON with base64-encoded audio.
+    """
+    try:
+        if not tts_request.stream:
+            result = await service.process_request(tts_request)
+            if tts_request.response_format.lower() in ("audio", "wav"):
+                # Return WAV bytes directly
+                if hasattr(result, "_wav_bytes") and result._wav_bytes:
+                    return Response(content=result._wav_bytes, media_type="audio/wav")
```
ztorlakTT: `_wav_bytes` isn't a private field, so you can rename it to `wav_bytes`

tt-media-server/model_services/text_to_speech_service.py
```diff
@@ -2,9 +2,87 @@
...
+
+    Args:
+        worker_context: Worker context (None for TTS, not used)
+        base64_audio: Base64-encoded audio string
+        response_format: Response format ("audio" or "wav" for WAV bytes, otherwise None)
+
+    Returns:
+        WAV bytes if response_format is "audio" or "wav", otherwise None
+    """
+    if response_format and response_format.lower() in ("audio", "wav"):
```
ztorlakTT: You can remove this condition because you have the same check on line 67 and it is better to leave that one there before using the CPU Workload Handler

fivanovicTT (approved):

PR #1815, opened 2026-01-19
feat(tts): Add new TTS client for benchmark and eval

vpetrovicTT:
Implementation of a complete TTS client with support for benchmark and eval workflows, including all necessary metrics (TTFT, RTR, P90/P95) and optimized metric calculations. Unit tests have been added

### DIFF
--- a/tests/utils/media_client/test_tts_client.py
+++ b/tests/utils/media_client/test_tts_client.py
@@ -702,9 +702,9 @@ class TestTtsClientStrategyRunEval(unittest.TestCase):
             "published_score",
             "score",
             "published_score_ref",
-            "accuracy_check",
             "rtr",
-            "wer",
+            "p90_ttft",
+            "p95_ttft",
         ]
         for key in required_keys:
             assert key in eval_result, f"Missing required key: {key}"
@@ -712,7 +712,7 @@ class TestTtsClientStrategyRunEval(unittest.TestCase):
         # Verify calculated averages
         assert eval_result["score"] == 150.0  # TTFT: (100 + 200) / 2 (in ms)
         assert abs(eval_result["rtr"] - 2.5) < 0.001  # (2.0 + 3.0) / 2
-        assert abs(eval_result["wer"] - 0.15) < 0.001  # (0.1 + 0.2) / 2
+        # WER is not calculated in run_eval (calculate_wer=False)
 
     @patch("utils.media_clients.tts_client.AutoTokenizer.from_pretrained")
     def test_run_eval_health_check_failed(self, mock_tokenizer):
@@ -795,7 +795,7 @@ class TestTtsClientStrategyRunBenchmark(unittest.TestCase):
         assert benchmarks["rtr"] == 2.5  # (2.0 + 3.0) / 2
         assert "ttft_p90" in benchmarks
         assert "ttft_p95" in benchmarks
-        assert "accuracy_check" in benchmarks
+        # accuracy_check is calculated in run_reports.py, not in tts_client.py
 
         assert report_data["model"] == "test_model"
         assert report_data["device"] == "test_device"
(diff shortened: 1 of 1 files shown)


utils/media_clients/tts_client.py
```diff
@@ -0,0 +1,715 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
```
fivanovicTT: Nit: now it is 2026 😄

utils/media_clients/tts_client.py
```diff
@@ -0,0 +1,715 @@
...
+from pathlib import Path
+from typing import Optional
+
+# Third-party imports
+import aiohttp
+from transformers import AutoTokenizer
+
+# Local imports
+from .base_strategy_interface import BaseMediaStrategy
+from .test_status import TtsTestStatus
```
fivanovicTT: I think we can remove this comments - Local imports, third-party imports etc

utils/media_clients/tts_client.py
```diff
@@ -0,0 +1,715 @@
...
+                return tts_default
+            else:
+                tts_default = 25
+                logger.info(
+                    f"Using TTS-specific benchmark default: {tts_default} calls (was {base_num_calls})"
+                )
+                return tts_default
+
+        logger.info(f"Using configured num_eval_runs: {base_num_calls} calls")
+        return base_num_calls
```
fivanovicTT: I would recommend using `guard` clauses for this - to have a nice flow which can easily followed (and to avoid having nested if statements). Something like this:
```suggestion
        if base_num_calls == 2 and is_eval:
                tts_default = 12
                logger.info(
                    f"Using TTS-specific eval default: {tts_default} calls (was {base_num_calls})"
                )
                return tts_default
          
       if base_num_calls == 2:
                tts_default = 25
                logger.info(
                    f"Using TTS-specific benchmark default: {tts_default} calls (was {base_num_calls})"
                )
                return tts_default

        logger.info(f"Using configured num_eval_runs: {base_num_calls} calls")
        return base_num_calls
```

utils/media_clients/tts_client.py
```diff
@@ -0,0 +1,715 @@
...
+        p90_ttft, p95_ttft = self._calculate_tail_latency(status_list)
+        logger.info(f"Extracted P90 TTFT: {p90_ttft:.2f}ms, P95 TTFT: {p95_ttft:.2f}ms")
+
+        benchmark_data["model"] = self.model_spec.model_name
+        benchmark_data["device"] = self.device.name
+        benchmark_data["timestamp"] = time.strftime(
+            "%Y-%m-%d %H:%M:%S", time.localtime()
+        )
+        benchmark_data["task_type"] = "text_to_speech"
+        # For eval workflow, all_params is always an object with tasks attribute
```
fivanovicTT: Since we are already in `run_eval` method, this is redundant: `- For eval workflow`

utils/media_clients/tts_client.py
```diff
@@ -0,0 +1,715 @@
...
+            else "PASS"
+            if accuracy_check == 2
+            else "FAIL"
+            if accuracy_check == 3
+            else "UNDEFINED"
+        )
+        logger.info(f"Accuracy Check: {accuracy_check} ({accuracy_status})")
+        logger.info("---------------------------\n")
+
+        logger.info(f"Report generated: {result_filename}")
```
fivanovicTT: This should be done in `summary_report.py` and `run_reports.py`

utils/media_clients/tts_client.py
```diff
@@ -0,0 +1,715 @@
...
+            and hasattr(self.all_params, "tasks")
+            and len(self.all_params.tasks) > 0
+        ):
+            if hasattr(self.all_params.tasks[0], "text"):
+                test_text = self.all_params.tasks[0].text
+            elif hasattr(self.all_params.tasks[0], "task_name"):
+                test_text = self.all_params.tasks[0].task_name
+
+        for i in range(num_calls):
+            if (i + 1) % 5 == 0 or i == 0 or i == num_calls - 1:
```
fivanovicTT: I am not sure why do have this calculation? Can you check how we do this in `audio_clien` for example?

PR #1817, opened 2026-01-19
Remove range download from video models

idjuricTT:


tests/server_tests/test_cases/video_generation_load_test.py
```diff
@@ -0,0 +1,156 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
```
fivanovicTT: Nit: make it 2026 😄

PR #1818, opened 2026-01-19
use model id in eval config

knovokmetTT:


mdobrosavljevicTT (changes requested):

vmaksimovicTT (commented):
Just for clarity, this PR is not meant to make any semantic changes, just code refactoring?

mjeremicTT (commented):
Are you sure your base branch: `ljovanovic/feature_add_forge_llm_model_specs` is correct?

knovokmetTT:
This PR is not actually needed until we start supporting different eval configs based on model impl.

PR #1841, opened 2026-01-21
QuietBox GE (P150X4) Support

stisiTT:
## Summary

Adds P150X4 (BH-QuietBox) hardware support for 3 model families and implements automatic RAM requirement calculation based on device topology.


bgoelTT (changes requested):
This PR should be scoped to just p150x4 changes. If we want to change other product-types' ModelSpecs then we need to run it through Models CI.

In general all ModelSpec changes need Models CI runs

PR #1893, opened 2026-01-26
fix llm streaming, dynamic batching

knovokmetTT:
Fixes a multitude of issues

- aligns memory_queue interface for put method with multiprocessing Queue
- fixes a bug in device_worker_dynamic_batch when handling non streaming requests
- simplifies and unifies vllm_runner's response structure

### DIFF
--- a/tt-media-server/tt_model_runners/base_device_runner.py
+++ b/tt-media-server/tt_model_runners/base_device_runner.py
@@ -17,7 +17,9 @@ class BaseDeviceRunner(ABC):
         self.settings = get_settings()
         self.ttnn_device = None
 
-        set_torch_thread_limits(num_torch_threads)
+        # Skip in main process when runner is only used for download_weights (device_id "-1")
+        if self.device_id != "-1":
+            set_torch_thread_limits(num_torch_threads)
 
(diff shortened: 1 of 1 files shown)


tt-media-server/domain/completion_response.py
```diff
@@ -21,22 +23,8 @@ def to_dict(self) -> Dict[str, Any]:
...
-    task_id: str
-    return_result: bool
-
+class CompletionOutput(TypedDict):
+    """Output from the model runner - used for both streaming and non-streaming."""
 
-# Union type for async generator yield type
-StreamingOutput = Union[StreamingChunkOutput, FinalResultOutput]
+    type: Literal["streaming_chunk", "final_result"]
+    data: CompletionResult
```
knovokmetTT: for llms, only type and data and necessary
data will contain text, type will tell us if the chunk is final or not
task_id is not necessary, as it is passed into the queues anyway, it was redundant info

knovokmetTT:
CI:
- https://github.com/tenstorrent/tt-shield/actions/runs/21513964712
- https://github.com/tenstorrent/tt-shield/actions/runs/21513976263

Issue #1819, opened 2026-01-19
upgrade tt-smi and tt-topology usage in run.py


tstescoTT:
Requirements: 
- tt-smi version and tt-topology versions
- tt-topology should only run on T3K machines

tstescoTT:
addressed in: https://github.com/tenstorrent/tt-inference-server/pull/1916

PR #1916, opened 2026-01-27
fix tt-smi and tt-topology versions for system sw validation, unify uv venv setup, decouple uv bootstrap

tstescoTT:
# change log
* addresses https://github.com/tenstorrent/tt-inference-server/issues/1819
* refactors venv system with uv to standardize venv setup
* decouples uv bootstrap into separate file for future needs (e.g. to have run.py run within a venv)

### DIFF
--- a/.cursor/rules/coding-standards.mdc
+++ b/.cursor/rules/coding-standards.mdc
@@ -11,3 +11,4 @@ alwaysApply: true
 - Minimize new dependencies that are not in the Python 3.8 standard library and not already used in the project.
 - Do not provide verbose explainations, writting code that correctly meets the requirements is more important, UNLESS there is a bug found in the code. If there is a bug it should be explained and then a fix to the bug proposed.
 - Use clear and professional names for all variables, functions, classes, and files.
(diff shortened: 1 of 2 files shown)


workflows/workflow_venvs.py
```diff
@@ -64,20 +65,33 @@ def _infer_data(self):
...
+            # --managed-python: explicitly use uv managed python versions
+            run_command(
+                f"{str(UV_EXEC)} venv --managed-python --python={self.python_version} {self.venv_path} --allow-existing",
+                logger=logger,
+            )
+        # uv will verify deps if venv exists
+        venv_setup_completed = self.setup_function(self, model_spec=model_spec)
+        if not venv_setup_completed:
+            logger.error(f"Failed to setup venv: {self.venv_type.name}")
+        return venv_setup_completed
```
bgoelTT: Is there a reason we don't raise an exception here? I could see this silently failing if the return state is unhandled like `run_system_software_validation.py` is set up to do

bgoelTT: Ah thanks for clarifying 🥇

PR #1936, opened 2026-01-29
Dmadic tt/prefill server poc

dmadicTT:


prefill-node-poc/main_prefill_real.py
```diff
@@ -0,0 +1,247 @@
...
+    p.add_argument("--max-num-batched-tokens", type=int, default=8192)
+    p.add_argument("--kv-cache-num-blocks", type=int, default=2048)
+    return p
+
+
+def greedy_sample(logits: torch.Tensor) -> torch.Tensor:
+    return torch.argmax(logits, dim=-1)
+
+
+def get_last_token_logits(
```
knovokmetTT: I guess this was that weird part where model generates logits(scores for each token in the llm vocabulary) for each token :)... the others are used in that speculative decoding

prefill-node-poc/model_runner.py
```diff
@@ -0,0 +1,341 @@
...
+        
+        from models.demos.deepseek_v3.utils.config_dataclass import KvCacheConfig as TtKvCacheConfig
+        
+        tt_kv_cache_config = TtKvCacheConfig(
+            kv_cache_shape=kv_cache_config.shape,
+            dtype=getattr(ttnn, kv_cache_config.dtype, ttnn.bfloat8_b),
+        )
+        
+        self._generator._prepare_run_configs("prefill", kv_cache_override=tt_kv_cache_config)
+        self._kv_cache = self._generator.get_kv_cache()
```
knovokmetTT: I guess this is the part we allocate memory in device RAM based on the number of layers, hidden dim, batch size etc? Altho it should already know it needs to prealloc 12 gigs.... from my research, vllm splits these 12 gigs into num_of_layers pools of cache blocks.. do we do this as well?

prefill-node-poc/model_runner.py
```diff
@@ -0,0 +1,341 @@
...
+                prompt_lens = [min(pl, cfg.prefill_max_tokens) for pl in prompt_lens]
+            
+            max_padded_len = ((max_prompt_len + ttnn.TILE_SIZE - 1) // ttnn.TILE_SIZE) * ttnn.TILE_SIZE
+            
+            all_logits = []
+            
+            # Build page_table from all sequences' block tables
+            # Shape: [batch_size, max_num_blocks_per_req]
+            max_blocks_per_req = max(len(seq.block_table) for seq in seqs)
+            page_table = torch.zeros((len(seqs), max_blocks_per_req), dtype=torch.int32)
```
knovokmetTT: int32 because each entry in the page table is the address?(or index?) of the physical block of device RAM?

prefill-node-poc/model_runner.py
dmadicTT: `_kv_cache` is actually a list of 61 objects, one per layer

prefill-node-poc/model_runner.py
dmadicTT: each entry is KV cache block ID

Issue #1938, opened 2026-01-29
Dead worker restart loop prevents device from becoming ready


ljovanovicTT:
## Problem
When a worker crashes, the restart mechanism enters an infinite restart loop. This prevents the device from fully initializing and reaching a warmed-up and ready state, leaving the service stuck and unavailable.

## Occurrence:
- Flaky

## Relevant configuration:
Issue was observed on Galaxy and T3K

## CI runs:
- https://github.com/tenstorrent/tt-shield/actions/runs/21447911793
- https://github.com/tenstorrent/tt-shield/actions/runs/21432009626

## Relevant logs:
[server_output.log](https://github.com/user-attachments/files/24938870/server_output.log)

ljovanovicTT:
# Issue 2 - Segmentation Fault

This issue is observed on different models on Galaxy, and more detail on previous investigation can be found here under `Segmentation fault`: [Fix BGE-Large-EN-v1.5 nightly failures: MTEB ModelMeta validation and Galaxy port conflicts #1651](https://github.com/tenstorrent/tt-inference-server/issues/1651)

CI link:
https://github.com/tenstorrent/tt-shield/actions/runs/21479308633/job/61875765488

Relevant logs:
```
*** Process received signal ***
[c0fdc6dea7d7:00279] Signal: Segmentation fault (11)
[c0fdc6dea7d7:00279] Signal code: Invalid permissions (2)
[c0fdc6dea7d7:00279] Failing at address: 0x7f1bc97fe000
[c0fdc6dea7d7:00279] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x7f1e99fb9520]
[c0fdc6dea7d7:00279] [ 1] /home/container_app_user/tt-metal/build/lib/libtt_metal.so(_ZN2tt8tt_metal16memcpy_to_deviceILb0EEEvPvPKvm+0x126)[0x7f1e6c163846]
[c0fdc6dea7d7:00279] [ 2] /home/container_app_user/tt-metal/build/lib/libtt_metal.so(_ZN2tt8tt_metal15buffer_dispatch43populate_sharded_buffer_write_dispatch_cmdsEPKvRNS0_13DeviceCommandILb1EEERNS0_6BufferERNS1_32ShardedBufferWriteDispatchParamsE+0x1a7)[0x7f1e6c012457]
[...]

PR #1953, opened 2026-01-30
Fix Dead Worker Infinite Restart Loop

ljovanovicTT:
## Issue
[#1938 Dead worker restart loop prevents device from becoming ready](https://github.com/tenstorrent/tt-inference-server/issues/1938)

## Problem
When `use_queue_per_worker` was enabled, worker restarts triggered an infinite device restart loop, preventing the server from ever reaching a ready state.

The root cause was in the `scheduler`’s `_start_worker` method, where the result queue was selected using:
```
worker_index = len(self.worker_info)
result_queue = self.result_queues_by_worker[worker_index]
```
which was correct on initialization, but on restart resulted in out of bounds error

### DIFF
--- a/tt-media-server/model_services/scheduler.py
+++ b/tt-media-server/model_services/scheduler.py
@@ -504,7 +504,7 @@ class Scheduler:
                         )
                         if self.settings.allow_deep_reset:
                             self.logger.info("Trying deep restart of all workers")
-                            self.deep_restart_workers()
+                            await self.deep_restart_workers()
 
                 await asyncio.sleep(self.settings.worker_check_sleep_timeout)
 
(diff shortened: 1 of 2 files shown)


ztorlakTT (approved):

tt-media-server/model_services/scheduler.py
```diff
@@ -152,8 +155,11 @@ def _start_worker(self, worker_id=None):
 
         self.logger.info(f"Starting worker {worker_id}")
 
-        # Get the queue index for this worker
-        worker_index = len(self.worker_info)
+        # Use existing queue index when restarting; otherwise assign by order
+        if queue_index is not None:
```
knovokmetTT: I guess this logic is only needed if we are using queue per worker?

PR #1957, opened 2026-01-30
Add pytest exit code checks for Tests workflow

mjeremicTT:
This PR fixes the crashing when pytest from `run_tests.py` returns a non-zero exit code by adding proper exit code handling. Previously, `run_command` with `check=True` would raise a `RuntimeError` immediately when tests failed, preventing the script from collecting and reporting results gracefully. Now the script uses `check=False` and explicitly handles pytest exit codes - allowing `OK` and `TESTS_FAILED` to continue while raising an error for actual pytest problems like `INTERNAL_ERROR`, `USAGE_ERROR`, or `NO_TESTS_COLLECTED`.

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -10,6 +10,7 @@ import logging
 import sys
 from pathlib import Path
 from typing import List
+from pytest import ExitCode
 
 # Add the script's directory to the Python path
 # this for 0 setup python setup script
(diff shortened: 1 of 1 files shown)


mjeremicTT:
Closing this one as this problem will be solved with: [https://github.com/tenstorrent/tt-inference-server/pull/1967](https://github.com/tenstorrent/tt-inference-server/pull/1967)

PR #1973, opened 2026-02-02
tstesco/model-types-refactor

tstescoTT:
# change log

* move ModelType ModelSource and InferenceEngine enum types to workflow_types.py
* update local testing python venv install instructions


bgoelTT (changes requested):
Just one comment about the dev venv creation with uv

PR #1989, opened 2026-02-04
Allow runners to set thread limits

ztorlakTT:
Thread limits were previously set at the device worker level with fixed values, but different model runners have different CPU threading requirements. 

What changed:
- Moved environment setup from device workers to `BaseDeviceRunner`, allowing runners to control their own thread configuration
- Default thread limits are now automatically determined based on workload type
- Runners can optionally override defaults via constructor parameters
- Removed redundant `num_torch_threads` parameter from runner constructors
- Created new `utils/runner_utils.py` to centralize environment setup logic


mcupacTT:
I went quickly through the PR, and here are my thoughts:

It is very good that now any runner is able to setup the number of threads arbitrarily. However, i have a couple of comments there
1. The setup_cup_threading_limits is now called later in the code (in the runner) compared to the previous version (in device worker)
    1. this might cause a problem since the setup_cup_threading_limits sets these env variables, which need to be set before the corresponding libraries are imported which read the envs
    1. meaning that maybe OMP_NUM_THREADS could be set to late, but the library which uses it is already imported by that time
    1. I am not sure if this is an issue or not, but wanted to point it out here - maybe somebody who has more experience can clarify it
2. For the torch, we dont have this issue since we also have set_torch_thread_limits which dinamically adjusts the thread pool after torch is imported
    1. however, it is a bit misleading that TORCH_NUM_THREADS is set to cpu_threads instead of num_threads - it will be anyway overwritten by set_torch_thread_limits after importing, but still misleading.

Off-topic
1. Is there any context why we set 16 threads for dynamic batch? If yes, maybe it could be commented

mcupacTT:
> @mcupacTT Thanks for the thorough review! These are good questions.
> 
> 1. The runner is instantiated very early - right at the start of initialize_device_worker() in worker_utils.py. This happens before any model-specific imports or library initialization. I discussed this with the team yesterday and we believe this should be early enough, but I'd like to verify with CI runs before merging. Currently our CI runs are failing for unrelated reasons, so I'll wait for those to stabilize before merging this PR.
> 2. Good catch - I'll update TORCH_NUM_THREADS to use num_torch_threads.
> 
> We use device_worker_dynamic_batch for LLM workloads, and VLLM performs better with this thread configuration

Thanks for the detailed explanations! It is clear now.

vmilosevic (approved):

vmilosevic:
@ztorlakTT changes related to tt-forge runner look good, but please write a brief PR description next time to help reviewers understand the change
- why are we making the change (motivation)
- what has changed
Thnx!

PR #2017, opened 2026-02-05
Flux motif eval test

vpetrovicTT:
Eval test for image generation models (Flux, Motif, SD3.5, etc.)

### DIFF
--- a/tests/server_tests/server_tests_config.json
+++ b/tests/server_tests/server_tests_config.json
@@ -2194,7 +2194,7 @@
                     "mock_mode": false
                 },
                 "targets": {
-                    "num_of_devices": 32,
+                    "num_of_devices": 1,
                     "num_inference_steps": 20,
                     "image_generation_time": 11
                 }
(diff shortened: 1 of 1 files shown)


tests/server_tests/server_tests_config.json
fivanovicTT: Can you also the following configurations:
`FLUX.1-dev` and `FLUX.1-schnell`:
- p150x4
- p150x8
- p300
- qbge

`Motif-Image-6B-Preview`:
- p150x4
- p150x8

tests/server_tests/test_cases/image_generation_eval_test.py
```diff
@@ -0,0 +1,346 @@
...
+        server_url=args.server_url,
+        num_inference_steps=args.num_inference_steps,
+    )
+
+    test = ImageGenerationEvalsTest(TestConfig.create_default(), {"request": request})
+    if args.service_port:
+        test.service_port = str(args.service_port)
+
+    result = test.run_tests()
+    raise SystemExit(0 if result.get("success") else 1)
```
fivanovicTT: This is not needed, will not run it using CLI

PR #2027, opened 2026-02-06
Update path to DiT models

sosborne-TT:
This metal [PR](https://github.com/tenstorrent/tt-metal/pull/34053) moved the location of tt_dit models so updating the paths here.


ljovanovicTT (approved):

bgoelTT (dismissed):
Please do not merge this yet until we make the v0.9.0 release, this will interfere with the fact that the tt-metal `stable` release branch has not been updated to accommodate the new path change

bgoelTT:
> > Please do not merge this yet until we make the v0.9.0 release, this will interfere with the fact that the tt-metal `stable` release branch has not been updated to accommodate the new path change
> 
> Hey @bgoelTT, unfortunately this is blocking for video models are expected to be finished by 25th of February (blackhole support) ☹️. And, if I am not mistaken, FLUX and Motif models are using DiT runner which have the same deadline.
> 
> Since we are currently working on multiple branches, and it will be kind of pain to have to merge this branch and revert from PR when merging to `dev`, could we, as an easiest option, just merge this PR to `dev`, and on Friday we make a revert PR and merge to `dev` and make a release without these changes? Couldn't think of the easier solution, maybe you can think of something better?
> 
> We should also think of introducing `stable` branch like metal, and use `stable` for making releases in the future 🤔
> 
> cc: @idjuricTT

Hey @fivanovicTT sorry for blocking this merge, I'll repeat here what we talked about offline for visibility.

We are planning to make the v0.9.0 release on February 9th, Monday so I don't expect this to block you for long. I completely agree this is justification for a tt-inference-server `stable` branch 👍 . @tstescoTT mentioned this a while back - to have a "time-delayed" `stable` branch that is really just a checkout of a previous `dev` commit

PR #2031, opened 2026-02-06
enable end to end streaming cpp server

knovokmetTT:


tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -234,55 +222,19 @@ void LLMController::handle_streaming(
                         sse.append(buf.GetString(), buf.GetSize());
                         sse.append("\n\n");
 
-                        std::lock_guard<std::mutex> lock(*buffer_mutex);
-                        buffer->append(sse);
-
-                        if (buffer->size() >= FLUSH_THRESHOLD) {
```
knovokmetTT: This breaks ITL/TPOT on vllm bench serve , will show up as 0

Issue #2019, opened 2026-02-05
Create cpp skeleton for LLM inference


dmadicTT:
<img width="593" height="425" alt="Image" src="https://github.com/user-attachments/assets/87cfd28b-baa5-4d1c-9de0-2edcb9859ea6" />

Goal is to create core components inspired by [vLLM](https://github.com/vllm-project/vllm) and [nano-vLLM](https://github.com/GeeeekExplorer/nano-vllm)
- Core engine
- Scheduler
- Block Manager
- Model runner

Python reference for prefill node poc https://github.com/tenstorrent/tt-inference-server/pull/1936/changes

PR #2035, opened 2026-02-06
Add cpp LLM engine

dmadicTT:
https://github.com/tenstorrent/tt-inference-server/issues/2019


tt-media-server/cpp_server/include/runners/llm_engine/engine/scheduler.hpp
```diff
@@ -0,0 +1,57 @@
...
+ */
+class Scheduler {
+ public:
+  explicit Scheduler(const Config& config);
+
+  /** @return true if there are no waiting or running sequences. */
+  bool is_finished() const;
+
+  /** Enqueues a sequence for prefill (waiting queue). */
+  void add(Sequence& seq);
```
idjuricTT: This one seems almost redundant to multiprocessing and dynamic batching schedulers. We should think how to combine those

tt-media-server/cpp_server/src/runners/llm_engine/engine/scheduler.cpp
```diff
@@ -0,0 +1,101 @@
...
+    running_.push_back(seq);
+    scheduled_seqs.push_back(seq);
+  }
+  if (!scheduled_seqs.empty()) {
+    LLM_ENGINE_LOG("scheduler") << "schedule prefill n=" << scheduled_seqs.size()
+                              << " batched_tokens=" << num_batched_tokens << std::endl;
+    return {scheduled_seqs, true};
+  }
+
+  while (!running_.empty() && num_seqs < max_num_seqs_) {
```
idjuricTT: It seems that in this paradigm scheduler is our device worker dynamic batch?
and engine is a model runner?

tt-media-server/cpp_server/src/runners/llm_engine/engine/scheduler.cpp
```diff
@@ -0,0 +1,101 @@
...
+    while (!block_manager_.can_append(*seq)) {
+      if (!running_.empty()) {
+        preempt(*running_.back());
+        running_.pop_back();
+      } else {
+        preempt(*seq);
+        break;
+      }
+    }
+    if (block_manager_.can_append(*seq)) {
```
knovokmetTT: this if seems redundant because the while above has an opposite cond?

knovokmetTT:
What do you think about adding this code under nanovllm/vllm folder inside the cpp folder, just so we can compartmentalize?

tt-media-server/cpp_server/src/runners/llm_engine/engine/scheduler.cpp
knovokmetTT: actually, there is a much worse bug here
if we self preempt in the while loop above, we can free up space, then add ourself again in the if block, this needs to be fixed before merging

knovokmetTT: https://github.com/tenstorrent/tt-inference-server/pull/2042

Proposed fix that worked on my dynamic batching branch

idjuricTT (approved):

PR #2042, opened 2026-02-08
nanovllm - fix self preempt logic 

knovokmetTT:


tt-media-server/cpp_server/src/runners/llm_engine/engine/scheduler.cpp
```diff
@@ -50,16 +50,18 @@ std::pair<std::vector<Sequence*>, bool> Scheduler::schedule() {
...
         preempt(*running_.back());
         running_.pop_back();
       } else {
         preempt(*seq);
+        self_preempt = true;
         break;
       }
     }
-    if (block_manager_.can_append(*seq)) {
+    if (block_manager_.can_append(*seq) && !self_preempt) {
```
knovokmetTT: must not add self to running if it self preempted to get space

PR #2050, opened 2026-02-09
Implement logic for canceling fine-tuning jobs and fix issues related to job workflow

### LINKED ISSUE #1616: Implement logic for canceling fine-tuning jobs
We need to implement "soft canceling" of fine-tuning jobs, so that the hardware returns in the correct state after canceling. 

It could be implemented with a flag which should be repeatedly checked at the end of each step during fine-tuning. In case of canceling, the flag is first turned to False, and the fine-tuning job is canceled once it ends the step.

mcupacTT:
Closes #1616 

This PR introduces: 
1. Cooperative cancellation support for training jobs - where the training loop repeatedly checks if the cancellation is initiated, and breaks if that is the case
2. Fix to the correct job workflow behavior: the job lands in _IN_PROGRESS_ state only after its corresponding runner started executing. Until then, it remains in _QUEUED_ state
3. Job metadata (all information related to job) are now deleted together with job generated data (e.g. fine-tuned model) - in `cleanup_old_jobs` only

### More detailed:

* Added `_start_event` and `_cancel_event` attributes - as `multiprocessing.Event` managed by `multiprocessing.Manager` - to `TrainingRequest` and `Job` 
* The job manager now uses these events to coordinate job state: it waits for `_start_event` before marking jobs as `in_progress`, and uses `_cancel_event` for cooperative cancellation.
* The `training_gemma_lora_runner.py` runner checks `_cancel_event` after each step for cancellation requests, breaks if initiated
* On shutdown, or force cancellation through CTRL+C - job manager force-cancels jobs without waiting for cooperative exit, ensuring prompt server shutdown.
* Corresponding tests are added to `test_job_manager.py`

### Note

* If training job is cooperatively cancelled, and server is turned off afterwards, `tt-smi -r` is not needed to start the server again - cooperative cancellation brings the hardware back to the correct state
* To show the correctness of starting a new training job after first is cancelled, here the loss curves of initial (left) and final job (right):

| First job | Second job |
| :---: | :---: |
|  |  |

_the curves slightly differ since the repro settings for setting the same random seed for parameter initialization are not implemented yet - the important part is that both curves start from the comparable loss and are able to optimize both training and validation loss towards zero_

### DIFF
--- a/tt-media-server/tt_model_runners/forge_runners/requirements.txt
+++ b/tt-media-server/tt_model_runners/forge_runners/requirements.txt
@@ -5,6 +5,7 @@ torchvision
 timm # input preprocessing
 flax # input preprocessing
 pytorchcv # for vovnet
+tabulate
 
 datasets # model training
 peft # model training


tt-media-server/utils/job_manager.py
```diff
@@ -358,41 +397,52 @@ def _cleanup_old_jobs(self):
...
-            except Exception as e:
-                self._logger.debug(f"Failed to delete file for job {job.id}: {e}")
+            if job._cancel_event and not force:
+                # runner handles cancellation
+                job._cancel_event.set()
+                running_task = job._task
+            else:
+                # runner does not handle cancellation, so we cancel the task ourselves
+                job._task.cancel()
+                running_task = job._task
```
```suggestion
            if job._cancel_event and not force:
                # runner handles cancellation
                job._cancel_event.set()
            else:
                # runner does not handle cancellation, so we cancel the task ourselves
                job._task.cancel()
            running_task = job._task
```

tt-media-server/utils/job_manager.py
```diff
@@ -358,41 +397,52 @@ def _cleanup_old_jobs(self):
...
-                        try:
-                            self.db.delete_job(job_id)
-                        except Exception as e:
-                            self._logger.error(
-                                f"Database deletion failed for job {job_id} during cleanup: {e}"
-                            )
+        if not jobs_to_remove:
+            return
+
+        # Cleanup outside the lock
```
ztorlakTT: I think you can remove this comment

PR #2062, opened 2026-02-10
Allow Flux to use inference steps 4

vpetrovicTT:
Enable Flux models to run with 4 inference steps and add unit tests for per-model validation.

### DIFF
--- a/tt-media-server/domain/image_generate_request.py
+++ b/tt-media-server/domain/image_generate_request.py
@@ -4,8 +4,14 @@
 
 from typing import List, Optional, Tuple, Union
 
+from config.settings import get_settings
 from domain.base_request import BaseRequest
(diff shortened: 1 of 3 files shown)


tt-media-server/domain/image_generate_request.py
```diff
@@ -11,7 +11,7 @@
 class ImageGenerateRequest(BaseRequest):
     prompt: str
     negative_prompt: Optional[str] = None
-    num_inference_steps: Optional[int] = Field(default=20, ge=12, le=50)
+    num_inference_steps: Optional[int] = Field(default=20, ge=4, le=50)
```
idjuricTT: Any chance you could read this from config? 
This would cause a change on SDXL side

PR #2098, opened 2026-02-12
Enable Apply Chat Template

ljovanovicTT:
## Problem
To fully support the `/chat/completions` endpoint, we need to enable an `apply_chat_template` mechanism that mirrors HuggingFace Transformers’ apply_chat_template behavior (https://github.com/huggingface/transformers).
This method relies on a model-specific `tokenizer_config.json`.

## Implementation
- Add an `apply_chat_template` method to the Tokenizer
- Default to the `Deepseek-V3` `tokenizer_config.json`, downloading it in `build.sh` if it is not already present
- Add unit test

## Important note
To fully support `apply_chat_template` across all models, we would need either a C++ tool with more complete Jinja support (current options are limited, see https://jinja2cpp.github.io/docs/j2_compatibility.html) or a custom Jinja parser.
For now, given our focus on minimizing server overhead, I suggest keeping the current DeepSeek-style implementation instead of using a Python-based Jinja parser.

### DIFF
--- a/tt-media-server/cpp_server/src/utils/tokenizer.cpp
+++ b/tt-media-server/cpp_server/src/utils/tokenizer.cpp
@@ -2,7 +2,6 @@
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
 #include "utils/tokenizer.hpp"
-#include "config/settings.hpp"
 
 #include <fstream>
 #include <sstream>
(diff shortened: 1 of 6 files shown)


tt-media-server/cpp_server/src/utils/tokenizer_config.cpp
```diff
@@ -0,0 +1,56 @@
...
+std::string extract_token(const Json::Value& v) {
+    if (v.isNull() || !v) return {};
+    if (v.isString()) return v.asString();
+    if (v.isObject() && v.isMember("content")) return v["content"].asString();
+    return {};
+}
+
+}  // namespace
+
+bool load_tokenizer_config(const std::string& path, TokenizerConfig& out) {
```
idjuricTT: Nitpick: This is more a fill_tokenizer_config since it works on an object

tt-media-server/cpp_server/src/utils/tokenizer.cpp
```diff
@@ -61,4 +62,61 @@ std::string Tokenizer::decode(const std::vector<int>& token_ids) const {
...
+const char* ASSISTANT_TAG = "<<|Assistant|>>";
+
+}  // namespace
+
+std::string Tokenizer::apply_chat_template(const std::vector<tt::domain::ChatMessage>& messages,
+    bool add_generation_prompt) {
+    static TokenizerConfig cfg = []() {
+        TokenizerConfig c;
+        std::string config_path = tt::config::tokenizer_config_path();
+        if (config_path.empty()) {
```
idjuricTT: Can we just have the load_tokenizer_config check both the path and loading an throw an error?
This way we're giving the tokenizer loading logic to apply_chat_template which just uses tokenizer + this might repeat if another method would use a tokenizer

PR #2105, opened 2026-02-12
Add v1 api prefix

vpetrovicTT:
Add /v1 prefix to all API endpoints for OpenAI API compatibility. Both versioned (/v1/...) and legacy (/...) paths are served during a deprecation period.

### DIFF
--- a/scripts/examples/example_audio_request.py
+++ b/scripts/examples/example_audio_request.py
@@ -14,7 +14,7 @@
 #
 # Notes:
 # - Uses Hugging Face Datasets Server HTTP APIs, no extra dependencies required.
-# - Intended to work with tt-media-server audio endpoint: /audio/transcriptions
+# - Intended to work with tt-media-server audio endpoint: /v1/audio/transcriptions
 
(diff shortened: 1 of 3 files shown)


tt-media-server/open_ai_api/__init__.py
```diff
@@ -21,27 +22,88 @@
...
+
+    for route in routes:
+        api_router.include_router(
+            route.router,
+            prefix=route.v1_prefix,
+            tags=route.tags,
+        )
+        logger.info(f"Registered: {route.v1_prefix} [{route.tags[0]}]")
+
+        if route.legacy_prefix != route.v1_prefix:
```
ztorlakTT: `legacy_prefix` and `v1_prefix` will always be different, right? Maybe it's better to check if `legacy_prefix` is ​​defined?

ljovanovicTT:
Can you also update media server strategies like https://github.com/tenstorrent/tt-inference-server/blob/dev/utils/media_clients/image_client.py#L362

PR #2126, opened 2026-02-18
FIX: default impl for t3k for gpt-oss-120b

stisiTT:
CI run test [here](https://github.com/tenstorrent/tt-shield/actions/runs/22148848836)

Set default implementation to true in DeviceModelSpec for T3K


bgoelTT (approved):
Approved, though the evals didn't actually complete due to the context length limitations.

PR #2128, opened 2026-02-19
Make entrypoint permission setup best-effort for restricted environments

ksriv001:
- `usermod`, `chown`, and `chmod` in `docker-entrypoint.sh` fail in Kubernetes pods with restricted security contexts (e.g. read-only `/etc/passwd`), causing container crash loops with `usermod: cannot lock /etc/passwd; Permission denied`.
- These permission setup operations are now best-effort: they attempt the operation, log a warning on failure, and continue startup instead of crashing.
- No behavior change for VM/bare-metal environments where permissions work

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -30,12 +30,18 @@ set_group_permissions() {
     
     # Set group if needed
     if [ "$current_group" != "$shared_group" ]; then
-        chown -R :"$shared_group" "$var_dir"
+        if ! chown -R :"$shared_group" "$var_dir" 2>/dev/null; then
+            echo "WARNING: could not chown ${var_dir} to group ${shared_group}"
+            return 0
+        fi
     fi
     
     # Set permissions if needed
     if [ "$current_perms" != "2775" ]; then
-        chmod -R 2775 "$var_dir"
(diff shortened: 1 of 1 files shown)


bgoelTT:
@ksriv001 this entrypoint script is only used in the `-dev` tagged vLLM docker images to faciliate local deployments. The ones for restricted security contexts are the ones tagged with `-cloud`

Are you able to use those images instead?

ztorlakTT (approved):

idjuricTT (approved):
Please just fork dev and open PR to dev.

@tstescoTT  and @bgoelTT  could you please double check the entry point code

ksriv001:
@bgoelTT I was trying ot run one of the shas for WAN2.2 with some bug fixes:
`ghcr.io/tenstorrent/tt-shield/tt-media-inference-server:0.8.0-91e687d09ce5c209133523cd18400576da027e4b-8aa88c2-62045074008`
cc: @idjuricTT 

I am still learning around the workflows in tt-inference-server. Is there a -cloud tagged version for this SHA that I can use? Or is there a way to generate one? As you said, we might not even need these changes then if there is a workflow that generates the corresponding -cloud images with the appropriate security context.

PR #2132, opened 2026-02-19
Make entrypoint permission setup best-effort for restricted environments

ksriv001:
- usermod, chown, and chmod in docker-entrypoint.sh fail in Kubernetes pods with restricted security contexts (e.g. read-only /etc/passwd), causing container crash loops with usermod: cannot lock /etc/passwd; Permission denied.
- These permission setup operations are now best-effort: they attempt the operation, log a warning on failure, and continue startup instead of crashing.
- No behavior change for VM/bare-metal environments where permissions work

**Context on why we need these changes:**
When running the container in Kubernetes, the entrypoint script tries to run usermod to add the app user to a shared group. That requires writing to /etc/passwd, which Kubernetes blocks in restricted security contexts. Because the script is set to exit on any error, that one failure kills the entire startup — the server never starts, and the pod crash-loops.
This fix wraps usermod, chown, and chmod so they warn on failure instead of crashing. The server continues to start because the cache directory (/tmp) is already world-writable, so the permission setup isn't actually needed for it to function.
On VMs and bare metal where the permissions aren't restricted, the commands still succeed and run exactly as before — the warning path is never hit. So existing deployments are unaffected.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -30,12 +30,18 @@ set_group_permissions() {
     
     # Set group if needed
     if [ "$current_group" != "$shared_group" ]; then
-        chown -R :"$shared_group" "$var_dir"
+        if ! chown -R :"$shared_group" "$var_dir" 2>/dev/null; then
+            echo "WARNING: could not chown ${var_dir} to group ${shared_group}"
+            return 0
+        fi
     fi
     
     # Set permissions if needed
     if [ "$current_perms" != "2775" ]; then
-        chmod -R 2775 "$var_dir"
+        if ! chmod -R 2775 "$var_dir" 2>/dev/null; then
+            echo "WARNING: could not chmod ${var_dir} to 2775"
+            return 0
+        fi
     fi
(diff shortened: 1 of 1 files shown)


idjuricTT (approved):
@tstescoTT  and @bgoelTT  please double check the changes, makes sense for me

ztorlakTT (approved):

tstescoTT:
Hi @ksriv001 I am removing the docker-entrypoint.sh script in https://github.com/tenstorrent/tt-inference-server/tree/tstesco/docker-interface-redesign as described in https://github.com/tenstorrent/tt-inference-server/issues/1253#issuecomment-3881945381

I'd rather avoid the conflicts and focus on getting the redesign finished unless this unblocks you today.

ksriv001:
@tstescoTT Thank you for the context. This is not blocking me as of now, as I was able to start the service manually from inside the container for testing. I'll close this PR for now. Looking forward to your refactoring 🙂

Issue #2136, opened 2026-02-19
Remove display_name from ModelSpec.py


fivanovicTT:
Currently, there is a tech debt `display_name` in `model_spec.py` which should be completely removed. Please check if it is used in the codebase (so we don't break anything)

PR #2142, opened 2026-02-19
Remove display_name from ModelSpec.py

fivanovicTT:
### Link to GitHub issue
[2136 - Remove display_name from ModelSpec.py](https://github.com/tenstorrent/tt-inference-server/issues/2136)

### Summary
Currently, there is a tech debt `display_name` in `model_spec.py` which should be completely removed. Please check if it is used in the codebase (so we don't break anything)


idjuricTT (approved):
@tstescoTT  would this change be fine with you?

PR #2154, opened 2026-02-19
move embedding service under base service

knovokmetTT:


tt-media-server/cpp_server/include/services/base_service.hpp
```diff
@@ -3,80 +3,81 @@
...
     }
 
+protected:
+    virtual ResponseType process_request(RequestType request) = 0;
+    virtual void pre_process(RequestType& request) const = 0;
+    virtual void post_process(ResponseType& response) const = 0;
+};
+
+template<std::derived_from<domain::BaseRequest> RequestType, std::derived_from<domain::BaseResponse> ResponseType>
+class Streamable {
```
idjuricTT: Nitpick: not sure is this practice in C++, but shouldn't this be a separate file?

tt-media-server/cpp_server/src/services/embedding_service.cpp
```diff
@@ -679,6 +690,11 @@ BaseService::SystemStatus EmbeddingService::get_system_status() const {
     status.device = "tenstorrent";
 
     for (size_t i = 0; i < impl_->num_workers_; ++i) {
+        WorkerInfo info;
```
idjuricTT: I think this is moved as well to health since it's common for all the workers

PR #2163, opened 2026-02-20
Benchmark cpp server using vllm bench serve

dmadicTT:


.github/workflows/test-gate.yml
```diff
@@ -565,49 +563,93 @@ jobs:
...
-          TEST_RUNNER_FREQUENCY_MS: "1"
+          OPENAI_API_KEY: 'your-secret-key'
         run: |
-          pytest performance_tests/test_llm_streaming.py -sv 2>&1 | tee test_output.txt
-          TEST_EXIT_CODE=${PIPESTATUS[0]}
+          RESULT_FILE="vllm-bench-result.json"
+          vllm bench serve \
+            --model 'deepseek-ai/DeepSeek-V3.2' \
+            --backend vllm \
+            --endpoint /v1/completions \
```
ljovanovicTT: Do we want to focus on `/v1/chat/completions` if that endpoint is the target?

Issue #2192, opened 2026-02-23
4-stage blitz decode pipeline on single BH


dmadicTT:
Rely on https://github.com/tenstorrent/tt-metal/pull/38305/changes#diff-ec961b848ea1e53b64bd1f8faa62600fdeca18db5610042c579325c04d438696

PR #2195, opened 2026-02-23
Multi-stage blitz pipeline

dmadicTT:
Issue https://github.com/tenstorrent/tt-inference-server/issues/2192

This PR adds integration with the blitz decode pipeline introduced with this PR in metal: https://github.com/tenstorrent/tt-metal/pull/38305

It allows us to run the inference server on 4 stages on single BH galaxy. Multi-host pipeline is not yet supported.

We used the following test as reference https://github.com/tenstorrent/tt-metal/blob/main/models/demos/deepseek_v3_b1/tests/unit_tests/test_multi_host_pipeline.py#L458

image

### DIFF
--- a/tt-media-server/cpp_server/src/runners/llm_runner/backend/device_backend.cpp
+++ b/tt-media-server/cpp_server/src/runners/llm_runner/backend/device_backend.cpp
@@ -8,10 +8,6 @@ namespace llm_engine::backend {
 std::unique_ptr<IDeviceBackend> make_device_backend_mock(const Config& config);
 std::unique_ptr<IDeviceBackend> make_device_backend_ttrun(const Config& config);
 
-#ifdef USE_METAL_CPP_LIB
-std::unique_ptr<IDeviceBackend> make_device_backend_sockets(const Config& config);
-#endif
-
 std::unique_ptr<IDeviceBackend> make_device_backend(const Config& config) {
   switch (config.device) {
     case DeviceBackend::TtRun:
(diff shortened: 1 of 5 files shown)


tt-media-server/cpp_server/src/runners/llm_runner/backend/device_backend.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <iostream>
+
+namespace llm_engine::backend {
+
+std::unique_ptr<IDeviceBackend> make_device_backend_mock(const Config& config);
+std::unique_ptr<IDeviceBackend> make_device_backend_ttrun(const Config& config);
+
+#ifdef USE_METAL_CPP_LIB
+std::unique_ptr<IDeviceBackend> make_device_backend_sockets(const Config& config);
```
dmadicTT: We don't need this one but maybe `make_device_backend_ttrun` should be behind this `ifdef`

tt-media-server/cpp_server/src/runners/llm_runner/backend/ttrun_device_backend.cpp
```diff
@@ -0,0 +1,157 @@
...
+      ttrunConfig.script.data(),
+      nullptr
+    };
+
+    pid_t pid = fork();
+    if (pid < 0) {
+      throw std::runtime_error("TtRunDeviceBackend: fork failed");
+    }
+    if (pid == 0) {
+      setpgid(0, 0);  // become process group leader so killpg covers all descendants
```
dmadicTT: Nice!

PR #2196, opened 2026-02-24
Remove DecodeResult and fix embedding service

ztorlakTT:
- Remove the `DecodeResult` struct
- Fix embedding service device ID
- Add auth headers to embedding tests
- Extract the inline lambda in `LLMService::start_workers` into the shared `make_worker_config_for_process` helper.


fivanovicTT (approved):
LGTM

idjuricTT (approved):

PR #2215, opened 2026-02-26
Move ci config to inference

vmaksimovicTT:
Move the config for model CI runs from tt-shield to tt-inference repo, so that developers can easily add new models to different model CI runs without opening a PR in tt-shield

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,7 @@ scripts/examples/example_data
 !benchmarking/benchmark_targets/*.json
 !model_specs_output.json
 !tt-media-server/static/demos/images.json
-!.github/workflows/*.json
+.github/workflows/*.json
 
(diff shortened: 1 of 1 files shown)


tstescoTT:
Does an accompanying PR need to land in tt-sheild to change the default location to be in tt-inference-server repo `dev`? https://github.com/tenstorrent/tt-shield/blob/4fcf2efa38f7eabf3ed69b8fab2f94871fbb2655/.github/scripts/generate_model_ci_workflows/generate_ci_matrix.py#L23

Also whenever someone changes `models-ci-config.json` does this mean someone will need to manually run `generate_ci_matrix.py` in tt-shield and PR the diffs?

tstescoTT (approved):

PR #2227, opened 2026-02-26
Add prefill/decode split mode for LLM service

ztorlakTT:
## Summary
Enables the LLM service to operate in three modes controlled by the `LLM_MODE` environment variable:
- **`regular`** (default) — existing behavior, single server handles both prefill and decode
- **`prefill`** — receives prompts via socket from a decode server, runs prefill + generates one token, returns the result with the updated token sequence
- **`decode`** — accepts HTTP requests, forwards prompts via socket to a prefill server, receives the first token back, then continues decode generation locally

Socket roles are auto-configured from `LLM_MODE` (decode = server, prefill = client). Regular mode is unaffected and does not use sockets.

### Usage

- Start Decode Server
`LLM_MODE=decode SOCKET_HOST=0.0.0.0 SOCKET_PORT=9000 ./build/tt_media_server_cpp -p 8001`
- Start Prefill Server
`LLM_MODE=prefill SOCKET_HOST=10.32.17.193 SOCKET_PORT=9000 ./build/tt_media_server_cpp -p 8002`
- Test example
```
curl -X POST http://bh-glx-b09u02:8001/v1/completions \
  -H "Authorization: Bearer your-secret-key" \
  -d '{"prompt": "Hello, how are you?", "max_tokens": 15}'
```


tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -379,4 +384,175 @@ void LLMService::post_process(domain::CompletionResponse&) const {
...
 }
 
+void LLMService::setup_socket_callbacks() {
+    if (mode_ == tt::config::LLMMode::DECODE_ONLY) {
+        socket_service_->setTaskResultCallback(
+            [this](const tt::sockets::TaskResultMessage& result) {
+                handle_socket_task_result(result);
+            });
+    } else if (mode_ == tt::config::LLMMode::PREFILL_ONLY) {
+        socket_service_->setTaskForwardCallback(
```
knovokmetTT: these namings are confusing... taskForwardCallback is the function that we will execute once we recieve the task?
Not very intuitive imo... I'd propose smth like TaskReceivedCallback and ResultReceivedCallback?
What do you think @idjuricTT @ztorlakTT

idjuricTT: Why not just prefillRequst and prefillResponse? 😄

tt-media-server/cpp_server/src/services/llm_service.cpp
dmadicTT: Thinking out loud:
I see the sockets as entry points into our system, similar to controllers.
Long-term I would think of having them on top of the service, where socket-controller can talk to socket-controller rather than service-to-service.
This would keep the service layer the same for both regular run and disaggregated run.

Image

PR #2255, opened 2026-03-03
fix: Auto-disable trace capture for models with built-in warmup

stisiTT:
## Summary

Extracted from #2129 for the v0.10.0 release.

- Auto-disable trace capture for models with `has_builtin_warmup=True` in the vLLM API server and evals, unless explicitly overridden by the user
- Auto-disable trace capture in stress tests for models with built-in warmup

Models with built-in warmup manage trace capture internally, so the background trace capture step is redundant and can cause issues.

## Files Changed

- `vllm-tt-metal-llama3/src/run_vllm_api_server.py` — skip background trace capture when `has_builtin_warmup` is set
- `evals/run_evals.py` — auto-add `--disable-trace-capture` flag when `has_builtin_warmup` is set
- `stress_tests/stress_tests_args.py` — auto-set `disable_trace_capture=True` for models with built-in warmup

## Test plan

- [ ] Verify models with `has_builtin_warmup=True` skip trace capture automatically
- [ ] Verify models without `has_builtin_warmup` are unaffected
- [ ] Verify explicit `--disable-trace-capture` flag still works as before

### DIFF
--- a/stress_tests/stress_tests_args.py
+++ b/stress_tests/stress_tests_args.py
@@ -89,7 +89,9 @@ class StressTestsArgs:
             device=runtime_config.device,
             service_port=runtime_config.service_port,
             # From RuntimeConfig (stress tests configuration)
-            disable_trace_capture=runtime_config.disable_trace_capture,
+            # Auto-disable trace capture if model has builtin warmup and user hasn't explicitly overridden
+            disable_trace_capture=runtime_config.disable_trace_capture
+            or getattr(model_spec, "has_builtin_warmup", False),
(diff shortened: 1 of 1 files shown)


bgoelTT:
Hmm okay for some reason when running just the `evals` workflow I still see trace capture occurring in both the server and from the evals workflow https://github.com/tenstorrent/tt-shield/actions/runs/22629542720/job/65575462510#step:21:880

@stisiTT can you please investigate?

bgoelTT:
Bypassing merge gate because only failure was due to cpp server threshold being off by <0.5ms

Given this PR does not change anything in `tt-media-server/` I think this is okay

Issue #2231, opened 2026-02-26
Throw 429 over max number of requests


idjuricTT:
Add max number of requests in task quue.

If task queue is full, throw an error on API level to not allow more users

dmadicTT:
@rdraskicTT I guess we can close?

rdraskicTT:
Closed with https://github.com/tenstorrent/tt-inference-server/pull/2316

PR #2316, opened 2026-03-06
Limit number of running requests in cpp_server

rdraskicTT:
https://github.com/tenstorrent/tt-inference-server/issues/2231

### DIFF
--- a/.github/workflows/test-gate.yml
+++ b/.github/workflows/test-gate.yml
@@ -554,7 +554,7 @@ jobs:
       - name: Start C++ server (test runner mode)
         run: |
           cd cpp_server/build
-          ./tt_media_server_cpp -p 8000 > ../../cpp_server.log 2>&1 &
+          MAX_QUEUE_SIZE=50 ./tt_media_server_cpp -p 8000 > ../../cpp_server.log 2>&1 &
           echo $! > ../../cpp_server.pid
           cd ../..
           for i in $(seq 1 30); do
@@ -604,7 +604,8 @@ jobs:
 
           COMPLETED=$(jq -r '.completed // 0' "$RESULT_FILE")
           FAILED=$(jq -r '.failed // 0' "$RESULT_FILE")
-          if [ "$COMPLETED" -eq 0 ] && [ "$FAILED" -gt 0 ]; then
+
+          if [ "$COMPLETED" -eq 0 ]; then
             echo "❌ All requests failed (completed=0, failed=$FAILED)"
             exit 1
           fi
@@ -630,6 +631,10 @@ jobs:
               echo "❌ mean_ttft_ms $MEAN_TTFT_MS exceeds threshold ${TTFT_THRESHOLD_MS}ms"
               FAIL=1
             fi
+          fi
+          if [ "$FAILED" -eq 0 ]; then
+            echo "❌ No requests were denied (expected some 429s with small MAX_QUEUE_SIZE)"
+            FAIL=1
           fi
 
           echo "MEAN_TPOT_MS=$MEAN_TPOT_MS" >> $GITHUB_ENV
--- a/.github/workflows/test-gate.yml
+++ b/.github/workflows/test-gate.yml
@@ -554,7 +554,7 @@ jobs:
       - name: Start C++ server (test runner mode)
         run: |
           cd cpp_server/build
-          MAX_QUEUE_SIZE=50 ./tt_media_server_cpp -p 8000 > ../../cpp_server.log 2>&1 &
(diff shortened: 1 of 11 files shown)


.github/workflows/test-gate.yml
```diff
@@ -554,7 +554,7 @@ jobs:
       - name: Start C++ server (test runner mode)
         run: |
           cd cpp_server/build
-          ./tt_media_server_cpp -p 8000 > ../../cpp_server.log 2>&1 &
+          MAX_QUEUE_SIZE=50 ./tt_media_server_cpp -p 8000 > ../../cpp_server.log 2>&1 &
```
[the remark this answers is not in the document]
rdraskicTT: That's a good question. In my understanding this shouldn't affect tpot significantly which is what we care more about, while ttft is obviously affected by the number queued requests which is why I think this is a good test for ttft (if it starts growing despite limited queue then something definitely went wrong).

tt-media-server/cpp_server/include/services/base_service.hpp
```diff
@@ -33,6 +33,7 @@ class IService {
     virtual void stop() = 0;
     virtual bool is_model_ready() const = 0;
     virtual SystemStatus get_system_status() const = 0;
+    virtual bool is_queue_full() const { return false; }
```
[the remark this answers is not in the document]
rdraskicTT: Done

tt-media-server/cpp_server/include/config/constants.hpp
```diff
@@ -102,6 +102,7 @@ namespace defaults {
     constexpr const char* LLM_MODE = "regular";  // "regular", "prefill", "decode"
     constexpr const char* SOCKET_HOST = "localhost";
     constexpr uint16_t SOCKET_PORT = 9000;
+    constexpr size_t MAX_QUEUE_SIZE = 1000;
```
madicd: Seems like it was 10k, now is 1k. https://github.com/tenstorrent/tt-inference-server/pull/2316/changes#diff-7bd89652bc12aa520a9af68592d46af3228caee79b13104ddfdc37709a84cc82R89
Is there a reason we lower it down?

rdraskicTT: Just applied a suggestion https://github.com/tenstorrent/tt-inference-server/pull/2316#discussion_r2904318446

idjuricTT: @dmadicTT  I was a bit worried about 10k, but if you guys want to keep it feel free to do that

dmadicTT: Alright, we need to find max queue size that has reasonable TTFT and ITL.
The problem is, we will have multiple targets, so we need to define what **reasonable** means for
* Llama
* Blitz decode with prefill on blitz decode

tt-media-server/cpp_server/include/services/base_service.hpp
```diff
@@ -5,13 +5,19 @@
...
 
 #include "domain/base_request.hpp"
 #include "domain/base_response.hpp"
 namespace tt::services {
 
+class QueueFullException : public std::runtime_error {
+public:
+    QueueFullException() : std::runtime_error("Request queue is full, please retry later") {}
+};
+
```
idjuricTT, line 21: Suggestion: You could add in base_service pre_process() method queue size check and make every service call super.pre_process() -> that would make a unique check everywhere

rdraskicTT: It looks like we need to put `std::lock_guard lock(impl_->queue_mutex_);` before accessing queue properties in `embedding_service.cpp`, am I correct?

idjuricTT: It should be fine to do, we need to review embedding flow anyway

tt-media-server/cpp_server/include/services/base_service.hpp
```diff
@@ -49,8 +56,13 @@ class BaseService : public IService {
 
 protected:
     virtual ResponseType process_request(RequestType request) = 0;
-    virtual void pre_process(RequestType& request) const = 0;
+    virtual void pre_process(RequestType& /*request*/) const {
+        if (current_queue_size() >= max_queue_size_) throw QueueFullException{};
```
idjuricTT, line 60: Nice :smile

idjuricTT (approved):

madicd (approved):

dmadicTT (approved):

PR #2319, opened 2026-03-06
Update whisper tt-metal commit from cce3da6 to 3b3f62b for WH

stisiTT:
## Summary

- Fixes whisper server startup failure: `No module named 'ttnn.model_preprocessing'`
- The Docker image built from `cce3da6` no longer ships `ttnn.model_preprocessing`; commit `3b3f62b` resolves this
- Updates `audio_tts_templates` in `workflows/model_spec.py` to point to the correct tt-metal commit

### DIFF
--- a/default_model_spec.json
+++ b/default_model_spec.json
@@ -17390,7 +17390,7 @@
           "model_name": "whisper-large-v3",
           "inference_engine": "media",
           "device_type": "N150",
-          "tt_metal_commit": "cce3da6",
+          "tt_metal_commit": "3b3f62b",
           "device_model_spec": {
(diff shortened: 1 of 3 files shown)


tstescoTT:
@stisiTT do we still need whisper on tt-metal `3b3f62b`? Also are the reports changes needed any more?

stisiTT:
> @stisiTT do we still need whisper on tt-metal `3b3f62b`? Also are the reports changes needed any more?

Reports changes still needed and yes we should uplift whisper models. I have successful runs on WH and BH for this commit.

PR #2324, opened 2026-03-07
Modify runner to handle mpi loopback from output rank to input rank

skhorasganiTT:


skhorasganiTT:
Closing as we kept the physical loopback design in the model and were able to remove the mpi loopback workaround.

PR #2354, opened 2026-03-10
pre-commit default_model_spec.json handling

tstescoTT:
# change log

* make git hooks automatically generate and stage `default_model_spec.json` to avoid precommit clashes
* add option for direct git hooks in docs
* add support for searching common locations for pre-commit python venv

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -36,6 +36,7 @@ git add default_model_spec.json
 STAGED_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep '\.py$')
 
 if [ -z "$STAGED_FILES" ]; then
+    echo "✅ No staged Python files, skipping Ruff checks"
     exit 0
 fi
 
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -60,6 +60,8 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+echo "✅ Ruff checks passed!"
+
 PYTEST_PYTHON=""
 for env_bin in $TOOL_ENV_BINS; do
     candidate="$env_bin/python3"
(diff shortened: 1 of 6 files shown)


scripts/hooks/pre-commit
```diff
@@ -48,3 +61,35 @@ if [ $? -ne 0 ]; then
...
+    echo "⚠️  No candidate virtualenv python found, skipping pytest"
+else
+    "$PYTEST_PYTHON" -m pytest -p no:warnings
+    if [ $? -ne 0 ]; then
+        echo "❌ pytest failed - fix errors above before committing"
+        exit 1
+    fi
+fi
+
+echo "✅ pytest passed!"
```
vpetrovicTT: Should this be logged only if the pytests are rly done? It's little confusing IMO, so maybe not to echo it if they are skipped? :)

scripts/hooks/pre-commit
```diff
@@ -3,13 +3,17 @@
...
+# Common virtualenv bin directories used by this repo.
+TOOL_ENV_BINS="./.pre-commit/bin ./.venv/bin /localdev/$USER/tt-metal/python_env/bin"
+
 RUFF=$(command -v ruff 2>/dev/null)
 if [ -z "$RUFF" ]; then
-    candidate="/localdev/$USER/tt-metal/python_env/bin/ruff"
-    if [ -x "$candidate" ]; then
-        RUFF="$candidate"
-    fi
+    for candidate in "$TOOL_ENV_BINS"; do
```
vpetrovicTT: Are we shure $TOOL_ENV_BINS should be quoted?? My concern is that it prevents word splitting in the string. Looking down to pytest section, I see that the var is without quotes in the loop.

tstescoTT: unquoted it

scripts/hooks/pre-commit
tstescoTT: I moved log into branch after pytests do run.

PR #2366, opened 2026-03-10
Add --bind-host for docker server port publishing

bonnieking:
Currently the docker server launched via `run.py --docker-server`
publishes the service port using:

    --publish <port>:<port>

Docker interprets this as binding to `0.0.0.0`, which exposes the
inference server on all host interfaces.

For local development or single-node setups it is often desirable to
restrict the API to localhost only.

This PR adds a `--bind-host` CLI option (default `0.0.0.0`) and uses it
when publishing the port:

    --publish <bind-host>:<port>:<port>

Example:

    python run.py --workflow server --docker-server --bind-host 127.0.0.1

This keeps the current default behavior unchanged while allowing users
to run the server in a localhost-only configuration.

### DIFF
--- a/run.py
+++ b/run.py
@@ -140,6 +140,12 @@ def parse_arguments():
         default=os.getenv("SERVICE_PORT", "8000"),
     )
     parser.add_argument(
+        "--bind-host",
+        type=str,
+        default="0.0.0.0",
+        help="Host interface to bind published docker service port to (default: 0.0.0.0)",
+    )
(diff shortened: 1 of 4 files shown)


tstescoTT:
Thank you for the contribution! Could you make the PR target branch `dev`? That is currently what we are using for development.

bonnieking:
Yes, thank you for looking!

bonnieking:
I will open a new PR

PR #2385, opened 2026-03-11
Fix TTFT Overhead

ljovanovicTT:
## Problem: 
Previously, the default constructor of `TaskID` invoked the expensive `boost::uuids::random_generator()` operation, even though the generated ID was immediately overwritten with the `task_id` value from the request JSON, causing ~8ms increase in TTFT


idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -434,8 +434,8 @@ void LLMService::submit_decode_continuation(
     std::vector<int64_t> token_ids(prompt.begin(), prompt.end());
 
     auto sequence = std::make_unique<llm_engine::Sequence>(
+        llm_engine::TaskID(task_id),
```
madicd, line 442: Can we keep the `task_id` in the constructor?
This would allow us to never have Task ID without a `task_id` (or with an empty one).

tt-media-server/cpp_server/tests/sequence_test.cpp
```diff
@@ -157,8 +157,7 @@ TEST(SequenceTest, SerializeDeserialize_RoundTrip_PreservesAllFields) {
 }
 
 TEST(SequenceTest, SerializeDeserialize_EmptyTokenIds) {
-  Sequence orig(256, {}, SamplingParams{.max_tokens = 10});
-  orig.task_id = TaskID();
+  Sequence orig(TaskID("seq-empty-tokens"), 256, {}, SamplingParams{.max_tokens = 10});
```
madicd, line 160: Detail: 
I tend to use realistic data in the tests for 2 reasons
* They serve as documentation for ppl that do not know the domain. If I am new to this codebase I would prefer seeing how Task ID really looks like.
* The quality of tests is impacted by the quality of the data used in the tests. If we don't use realistic data, are the test results realistic?

Both are very much details for this use case, but I have built this habit over time and thought of sharing it.
Does it make sense?

dmadicTT (approved):

PR #2397, opened 2026-03-11
Initial: Multihost deployment support

### LINKED ISSUE #2411: [Multihost] Initial Introduction of Multi-Host Support
Introduce initial multi-host support based on the design document.

Design Doc is [Multi-Host Support Design for tt-inference-server](https://docs.google.com/document/d/1C5gZ5avDb-gokC7mLPwfWsnEgO6UHfAlChlcRoZ9itA/edit?tab=t.0#heading=h.zijz6zjdbbbo)

NOTE: This document is only accessible to internal TT members.

## Support Models
- DeepSeek R1

## Support Multi devices
- Dual Galaxy (8x8)
- Quad Galaxy (16x8)

sott0n:
Add multi-host deployment capability for running vLLM across multiple nodes like Dual/Quad Galaxy. Enables container orchestration and lifecycle management via MPI.

Design Doc is [Multi-Host Support Design for tt-inference-server
](https://docs.google.com/document/d/1C5gZ5avDb-gokC7mLPwfWsnEgO6UHfAlChlcRoZ9itA/edit?tab=t.0#heading=h.zijz6zjdbbbo)
This document is only accessible to internal TT members.

Close #2411

## Core Components

- **Multihost Orchestrator**: Container lifecycle management (start/stop/signal handling)
- **Multihost Config**: MPI config, SSH config, rankfile generation
- **Multihost Validation**: Pre-flight validation for Multi node (SSH connectivity for MPI, Docker, FW/KMD compatibility, permissions to read/write to shared storage)
- **New a docker image for Multihost**: This multihost deployment needs Controller and Worker containers
  - named: `ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-multihost-ubuntu-22.04-amd64`
  - It is based on `vllm-tt-metal-src-release-ubuntu-22.04-amd64` and extended with SSH configuration for multi-host support.

## Support Model and device
- Model: DeepSeek-R1
- Device: Dual WH Galaxy6U
- Device: Quad WH Galaxy6U

### User Configuration Required for Multi-Host Deployment

Users must configure the following environment variables in `.env` file or provide them interactively when running `run.py` with a multihost device:

### Required Settings
| Variable | Description | Example |
|----------|-------------|---------|
| `MULTIHOST_HOSTS` | Comma-separated list of hostnames | `host1,host2` (Dual Galaxy) or `host1,host2,host3,host4` (Quad Galaxy) |
| `MPI_INTERFACE` | Network interface for MPI communication | `cnx1`, `eth0` |
| `SHARED_STORAGE_ROOT` | Shared storage path accessible from all hosts | `/mnt/shared` |

### Optional
| Variable | Description | Default |
|----------|-------------|---------|
| `CONFIG_PKL_DIR` | Directory for vLLM config pickle files | Auto-generated under `SHARED_STORAGE_ROOT` |
| `TT_SMI_PATH` | Path to `tt-smi` binary on remote hosts | `tt-smi` |

### DeepSeek Models Only
| Variable | Description |
|----------|-------------|
| `DEEPSEEK_V3_HF_MODEL` | Path to DeepSeek-V3 HuggingFace model (must be under `SHARED_STORAGE_ROOT`) |
| `DEEPSEEK_V3_CACHE` | Path to DeepSeek-V3 cache directory (must be under `SHARED_STORAGE_ROOT`) |

## Limitations
- Only the **server** workflow is currently confirmed to run stably. The Eval/Benchmark workflows have **not yet** been validated due to a hang issue on the tt-metal side.
- Validation has only been performed on **Dual Galaxy**. Testing on Quad Galaxy is planned once the environment is prepared.

## Todos
- Once PR is approved, I will push a image: `ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-multihost-ubuntu-22.04-amd64`.

---
## Follow-up work
The following items will be filed as Issues once agreement is reached.

#### 1. Integration with utils.prompt_client for monitoring

`prompt_client` monitors cache generation progress using a cache path derived either from `TT_CACHE_PATH` or from the ModelSpec. However, DeepSeek on multi-host uses its own environment variable and a cache directory structure that differs slightly from what `tt-transformers` expects. As a result, `CacheMonitor` cannot currently detect the DeepSeek cache.

Since supporting this behavior is out of scope for this PR, the cache is treated as not configured and the following logs appear. To properly support this case, the **multi-host cache directory structure will need to be incorporated into CacheMonitor in the future.

At the moment, the following logs are produced:
```
2026-03-12 05:18:56,927 - utils.prompt_client - INFO - Waiting for vLLM service to become healthy (base timeout: 1200.0s)
2026-03-12 05:18:56,927 - utils.cache_monitor - INFO - 🔍 No cache directory configured - cache monitoring disabled
2026-03-12 05:18:56,927 - utils.prompt_client - INFO - 📁 No active cache generation detected, using standard timeout:=1200.0s
2026-03-12 05:18:56,928 - utils.prompt_client - INFO - ⏳ Service not ready after 0.0s, waiting 10.0s before polling (timeout: 1200.0s)
```

#### 2. DeepSeek Benchmark/Eval Workflow Hang Issue (this is tt-metal's issue)

DeepSeek hangs when running the benchmark/Eval workflow. During investigation, the issue was confirmed to be reproducible when launching `tt-run` directly, indicating that the root cause is likely in `tt-metal` rather than the workflow itself. An issue will be filed on the `tt-metal` side for further investigation.

For now, it is recommended to use the **server workflow** only as a workaround.


docs/multihost_deployment.md
```diff
@@ -0,0 +1,403 @@
...
+|------|----------|---------|
+| 2200 | TCP | SSH for MPI process spawning (Worker containers) |
+| 8000 | TCP | vLLM API server (Controller container) |
+| MPI ports | TCP | MPI communication (dynamic, via `MPI_INTERFACE`) |
+
+### Environment Variables
+
+| Variable | Required | Description |
+|----------|----------|-------------|
+| `HF_TOKEN` | Yes | HuggingFace token for model downloads |
```
bgoelTT:
```suggestion
| `HF_TOKEN` | Yes | HuggingFace token for gated model downloads |
```

docs/multihost_deployment.md
```diff
@@ -0,0 +1,403 @@
...
+    --workflow server \
+    --docker-server \
+    --tt-device <dual_galaxy|quad_galaxy>
+```
+
+### Required Environment Variables
+
+| Variable | Description | Example |
+|----------|-------------|---------|
+| `MULTIHOST_HOSTS` | Comma-separated list of hostnames | `host1,host2` |
```
bgoelTT, line 130: Is the first host in this list automatically defined as rank-0 in the generated rankfile?

docs/multihost_deployment.md
```diff
@@ -0,0 +1,403 @@
...
+    --name tt-worker-<rank> \
+    --user root \
+    --net host \
+    --pid host \
+    --device /dev/tenstorrent:/dev/tenstorrent \
+    --mount type=bind,src=/dev/hugepages-1G,dst=/dev/hugepages-1G \
+    --mount type=bind,src=${CONFIG_PKL_DIR},dst=${CONFIG_PKL_DIR} \
+    --mount type=bind,src=<pubkey-path>,dst=/tmp/authorized_keys.pub,readonly \
+    -e MESH_DEVICE=<mesh-device> \
+    -e VLLM_TARGET_DEVICE=tt \
```
bgoelTT: Don't both of these get passed through by `ttrun` from the Controller?

tests/test_multihost_config.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+class TestGetRankBindingPath:
+    def test_dual_galaxy(self):
+        path = get_rank_binding_path(2)
+        assert "dual_galaxy_rank_bindings.yaml" in path
+        assert "tt-metal" in path
+
+    def test_quad_galaxy(self):
+        path = get_rank_binding_path(4)
```
bgoelTT: Can this instead take a `workflows/workflow_types::DeviceTypes` instead of an int?

vllm-tt-metal/src/run_vllm_api_server.py
```diff
@@ -492,6 +492,16 @@ def runtime_settings(model_spec_json, no_auth=False):
     logger.info(f"using model: {model_spec_json['model_id']}")
     handle_secrets(no_auth=no_auth)
 
+    # In multihost deployments, model weights are on shared storage and accessed
+    # via model-specific environment variables (e.g., DEEPSEEK_V3_HF_MODEL).
```
bgoelTT, line 496: As a follow-up, we should ask the deepseek model impl to use the standard `HF_MODEL` env var so that we may re-use the existing model setup and allow us to use the standard way of mounting weights / cache path in tt-inference-server.

docs/multihost_deployment.md
```diff
@@ -0,0 +1,403 @@
...
+    --mount type=bind,src=${CONFIG_PKL_DIR},dst=${CONFIG_PKL_DIR} \
+    -e MESH_DEVICE=<mesh-device> \
+    -e VLLM_TARGET_DEVICE=tt \
+    -e NCCL_SOCKET_IFNAME=${MPI_INTERFACE} \
+    -e ETH=${MPI_INTERFACE} \
+    -e HOSTS=${MULTIHOST_HOSTS} \
+    -e CONFIG_JSON='<override_tt_config_json>' \
+    <docker-image> \
+    --model <model> \
+    --tt-device <device>
```
bgoelTT, line 325: I believe this is missing the `MULTIHOST_ROLE` env var

docs/multihost_deployment.md
bgoelTT: I believe this is missing the `MULTIHOST_ROLE` env var?

workflows/multihost_config.py
```diff
@@ -0,0 +1,175 @@
...
+        rankfile_path: Path to the rankfile inside the container
+
+    Returns:
+        mpi_args string for vLLM
+    """
+    host_list = ",".join(hosts)
+    return f"--host {host_list} --map-by rankfile:file={rankfile_path} --bind-to none"
+
+
+def get_rank_binding_path(num_hosts: int) -> str:
```
bgoelTT: See above regarding taking the `DeviceTypes` enum

workflows/validate_multihost.py
```diff
@@ -0,0 +1,797 @@
...
+        from workflows.model_spec import get_runtime_model_spec
+
+        # Derive device type from number of hosts
+        device_map = {2: "dual-galaxy", 4: "quad-galaxy"}
+        if len(hosts) not in device_map:
+            logger.error(
+                f"Unsupported number of hosts: {len(hosts)}. Supported: 2 or 4"
+            )
+            sys.exit(1)
+        device = device_map[len(hosts)]
```
bgoelTT: Can we instead try to pass the ModelSpec in the `main()` argparsing like other workflows? Then we don't have to try to re-derive it

bgoelTT (changes requested):
This is my first-pass. I have reviewed all files except for `workflows/multihost_orchestrator.py` because I see that as the most complex and want to discuss some low-priority details first.

I will review this file in isolation to be extra-thorough

docs/multihost_deployment.md
sott0n: Yes. However, this is a good catch. After receiving this comment, I investigated what happens if the host that launches run.py is configured as something other than rank-0 (means swapped `host2,host1`), and found that there is indeed an issue.

vLLM uses Torch Distributed, and it turns out that the TCP rendezvous cannot be established unless the Controller launch host (the Controller always starts on the host that launches run.py) matches rank-0.

vLLM has validation for this and triggers an assertion failure, but the failure happens after the container has already started. Therefore, I’m planning to add an additional validation on the inference-server side to ensure that the host kicking off run.py is configured as rank-0 before startup.

sott0n: addressed: https://github.com/tenstorrent/tt-inference-server/pull/2397/commits/ccb94873ee689b6e799b89794cf9a3fc5ec296ed https://github.com/tenstorrent/tt-inference-server/pull/2397/changes/7f09036b5dbf8a67c193b1676de191a98694b2a4

vllm-tt-metal/src/run_vllm_api_server.py
sott0n: agree with you. I added Todo comment.

docs/multihost_deployment.md
sott0n: Good catch! addressed it.

sott0n:
Tracked some follow-up works at https://github.com/tenstorrent/tt-inference-server/issues/2410

docs/multihost_deployment.md
```diff
@@ -0,0 +1,430 @@
...
+│   │  └──────────────────────────────────────┘  │                              │
+│   └────────────────────────────────────────────┘                              │
+│              │                                                                │
+│              │ SSH (port 2200) to host2, host3, ...                           │
+│              ▼                                                                │
+│   ┌─────────────────────────────────┐  ┌─────────────────────────────────┐    │
+│   │   Host 1: Worker (rank 1)       │  │   Host 2, 3: Workers            │    │
+│   │  ┌───────────────────────────┐  │  │   (Quad Galaxy only)            │    │
+│   │  │  Worker Container         │  │  │                                 │    │
+│   │  │  - sshd (port 2200)       │  │  │  ┌───────────────────────────┐  │    │
```
idjuricTT: Does it really work this way?
For DS we're setting this up the way that every rank has device access but rank 0 has an extra interface towards the inference server.

I know you'r not changing this, just want to understand  the whole picture

idjuricTT (approved):
I honestly think the initial architecture is a bit too complex and it needs reviewing.

I understand  you guys need this to work for customers and for that purpose I think this is the best we can offer, but I think we should go back to the original design and simplify it

docs/multihost_deployment.md
sott0n: This diagram might be misleading. On the host responsible for rank-0, two containers are launched: a Controller container (as vLLM inference server) and a rank-0 Worker container. Therefore, the rank-0 worker container itself does not start the inference server.

In the current diagram, these are shown as a single component, which could cause confusion. We need to update the diagram to clearly show that two separate containers run on Host0.

idjuricTT: Thanks @sott0n ! This completely makes sense now 😄

sott0n: Thank you for this comment, updated this picture.

sott0n:
@idjuricTT We discussed this once on chat, and I agree with your point that it is complex. The objectives of implementing this functionality in tt-inference-server are, I believe, the following:

- To provide an inference server that can run immediately once a multi-host environment is ready (enabled by containerization and validation features).
- To allow users to quickly obtain model evaluation results through benchmarks/evaluations.
- To serve as a reference model for system design, such as for k8s-based deployments.

The goal is to hide as much of the environment configuration required to run TT-vLLM in a multi-host setup as possible from the user, and I believe this is the main motivation behind containerization.

That said, if there are ideas to make the design simpler, they are very welcome. If changes should be made at the timing of this PR, then we should fix them now. Alternatively, we could refactor it in the next phase to make the design simpler.

In any case, I’d be happy to discuss ideas for a simpler design.

workflows/validate_multihost.py
```diff
@@ -0,0 +1,835 @@
...
+            logger.error(f"❌ Hugepages /dev/hugepages-1G on {host}: FAILED")
+    return errors
+
+
+# =============================================================================
+# System Software Validation
+# =============================================================================
+
+
+def validate_host_system_software(
```
bgoelTT, line 398: No need to fix here, but as a TODO we should consider re-factoring and re-using as many parts from the system SW validation performed here https://github.com/tenstorrent/tt-inference-server/blob/dev/workflows/run_system_software_validation.py

workflows/multihost_orchestrator.py
```diff
@@ -0,0 +1,1105 @@
...
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+
+"""Multi-host orchestration for distributed vLLM inference.
+
+This module manages the lifecycle of Worker and Controller containers
+across multiple hosts for MPI-based distributed inference.
+
+Architecture:
+- Controller (rank-0): Runs vLLM API server, initiates MPI processes
+- Workers (rank-1+): Run sshd, MPI processes spawned by Controller via SSH
```
bgoelTT: Reflecting that rank-0 runs both the Controller and Worker-0
```suggestion
- Workers (rank-0+): Run sshd, MPI processes spawned by Controller via SSH
```

workflows/multihost_orchestrator.py
```diff
@@ -0,0 +1,1105 @@
...
+
+Execution Flow:
+1. Orchestrator starts Worker containers on remote hosts (sshd mode)
```
bgoelTT: Reflecting that rank-0 is where the orchestrator is executed from
```suggestion
1. Orchestrator starts Worker containers on local and remote hosts (sshd mode)
```

workflows/multihost_orchestrator.py
```diff
@@ -0,0 +1,1105 @@
...
+
+    hosts: list[str]
+    mpi_interface: str
+    shared_storage_root: str
+    config_pkl_dir: str
+    tt_smi_path: str = "tt-smi"
+    deepseek_hf_model: Optional[str] = None
+    deepseek_cache: Optional[str] = None
+    # Internal: tracks if config_pkl_dir was auto-generated (for cleanup)
+    _auto_generated_config_pkl_dir: bool = field(default=False, repr=False)
```
bgoelTT: This appears to be a duplicate of the class defined in https://github.com/tenstorrent/tt-inference-server/pull/2397/changes#diff-57018f3feb9f842e34add8b24d177369afa07340d7f2bf9eff735f4aa9cf8386R37

Can we re-use this?

workflows/multihost_orchestrator.py
```diff
@@ -0,0 +1,1105 @@
...
+        if self.setup_config:
+            if self.setup_config.host_model_volume_root:
+                cmd.extend(
+                    [
+                        "--mount",
+                        f"type=bind,src={self.setup_config.host_model_volume_root},dst={self.setup_config.cache_root}",
+                    ]
+                )
+            else:
+                volume_name = _generate_docker_volume_name(self.model_spec)
```
bgoelTT: I believe this can be read from `self.setup_config` https://github.com/tenstorrent/tt-inference-server/blob/4e33a9b4fbbb5ebe889d311124148947d52e504f/workflows/setup_host.py#L73

And therefore we can delete the `_generate_docker_volume_name` function

bgoelTT (approved):
Approving because I do not want to block, but please see if my last comments make sense

workflows/multihost_orchestrator.py
sott0n: Thank you for good catch, I addressed it.

workflows/validate_multihost.py
bgoelTT: Filed issue here https://github.com/tenstorrent/tt-inference-server/issues/2436

sott0n:
Pushed an image:
https://github.com/tenstorrent/tt-inference-server/pkgs/container/tt-inference-server%2Fvllm-tt-metal-src-multihost-ubuntu-22.04-amd64

PR #2400, opened 2026-03-12
Warmup method for runners

idjuricTT:
Currently just a default warmup method for runners that returns true every time

### DIFF
--- a/tt-media-server/cpp_server/include/runners/runner_interface.hpp
+++ b/tt-media-server/cpp_server/include/runners/runner_interface.hpp
@@ -16,12 +16,6 @@ public:
     virtual ~IRunner() = default;
 
     /**
-     * Start the runner and begin processing.
-     * This method should run the main inference loop.
-     */
-    virtual void run() = 0;
-
-    /**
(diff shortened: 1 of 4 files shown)


tt-media-server/cpp_server/include/runners/runner_interface.hpp
```diff
@@ -26,6 +26,22 @@ class IRunner {
...
+        return true;
+    }
+
+    void start() {
+        // Initialize resources and prepare for inference.
+        bool is_warmed_up = warmup();
+        if (!is_warmed_up) {
+            throw std::runtime_error(std::string(runner_type()) + " warmup failed");
+        }
+        run();
```
dmadicTT, line 36: Do we make this method private then?

dmadicTT (approved):

tt-media-server/cpp_server/include/runners/runner_interface.hpp
idjuricTT: run() ? Yep, it should be private, making the change

PR #2423, opened 2026-03-13
Add spec test for race condition

### LINKED ISSUE #2420: Race condition in tt-inference-server
**Type**: CI Failure / Regression
**Priority**: P0
**Related**:
- https://github.com/tenstorrent/tt-shield/actions/runs/22738987835/job/65950604345#step:21:49
- https://github.com/tenstorrent/tt-shield/actions/runs/22808169828/job/66161827150#step:21:49
- https://github.com/tenstorrent/tt-shield/actions/runs/22830886393/job/66221045871#step:21:47
- https://github.com/tenstorrent/tt-shield/actions/runs/22997316915/job/66778168301#step:21:136

**What**: We had race condition in tt-inference-server regardless of the model/HW used. It was caused by **`logger`**. There is a "fake" deadlock (fork + logging deadlock) between the child and parent for logging. Forked child process inherits a copy of the lock in an acquired state on a thread that doesn't exist. When it goes to call `acquire()` to log, it just stays waiting forever as there is no other thread to release it. Sometimes when the child process is forked, the lock is not in the `acquired` state, so things work luckily.
**Why**: To prevent future race conditions with logger
**Root Cause**: It was observed somewhat before, but it was thought of some HW hangs of the chips (and it wasn't that frequent at all). 

**Acceptance Criteria**:
- [x] Test run in CI

fivanovicTT:


tt-media-server/tests/test_logger_fork_safety.py
```diff
@@ -0,0 +1,203 @@
...
+
+            os.close(write_fd)
+            release.set()
+            bg.join()
+            _, status = os.waitpid(pid, 0)
+            result = os.read(read_fd, 16)
+            os.close(read_fd)
+            assert result == b"OK", (
+                f"Child failed to log via TTLogger — likely deadlocked "
+                f"(got {result!r}, exit status {status}). "
```
idjuricTT: Can this be a unit test on the PR gate? It doesn’t seem to interfere with the server

fivanovicTT: I wanted actually to have both - but there were some issues with CI gate:

FAILED tests/test_logger_fork_safety.py::TestDeadlockWithoutFix::test_child_deadlocks_without_lock_reset - subprocess.TimeoutExpired: Command '['/opt/hostedtoolcache/Python/3.10.19/x64/bin/python', '-c', "import logging, os, signal, threading\nhandler = logging.StreamHandler()\nlogger = logging.getLogger('deadlock_proof')\nlogger.propagate = False\nlogger.setLevel(logging.DEBUG)\nlogger.addHandler(handler)\nheld = threading.Event()\ndef hold():\n    handler.lock.acquire()\n    held.set()\n    threading.Event().wait()\nt = threading.Thread(target=hold, daemon=True)\nt.start()\nheld.wait()\npid = os.fork()\nif pid == 0:\n    signal.alarm(2)\n    logger.info('deadlock')\n    os._exit(0)\n_, status = os.waitpid(pid, 0)\nexit(0 if os.WIFSIGNALED(status) else 1)"]' timed out after 10 seconds

fivanovicTT: I will check how to add this also as test in CI gate

idjuricTT (approved):

PR #2432, opened 2026-03-13
cpp server - fix memory leak in scheduler when preempting

knovokmetTT:


tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -97,7 +97,7 @@ void Scheduler::try_schedule_decode(std::vector<Sequence*>& scheduled_seqs,
         break;
       }
     }
-    if (block_manager_.can_append(*seq) && !self_preempt) {
+    if (!self_preempt && block_manager_.can_append(*seq)) {
```
knovokmetTT, line 100: swap conditions so we never access can_append(*seq) after freeing it

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -137,6 +137,7 @@ void Scheduler::preempt(Sequence& seq) {
   seq.status_ = SequenceStatus::WAITING;
   block_manager_.deallocate(seq);
   prefill_queue_->push(seq);
+  sequences_.erase(seq.task_id);
```
knovokmetTT, line 140: another worker can pick up the task after we preempt, we will have a dangling pointer in that case here

idjuricTT (approved):

dmadicTT (approved):

ljovanovicTT (approved):

PR #2450, opened 2026-03-16
Fix various minor issues regarding cancellation logic

### LINKED ISSUE #2444: Fix various minor issues regarding cancellation logic
- Training jobs might be cancelled due to request processing timeout set in `base_service.py` which is inappropriate for the training use case - this timeout may be removed for the training service
- the `_process_job` function may need to run `cleanup_job` if it reaches the `asyncio.CancelledError`
- force logic in `cleanup_job` should be revisited
- test if job deletion deletes the corresponding metrics related to the job
- the `QUEUED` job if cancelled, should be cancelled immediately and not remain in `CANCELLING` state
- `cancel_event` may also be set inside validation loop
- in the `lifespan` function, `await get_job_manager` and `stop_workers` may be reversed

- mark completed logic should be added for `in_progress` jobs which we want to finish but not cancel
- delete logic should be implemented, in case we want to delete a model and not to wait for `cleanup_old_jobs` to do it
- both changes should upgrade the UI as well

mcupacTT:
Closes #2444 

### Changes
- If job is in `QUEUED` state, it can be cancelled immediately without waiting to become `IN-PROGRESS`
- if training is in the middle of validation, it can be cancelled immediately without waiting for the validation cycle to complete
- force cancel logic is rewritten
- if task reaches `CancellationError`, `_cleanup_job` is also run to set the `cancel_event` and softly stop the worker
- if the server is cancelled with ctrl+C, `job_manager.shutdown` is awaited before the workers are stopped (first soft stop before the hard stop)
- add `_cleanup_job` to exceptions in `_process_job` so that if the expcetion is reached, the job is softly stopped and properly cleaned up

### Other
- added `_sync_status_to_db` to unify db updates code and make the job_manager class easier for read
- fix UI not refreshing the job status after change in the job detail section


ztorlakTT (approved):

tt-media-server/utils/job_manager.py
```diff
@@ -225,14 +225,16 @@ def cancel_job(self, job_id: str) -> bool:
                 )
                 return None
 
+            # if the job is queued, we can cancel it immediately
+            if job.status == JobStatus.QUEUED:
```
idjuricTT, line 229: @ztorlakTT  any objections for video logic?
I don't think this helps us with jobs in task queue

idjuricTT (approved):

tt-media-server/utils/job_manager.py
ztorlakTT: I think it looks good

PR #2484, opened 2026-03-17
cpp - server lock free spsc queue

knovokmetTT:


tt-media-server/cpp_server/include/runners/llm_runner.hpp
```diff
@@ -34,7 +33,6 @@ class LLMRunner : public IRunner {
   ipc::TokenRingBuffer<65536>* result_queue_;
   std::unique_ptr<IModelRunner> model_runner_;
   std::unique_ptr<Scheduler> scheduler_;
-  DecodeQueue decode_queue_;
```
knovokmetTT, line 37: dead code

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -20,6 +20,7 @@ SpPipelineRunner::SpPipelineRunner(const tt::config::LLMConfig& config,
                       config.stop_token_ids.end()),
       result_queue_(resultQueue),
       task_queue_(taskQueue),
+      decode_queue_(config.max_in_flight_count),
```
knovokmetTT: not sure this queue size makes sense

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -64,8 +65,9 @@ bool SpPipelineRunner::warmup() {
   bool receivedToken = false;
 
   while (attempts < MAX_ATTEMPTS && !receivedToken) {
-    // Drain decode queue to check for results
-    for (const auto& dr : decode_queue_.drain()) {
+    std::vector<llm_engine::TokenResult> results;
+    decode_queue_.popMany(results, max_in_flight_count_);
```
knovokmetTT, line 69: same here, not sure what the optimal pop count should be

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
idjuricTT: We probably need more since token creation might be much higher than token pickup. I'd set like 30 x max_in_flight to be on the safe side, anyway it's not much (i.e. ~2000 tokens for 64 users)

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
idjuricTT: This one makes sense IMO

idjuricTT (approved):

Issue #2441, opened 2026-03-16
Visualize sweep benchmark


dmadicTT:
Depends on https://github.com/tenstorrent/tt-inference-server/issues/2247

- Run benchmarks nightly against `dev`
- Visualize them somewhere, e.g. Streamlit
- Possible to run benchmarks on-demand against any branch and compare benchmarks on the same charts `dev` vs `branch`

PR #2500, opened 2026-03-18
Nightly benchmarks against blaze DeepSeek mock

dmadicTT:
Issue: https://github.com/tenstorrent/tt-inference-server/issues/2441

We want to understand how TTFT, TPOT and throughput is impacted by ISL, OSL and concurrency.

To achieve that we run nightly tests:

- 1 user with changing ISL, OSL fixed
- 1 user with changing OSL, ISL fixed
- changing concurrency, ISL and OSL fixed

The report created by this nightly job will be visualized in Shield Streamlit 
PR: https://github.com/tenstorrent/tt-github-actions/pull/114/changes

This will allow us to track how changes on the inference server impact TTFT, TPOT and throughput.
We will be able to compare reports between `dev` and feature branches, as well as between different dates on the `dev`.
The biggest value comes from seeing do metrics grow linear or exponential and did this change over time.

For example:
phase3_concurrency_impact
phase2_osl_impact
phase1_isl_impact


.github/workflows/produce_data.yml
```diff
@@ -0,0 +1,28 @@
+name: "[internal] Collect workflow data"
```
dmadicTT, line 1: We can do this as part of the nightly cpp benchmarks instead as separate workflow

idjuricTT (approved):

idjuricTT (approved):

.github/workflows/produce_data.yml
```diff
@@ -0,0 +1,40 @@
...
+      - completed
+
+jobs:
+  produce-cicd-data:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: Collect CI/CD data
+        uses: knovokmetTT/tt-github-actions/.github/actions/collect_data@knovokmetTT/cpp-server-benchmark-visualization
```
mjeremicTT: NOTE: Don't forget to switch to official tenstorrent/tt-github-actions collect_data once the PR is merged on collect_data side

PR #2508, opened 2026-03-18
acceptance_criteria.py to disambiguate between job failures and acceptance thresholds not being met

tstescoTT:
# change log

* acceptance_criteria.py to disambiguate between job failures and acceptance thresholds not being met.Aacceptance criteria checks and publishes acceptance status independent of job fail/success
* run.py no longer errors out and sends exit code 0 on run completed

### DIFF
--- a/run.py
+++ b/run.py
@@ -624,17 +624,19 @@ def main():
         run_local_server(model_spec, runtime_config, json_fpath, setup_config)
 
     # step 5: run workflows
-    main_return_code = 0
-
     skip_workflows = {WorkflowType.SERVER}
     if WorkflowType.from_string(runtime_config.workflow) not in skip_workflows:
-        return_codes = run_workflows(model_spec, runtime_config, json_fpath)
-        if all(return_code == 0 for return_code in return_codes):
-            logger.info("Completed run.py successfully.")
+        workflow_results = run_workflows(model_spec, runtime_config, json_fpath)
+        if all(result.return_code == 0 for result in workflow_results):
+            logger.info("Completed run.py.")
         else:
-            main_return_code = 1
+            failed_workflows = [
+                f"{result.workflow_name} ({result.return_code})"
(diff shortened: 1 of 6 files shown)


mjeremicTT (approved):
As this acceptance criteria is new workflow, should we follow the same pattern as for the other workflows when writing to `workflow_logs` instead of writing directly to report.json?
1. Write acceptance_criteria results to separate folder like for `evals, benchmarks`, etc.
2. Parse that result in final report.json

mjeremicTT:
One more note:
Final JSON will be structured something like this:
```
{
  "benchmarks_summary": [...],
  "evals": [...],
  "parameter_support_tests": {...},
  "acceptance_criteria": true,
  "acceptance_blockers": {},
  "acceptance_summary_markdown": "..."
}
```

Does it make sense to restructure it under `acceptance_summary` field and have the following structure, so we continue with the same pattern:
```
{
  "benchmarks_summary": [...],
  "evals": [...],
  "parameter_support_tests": {...},
  "acceptance_summary": {
    "acceptance_criteria": true,
    "acceptance_blockers": {},
    "acceptance_summary_markdown": "..."
  }
}
```
Also, just to keep in mind, we need to parse this information and add it to database, so this would be much easier than current solution as we need to fit it to table schemas.

tstescoTT:
I'd like to address the schema of the JSON and refactor the workflows/run_reports.py code in another PR. Tracking that here: https://github.com/tenstorrent/tt-inference-server/issues/2517

tstescoTT:
> As this acceptance criteria is new workflow, should we follow the same pattern as for the other workflows when writing to `workflow_logs` instead of writing directly to report.json?
> 
> 1. Write acceptance_criteria results to separate folder like for `evals, benchmarks`, etc.
> 2. Parse that result in final report.json

The acceptance criteria is in `reports` workflow, not a new workflow.

Issue #2505, opened 2026-03-18
Set tensor cache generation timeout in model_spec.py


tstescoTT:
For example evals failed in this Models CI job due to tensor cache generation timeout https://github.com/tenstorrent/tt-shield/actions/runs/23258046260/job/67619680661#step:10:1203
```
2026-03-18 17:57:23,051 - utils.prompt_client - INFO - ⏳ Service not ready after 1181.1s, waiting 10.0s before polling (timeout: 1200.0s)
[SERVER] (APIServer pid=1) DEBUG 03-18 17:57:27 utils.py:1031] Waiting for 1 local, 0 remote core engine proc(s) to start.
2026-03-18 17:57:33,057 - utils.cache_monitor - INFO - 🔍 No cache directory configured - cache monitoring disabled
2026-03-18 17:57:33,059 - utils.prompt_client - INFO - ⏳ Service not ready after 1191.1s, waiting 10.0s before polling (timeout: 1200.0s)
[SERVER] (APIServer pid=1) DEBUG 03-18 17:57:37 utils.py:1031] Waiting for 1 local, 0 remote core engine proc(s) to start.
2026-03-18 17:57:43,066 - utils.cache_monitor - INFO - 🔍 No cache directory configured - cache monitoring disabled
2026-03-18 17:57:43,066 - run_evals.py:442 - ERROR: ⛔️ vLLM server is not healthy. Aborting evaluations.
2026-03-18 17:57:43,066 - utils.prompt_client - ERROR - ⛔ Service did not become healthy within 1200.0s
2026-03-18 17:57:43,938 - utils.py:288 - ERROR: ⛔ subprocess.Popen command failed with return code: 1
```

This needs to be configurable and set in model_spec.py so we can size it appropriately per model impl. Blanket 1200s is not enough for Llama-3.3-70B-Instruct in this case.

PR #2509, opened 2026-03-18
add tensor_cache_timeout and tensor cache stall handling.

tstescoTT:
# change log 

* addresses https://github.com/tenstorrent/tt-inference-server/issues/2505
* add tensor_cache_timeout to model_spec.py configurable per model-hardware
* make cache_monitor.py track stalled cache generation. After 180 seconds of no cache growth, the workflow aborts startup waiting, logs the stall, returns exit code 1, and run.py ultimately reports failure rather than waiting all the way to tensor_cache_timeout.

### DIFF
--- a/tests/test_cache_monitor.py
+++ b/tests/test_cache_monitor.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+from types import SimpleNamespace
+
+import pytest
+
+from utils.cache_monitor import CacheGenerationStatus, CacheMonitor
+
+
+def test_existing_tensor_cache_without_markers_is_not_treated_as_first_run(tmp_path):
+    tensor_file = tmp_path / "tensor.bin"
+    tensor_file.write_bytes(b"cached-tensor-data")
+
+    monitor = CacheMonitor(cache_dir=tmp_path)
+
+    status = monitor.get_cache_generation_status()
+
+    assert status.is_generating is False
+    assert status.is_first_run is False
+    assert status.is_stalled is False
+    assert status.file_count == 1
+    assert status.total_size_bytes == tensor_file.stat().st_size
+
+
+def test_tensor_cache_stall_detection_trips_after_no_growth_timeout(
+    tmp_path, monkeypatch
+):
+    fake_time = [1000.0]
+    monkeypatch.setattr("utils.cache_monitor.time.time", lambda: fake_time[0])
+
+    monitor = CacheMonitor(cache_dir=tmp_path)
+    tensor_file = tmp_path / "cache.bin"
+    tensor_file.write_bytes(b"1234")
+    assert monitor.mark_cache_first_run_started() is True
+
+    initial_status = monitor.get_cache_generation_status()
+    assert initial_status.is_generating is True
+    assert initial_status.is_stalled is False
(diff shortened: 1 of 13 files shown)


tests/test_cache_monitor.py
```diff
@@ -0,0 +1,74 @@
...
+    assert initial_status.is_generating is True
+    assert initial_status.is_stalled is False
+    assert initial_status.file_count == 1
+
+    fake_time[0] += 100.0
+    mid_status = monitor.get_cache_generation_status()
+    assert mid_status.is_stalled is False
+    assert mid_status.no_progress_duration == pytest.approx(100.0)
+
+    fake_time[0] += 81.0
```
bgoelTT, line 103: Let's reference `utils/cache_monitor::TENSOR_CACHE_NO_CHANGE_TIMEOUT` and derive the fake_time values from it, just in case we ever decide to change this constant

workflows/model_spec.py
```diff
@@ -291,6 +291,7 @@ class DeviceModelSpec:
     vllm_args: Dict[str, str] = field(default_factory=dict)
     override_tt_config: Dict[str, str] = field(default_factory=dict)
     env_vars: Dict[str, str] = field(default_factory=dict)
+    tensor_cache_timeout: float = 3600.0
```
bgoelTT, line 294: Can you share why you chose this as the default? I'm not disagreeing but want to make sure we're considering the avg case properly:
* n150, n300, p100, p150 should all be <30min
* n150x4, T3K, p150x8, p300x2, p150x4 should all be <60min
* single-Galaxy should also be <60min (can just override as needed)
* anything else will be overridden

workflows/model_spec.py
```diff
@@ -870,6 +871,8 @@ def expand_to_specs(self) -> List["ModelSpec"]:
                     vllm_args=device_model_spec.vllm_args,
                     override_tt_config=device_model_spec.override_tt_config,
                     env_vars=device_model_spec.env_vars,
+                    tensor_cache_timeout=device_model_spec.tensor_cache_timeout,
+                    system_requirements=device_model_spec.system_requirements,
```
bgoelTT, line 875: Ah nice catch!

bgoelTT (approved):
Looks great, thank you for saving so much CI time.

Just a few questions

workflows/model_spec.py
tstescoTT: with the stalling handling we can default to a higher max, i.e. tensor_cache_timeout shouldnt need to be a minimum but be a maximum timeout assuming the cache files are still being written, e.g. if for some reason a particular system has poor disk I/O compared to most machines we test on, e.g. due to network FS.

bgoelTT: 👍 okay sounds good.

tstescoTT:
@bgoelTT pointed out yesterday when default docker volume is used to store the tensor caches additional handling is needed.

WIP on that.

tstescoTT:
Need to bump tensor cache stalled timeout to 300s to match tt-metal op hang timeout.

tstescoTT:
> @bgoelTT pointed out yesterday when default docker volume is used to store the tensor caches additional handling is needed.
> 
> WIP on that.

Addressing this in last commit.

Tested locally.

tstescoTT:
Dispatch run: https://github.com/tenstorrent/tt-shield/actions/runs/23415756912

also ran https://github.com/tenstorrent/tt-shield/actions/runs/23413525027

taking a while to get a runner.

PR #2510, opened 2026-03-19
add --limit-samples-mode smoke-test support in benchmarking

tstescoTT:
# change log

* add --limit-samples-mode smoke-test support in benchmarking and single sample for eval
* add docs for --limit-samples-mode smoke-test usage

### DIFF
--- a/benchmarking/benchmark_config.py
+++ b/benchmarking/benchmark_config.py
@@ -3,7 +3,7 @@
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Dict, Iterable, List, Tuple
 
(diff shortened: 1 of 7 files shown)


tstescoTT:
## Automated PR review from GPT-5.4 High

**Regression risk low:**
no limit_samples_mode at all: behavior looks unchanged
other modes like ci-nightly / ci-long / ci-commit: behavior also looks unchanged
This PR does refactor limit handling, so if some external script imports these helpers directly, it may be changed.

**Feature implementation:**
The biggest benchmark risk is that smoke mode still runs full perf targets instead of one lightweight target.
The biggest eval risk is inconsistent sample limiting: text/meta/audio lm_eval tasks get --limit 3, while media evals still use their old hard-coded call counts. I do not see an obvious regression for non-smoke modes from this diff alone.

PR #2527, opened 2026-03-19
Move worker management responsibility out of base service

dmadicTT:
- Extract worker lifecycle management from BaseService into a dedicated WorkerManager component, applying the single responsibility principle
- Replace inheritance-based worker management with composition -- services own a worker manager rather than inheriting worker state
- Centralize process spawning, warmup signal handling, crash detection, restart, and worker status reporting behind a single cohesive API
- Validate worker configuration at construction time, rejecting invalid setups with actionable error messages
- Clean up BaseService to be a pure service interface focused solely on request processing and queue management


tt-media-server/cpp_server/include/services/llm_service.hpp
```diff
@@ -61,6 +63,8 @@ class LLMService
   domain::CompletionResponse processRequest(
       domain::CompletionRequest request) override;
 
+  std::vector<tt::worker::WorkerInfo> getWorkerInfo() const override;
```
idjuricTT, line 66: Do we need workerinfo and isModelReady overrides here?

tt-media-server/cpp_server/include/worker/worker_manager.hpp
```diff
@@ -0,0 +1,92 @@
...
+  void restartWorker(size_t workerIdx);
+
+ private:
+  bool isWorkerWarmed(int workerId) const;
+
+  void startWorkers();
+  void startWarmupListenerThread();
+  void waitForFirstWarmup();
+  void stopWarmupListener();
+  void stopProcesses();
```
idjuricTT, line 67: StopProcesses seems to be doing the same as stop except closing the warmup queue

tt-media-server/cpp_server/include/worker/worker_manager.hpp
```diff
@@ -0,0 +1,92 @@
...
+  bool checkWorkerAlive(size_t workerIdx);
+
+  /** Re-fork a crashed worker process and update the workers entry. */
+  void restartWorker(size_t workerIdx);
+
+ private:
+  bool isWorkerWarmed(int workerId) const;
+
+  void startWorkers();
+  void startWarmupListenerThread();
```
idjuricTT: Nitpick: probably Thread in name is too much information for the consumer

idjuricTT (approved):

tt-media-server/cpp_server/include/services/llm_service.hpp
dmadicTT: Yeah, because it's the LLM service that starts the workers, not the base service.

In future we can create worker manager independently from the service, and inject it to the service as dependency.

Then the health controller will depend on worker manager, not on base service.

tt-media-server/cpp_server/include/worker/worker_manager.hpp
dmadicTT: Yeah, just a private method to encapsulate behavior
public method `stop` calls `stopWarmupListener` and `stopProcesses`

PR #2551, opened 2026-03-23
Add /v1/chat/completions endpoint and fix streaming artifacts

### LINKED ISSUE #2544: Support v1/models api
For both media and cpp server we need an info of which model is running currently

kmabeeTT:
## Issue

https://github.com/tenstorrent/tt-inference-server/issues/2544

## Summary
- Add OpenAI-compatible `/v1/chat/completions` endpoint that applies the model's chat template via the tokenizer, supporting both streaming SSE and non-streaming responses with real token usage stats
- Fix streaming artifact where `"final_text"` literal was emitted as the last token in both `vllm_runner` and `vllm_forge_llama_70b` runner
- Fix `/v1/models` returning HF cache path instead of model name for LLM models (hugging_face_utils overwrites `model_weights_path` with local snapshot path; use `settings.vllm.model` which retains the HuggingFace ID)

## What's changed
- `open_ai_api/chat.py` — New chat completions endpoint with chat template support
- `open_ai_api/__init__.py` — Register chat router in LLM service routes
- `open_ai_api/models.py` — Fix model ID for LLM models (use `vllm.model` instead of resolved cache path)
- `domain/chat_completion_request.py` — Request schema for chat completions
- `tt_model_runners/vllm_runner.py` — Fix final_text artifact in streaming
- `tt_model_runners/vllm_forge_llama_70b.py` — Fix final_text artifact in streaming
- `tests/test_chat_api.py` — Tests for streaming, non-streaming, validation

## Test plan
- [x] Verified `/v1/chat/completions` works with streaming and non-streaming requests
- [x] Verified `final_text` no longer appears in streaming output
- [x] Verified `/v1/models` returns correct model name (not HF cache path)
- [x] Tested with Llama-3.2-3B-Instruct and Llama-3.1-8B-Instruct on Blackhole (single device)
- [x] Model served on hosted tt-cloud-console using these changes
- [x] Added unit tests for chat completions endpoint (non-streaming format, usage stats, streaming SSE, validation)

### DIFF
--- a/tt-media-server/open_ai_api/models.py
+++ b/tt-media-server/open_ai_api/models.py
@@ -38,7 +38,10 @@ def list_models():
     List current model. OpenAI-compatible endpoint.
     See: https://platform.openai.com/docs/api-reference/models/list
     """
-    model_id = settings.model_weights_path
+    if settings.model_service == "llm":
+        model_id = settings.vllm.model
+    else:
+        model_id = settings.model_weights_path
     if not model_id:
         return {"object": "list", "data": []}
 
--- a/tt-media-server/domain/chat_completion_request.py
+++ b/tt-media-server/domain/chat_completion_request.py
@@ -1,0 +1,36 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+from typing import Union
+
+from pydantic import BaseModel
+
+
+class ChatMessage(BaseModel):
+    """A single message in a chat conversation."""
+
+    role: str
+    content: str
+
+
+class ChatCompletionRequest(BaseModel):
+    """
+    OpenAI-compatible chat completion request.
+    See: https://platform.openai.com/docs/api-reference/chat/create
+    """
+
+    model: str | None = None
+    messages: list[ChatMessage]
+    max_tokens: int | None = 2048
+    temperature: float | None = None
+    top_p: float | None = None
+    top_k: int | None = None
+    repetition_penalty: float | None = None
+    frequency_penalty: float | None = 0.0
+    presence_penalty: float | None = 0.0
+    stream: bool | None = False
+    stop: Union[str, list[str], None] = []
+    n: int = 1
+    seed: int | None = None
+    user: str | None = None
--- a/tt-media-server/open_ai_api/__init__.py
+++ b/tt-media-server/open_ai_api/__init__.py
@@ -11,6 +11,7 @@ from fastapi import APIRouter
 
 from open_ai_api import (
     audio,
+    chat,
     cnn,
     embedding,
     fine_tuning,
@@ -46,6 +47,7 @@ SERVICE_ROUTER_MAP: dict[str, list[ServiceRoute]] = {
     ModelServices.LLM.value: [
         ServiceRoute(tokenizer.router, "/v1", "", ["Tokenizer"]),
         ServiceRoute(llm.router, "/v1", None, ["Text processing"]),
+        ServiceRoute(chat.router, "/v1", None, ["Chat completions"]),
     ],
     ModelServices.CNN.value: [
         ServiceRoute(cnn.router, "/v1/cnn", "/cnn", ["CNN processing"]),
--- a/tt-media-server/open_ai_api/chat.py
+++ b/tt-media-server/open_ai_api/chat.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import json
+import time
+import uuid
+from functools import lru_cache
+
+from config.settings import settings
+from domain.chat_completion_request import ChatCompletionRequest
+from domain.completion_request import CompletionRequest
+from fastapi import APIRouter, Depends, HTTPException, Security
+from fastapi.responses import JSONResponse, StreamingResponse
+from model_services.base_service import BaseService
+from resolver.service_resolver import service_resolver
+from security.api_key_checker import get_api_key
+from transformers import AutoTokenizer
+from utils.logger import TTLogger
+
+router = APIRouter()
+logger = TTLogger()
+
+
+@lru_cache(maxsize=1)
+def _get_tokenizer():
+    model_name = settings.model_weights_path
+    logger.info(f"Loading tokenizer for chat template: {model_name}")
+    return AutoTokenizer.from_pretrained(model_name)
+
+
+def _apply_chat_template(messages: list[dict]) -> str:
+    tokenizer = _get_tokenizer()
+    return tokenizer.apply_chat_template(
+        messages,
+        tokenize=False,
+        add_generation_prompt=True,
+    )
+
+
+def _count_tokens(text: str) -> int:
+    tokenizer = _get_tokenizer()
+    return len(tokenizer.encode(text))
+
+
+def _build_completion_request(
+    chat_request: ChatCompletionRequest, prompt: str
+) -> CompletionRequest:
+    return CompletionRequest(
+        model=chat_request.model,
+        prompt=prompt,
+        max_tokens=chat_request.max_tokens,
+        temperature=chat_request.temperature,
+        top_p=chat_request.top_p,
+        top_k=chat_request.top_k,
+        repetition_penalty=chat_request.repetition_penalty,
+        frequency_penalty=chat_request.frequency_penalty,
+        presence_penalty=chat_request.presence_penalty,
+        stream=chat_request.stream,
+        stop=chat_request.stop,
+        n=chat_request.n,
+        seed=chat_request.seed,
+    )
+
+
+@router.post("/chat/completions")
+async def chat_completions(
+    chat_request: ChatCompletionRequest,
+    service: BaseService = Depends(service_resolver),
+    api_key: str = Security(get_api_key),
+):
+    """
+    Create a chat completion. OpenAI-compatible endpoint.
+
+    Accepts messages in chat format, applies the model's chat template,
+    and returns completions in the chat completions response format.
+    See: https://platform.openai.com/docs/api-reference/chat/create
+    """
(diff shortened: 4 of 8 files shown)


kmabeeTT:
### Implementation notes

- `/v1/chat/completions` uses the HuggingFace tokenizer's `apply_chat_template()` to convert the chat messages array into a prompt string, then delegates to the existing VLLM runner infrastructure. The tokenizer is loaded once and cached via `lru_cache`.
- `/v1/models` reads the model name from `settings.vllm.model` so clients can auto-discover what model the server is running without hardcoding it.
- The `final_text` fix changes the sentinel `CompletionResult(text="final_text")` to `CompletionResult(text="")` in both VLLM runners, preventing the literal string from leaking into streaming output.


### More Testing

Screenshot of working integration w/ console.

image


```
# /v1/models - model discovery
curl -s http://localhost:8000/v1/models \
    -H "Authorization: Bearer your-secret-key" | python3 -m json.tool
{
    "object": "list",
    "data": [
        {
            "id": "meta-llama/Llama-3.1-8B-Instruct",
            "object": "model",
            "created": 1700000000,
            "owned_by": "tenstorrent"
        }
    ]
}

# /v1/chat/completions - non-streaming
curl -s http://localhost:8000/v1/chat/completions \
    -H "Authorization: Bearer your-secret-key" \
    -d '{"messages": [{"role": "user", "content": "What is 2+2?"}], "max_tokens": 32}' | python3 -m json.tool

{
    "id": "chatcmpl-f1141f6e606c400f87ac5b48",
    "object": "chat.completion",
    "created": 1774247226,
    "model": "meta-llama/Llama-3.1-8B-Instruct",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "2 + 2 = 4."
            },
            "finish_reason": "stop"
        }
    ],
    "usage": {
        "prompt_tokens": 43,
        "completion_tokens": 9,
        "total_tokens": 52
    }
}


# /v1/chat/completions - streaming
curl -N http://localhost:8000/v1/chat/completions \
    -H "Authorization: Bearer your-secret-key" \
    -d '{"messages": [{"role": "user", "content": "What is 2+2?"}], "max_tokens": 32, "stream": true}'

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {"content": "The"}, "finish_reason": null}]}

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {"content": " answer"}, "finish_reason": null}]}

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {"content": " is"}, "finish_reason": null}]}

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {"content": " "}, "finish_reason": null}]}

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {"content": "4"}, "finish_reason": null}]}

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {"content": "."}, "finish_reason": null}]}

data: {"id": "chatcmpl-05d3ce64ec3c4a4a8261fbc0", "object": "chat.completion.chunk", "created": 1774247263, "model": "meta-llama/Llama-3.1-8B-Instruct", "choices": [{"index": 0, "delta": {}, "finish_reason": "stop"}], "usage": {"prompt_tokens": 43, "completion_tokens": 7, "total_tokens": 50}}

data: [DONE]

```

tt-media-server/open_ai_api/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@router.get("/models")
+async def list_models(api_key: str = Security(get_api_key)):
+    """
+    List available models. OpenAI-compatible endpoint.
+    See: https://platform.openai.com/docs/api-reference/models/list
+    """
+    model_id = (
+        settings.vllm.model
```
idjuricTT: This will work for vllm models only. @fivanovicTT could you please help with this part?

kmabeeTT: This was updated based off prev feedback, so it works for both I think (hasattr(settings, 'vllm') falls back to settings.model_weights_path for non-VLLM models.)

ayerofieiev-tt (approved):

tt-media-server/open_ai_api/models.py
```diff
@@ -0,0 +1,33 @@
...
+        "object": "list",
+        "data": [
+            {
+                "id": model_id,
+                "object": "model",
+                "created": 1700000000,
+                "owned_by": "tenstorrent",
+            }
+        ],
+    }
```
fivanovicTT, line 53: Updated code:
```
@router.get("/models")
async def list_models(api_key: str = Security(get_api_key)):
  """
  List current model. OpenAI-compatible endpoint.
  See: https://platform.openai.com/docs/api-reference/models/list
  """
  model_id = settings.model_weights_path
  if not model_id:
      return {"object": "list", "data": []}
  return {
      "object": "list",
      "data": [
          {
              "id": model_id,
              "object": "model",
              "created": 1700000000,
              "owned_by": "tenstorrent",
          }
      ],
  }
```

fivanovicTT (changes requested):

tt-media-server/open_ai_api/models.py
fivanovicTT: @idjuricTT is right here. Vllm settings is always loaded and will auto default to QWEN_3_4B for non-vllm models. Will add a fix in the next comment

idjuricTT (approved):

tt-media-server/open_ai_api/models.py
kmabeeTT: Hmm @fivanovicTT ... But with the suggested code that was checked in other PR https://github.com/tenstorrent/tt-inference-server/pull/2703 yesterday I still face the problem that prompted my code in the first place where `settings.model_weights_path` is an actual path:

```
kmabee@Kyle-Mabee's-Mac ~ % curl -s http://10.32.48.16:8000/v1/models -H "Authorization: Bearer your-secret-key" | python3 -m json.tool
{
    "object": "list",
    "data": [
        {
            "id": "/home/ttuser/.cache/huggingface/hub/models--meta-llama--Llama-3.1-8B-Instruct/snapshots/0e9e39f249a16976918f6564b8830bc894c89659",
            "object": "model",
            "created": 1700000000,
            "owned_by": "tenstorrent"
        }
    ]
}
```

It was the reason for my change, so I need to keep this locally unless you have alternative fix?

```
 41 -    model_id = settings.model_weights_path
 41 +    model_id = (
 42 +        settings.vllm.model
 43 +        if hasattr(settings, "vllm") and settings.vllm.model
 44 +        else settings.model_weights_path
 45 +    )
 ```

Comment (kmabeeTT) [reply tt-media-server/open_ai_api/models.py]: Given vllm settings is always loaded, how about this as the fix?

```
  if settings.model_service == "llm":
      model_id = settings.vllm.model
  else:
      model_id = settings.model_weights_path
```

Some context from claude about what was overwriting `settings.model_weights_path` :


```
● Line 52 and 67 in hugging_face_utils.py — that's the culprit:

  Read 1 file (ctrl+o to expand)

● There it is. Lines 52 and 67:

  settings.model_weights_path = cached_path      # line 52
  settings.model_weights_path = downloaded_path   # line 67

  snapshot_download returns the full local cache path (/home/ttuser/.cache/huggingface/hub/models--meta-llama--Llama-3.1-8B-Instruct/snapshots/0e9e39f...), and this overwrites model_weights_path which started as "meta-llama/Llama-3.1-8B-Instruct".

  This is needed for model loading (runners need the local path), but it breaks /v1/models which should show the HuggingFace model ID.

  The clean fix: the /v1/models endpoint should use a different source for the display name. For LLM models, settings.vllm.model is the right one — it never gets overwritten. The author's concern about it defaulting to QWEN_3_4B is valid for non-LLM models, but you can guard on model_service:

  if settings.model_service == "llm":
      model_id = settings.vllm.model
  else:
      model_id = settings.model_weights_path

  This is safe because settings.vllm.model gets explicitly set to the correct value for LLM models at settings.py line 312. For non-LLM models, model_weights_path (even if expanded) is acceptable since they don't go through the HF download path the same way.

```

kmabeeTT:
### Tests added

**`tests/test_chat_api.py`** — covers the new `/v1/chat/completions` endpoint:

**Non-streaming:**
- Response format: `chat.completion` object, assistant role, `finish_reason`
- Usage stats: `prompt_tokens`, `completion_tokens`, `total_tokens`
- Completion ID format (`chatcmpl-*`)
- Chat template application via tokenizer
- Parameter passthrough: `temperature`, `top_p`, `max_tokens`

**Streaming:**
- SSE format with correct `text/event-stream` content-type
- Final chunk with `finish_reason=stop` and usage stats
- `[DONE]` sentinel

**Validation:**
- 422 on missing `messages` field
- Empty messages list accepted

### Note

`/v1/models` changes from earlier revision have been dropped — that support was merged separately via #2703.

tt-media-server/open_ai_api/models.py
kmabeeTT: Pushed this change, cleaned up, rebased, included tests and would like to merge :)

fivanovicTT (approved):

Issue #2264, opened 2026-03-03
Add cancellation feature


idjuricTT:
Add an option to stop decoding

rdraskicTT:
Initial support finalized, remaining work will be tracked here https://github.com/tenstorrent/tt-inference-server/issues/2732

PR #2555, opened 2026-03-23
Add request cancellation support

rdraskicTT:
https://github.com/tenstorrent/tt-inference-server/issues/2264

  Summary

  - Introduce per-worker IPC cancel queues (BoostIpcCancelQueue) so the main process can broadcast abort signals to worker processes when a streaming client disconnects                       
  - Add Scheduler::abortRequest() which removes sequences from prefill/decode queues, frees KV cache blocks, and is idempotent for repeated calls; includes pending_aborts_ set to handle races
   where abort arrives before the sequence is dequeued                                                                                                                                         
  - Add LLMService::abortRequest() which sends a final finish_reason="abort" chunk to unblock synchronous waiters, then broadcasts to all worker cancel queues
  - Register POST /v1/responses/{response_id}/cancel endpoint
                                                                                                                                                                                               
  Test plan                                                                                                                                                                                    
                                                                                                                                                                                               
  - Unit tests (cancellation_test.cpp): scheduler abort removes sequences and frees blocks, IPC cancel queue push/drain round-trip, LLMRunner cancel stops token emission                      
  - E2e test (test_cancellation_e2e.py): streaming client disconnect doesn't crash server, subsequent requests succeed, multiple rapid disconnects handled
  - E2e test harness (run_e2e_with_server.sh): starts mock server, runs e2e suite, tears down

### DIFF
--- a/tt-media-server/cpp_server/include/ipc/boost_ipc_cancel_queue.hpp
+++ b/tt-media-server/cpp_server/include/ipc/boost_ipc_cancel_queue.hpp
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include "domain/task_id.hpp"
+#include "ipc/cancel_queue.hpp"
 
 namespace tt::ipc {
 
@@ -25,7 +26,7 @@ constexpr size_t CANCEL_QUEUE_CAPACITY = 1024;
  * push() is non-blocking (drops the message and logs a warning when full).
  * tryPopAll() drains every available message without blocking.
  */
-class BoostIpcCancelQueue {
+class BoostIpcCancelQueue : public ICancelQueue {
  public:
   /** Create the queue (main-process side). */
   BoostIpcCancelQueue(const std::string& name, size_t capacity);
@@ -33,22 +34,15 @@ class BoostIpcCancelQueue {
   /** Open an existing queue (worker-process side). */
   explicit BoostIpcCancelQueue(const std::string& name);
 
-  ~BoostIpcCancelQueue();
+  ~BoostIpcCancelQueue() override;
 
-  /**
-   * Non-blocking push. Silently drops the message if the queue is full
-   * (the request will finish on its own shortly anyway).
-   * Thread-safe.
-   */
-  void push(const tt::domain::TaskID& taskId);
+  void push(const tt::domain::TaskID& taskId) override;
 
-  /**
-   * Drain all available cancel messages into out without blocking.
-   * Intended to be called once per scheduler step from the worker thread.
-   */
-  void tryPopAll(std::vector<tt::domain::TaskID>& out);
+  void tryPopAll(std::vector<tt::domain::TaskID>& out) override;
 
-  static void remove(const std::string& name);
+  void remove() override;
+
+  static void removeByName(const std::string& name);
 
  private:
   std::string name_;
--- a/tt-media-server/cpp_server/include/ipc/boost_ipc_cancel_queue.hpp
+++ b/tt-media-server/cpp_server/include/ipc/boost_ipc_cancel_queue.hpp
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+#pragma once
+
+#include <boost/interprocess/ipc/message_queue.hpp>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "domain/task_id.hpp"
+
+namespace tt::ipc {
+
+constexpr const char* CANCEL_QUEUE_NAME = "tt_cancels";
+constexpr size_t CANCEL_QUEUE_CAPACITY = 1024;
+
+/**
+ * Lightweight IPC message queue for carrying request-cancel signals from the
+ * main process to worker processes. Each message is exactly one serialized
+ * TaskID (36 bytes).
+ *
+ * One queue per worker; the main process creates each queue and the
+ * corresponding worker process opens it to drain cancel signals.
+ * push() is non-blocking (drops the message and logs a warning when full).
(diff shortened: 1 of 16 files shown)


tt-media-server/cpp_server/include/api/llm_controller.hpp
```diff
@@ -42,6 +44,16 @@ class LLMController : public drogon::HttpController<LLMController> {
       const drogon::HttpRequestPtr& req,
       std::function<void(const drogon::HttpResponsePtr&)>&& callback) const;
 
+  /**
+   * POST /v1/responses/{response_id}/cancel
```
dmadicTT: This does not seem fully compatible with OpenAI.
We implement Chat Completions API, not the Responses API.

Also Reponses API supports cancellation only for background requests: https://developers.openai.com/api/reference/resources/responses/methods/cancel

My quick research shows Chat Completions API does not support cancellation endpoint.
We probably need to detect the cancellation from the HTTP session disconnect.
How does vLLM/SGLang handle this for Chat Completions?

tt-media-server/cpp_server/include/ipc/boost_ipc_cancel_queue.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * Lightweight IPC message queue for carrying request-cancel signals from the
+ * main process to worker processes. Each message is exactly one serialized
+ * TaskID (36 bytes).
+ *
+ * One queue per worker; the main process creates each queue and the
+ * corresponding worker process opens it to drain cancel signals.
+ * push() is non-blocking (drops the message and logs a warning when full).
+ * tryPopAll() drains every available message without blocking.
+ */
+class BoostIpcCancelQueue {
```
dmadicTT: I would make abstract interface for all queues, as we might want to change the implementation from Boost to Shared Memory to something else.
I have done this in the warmup signals queue https://github.com/tenstorrent/tt-inference-server/blob/ljovanovic/feature_handle_multitoken_representation/tt-media-server/cpp_server/include/ipc/warmup_signal_queue.hpp

Do you agree @ztorlakTT @knovokmetTT ?

knovokmetTT: In my opinion, the direction we should strive for is a unified , templated Queue interface for ipc communication

tt-media-server/cpp_server/include/utils/concurrent_map.hpp
```diff
@@ -54,6 +54,15 @@ class ConcurrentMap {
     map_.clear();
   }
 
+  template <typename Func>
```
knovokmetTT: can we use concepts here to safeguard

```
#include <concepts>
......
......
template <std::invocable<Value&> Func>
bool update(const Key& key, Func&& func) {
    std::lock_guard lock(mutex);
    auto it = map_.find(key);
    if (it == map_.end()) return false;
    
    func(it->second); 
    return true;
}
```

rdraskicTT: Good idea, wasn't aware of this, will take a look

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -99,6 +101,30 @@ void SpPipelineRunner::stop() {
 }
 
 void SpPipelineRunner::step() {
+  // Drain cancel queue before scheduling so aborted sequences are excluded.
+  if (cancel_queue_) {
```
knovokmetTT: can this cancel logic/check be encapsulated in its own function?

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -15,11 +15,13 @@ namespace tt::runners {
...
+                                   llm_engine::ITaskQueue* taskQueue,
+                                   ipc::BoostIpcCancelQueue* cancelQueue)
     : config(config),
       stopTokenIds(config.stop_token_ids.begin(), config.stop_token_ids.end()),
       resultQueue(resultQueue),
       taskQueue(taskQueue),
+      cancel_queue_(cancelQueue),
       decodeQueue(config.max_in_flight_count),
       maxInFlightCount(config.max_in_flight_count * 30) {
   auto decodeCb = [this](const llm_engine::TokenResult& result) {
```
knovokmetTT, line 27: not sure if this is needed, but if we canceled a request, and pipeline still returns a token result here, this callback should ignore it? please check :)

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -174,4 +193,40 @@ void Scheduler::postprocess(std::vector<Sequence*>& seqs,
 }
 
 void Scheduler::removeSequence(TaskID taskId) { sequences_.erase(taskId); }
+
+void Scheduler::abortRequest(TaskID taskId) {
+  auto* seq = findSequence(taskId);
+  if (!seq || seq->isAborted() || seq->isFinished()) {
```
knovokmetTT, line 199: this is very messy, proposed refactor:

```
if (seq && (seq->isAborted() || seq->isFinished())) {
    return;
}

if (!seq) {
    if (pending_aborts_.size() < 2 * max_in_flight_count_) {
        pending_aborts_.insert(taskId);
    }
    return;
}
```

tt-media-server/cpp_server/include/api/llm_controller.hpp
rdraskicTT: I think you are right, session disconnect seems to be only way to cancel these requests according to the api and other servers.

PR #2569, opened 2026-03-24
cpp server ds runner pos id reset

knovokmetTT:


tt-media-server/cpp_server/src/runners/runner.py
```diff
@@ -146,6 +145,10 @@ def _open_mesh_device(
         return ttnn.open_mesh_device(mesh_shape=ttnn.MeshShape(4, 2))
 
 
+def _run_dummy_inference(model_pipeline: ModelPipeline) -> None:
+    while not _shutdown:
+        model_pipeline.run_inference(prompt_token_ids=None, max_new_tokens=1)
```
idjuricTT, line 151: Should we have some sleep in between? Not sure how this will affect the pipeline

idjuricTT (approved):

skhorasganiTT (approved):

tt-media-server/cpp_server/src/runners/runner.py
knovokmetTT: it has a barrier inside run_infernece, non rank 0 will block immediately, until rank 0 also calls run_infenrece

skhorasganiTT: Yeah the barrier inside run_inference should be sufficient. But maybe worth adding a comment about it here @knovokmetTT.

PR #2580, opened 2026-03-24
cpp server - encapsulate disaggregated service logic

knovokmetTT:


tt-media-server/cpp_server/include/utils/service_factory.hpp
```diff
@@ -3,50 +3,15 @@
...
-
-/**
- * Returns the currently configured service (LLM or Embedding) as IService.
- * Use when only interface methods are needed (e.g. get_system_status).
+ * Build, wire, and start all services based on current configuration.
+ * Populates the ServiceContainer singleton.
+ * Called once from main(), before Drogon starts.
  */
-std::shared_ptr<services::IService> getConfiguredService();
+void initializeServices();
```
idjuricTT, line 16: We might not need / want all the services at once.
i.e. in this case Embedding service does nothing. We might get more "dead" services once we add video and image.

Maybe it'd be better to have service pairs running

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -314,131 +316,134 @@ void LLMController::handleStreaming(
...
-          }
-        });
+          (*streamPtr)->send("data: [DONE]\n\n");
+          (*streamPtr)->close();
+        }
+      });
+    }
+  };
+  try {
+    if (tt::config::llmMode() == tt::config::LLMMode::REGULAR) {
```
idjuricTT, line 438: I found this to be a bit confusing at the beginning since we have 
llm controller -> disaggregation service -> socket service & llm service but it actually makes sense 😄 

For the future work I'd add a small filter here named "shouldCallPrefill" that would just pass the request is: 

- over certain configurable size to cover for the request to not send all to prefill and
- only first request in the conversation (requests 1+ in conversation always go to decode only)

idjuricTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
knovokmetTT: I think that filter should be in disaggregation service actually, not here (deciding if we should go to prefill or decode)

tt-media-server/cpp_server/include/utils/service_factory.hpp
knovokmetTT: we have if checks in service_factory, not to create services that we will not use

tt-media-server/cpp_server/include/utils/service_container.hpp
```diff
@@ -0,0 +1,51 @@
...
+      std::shared_ptr<services::EmbeddingService> embedding,
+      std::shared_ptr<sockets::InterServerService> socket,
+      std::shared_ptr<services::DisaggregationService> disaggregation);
+
+  std::shared_ptr<services::IService> configuredService() const;
+
+  std::shared_ptr<services::LLMService> llm;
+  std::shared_ptr<services::EmbeddingService> embedding;
+  std::shared_ptr<sockets::InterServerService> socket;
+  std::shared_ptr<services::DisaggregationService> disaggregation;
```
ztorlakTT: Can we enforce that only ServiceContainer itself can set the pointers:
```suggestion
  std::shared_ptr<services::LLMService> llm() const { return llm_; }
  std::shared_ptr<services::EmbeddingService> embedding() const { return embedding_; }
  std::shared_ptr<sockets::InterServerService> socket() const { return socket_; }
  std::shared_ptr<services::DisaggregationService> disaggregation() const { return disaggregation_; }
```

tt-media-server/cpp_server/include/services/llm_service.hpp
```diff
@@ -31,8 +27,8 @@ class LLMService
       public Streamable<domain::CompletionRequest,
                         domain::StreamingChunkResponse> {
  public:
-  using PrefillRequestCallback =
-      std::function<bool(const domain::PrefillRequest&)>;
+  using StreamCallback =
+      std::function<void(domain::StreamingChunkResponse&, bool)>;
```
ztorlakTT: Why don't we have the same signature as in `DisaggregationService`
```suggestion
  using StreamCallback =
      std::function<void(const domain::StreamingChunkResponse&, bool)>;
```

ztorlakTT (approved):

Issue #2578, opened 2026-03-24
tt-inference-server tt-metal vLLM server startup script pass vLLM args if overriding


tstescoTT:
Users need to be able to naturally pass vllm serve CLI args to Docker interface: https://docs.vllm.ai/en/stable/cli/serve/?h=serve#frontend

default case:
- release_model_spec.json.vllm_args will be overriden by any CLI variables. Currently when there is a default it overrides the given CLI vars. This needs to be reversed.

Special handling because Docker run CLI collides 
- `--service-port`: needs to remap internally to `--port`
  - vllm: `--port`
  - Docker: `--port`

tstescoTT:
example:
```
docker run \
  --env "HF_TOKEN=$HF_TOKEN" \
  --ipc host \
  --publish 7000:7000 \
  --device /dev/tenstorrent \
  --mount type=bind,src=/dev/hugepages-1G,dst=/dev/hugepages-1G \
  --volume volume_id_Qwen3-32B:/home/container_app_user/cache_root \
  ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64:0.11.0-bac8b34-7c6685a \
  --model Qwen3-32B \
  --tt-device galaxy \
  --service-port 7000 \
  --max-model-len 16384
```

tstescoTT:
addressed in https://github.com/tenstorrent/tt-inference-server/pull/2584

PR #2584, opened 2026-03-24
adding vLLM arg parsing allow for default overrides and --service-por…

tstescoTT:
# change log

* address https://github.com/tenstorrent/tt-inference-server/issues/2578
* adding vLLM arg parsing allow for default overrides 
* special handling for both --service-port and --port
* printing raw vLLM args sent to runpy

### DIFF
--- a/vllm-tt-metal/src/run_vllm_api_server.py
+++ b/vllm-tt-metal/src/run_vllm_api_server.py
@@ -8,6 +8,7 @@ import logging
 import multiprocessing
 import os
 import runpy
+import shlex
 import sys
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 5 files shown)


vllm-tt-metal/src/run_vllm_api_server.py
```diff
@@ -574,7 +570,7 @@ def start_trace_capture(
         return
 
     if service_port is None:
-        service_port = int(os.getenv("SERVICE_PORT", "8000"))
+        service_port = int(os.getenv("SERVICE_PORT", DEFAULT_VLLM_SERVER_PORT))
```
bgoelTT: Since `start_trace_capture()` is called with `service_port=resolve_service_port()` do we actually need this block, and then service_port becomes a position arg with no default?

tstescoTT: I believe we can remove SERVICE_PORT usage entirely, and only pass `--service-port` to the Docker image, and `--port` to vLLM when its != DEFAULT_VLLM_SERVER_PORT. I didn't want to do this in this PR for simplicity but could add that.

tstescoTT: Looks like removing `SERVICE_PORT` entirely from the code base needs https://github.com/tenstorrent/tt-inference-server/issues/2588 to improve the existing example scripts.

But I can remove it from `vllm-tt-metal/src/run_vllm_api_server.py` and `workflows/run_local_server.py`

tstescoTT: Added that in https://github.com/tenstorrent/tt-inference-server/pull/2584/commits/f7ac109ddd0287d8c4ad47168e51ba43b05fce80

bgoelTT (approved):

PR #2585, opened 2026-03-24
Updated docs and model_spec with release 0.11.1

vmaksimovicTT:


workflows/model_spec.py
```diff
@@ -2756,36 +2710,6 @@ def expand_to_specs(self) -> List["ModelSpec"]:
         ],
         status=ModelStatusTypes.COMPLETE,
     ),
-    ModelSpecTemplate(
```
tstescoTT, line 2759: This should not be removed.

workflows/model_spec.py
```diff
@@ -1788,8 +1772,8 @@ def expand_to_specs(self) -> List["ModelSpec"]:
     ModelSpecTemplate(
         weights=["meta-llama/Llama-3.1-8B", "meta-llama/Llama-3.1-8B-Instruct"],
         impl=tt_transformers_impl,
-        tt_metal_commit="3f72232",
```
tstescoTT, line 1791: why is this being set to older commit?

workflows/model_spec.py
```diff
@@ -871,8 +870,6 @@ def expand_to_specs(self) -> List["ModelSpec"]:
                     vllm_args=device_model_spec.vllm_args,
                     override_tt_config=device_model_spec.override_tt_config,
                     env_vars=device_model_spec.env_vars,
-                    tensor_cache_timeout=device_model_spec.tensor_cache_timeout,
```
tstescoTT, line 874: these should not be reverted.

bgoelTT (changes requested):
* ~~the release_model_spec.json is now in docs/ but in v0.11.0 it exists at the root https://github.com/tenstorrent/tt-inference-server/blob/v0.11.0/release_model_spec.json~~
  * ~~can we please keep this?~~
* is there a reason to remove the docs around `--limit-samples-mode`?
* it seems a lot of changes are made to ModelSpecs regarding the cache timeout, is this a merge gone wrong?

bgoelTT (approved):

tstescoTT (approved):

PR #2591, opened 2026-03-25
Expose metrics to Grafana

dmadicTT:
https://github.com/user-attachments/assets/f733ee62-4537-42bb-a8ff-9693d16c0006


knovokmetTT (approved):

tt-media-server/cpp_server/monitoring/grafana/dashboards/tt_media_server.json
```diff
@@ -0,0 +1,380 @@
+{
+  "annotations": {
+    "list": [
+      {
```
idjuricTT, line 4: Should we call this file differently since it's cpp based?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -180,6 +207,30 @@ void LLMService::consumerLoopForWorker(size_t workerIdx) {
...
+          ms.prev_token_time = std::chrono::steady_clock::now();
+        } else if (ms.token_count % kItlSampleStride == 0) {
+          auto now = std::chrono::steady_clock::now();
+          double itl =
+              std::chrono::duration<double>(now - ms.prev_token_time).count() /
+              kItlSampleStride;
+          tt::metrics::ServerMetrics::instance().onITLSample(taskId, itl);
+          ms.prev_token_time = now;
+        }
+        ms.token_count++;
```
idjuricTT, line 231: We already have a calculation of ttft and tps, could we use those?
Also would it be better to stream on task completion? This might affect perf

idjuricTT (approved):

Issue #2586, opened 2026-03-25
Emoji decoding is broken


idjuricTT:
Example:

{
  "frequency_penalty": 0,
  "max_tokens": 1600,
  "messages": [
    {
      "content": "Send me a stop sign emoji",
      "role": "user"
    }
  ],
  "model": "test-model",
  "presence_penalty": 0,
  "seed": 0,
  "stop": "string",
  "stream": true,
  "stream_options": {
    "continuous_usage_stats": false,
    "include_usage": true
  },
  "temperature": 2,
  "top_p": 1,
  "user": "string"

Response:

data: {"choices":[{"delta":{"content":" \ufffd"},"finish_reason":null,"index":0,"logprobs":null}],"created":1774396622,"id":"chatcmpl-20e1dfe927ec27204289406c","model":"test-model","object":"chat.completion.chunk"}

data: {"choices":[{"delta":{"content":"\ufffd"},"finish_reason":null,"index":0,"logprobs":null}],"created":1774396622,"id":"chatcmpl-20e1dfe927ec27204289406c","model":"test-model","object":"chat.completion.chunk"}

PR #2594, opened 2026-03-25
Handle Multi-Token Representation

ljovanovicTT:
## Issue
https://github.com/tenstorrent/tt-inference-server/issues/2586

## Problem
For both streaming and non-streaming paths, we are decoding one token at a time. This is an issue for special characters, like emojis, which require multiple tokens for their representation.

With this PR, we are doing cumulative decode if we find `[UNK] token` (�).
This approach is inspired with `SlowIncrementalDetokenizer` in vLLM project [codepoint](https://github.com/vllm-project/vllm/blob/34d317dcec3935e588d6c8ee8a7a57abb7a3e731/vllm/tokenizers/detokenizer_utils.py#L194-L199)

Additionally, StreamingResponses were changed to use UTF-8 encoding.

## Example
For the following output of the mock_sp_pipeline:
```
const std::vector<int64_t> kReasoningStressReplayTokens = {3574LL, 252LL, 242LL,    1LL};
```
We get following responses:

### Before
```
data: {"choices":[{"delta":{"content":" \ufffd"},"finish_reason":null,"index":0,"logprobs":null}],"created":1774396622,"id":"chatcmpl-20e1dfe927ec27204289406c","model":"test-model","object":"chat.completion.chunk"}

data: {"choices":[{"delta":{"content":"\ufffd"},"finish_reason":null,"index":0,"logprobs":null}],"created":1774396622,"id":"chatcmpl-20e1dfe927ec27204289406c","model":"test-model","object":"chat.completion.chunk"}
```

### After
```
data: {"id":"cmpl-212fbb4276b68a1b0d1587c6","object":"text_completion","created":1774437647,"model":"default","choices":[{"text":"🛑","index":0,"logprobs":null,"finish_reason":null}]}

data: {"id":"cmpl-212fbb4276b68a1b0d1587c6","object":"text_completion","created":1774437647,"model":"default","choices":[],"usage":{"prompt_tokens":7,"completion_tokens":2,"total_tokens":2,"ttft_ms":9.080000,"tps":83333.333000}}
```

## Perf Impact
There is no visible real perf impact currently, since each token is still decoded in O(1) complexity


tt-media-server/cpp_server/src/utils/tokenizer.cpp
dmadicTT, line 1: Some unit tests would be nice to have, for this but also previous improvements you have made to skip special tokens

knovokmetTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -227,7 +235,7 @@ void LLMService::consumerLoopForWorker(size_t workerIdx) {
       callback(response, isFinal);
 
       if (isFinal) {
-        // Clean up reasoning parser state for this task
+        streamDecoders.erase(taskId);
```
dmadicTT, line 229: We should also do this on connection lost, and on the cancellation @rdraskicTT is working on

tt-media-server/cpp_server/src/utils/tokenizer.cpp
```diff
@@ -117,6 +117,46 @@ std::string Tokenizer::decode(const std::vector<int>& tokenIds) const {
...
+
+}  // namespace
+
+Tokenizer::StreamDecoder::StreamDecoder(const Tokenizer& tokenizer)
+    : tokenizer_(tokenizer) {}
+
+std::string Tokenizer::StreamDecoder::step(int tokenId) {
+  pending_.push_back(tokenId);
+  std::string decoded = tokenizer_.decode(pending_);
+  if (!endsWithReplacementChar(decoded)) {
```
madicd, line 163: We saw the model behaving unexpectedly, for example publishing `<think>` without matching `</think>`.
What would happen if the model has a bug and it never publishes this char?

madicd (approved):

dmadicTT (approved):

idjuricTT (approved):

tt-media-server/cpp_server/src/utils/tokenizer.cpp
ljovanovicTT: Currently, model sending invalid tokens has an impact on streaming responses - we will wait until we are able to detokenize the output (or worst case, when the model is finished generating) and dump the response. 
Whilst this is not optimal, this is the approach and risk vLLM is taking.

One solution we can do is to cap the buffer size to take max 4 tokens at a time, since the longest UTF-8 character is 4 bytes, but I would like to implement and test this in a followup PR

tt-media-server/cpp_server/src/utils/tokenizer.cpp
ljovanovicTT: Added some multi-token char tests in this PR, and tests for special tokens will be added as a follow-up PR!

tt-media-server/cpp_server/src/services/llm_service.cpp
ljovanovicTT: Great catch! Done, and we'll look out for cancellation

PR #2618, opened 2026-03-26
Add MemoryManager for Llama (LLMRunner)

### LINKED ISSUE #2455: Make a Memory manager for Llama
Prerequisite: #2454 

When running Llama memory manager should peek in block manager and preallocate memory locations

ztorlakTT:
MemoryManager now peeks into BlockManager to allocate and release KV cache blocks for the LLMRunner.

- Added read-only accessors (blockSize, numFreeBlocks, freeBlockIds) and state-mutation primitives (claimBlock, releaseBlock) to BlockManager
- MemoryManager accepts a BlockManager reference, computes required blocks from input sequence length, claims them from the free list, and returns WAITING when capacity is insufficient
- Wired MemoryManager + Boost IPC queues into LLMRunner with a dedicated background memory loop (matching SpPipelineRunner's pattern) so memory requests are processed continuously


idjuricTT (approved):

tt-media-server/cpp_server/include/services/memory_manager.hpp
```diff
@@ -9,11 +9,16 @@
 #include "domain/manage_memory.hpp"
 #include "utils/concurrent_map.hpp"
 
+namespace llm_engine {
+class BlockManager;
+}
+
 namespace tt::services {
 
 class MemoryManager {
```
knovokmetTT, line 16: what about having a mem manager iface, since we will have different impl for llama vs sp_pipeline, or a strategy pattern? thoughts?

dmadicTT: I like it. We probably need something abstract like `allocate(sequence)`. This will then be different for models with and without paged attention.

@knovokmetTT But the more important question is what did you learn about the future interface of memory manager after your work on pipeline manager

tt-media-server/cpp_server/src/runners/llm_runner/block_manager.cpp
```diff
@@ -69,6 +69,14 @@ void BlockManager::deallocateBlock(int blockId) {
       << " free=" << free_block_ids_.size() << std::endl;
 }
 
+void BlockManager::claimBlock(int blockId) { allocateBlock(blockId); }
```
dmadicTT: Why not keep using the `allocateBlock`?

tt-media-server/cpp_server/src/services/memory_manager.cpp
```diff
@@ -24,18 +26,52 @@ ManageMemoryResult makeResult(const ManageMemoryTask& task,
...
+  int blkSize = blockManager->blockSize();
+  size_t numBlocks =
+      (static_cast<size_t>(inputSeqLen) + static_cast<size_t>(blkSize) - 1) /
+      static_cast<size_t>(blkSize);
+
+  if (static_cast<size_t>(blockManager->numFreeBlocks()) < numBlocks) {
+    return ManageMemoryStatus::WAITING;
+  }
+
+  const auto& freeIds = blockManager->freeBlockIds();
```
dmadicTT: Is block manager thread-safe?
We now introduce a concurrent consumer of the block manager next to the scheduler.
Can memory manager and scheduler get into race condition and get same block IDs?

tt-media-server/cpp_server/src/services/memory_manager.cpp
```diff
@@ -24,18 +26,52 @@ ManageMemoryResult makeResult(const ManageMemoryTask& task,
...
+  std::vector<int> picked(
+      freeIds.begin(),
+      freeIds.begin() + static_cast<std::ptrdiff_t>(numBlocks));
+
+  outBlockIds.reserve(numBlocks);
+  outLocations.reserve(numBlocks);
+  for (int blockId : picked) {
+    blockManager->claimBlock(blockId);
+    outBlockIds.push_back(blockId);
+    // TODO: map blockId to actual device DRAM address
```
dmadicTT: As we are host-to-host for Llama should we even care about DRAM address?
Llama says
* For this prompt you got block IDs 1, 7, 13
* Prefill calculates the KV cache and sends it to decode
* Decode propagates the KV cache and block IDs 1, 7 and 13 to the model

So the block IDs stay as memory abstractions

tt-media-server/cpp_server/src/runners/llm_runner.cpp
```diff
@@ -90,6 +102,30 @@ void LLMRunner::run() {
...
+    if (!retryQueue.empty()) {
+      auto result = memoryManager->handle_task(retryQueue.front());
+      if (result.status != domain::ManageMemoryStatus::WAITING) {
+        memoryResults.push(result);
+        retryQueue.erase(retryQueue.begin());
+      }
+    } else if (memoryRequests.tryPop(task)) {
+      auto result = memoryManager->handle_task(task);
+      if (result.status == domain::ManageMemoryStatus::WAITING) {
+        retryQueue.push_back(task);
```
dmadicTT: Why not pushing it back to `memoryRequests` queue?

tt-media-server/cpp_server/src/runners/llm_runner/block_manager.cpp
```diff
@@ -69,6 +69,14 @@ void BlockManager::deallocateBlock(int blockId) {
...
+
+void BlockManager::releaseBlock(int blockId) {
```
dmadicTT: Same, why not using existing `deallocateBlock`?

tt-media-server/cpp_server/src/services/memory_manager.cpp
```diff
@@ -24,18 +26,52 @@ ManageMemoryResult makeResult(const ManageMemoryTask& task,
...
+  if (!blockManager) {
+    return ManageMemoryStatus::SUCCESS;
+  }
+
+  int blkSize = blockManager->blockSize();
+  size_t numBlocks =
+      (static_cast<size_t>(inputSeqLen) + static_cast<size_t>(blkSize) - 1) /
+      static_cast<size_t>(blkSize);
+
+  if (static_cast<size_t>(blockManager->numFreeBlocks()) < numBlocks) {
```
dmadicTT: I think block manager should stay abstract, here we are looking into it's internals, a bit trying to do it's job.
Ideally you give it a `Sequence`, like "Hey reserve some blocks for this", and it gives them back to you.

This means these methods should stay private
-  allocateBlock
-  deallocateBlock

While we use more abstract methods like `allocate(Sequence)` or similar

tt-media-server/cpp_server/src/runners/llm_runner/block_manager.cpp
dmadicTT: More important https://github.com/tenstorrent/tt-inference-server/pull/2618#discussion_r3004643324

tt-media-server/cpp_server/src/runners/llm_runner/block_manager.cpp
dmadicTT: More important https://github.com/tenstorrent/tt-inference-server/pull/2618#discussion_r3004643324

tt-media-server/cpp_server/include/services/memory_manager.hpp
ztorlakTT: I'll address this in the following PR

PR #2624, opened 2026-03-26
Add API_KEY env variables support for inference server

kluongTT:
This aligns all three inference servers together to provide support for using the API_KEY env variable for auth.

* vllm inference-server
* cpp inference-server
* media-server

We can leave VLLM_API_KEY in for the vllm based servers to keep compatibility with stock vllm. The ordering is here:

    1. ``VLLM_API_KEY`` — use as-is if set
    2. ``API_KEY`` — copy to ``VLLM_API_KEY`` (parity with tt-media-server)
    3. ``JWT_SECRET`` — derive a bearer token and set ``VLLM_API_KEY``


bgoelTT:
Thanks @kluongTT I'll review

PR #2628, opened 2026-03-27
Align training api for tt-console

mcupacTT:
This PR introduces following changes
- new enums in `constants.py` to mark training trainers, training optimizers, and model display names
- add support for sending **training logs** to the client - for this `add_list_handler` is implemented in `TTLogger`, and `TTLogger` functions are extended to accept `kwargs`
    - corresponding tests are implemented
- modify training api endpoints to align with openapi specification and tt-train api endpoints 
- `/catalog` endpoint is introduced to list the available models, datasets, trainers etc
- ----
- nit changes: dataset template


mmilosevicTT (approved):
LGTM, from perspective of training needs

tt-media-server/utils/dataset_loaders/sst2/sst2_utils.py
```diff
@@ -4,7 +4,8 @@
 from string import Template
 
 
-PROMPT_TEMPLATE = Template("""Review: $input\nOutput:""")
+PROMPT_TEMPLATE = Template("""Review: $input
```
mmilosevicTT: nit: Maybe use cleandoc, like we had in one TT-Blacksmith PR. It just keeps it clean
```
PROMPT_TEMPLATE = Template(
    cleandoc(
        """
        Below is an instruction that describes a task.
        Write a response that appropriately completes the request.

        ### Instruction:
        $instruction

        ### Response:
        $response
        """
    )
)
```

ztorlakTT (approved):

tt-media-server/utils/dataset_loaders/sst2/sst2_utils.py
mcupacTT: rewritten

Issue #1470, opened 2025-12-15
[Model Readiness Support]: DeepSeek-R1


tstescoTT:
### Model Name

DeepSeek-R1-0528, DeepSeek-R1, DeepSeek-V3

### Model URL

https://huggingface.co/deepseek-ai/DeepSeek-R1-0528, https://huggingface.co/deepseek-ai/DeepSeek-V3

### Model Tenstorrent code implementation

https://github.com/tenstorrent/tt-metal/tree/main/models/demos/deepseek_v3

### Hardware

- 1x WH Galaxy (for testing)
- 2x WH Galaxy interconnected (Dual Galaxy setup)
- 4x WH Galaxy interconnected (Quad Galaxy setup)

### What accuracy evals are needed?

Evals already implemented that there are DeepSeek-R1-0528 published results for https://huggingface.co/deepseek-ai/DeepSeek-R1-0528#deepseek-r1-0528-1 
- MMLU-Pro (EM): 	85.0
- GPQA-Diamond (Pass@1): 81.0
- LiveCodeBench (2408-2505) (Pass@1): 73.3
- AIME 2024 (Pass@1): 91.4

### Model Readiness Support Preconditions

- [ ] Tenstorrent code implementation has all runtime preconditions and configurations documented.
- [ ] vLLM or other inference server integration complete.
- [ ] [Non-blocking] Performance targets (Top Perf) set in benchmarking/benchmark_targets/model_performance_reference.json.
- [ ] [Non-blocking] Required evals are implemented and integrated for model in evals/eval_config.py.
- [ ] [Non-blocking] GPU accuracy reference scores collected and set for model in evals/eval_config.py.

### Model Readiness Support Checklist

- [ ] Inference server can serve model via run.py server workflow.
- [ ] Inference server can run release workflow. Release markdown report for target hardware pasted in comment below.
- [ ] Specific inference server docs cover how to serve model and run release workflow.
- [ ] Model is added to top level README table.
- [ ] Model is added to tt-shield Model Readiness CI nightly run producing release report data.

bgoelTT:
Re-opening as the containerization needs to be completed

PR #2631, opened 2026-03-27
Add DeepSeek-R1-0582 to nightly on DUAL_ and QUAD_GALAXY

bgoelTT:
This PR adds `deepseek-ai/DeepSeek-R1-0528` to Models CI nightly #1470

### DIFF
--- a/.github/workflows/models-ci-config.json
+++ b/.github/workflows/models-ci-config.json
@@ -5,8 +5,7 @@
       "ci": {
         "nightly": {
           "devices": [
-            "DUAL_GALAXY",
-            "QUAD_GALAXY"
+            "DUAL_GALAXY"
           ]
         }
       }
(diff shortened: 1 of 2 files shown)


.github/workflows/models-ci-config.json
```diff
@@ -1,5 +1,16 @@
 {
   "models": {
+    "DeepSeek-R1-0528": {
+      "inference_engine": "vLLM",
+      "ci": {
+        "nightly": {
+          "devices": [
+            "DUAL_GALAXY",
+            "QUAD_GALAXY"
```
acvejicTT: @vmaksimovicTT we don't have a quad at the moment, can we blacklist it on our side until we get one? 
@bgoelTT actually we will transform this dual to quad, so maybe it is better to remove quad in this PR, as we will need later to update it once more by replacing dual with quad

acvejicTT (approved):
LGTM

bgoelTT:
Force-merging due to known C++ server perf regression

Issue #2630, opened 2026-03-27
Allow the handling of >1 TestTask in TestConfig


bgoelTT:
Currently, we have a restriction in the report generation that we can only process 1 test result from `--workflow tests` when executing `--workflow reports`.

We should be able to execute multiple TestTasks and generate a full report for all.

PR #2633, opened 2026-03-27
Allow the handling of >1 TestTask in TestConfig

lukezTT:
closes https://github.com/tenstorrent/tt-inference-server/issues/2630

workflow:
- now one test_{model_id}__{timestamp}/ directory before the task loop instead of per-task directories
- build_test_command now takes output_dir_path directly (no longer creates its own), passes --task-name to conftest to generate the parameter_reports.json

CI test run:
https://github.com/tenstorrent/tt-shield/actions/runs/23817543061

### DIFF
--- a/server_tests/conftest.py
+++ b/server_tests/conftest.py
@@ -89,7 +89,6 @@ def results_report(request, output_path):
 
     # 4. This code runs after the session finishes
     report_filename = f"parameter_report_{task_name}.json"
-    print(f"Generating {report_filename}...")
     filename = output_path / report_filename
     with open(filename, "w") as f:
         json.dump(report_data, f, indent=2)
--- a/server_tests/conftest.py
+++ b/server_tests/conftest.py
@@ -46,7 +46,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--task-name",
         action="store",
-        default="default",
+        default="unknown-task",
         help="Name of the test task, used to name the output report file",
     )
 
@@ -77,16 +77,17 @@ def output_path(request):
 # 3. Fixture for the session-wide report dictionary (now with metadata)
 @pytest.fixture(scope="session")
 def results_report(request, output_path):
+    task_name = request.config.getoption("--task-name")
     report_data = {
         "endpoint_url": request.config.getoption("--endpoint-url"),
         "model_name": request.config.getoption("--model-name"),
         "model_impl": request.config.getoption("--model-impl"),
+        "task_name": task_name,
         "results": {},
     }
     yield report_data
 
     # 4. This code runs after the session finishes
-    task_name = request.config.getoption("--task-name")
     report_filename = f"parameter_report_{task_name}.json"
     print(f"Generating {report_filename}...")
     filename = output_path / report_filename
--- a/server_tests/conftest.py
+++ b/server_tests/conftest.py
@@ -43,6 +43,12 @@ def pytest_addoption(parser):
         type=int,
         help="Maximum context length for the model",
     )
+    parser.addoption(
+        "--task-name",
+        action="store",
+        default="default",
+        help="Name of the test task, used to name the output report file",
+    )
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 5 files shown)


server_tests/run_tests.py
```diff
@@ -36,27 +35,21 @@ def build_test_command(
     task: TestTask,
     model_spec,
     device,
-    output_path,
+    output_dir_path,
     service_port,
-) -> List[str]:
+):
```
bgoelTT: Please preserve the type-hinting
```suggestion
) -> List[str]:
```

server_tests/conftest.py
```diff
@@ -43,6 +43,12 @@ def pytest_addoption(parser):
         type=int,
         help="Maximum context length for the model",
     )
+    parser.addoption(
+        "--task-name",
+        action="store",
+        default="default",
```
bgoelTT:
```suggestion
        default="unknown",
```

server_tests/conftest.py
bgoelTT, line 84: If we add `"task_name"` to the `report_data` then we have the opportunity to greatly simplify the report generation in `server_tests/utils/vllm_parameter_json_to_md.py`.

We could parse the field and enforce a simple contract to JSONs passed to `main()` that says "this JSON contains all the data needed to produce a report and remove our need to pass the `task_name` in from `run_reports.py`

server_tests/utils/vllm_parameter_json_to_md.py
```diff
@@ -155,24 +170,45 @@ def format_detailed_results_table(summary):
...
+            sections.append(_generate_single_report(report_data, task_name=task_name))
+        return "\n---\n\n".join(sections)
+
+    # Accept a single file path (backwards compatible)
+    if isinstance(report_file, list):
+        # list of file paths without task names — fall back to single merged report
+        report_file = report_file[0] if len(report_file) == 1 else report_file[0]
+
+    report_data = _load_report(report_file)
+    return _generate_single_report(report_data)
```
bgoelTT: With the suggestion below to `server_tests/conftest.py` we should be able to simplify this `main()` to just take a big list of JSON file paths and naively generate the report(s) as required

workflows/run_reports.py
```diff
@@ -2601,76 +2601,108 @@ def evals_generate_report(args, server_mode, model_spec, report_id, metadata={})
...
+    dir_pattern = f"test_{model_spec.model_id}__*"
+    matching_dirs = sorted(tests_output_dir.glob(dir_pattern))
+    if not matching_dirs:
+        logger.info(
+            f"No test output directories matching '{dir_pattern}' "
+            f"in {tests_output_dir}. Skipping."
         )
-    # When multiple test runs exist, use only the most recent result
-    files = max(files, key=lambda f: Path(f).stat().st_mtime)
-    logger.info(f"Selected most recent test report: {files}")
```
bgoelTT, line 2633: I don't think indexing on the `-1` element is a strong enough guarantee to get the most recent output directory. Let's use the one that previously existed

workflows/run_reports.py
```diff
@@ -2601,76 +2601,108 @@ def evals_generate_report(args, server_mode, model_spec, report_id, metadata={})
...
+        return empty_result
+
+    logger.info(f"Processing {len(report_entries)} parameter report(s)")
+
+    # Generate vLLM parameter coverage report from all report files
+    # Pass (task_name, file_path) tuples for per-task sections when multiple tasks
+    if len(report_entries) == 1:
+        report_file_arg = report_entries[0][1]
+    else:
+        report_file_arg = report_entries
```
bgoelTT: With the addition of `"task_name"` into the parameter report JSON we can greatly simplify this logic.

We can do `report_entries = sorted(latest_dir.glob("parameter_report_*.json"))` and then pass `report_entries` to `generate_vllm_parameter_report()`

bgoelTT (changes requested):
Looks good, requesting one simplification for report generation

bgoelTT (approved):
Looks good, can you please run a Llama-3.1-8B on p150x8 dispatch in Models CI to verify the single TestTask handling works?

Issue #2606, opened 2026-03-25
Session manager


idjuricTT:
Extend llm_api with function:

- Open session - returns session id (uuid)
- Close session - takes session id returns ok

If LLM api is called without session id it creates a session and returns it to the user

Create a new service - SessionManager that will:

- Record sessions 
- Merge sessions with slots
- Have a method to get a slot via session id 

Flow: LLM service calls session manager to create a session if it doesn't exist.
If it exist LLM Service calls session manager to get a slot, if it exists it's a multi turn request and prefill must be done on decode
If slot id doesn't exist:

- If request is short enough (<1k) it gets slot and goes to decode (and prefill via decode)
- If request is long it needs a slot id and goes to prefill

Active request tracking should be in format:

<img width="1210" height="388" alt="Image" src="https://github.com/user-attachments/assets/269500cd-18ac-42f4-9e74-7edf4a210054" />

idjuricTT:
Reassigning to @dmadicTT  instead of active request tracker item (basically the same item with more info)

idjuricTT:
Implemented

PR #2637, opened 2026-03-27
Initial session manager

idjuricTT:
Features:

New API methods:

- Create session
- Close session
- Get slot by session id (can be removed, just for testing)

Session manager that manages session_id (uuid) and associated slot_id (int) is introduced.

LLM service will check if the requests have session_id and if not a new session_id will be created and returned to the user:

image

Remaining scope:

- Get slot_id if not assigned
- If session_id is just created it might go to prefill, otherwise if exists it's always prefill by decode

Issues:

- #2606
- #2646

### DIFF
--- a/tt-media-server/cpp_server/include/domain/session.hpp
+++ b/tt-media-server/cpp_server/include/domain/session.hpp
@@ -5,6 +5,7 @@
 
 #include <json/json.h>
 
+#include <chrono>
 #include <optional>
 #include <random>
 #include <sstream>
@@ -45,6 +46,20 @@ class Session {
   bool hasSlot() const { return slot_id_ != -1; }
 
   /**
+   * Get the last activity time.
+   */
+  std::chrono::system_clock::time_point getLastActivityTime() const {
+    return last_activity_time_;
+  }
+
+  /**
+   * Update the last activity time to now.
+   */
+  void updateActivityTime() {
+    last_activity_time_ = std::chrono::system_clock::now();
+  }
+
+  /**
    * Convert to JSON representation.
    */
   Json::Value toJson() const {
@@ -57,6 +72,7 @@ class Session {
  private:
   std::string session_id_;
   int slot_id_;
+  std::chrono::system_clock::time_point last_activity_time_;
 
   /**
    * Generate a UUID v4 string.
--- a/tt-media-server/cpp_server/include/domain/session.hpp
+++ b/tt-media-server/cpp_server/include/domain/session.hpp
@@ -6,6 +6,8 @@
 #include <json/json.h>
 
 #include <chrono>
+#include <cstdint>
+#include <limits>
 #include <optional>
 #include <random>
 #include <sstream>
@@ -20,9 +22,9 @@ class Session {
  public:
   /**
    * Create a new session with a generated UUID.
-   * @param slotId Optional slot ID (-1 means unassigned)
+   * @param slotId Optional slot ID (max uint32_t means unassigned)
    */
-  explicit Session(int slotId = -1);
+  explicit Session(uint32_t slotId = std::numeric_limits<uint32_t>::max());
 
   /**
    * Get the session ID (UUID).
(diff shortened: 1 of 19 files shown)


tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -150,6 +151,15 @@ void LLMController::completions(
     handleStreaming(request, std::move(callback), false);
   } else {
     try {
+      // Create session if not provided
+      if (!request->session_id.has_value() && sessionManager) {
```
dmadicTT, line 155: I assume `sessionManager` will never be null, I would not double-check it everywhere

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -477,4 +528,107 @@ void LLMController::handleStreaming(
   callback(resp);
 }
 
+void LLMController::createSession(
+    const drogon::HttpRequestPtr& req,
+    std::function<void(const drogon::HttpResponsePtr&)>&& callback) const {
+  if (!sessionManager) {
```
dmadicTT: Let's make `sessionManager` be never null

tt-media-server/cpp_server/include/domain/session.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Assign a slot ID to this session.
+   */
+  void setSlotId(int slotId) { slot_id_ = slotId; }
+
+  /**
+   * Check if a slot is assigned.
+   */
+  bool hasSlot() const { return slot_id_ != -1; }
```
dmadicTT: Can we ever have session without a slot?
I think `Session` could be immutable object that starts living once it has a `slot_id`.

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -237,6 +252,15 @@ void LLMController::chatCompletions(
     handleStreaming(request, std::move(callback), true);
   } else {
     try {
+      // Create session if not provided
+      if (!request->session_id.has_value() && sessionManager) {
```
dmadicTT, line 256: I would create a session only once we have `slot_id`

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  auto it = sessions_.find(sessionId);
+  if (it == sessions_.end()) {
+    return std::nullopt;
+  }
+
+  return it->second;
+}
+
+size_t SessionManager::getActiveSessionCount() const {
```
dmadicTT, line 94: We should expose this via prometheus

dmadicTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
idjuricTT: This is already done, I'll remove the unnecessary checks

tt-media-server/cpp_server/src/api/llm_controller.cpp
idjuricTT: Yep, checks are not needed, thanks!

tt-media-server/cpp_server/src/services/session_manager.cpp
idjuricTT: I'll add a follow up to add active session count once your PR is merged

tt-media-server/cpp_server/include/domain/session.hpp
idjuricTT: I kind a left that to work both directions before we have the integration but I agree overall.
I'll create a follow up

tt-media-server/cpp_server/include/services/session_manager.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Close a session without locking (assumes mutex is already locked).
+   * @param sessionId The session ID to close
+   * @return true if session was found and closed, false otherwise
+   */
+  bool closeSessionLocked(const std::string& sessionId);
+
+  mutable std::mutex mutex_;
+  std::unordered_map<std::string, domain::Session> sessions_;
```
knovokmetTT, line 92: can we use, ConcurrentMap , no  need for mutex

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -0,0 +1,160 @@
...
+  if (sessions_.empty()) {
+    return {};
+  }
+
+  // Create a vector of (sessionId, activityTime) pairs
+  std::vector<std::pair<std::string, std::chrono::system_clock::time_point>>
+      sessionTimes;
+  sessionTimes.reserve(sessions_.size());
+
+  for (const auto& [sessionId, session] : sessions_) {
```
knovokmetTT, line 140: can we do a heap here?

knovokmetTT (approved):

PR #2638, opened 2026-03-27
Prefill runner and prefill SHM

idjuricTT:
A dedicated runner for prefill that always returns one token.
In order to use it set: 

        "LLM_DEVICE_BACKEND": "prefill",

also it uses the same SHM variables as decode:

        "TT_IPC_SHM_C2P": "tt_ipc_c2p_12345",
        "TT_IPC_SHM_P2C": "tt_ipc_p2c_12345",

but it uses mock_prefill_runner that can run 4 workers that can be in sync and returns a single token

From mock prefill runner:

image

### DIFF
--- a/tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
+++ b/tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
@@ -39,7 +39,6 @@ class SpPrefillRunner : public IRunner {
   llm_engine::ITaskQueue* taskQueue;
   std::unique_ptr<sp_prefill::ISpPrefillModelRunner> modelRunner;
   sp_prefill::PrefillQueue prefillQueue;
-  std::unique_ptr<llm_engine::Sequence> activeSequence;
   std::atomic<bool> stopped{false};
 };
 
--- a/tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
+++ b/tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
@@ -5,18 +5,14 @@
 
 #include <atomic>
 #include <memory>
-#include <thread>
-#include <unordered_map>
 #include <unordered_set>
 
 #include "config/runner_config.hpp"
-#include "ipc/boost_ipc_memory_queue.hpp"
 #include "ipc/shared_memory.hpp"
 #include "runners/llm_runner/sequence.hpp"
 #include "runners/llm_runner/task_queue.hpp"
 #include "runners/runner_interface.hpp"
 #include "runners/sp_prefill_runner/i_sp_prefill_model_runner.hpp"
-#include "services/memory_manager.hpp"
 
 namespace tt::runners {
 
@@ -33,9 +29,6 @@ class SpPrefillRunner : public IRunner {
   const char* runnerType() const override { return "SpPrefillRunner"; }
 
  private:
-  void step();
-  void drainPrefillResults();
-  void memoryLoop();
   void pushToken(const llm_engine::TaskID& taskId, uint64_t tokenId,
                  bool finished);
   void pushErrorToken(const llm_engine::TaskID& taskId);
@@ -48,13 +41,6 @@ class SpPrefillRunner : public IRunner {
   sp_prefill::PrefillQueue prefillQueue;
   std::unique_ptr<llm_engine::Sequence> activeSequence;
   std::atomic<bool> stopped{false};
-
-  tt::services::MemoryManager memoryManager;
-  ipc::MemoryRequestQueue memoryRequests{ipc::k_memory_request_queue_name,
-                                         ipc::MEMORY_QUEUE_CAPACITY};
-  ipc::MemoryResultQueue memoryResults{ipc::k_memory_result_queue_name,
-                                       ipc::MEMORY_QUEUE_CAPACITY};
-  std::thread memoryThread;
 };
 
 }  // namespace tt::runners
--- a/tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
+++ b/tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+#pragma once
+
+#include <atomic>
+#include <memory>
+#include <thread>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "config/runner_config.hpp"
+#include "ipc/boost_ipc_memory_queue.hpp"
+#include "ipc/shared_memory.hpp"
+#include "runners/llm_runner/sequence.hpp"
+#include "runners/llm_runner/task_queue.hpp"
+#include "runners/runner_interface.hpp"
+#include "runners/sp_prefill_runner/i_sp_prefill_model_runner.hpp"
+#include "services/memory_manager.hpp"
+
+namespace tt::runners {
+
+class SpPrefillRunner : public IRunner {
+ public:
+  SpPrefillRunner(const tt::config::LLMConfig& config,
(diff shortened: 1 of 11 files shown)


tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
```diff
@@ -0,0 +1,207 @@
...
+  TT_LOG_INFO("SpPrefillRunner: Warmup successful");
+  return true;
+}
+
+void SpPrefillRunner::stop() {
+  TT_LOG_INFO("SpPrefillRunner: Stopping");
+  stopped.store(true, std::memory_order_relaxed);
+}
+
+void SpPrefillRunner::step() {
```
madicd: This looks like async scheduling we used for decode, but to my knowledge prefill will be sync.
So we can, or should, do

- get tasks
- send to queue
- wait for result
- repeat

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
```diff
@@ -0,0 +1,207 @@
...
+    pushToken(dr.taskId, dr.tokenId, true);  // Always finished after prefill
+
+    // Clear active sequence if it matches
+    if (activeSequence && activeSequence->taskId == dr.taskId) {
+      activeSequence.reset();
+    }
+  }
+}
+
+void SpPrefillRunner::memoryLoop() {
```
madicd: I think we won't need this for a longer time
prefill will always have enough memory
it will just keep reusing one KV slot as it has b=1
even with larger batches, we are quite good for a longer time

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
idjuricTT: Simplified, thanks 😄

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
idjuricTT: Removed, thank you!

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_model_runner.cpp
```diff
@@ -0,0 +1,50 @@
...
+  deviceInput.write(taskId, tokenIds, 1);
+}
+
+void SpPrefillModelRunner::exit() {
+  if (stop.exchange(true)) return;
+  if (readerThread.joinable()) readerThread.join();
+  LLM_ENGINE_LOG("sp_prefill") << "model runner exit" << std::endl;
+}
+
+void SpPrefillModelRunner::readerLoop() {
```
dmadicTT: Due to sync nature, we probably don't need a separate thread this one, too.
Just a blocking read, waiting for the result to come

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
```diff
@@ -0,0 +1,161 @@
...
+void SpPrefillRunner::run() {
+  while (!stopped.load(std::memory_order_relaxed)) {
+    // Get next sequence from task queue
+    auto* seq = taskQueue->tryPop();
+    if (!seq) {
+      std::this_thread::yield();
+      continue;
+    }
+
+    activeSequence.reset(seq);
```
dmadicTT: What's the motive for having `activeSequence`?
If we go the sync way, I think we might not need it

tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
```diff
@@ -0,0 +1,46 @@
...
+class SpPrefillRunner : public IRunner {
+ public:
+  SpPrefillRunner(const tt::config::LLMConfig& config,
+                  ipc::TokenRingBuffer<65536>* resultQueue,
+                  llm_engine::ITaskQueue* taskQueue);
+  ~SpPrefillRunner() override;
+
+  void run() override;
+  void stop() override;
+  bool warmup();
```
dmadicTT, line 28: We have a funny bug I forgot to share.
I think the `warmup` you define here will not be invoked as it's not declaraed as `virutal` method in `IRunner`.
Instead it will just invoke the method in `IRunner` returning `true`.

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
```diff
@@ -0,0 +1,161 @@
...
+                   result.tokenId, result.taskId.id);
+      pushToken(result.taskId, result.tokenId, true);  // Always finished
+    }
+
+    // Clear active sequence
+    activeSequence.reset();
+  }
+}
+
+bool SpPrefillRunner::warmup() {
```
dmadicTT, line 84: Some blaze decode experience: If the shared memory will be created by the python side as in blaze decode, should we wait in the warmup for the shared memory files to show up before sending the warmup task?

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_model_runner.cpp
```diff
@@ -0,0 +1,50 @@
...
+                                 const std::vector<int64_t>& tokenIds) {
+  // For prefill, we always send tokens but don't need maxTokens since we only
+  // get 1 token back
+  deviceInput.write(taskId, tokenIds, 1);
+}
+
+void SpPrefillModelRunner::exit() {
+  if (stop.exchange(true)) return;
+  if (readerThread.joinable()) readerThread.join();
+  LLM_ENGINE_LOG("sp_prefill") << "model runner exit" << std::endl;
```
dmadicTT: We should kick out the `LLM_ENGINE_LOG` macro :D

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_model_runner.cpp
idjuricTT: Makes sense, adding the change

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_model_runner.cpp
idjuricTT: I'll create a follow up item, this macro  is everywhere 😄

tt-media-server/cpp_server/include/runners/sp_prefill_runner/sp_prefill_runner.hpp
idjuricTT: I'll create an item to fix the logger and do it properly - only when shared memory is ready, thanks!

idjuricTT: Follow up item:

https://github.com/tenstorrent/tt-inference-server/issues/2641

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
idjuricTT: Follow up item for a proper warmup:

https://github.com/tenstorrent/tt-inference-server/issues/2641

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
idjuricTT: Fair, no need for it, I'll simplify.
I kind a left it to grab multiple items but it still can be simplified

idjuricTT: Change added

dmadicTT:
Great for first version!
We still should make some improvements for making the read from the socket purely synchronous.

dmadicTT (approved):

PR #2639, opened 2026-03-27
cpp server - pipeline manager initial setup

knovokmetTT:


knovokmetTT:
build will fail because we lack the tt-blaze repo files atm

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+std::optional<pm::OutputMessage> SpPipelineRunner::getOutput() {
+  pm::OutputMessage output;
+  if (pipelineManager->try_pop_output(output)) {
+    return output;
+  }
+  return std::nullopt;
+}
+
+std::unique_ptr<llm_engine::Sequence> SpPipelineRunner::getRequest() {
+  if (inFlightCount >= maxInFlightCount) return nullptr;
```
dmadicTT: Don't we double this logic?
There is already max limit in the pipeline manager

tt-media-server/cpp_server/include/runners/sp_pipeline_runner/sp_pipeline_runner.hpp
```diff
@@ -35,20 +36,28 @@ class SpPipelineRunner : public IRunner {
...
-  std::unique_ptr<sp_pipeline::ISpPipelineModelRunner> modelRunner;
-  sp_pipeline::DecodeQueue decodeQueue;
-  std::unordered_map<llm_engine::TaskID, std::unique_ptr<llm_engine::Sequence>>
-      activeSequences;
+  std::unique_ptr<tt_blaze::pipeline_manager::PipelineInterface> pipeline;
+  std::unique_ptr<tt_blaze::pipeline_manager::PipelineManager> pipelineManager;
+  std::unordered_map<uint32_t, std::unique_ptr<llm_engine::Sequence>>
+      slotToSequence;
+  std::unordered_map<uint32_t, std::unique_ptr<llm_engine::Sequence>>
+      requestToSequence;
```
dmadicTT: This naming is more technical. It would be easier to understand if they are named based on semantics.

- `requestToSequence` -> `allocating`
-  `slotToSequence` -> `decoding`

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -129,72 +177,21 @@ void SpPipelineRunner::memoryLoop() {
 }
 
 void SpPipelineRunner::step() {
```
dmadicTT, line 126: I love the readability! Clean!

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
dmadicTT: Should we do a try-catch here?

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+  auto seq = std::move(it->second);
+  requestToSequence.erase(it);
+  if (response.error_code != 0) {
+    TT_LOG_ERROR(
+        "SpPipelineRunner: Response for request {} returned error code {}",
+        response.request_id, response.error_code);
+    pushErrorToken(seq->taskId);
+    return;
+  }
+  pipelineManager->push_request(makeSubmitRequest(response.slot_id, *seq));
```
dmadicTT: `push_request` and other pipeline manager methods return true/false.
We should handle that in future

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+  if (it == slotToSequence.end()) {
+    TT_LOG_ERROR("SpPipelineRunner: Output for unknown slot");
+    return;
+  }
+  auto& seq = *it->second;
+  seq.appendToken(output.token_id);
+  bool finished = output.is_complete || stopTokenIds.count(output.token_id);
+  pushToken(seq.taskId, output.token_id, finished);
+  if (finished) {
+    // We need a mechanism here for when to evict in the future, we dont support
```
dmadicTT: I would rather not do any eviction and say "We support max 64 users".
This way we enable an important feature but with a limit.

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+}
+
+std::unique_ptr<llm_engine::Sequence> SpPipelineRunner::getRequest() {
+  if (inFlightCount >= maxInFlightCount) return nullptr;
+  auto requestRaw = taskQueue->tryPop();
+  if (!requestRaw) return nullptr;
+  return std::unique_ptr<llm_engine::Sequence>(requestRaw);
+}
+
+void SpPipelineRunner::handleResponse(pm::PMResponse& response) {
```
dmadicTT: We should check if `response.slot_id == -1` and make sure this propagates as `503` or `429` to the user

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+  auto it = requestToSequence.find(response.request_id);
+  if (it == requestToSequence.end()) {
+    TT_LOG_ERROR("SpPipelineRunner: Response for unknown request");
+    return;
+  }
+  auto seq = std::move(it->second);
+  requestToSequence.erase(it);
+  if (response.error_code != 0) {
```
tt-asaigal: I think its okay to initially support up to as many users as the pipeline manager supports (currently 64), however as a next step we'd like to have a waiting/retry mechanism here - instead of erroring out.

This is related to your comment in `handleOutput` where we cancel sessions after the first turn.

Something like this could work in this function:

1. If Allocation fails in `handleResponse`, track the session/request separately and issue up to n allocations (can maintain a `num_allocations_tried` table, indexed by session id or task id)
2. If we we fail after all allocation attempts, we can cancel a session for another user (for example the least active one), wait for the cancellation to be processed and then reallocate this user

Eventually, we will evict users from the pipeline instead of cancelling them.

Doesn't need to be implemented now, since multi-user up to 64 is okay initially. Just flagging for next steps.

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+    return;
+  }
+  auto& seq = *it->second;
+  seq.appendToken(output.token_id);
+  bool finished = output.is_complete || stopTokenIds.count(output.token_id);
+  pushToken(seq.taskId, output.token_id, finished);
+  if (finished) {
+    // We need a mechanism here for when to evict in the future, we dont support
+    // multi-turn conversations yet because we cancel the slot after the first
+    // prompt.
```
tt-asaigal: Would be preferable to not cancel users and support actual multi-turn.

For the initial demo and testing, we can either restrict to max users at a time or limit the number of requests a user can submit. Multi-turn should be supported though.

The generic approach for this has been described in an earlier comment.

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -219,4 +216,81 @@ void SpPipelineRunner::pushErrorToken(const llm_engine::TaskID& taskId) {
...
+void SpPipelineRunner::handleOutput(pm::OutputMessage& output) {
+  auto it = slotToSequence.find(output.slot_id);
+  if (it == slotToSequence.end()) {
+    TT_LOG_ERROR("SpPipelineRunner: Output for unknown slot");
+    return;
+  }
+  auto& seq = *it->second;
+  seq.appendToken(output.token_id);
+  bool finished = output.is_complete || stopTokenIds.count(output.token_id);
+  pushToken(seq.taskId, output.token_id, finished);
```
tt-asaigal: is the `taskID` mapped to a session id? The original idea was to maintain a `slot_to_session` map. `slotToSequence` is fine as well, as long as we can map tasks to sessions.

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -3,14 +3,61 @@
...
+}
+
+inline pm::ISRequest makeCancelRequest(uint32_t slotId) {
+  return {.type = pm::RequestType::CANCEL, .slot_id = slotId};
+}
+
+inline void fillSequenceFields(pm::ISRequest& req,
+                               const llm_engine::Sequence& seq) {
+  req.token_count = static_cast<uint32_t>(seq.tokenIds.size());
+  for (uint32_t i = 0; i < req.token_count && i < pm::MAX_SEQ_LEN; i++) {
```
tt-asaigal: I've changed the interface slightly on the pipeline manager side, the `ISRequest` now looks like:

```
struct ISRequest {
    RequestType type = RequestType::ALLOCATE;
    uint32_t request_id = 0;
    uint32_t slot_id = INVALID_SLOT;
    std::vector<uint32_t> tokens;
    uint32_t max_new_tokens = 0;
    bool spec_decode = false;
    float temperature = 1.0f;
    float top_p = 1.0f;
    int32_t top_k = -1;
};
```

where `tokens` is a vector. You should just be able to do a direct move or data-copy.

tt-asaigal (commented):
Interaction with `PipelineManager` looks good overall. Would be good to add some unit tests where we can simulate multiple users submitting requests and getting responses through the pipeline.

Next step would be to integrate this with the Pipeline Manager device backend, which I can help you with.

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -97,104 +142,33 @@ bool SpPipelineRunner::warmup() {
...
 void SpPipelineRunner::step() {
-  drainDecodeResults();
+  // an open question: do we want to drain the task, response and output queues
+  // here? or just pop each one once every iteration? I am afraid of starvation.
 
-  if (inFlightCount >= maxInFlightCount) {
-    return;
+  auto memoryRequest = getMemoryRequest();
+  if (memoryRequest.has_value()) {
+    handleMemoryRequest(*memoryRequest);
```
knovokmetTT, line 129: idea here is that session manager will send a memory alloc request when a new session is created, we allocate then return the allocated slot in an async way via queues

only then will the actual task come in

that means that when the request comes in, it will always have a slot id attached

when session manager decides to evit someone, a new memory request will come it , but this time it will be dealloc

tt-asaigal: I see that we currently throw an error when an allocation fails. Is the idea to implement an internal retry mechanism through a retry queue?
- `SPPipelineRunner` can try to allocate a user up to `n` times.
- After waiting long enough, it notifies the session manager (the one pushing requests to the IPC queues) which decides to evict

Do we also plan on adding support for the `MemoryManagementAction::ALLOCATE` request?

knovokmetTT: > Do we also plan on adding support for the MemoryManagementAction::ALLOCATE request?

yes, I think its already added

> I see that we currently throw an error when an allocation fails. Is the idea to implement an internal retry mechanism through a retry queue?
> 
> SPPipelineRunner can try to allocate a user up to n times.
> After waiting long enough, it notifies the session manager (the one pushing requests to the IPC queues) which decides to evict



yeah, but I dont think it will be in the scope of this PR

idjuricTT (approved):

PR #2642, opened 2026-03-28
fix: correct false PASS on whisper WER=100 in evals release report

### LINKED ISSUE #2636: bug: Whisper audio evals always 401 — JWT_SECRET silently shadows API_KEY and --jwt-secret never forwarded to run_evals.py
## Summary

When running `--workflow evals` for Whisper (or any audio/media server model), all transcription requests from `lmms-eval` fail with `HTTP 401 Unauthorized`, producing zero valid eval results. The root cause is a two-part bug in how `OPENAI_API_KEY` is resolved for media server authentication.

## Observed behaviour

All 2939 `librispeech_test_other` samples fail:
```
❌ Audio transcription failed with status: 401  (×2939)
```
The server liveness check passes successfully (`runner_in_use: tt-whisper`, 1/1 workers ready), so there is no early warning before every sample fails.

### Bug 1 — `JWT_SECRET` silently shadows `API_KEY` for media server auth

[`evals/run_evals.py` L141–L148](https://github.com/tenstorrent/tt-inference-server/blob/a3fadb7ddd95c922d6371316cdae21ea5cc0c69f/evals/run_evals.py#L141-L148):

```python
def _setup_openai_api_key(args, logger):
    """Setup OPENAI_API_KEY environment variable based on JWT secret or API key."""
    api_key = args.jwt_secret or os.getenv("API_KEY", "your-secret-key")
    os.environ["OPENAI_API_KEY"] = api_key
```

`args.jwt_secret` defaults to `os.getenv("JWT_SECRET", "")` ([L185](https://github.com/tenstorrent/tt-inference-server/blob/a3fadb7ddd95c922d6371316cdae21ea5cc0c69f/evals/run_evals.py#L185)), so the effective expression is:

```python
api_key = os.getenv("JWT_SECRET", "") or os.getenv("API_KEY", "your-secret-key")
```

`JWT_SECRET` is a **signing secret for vLLM JWT tokens** — it is not the media server's bearer token. If `JWT_SECRET` is present in the environment (e.g. from a prior LLM server workflow in the same shell session), it silently takes priority over `API_KEY`, sending the wrong credential to the media server → 401.

If neither env var is set, the placeholder `"your-secret-key"` is sent → 401.

### Bug 2 — `--jwt-secret` is never forwarded to `run_evals.py`

[`workflows/run_workflows.py` L96–L103](https://github.com/tenstorrent/tt-inference-server/blob/a3fadb7ddd95c922d6371316cdae21ea5cc0c69f/workflows/run_workflows.py#L96-L103):

```python
cmd = [
    str(self.workflow_venv_config.venv_python),
    str(self.workflow_config.run_script_path),
    "--runtime-model-spec-json", str(self.runtime_model_spec_json_path),
    "--output-path", str(self.get_output_path()),
    "--model", self.model_spec.model_name,
    "--device", self.runtime_config.device,
]
```

`--jwt-secret` is never included in this command, so `args.jwt_secret` in `run_evals.py` always comes from the `JWT_SECRET` environment variable — not from anything `run.py` controls. This makes Bug 1 worse: even if a user passes `--jwt-secret` to `run.py`, it has no effect on the audio eval auth.

### How `lmms-eval` uses the credential

`whisper_tt` (the lmms-eval model class) reads `OPENAI_API_KEY` from the environment at init time and sends it as `Authorization: Bearer <key>` on every request. It does **not** accept `api_key` via `--model_args`:

```python
# lmms_eval/models/simple/whisper_tt.py:63
self.api_key = os.getenv("OPENAI_API_KEY", "your-secret-key")

# L154–155 / L223–224
if self.api_key:
    headers["Authorization"] = f"Bearer {self.api_key}"
```

The `EVALS_AUDIO` command constructed in [`evals/run_evals.py` L283–L296](https://github.com/tenstorrent/tt-inference-server/blob/a3fadb7ddd95c922d6371316cdae21ea5cc0c69f/evals/run_evals.py#L283-L296) also does not pass `api_key` in `--model_args`, so the env-var path is the only way to inject the credential.

## Proposed fix

1. In `_setup_openai_api_key`, remove the `JWT_SECRET` / `args.jwt_secret` fallback for audio/media models — these models use `API_KEY` directly, not JWT tokens:
   ```python
   # Before
   api_key = args.jwt_secret or os.getenv("API_KEY", "your-secret-key")
   # After
   api_key = os.getenv("API_KEY", "your-secret-key")
   ```

2. In `run_workflows.py` `run_workflow_script`, forward `--jwt-secret` from `runtime_config` so `run_evals.py` can use it for LLM models if ever needed:
   ```python
   if runtime_config.jwt_secret:
       cmd += ["--jwt-secret", runtime_config.jwt_secret]
   ```

3. Consider logging a warning in `run_evals.py` when `API_KEY` is unset or still the default placeholder, to make this failure mode visible before all eval samples are attempted.

## Workaround

Export `API_KEY` in the shell to the value the media server was started with before running the evals workflow:
```bash
export API_KEY=<your-media-server-api-key>
python run.py --model whisper-large-v3 --device p150 --workflow evals
```

## Environment

- Version: `0.12.0` (SHA `a3fadb7d`)
- Device: P150
- Model: `whisper-large-v3`
- Workflow: `evals`

stisiTT:
## Summary

Fix two bugs in `evals_release_report_data()` that caused WER=100 (total failure) to be reported as PASS:

1. **kwargs mutation**: `score_func_kwargs` was assigned by reference, so `kwargs["task_name"] = t_key` permanently mutated the shared config dict, corrupting scoring for subsequent tasks in the same process. Fixed with a shallow copy: `dict(...)`.

2. **WER exception fallback**: on any scoring exception, `score` defaulted to `0.0`, which the WER transform (`100 - score`) flipped to `100.0` — passing the tolerance check and masking catastrophic failure as PASS. Fixed by setting `score=100.0` in the exception path when `unit=="WER"`, so the transform yields `0.0` (worst possible accuracy) instead.

**Note:** The tuple key mismatch issue (multi-level result keys not matched in `key_found`) is a separate upstream concern tracked independently.

---

This PR was originally an omnibus (split per review feedback from bgoelTT). Other scopes:
- Auth fix (Closes #2636 Bug 2): #2701
- tt-metal bump (draft, blocked on tenstorrent/tt-metal#40970): #2702
- Logging/lmms-eval cleanup: #2704

### DIFF
--- a/workflows/run_reports.py
+++ b/workflows/run_reports.py
@@ -2215,7 +2215,9 @@ def evals_release_report_data(args, results, meta_data, model_spec):
                 # do NOT extract results[t_key] here.
                 # The score_func expects the ROOT results dict so it can do results[task_name].
 
-                kwargs = task.score.score_func_kwargs
+                # Shallow-copy to avoid mutating the shared config dict: kwargs["task_name"] = t_key
+                # below would permanently alter score_func_kwargs for all subsequent tasks in this process.
+                kwargs = dict(task.score.score_func_kwargs)
                 # Update task_name so the score function looks up the specific subtask (e.g. longbench_2wikimqa)
                 kwargs["task_name"] = t_key
                 configured_keys = kwargs.get("result_keys", [])
@@ -2243,7 +2245,12 @@ def evals_release_report_data(args, results, meta_data, model_spec):
                     )
                 except Exception as e:
                     logger.warning(f"  Could not calculate score for {t_key}: {e}")
-                    score = 0.0
+                    if kwargs.get("unit") == "WER":
+                        # WER=100 is worst-case (100% word error rate). Using score=0.0 here
+                        # would be transformed to 100-0=100 and incorrectly pass the tolerance check.
+                        score = 100.0
+                    else:
+                        score = 0.0
                 if kwargs.get("unit") == "WER":
                     score = 100 - score
 


workflows/workflow_venvs.py
bgoelTT, line 1: Let's break this into an isolated upstream issue in lmms-evals

workflows/model_spec.py
bgoelTT, line 1: Need this merged into tt-metal `main` before we can bump these https://github.com/tenstorrent/tt-metal/pull/40970

workflows/run_docker_server.py
bgoelTT, line 1: The media-server should only use `API_KEY` and it should not be inferred from `JWT_SECRET`. See https://github.com/tenstorrent/tt-inference-server/blob/6070ddbc3db32c2bced6d8f7786a291c64589096/evals/run_evals.py#L141-L149 for how media-server gets special handling

workflows/run_local_server.py
bgoelTT, line 1: See above

bgoelTT (changes requested):

workflows/run_reports.py
bgoelTT: Need to handle special case for Whisper
```suggestion
                if kwargs.get("unit") == "WER":
                    score = 100.0
```

bgoelTT (changes requested):
Let's separate the uplift from fixing evals so that we can make Models CI accurately represent the state of evals.

That way we don't need to block this PR

stisiTT:
Splitting this omnibus PR into focused PRs per @bgoelTT's review feedback:

1. **Scoring fix** (this PR — #2642): `fix: correct false PASS on whisper WER=100`
   - dict() copy to prevent kwargs mutation
   - WER exception handler set to 100.0 (not 0.0) so transform yields 0 → FAIL

2. **Auth fix** — #2701: `fix: pass API_KEY to media and forge server engines`
   - Passes `API_KEY` directly (not inferred from `JWT_SECRET`) per bgoelTT's feedback
   - Closes #2636 (Bug 2)

3. **tt-metal bump** — #2702 (draft): `fix: bump tt_metal_commit for whisper and speecht5`
   - Blocked on tenstorrent/tt-metal#40970

4. **Cleanup** — #2704: logging + lmms-eval branch update

Note: The tuple key mismatch fix (originally in this PR) is deferred as a separate upstream issue per bgoelTT's suggestion.

bgoelTT (approved):

idjuricTT (approved):

PR #2650, opened 2026-03-30
Implement training checkpoint retrieval from the server

mcupacTT:
Checkpoint related
- add logic to save model checkpoints (lora adapters), each to a new directory within the current job directory, located in `model_store` dir
- implement endpoint to list the available checkpoints (validating them if they are indeed located on the disk), and endpoint to download a desired checkpoint
- _implement corresponding tests_

Logs, metrics, learning rate persistance:
- implement logs persistance in the database
- implement the same logic as metrics - of periodically querying for new logs/checkpoints and persisting them in the database
- persist learning rate related data in the database together with metrics - new column in the metrics table
- _implement corresponding tests_

Nits:
- rename `model` to `compiled_model` in the training runner
- reformat logger message which is persisted to exclude datetime since it is saved in another field 
- add proper parametrization for existing tests which are only meant for persistance-enabled case (previously we had a `break` if persistance is not enabled)


tt-media-server/utils/job_database.py
```diff
@@ -209,7 +245,74 @@ def get_metrics_flat(self, job_id: str) -> List[Dict[str, Any]]:
...
+                    "created_at": r["created_at"],
+                }
+                for r in cursor.fetchall()
+            ]
+
+    # ------------- LOGS -------------
+    def insert_log(
+        self,
+        job_id: str,
+        log_index: str,
```
ztorlakTT:
```suggestion
        log_index: int,
```

tt-media-server/utils/job_manager.py
```diff
@@ -479,6 +451,117 @@ def _sync_status_to_db(self, job: Job, **overrides):
...
+                    message=log_entry["message"],
+                )
+            except IntegrityError:
+                self._logger.warning(
+                    f"Duplicate log for job {job.id} at index {i}, skipping"
+                )
+            except Exception as e:
+                self._logger.error(f"Failed to persist log for job {job.id}: {e}")
+                return -1
+        return current_len
```
ztorlakTT: Maybe instead of three similar methods (`_persist_new_metrics`, `_persist_new_checkpoints`, `_persist_new_logs`) you can have something like:
```
def _persist_new_items(
    self,
    job: Job,
    items_list: list,
    last_seen: int,
    insert_fn: Callable[[str, Any], None],
    item_label: str,
) -> int:
    current_len = len(items_list)
    for i in range(last_seen, current_len):
        try:
            insert_fn(job.id, items_list[i], i)
        except IntegrityError:
            self._logger.warning(
                f"Duplicate {item_label} for job {job.id} at index {i}, skipping"
            )
        except Exception as e:
            self._logger.error(
                f"Failed to persist {item_label} for job {job.id}: {e}"
            )
            return -1
    return current_len

def _insert_metric(self, job_id: str, metric: dict, _index: int):
    self.db.insert_metric(
        ...
    )
def _insert_checkpoint(self, job_id: str, ckpt: dict, _index: int):
    self.db.insert_checkpoint(
        ...
    )
def _insert_log(self, job_id: str, log_entry: dict, index: int):
    self.db.insert_log(
        ...
    )
```

And _persist_job_data_to_db becomes:
```
async def _persist_job_data_to_db(self, job: Job):
    if job.job_type != JobTypes.TRAINING.value:
        return
    streams = [
        (self.get_job_metrics(job.id), self._insert_metric, "metric"),
        (self.get_job_checkpoints(job.id), self._insert_checkpoint, "checkpoint"),
        (self.get_job_logs(job.id), self._insert_log, "log"),
    ]
    last_seen = [0] * len(streams)
    failed = [items is None for items, _, _ in streams]
    while True:
        for i, (items, insert_fn, label) in enumerate(streams):
            if failed[I]:
                continue
            last_seen[i] = self._persist_new_items(
                job, items, last_seen[i], insert_fn, label
            )
            if last_seen[i] < 0:
                failed[i] = True
        if job.is_terminal():
            break
        await asyncio.sleep(1.0)

```

tt-media-server/domain/training_request.py
```diff
@@ -31,8 +31,11 @@ class TrainingRequest(BaseRequest):
     device_type: str = DeviceTypes.P150.value
     optimizer: str = TrainingOptimizers.ADAMW.value
 
+    save_interval: int = 100
```
ztorlakTT: Can you add a validator for this field to check if it is > 0?

```
@field_validator("save_interval")
@classmethod
def validate_save_interval(cls, v):
    if v <= 0:
        raise ValueError("save_interval must be > 0")
    return v
```

tt-media-server/config/constants.py
```diff
@@ -282,6 +282,9 @@ class TrainingOptimizers(Enum):
     ADAMW = "adamw"
 
 
+TRAINING_STORE_ADAPTERS_DIR = "model_store/"
```
ztorlakTT, line 316: Can you add a comment for this const to describe its usage?

ztorlakTT (approved):

mcupacTT:
Resolved all comments, thanks a lot for suggestions!

PR #2675, opened 2026-04-01
Abort cancelled requests

rdraskicTT:
https://github.com/tenstorrent/tt-inference-server/issues/2264

Support request abortion for streaming requests.
Currently not supporting sp runner.

### DIFF
--- a/tt-media-server/cpp_server/src/api/llm_controller.cpp
+++ b/tt-media-server/cpp_server/src/api/llm_controller.cpp
@@ -8,6 +8,7 @@
 
 #include <chrono>
 #include <cmath>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <optional>
@@ -28,18 +29,17 @@ namespace {
 void sseSend(const std::string& sse, trantor::EventLoop* loop,
              const std::shared_ptr<drogon::ResponseStreamPtr>& streamPtr,
              const std::shared_ptr<ConcurrentQueue<std::string>>& accumulator,
-             const std::shared_ptr<std::atomic<bool>>& cancelled,
-             const std::string& taskId) {
-  if (cancelled->load()) return;
+             const std::shared_ptr<std::vector<std::string>>& earlyBuffer,
+             std::function<void()> onDisconnect = nullptr) {
   if (!accumulator) {
-    loop->queueInLoop([streamPtr, sse, cancelled, taskId]() {
-      if (*streamPtr && !cancelled->load()) {
-        if (!(*streamPtr)->send(sse)) {
-          cancelled->store(true);
-          TT_LOG_WARN(
-              "[LLMController] Client disconnected mid-stream task_id={}",
-              taskId);
-        }
+    loop->queueInLoop([streamPtr, earlyBuffer, sse,
+                       onDisconnect = std::move(onDisconnect)]() {
+      if (*streamPtr) {
+        bool ok = (*streamPtr)->send(sse);
+        if (!ok && onDisconnect) onDisconnect();
+      } else if (earlyBuffer) {
+        // Stream not ready yet — buffer for flush when it opens.
+        earlyBuffer->push_back(sse);
       }
     });
     return;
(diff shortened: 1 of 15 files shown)


tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -49,35 +49,27 @@ void sseSend(const std::string& sse, trantor::EventLoop* loop,
...
   if (!accumulated.empty()) {
     std::string batch;
     for (auto& s : accumulated) batch.append(s);
-    if (!(*streamPtr)->send(batch)) {
-      cancelled->store(true);
-      TT_LOG_WARN("[LLMController] Client disconnected mid-stream task_id={}",
-                  taskId);
-    }
+    if (*streamPtr) (*streamPtr)->send(batch);
   }
```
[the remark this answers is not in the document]
rdraskicTT: Claude reply:

The comment has a point, but I think it's fine to leave as-is. Here's why:

  flushAccumulated is only called inside the isFinal handler — at that point the request is already completing naturally. If send() fails here, the sequence is already on its final token and about to be cleaned up anyway.  
  There's no "server-side work running longer than necessary" because the work is already done.
                                                                                                                                                                                                                               
  The onDisconnect abort path handles the important case: disconnect during token generation, which is where we save real resources by stopping early. A send failure on the final flush is a no-op from a resource            
  perspective.
                                                                                                                                                                                                                               
  Adding disconnect handling here would also be tricky since flushAccumulated runs inside a queueInLoop lambda where done is already being exchanged to true — calling onDisconnect would be redundant.

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -28,18 +29,17 @@ namespace {
...
-        }
+    loop->queueInLoop([streamPtr, earlyBuffer, sse,
+                       onDisconnect = std::move(onDisconnect)]() {
+      if (*streamPtr) {
+        bool ok = (*streamPtr)->send(sse);
+        if (!ok && onDisconnect) onDisconnect();
+      } else if (earlyBuffer) {
+        // Stream not ready yet — buffer for flush when it opens.
+        earlyBuffer->push_back(sse);
       }
```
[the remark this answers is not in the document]
rdraskicTT: Claude reply:

This is also a low-risk scenario. The earlyBuffer only accumulates between submitStreamingRequest() and the newAsyncStreamResponse lambda executing — that's a tiny window within the same IO thread turn. In practice it
  holds 0-2 events at most.
                                                                                                                                                                                                                               
  The unbounded growth concern is theoretical: the stream setup happens synchronously right after the submit call (same function, lines 541-555), so the buffer is flushed almost immediately. A client disconnecting before   
  the stream opens would mean the HTTP response was never even sent — Drogon would clean up the connection without our involvement.

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -172,4 +191,40 @@ void Scheduler::postprocess(std::vector<Sequence*>& seqs,
 }
 
 void Scheduler::removeSequence(TaskID taskId) { sequences_.erase(taskId); }
+
+void Scheduler::abortRequest(TaskID taskId) {
```
knovokmetTT, line 195:
```
void Scheduler::abortRequest(TaskID taskId) {
  auto* seq = findSequence(taskId);

  // If the task isn't tracked or is still waiting, it might have a stale 
  // copy in the IPC queue. Mark it for skipping.
  bool is_waiting = seq && seq->status == SequenceStatus::WAITING;
  if (!seq || is_waiting) {
    if (pending_aborts_.size() < 2 * max_in_flight_count_) {
      pending_aborts_.insert(taskId);
    }
  }

  // If the sequence doesn't exist or is already terminal, we're done.
  if (!seq || seq->isAborted() || seq->isFinished()) {
    return;
  }

  // Clean up resources for active sequences
  seq->status = SequenceStatus::ABORTED;
  block_manager_.deallocate(*seq);

  std::erase_if(decode_queue_, [&](Sequence* s) { return s->taskId == taskId; });
  sequences_.erase(taskId);
}
```

more concise version , please check

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -134,6 +144,15 @@ std::pair<std::vector<Sequence*>, bool> Scheduler::schedule() {
   }
 
   tryScheduleDecode(scheduledSeqs, numSeqs);
```
knovokmetTT, line 146: is it possible we schedule a decode on a sequence which was aborted?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -295,7 +302,7 @@ void LLMService::consumerLoopForWorker(size_t workerIdx) {
     }
 
     if (!anyActivity) {
-      std::this_thread::yield();
+      std::this_thread::sleep_for(std::chrono::microseconds(100));
```
knovokmetTT: why is this change necessary?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -179,26 +179,33 @@ void LLMService::consumerLoopForWorker(size_t workerIdx) {
...
-      auto callback = val.value();
-
       bool isFinal = token.isFinal();
 
+      // For final tokens, atomically take ownership of the callback so that
+      // only one of {consumer, abortRequest} decrements pending_tasks_.
+      std::function<void(domain::StreamingChunkResponse&, bool)> callback;
       if (isFinal) {
-        stream_callbacks_.erase(token.task_id);
+        auto val = stream_callbacks_.take(token.task_id);
```
knovokmetTT, line 188: are these checks here for when a token comes back for a cancelled request?

tt-media-server/cpp_server/src/services/llm_service.cpp
rdraskicTT: Good catch, used during testing, removed now

tt-media-server/cpp_server/src/services/llm_service.cpp
rdraskicTT: Claude response:   Yes, exactly. There's a race between the consumer thread (here) and abortRequest() — both can try to claim ownership of the callback for the same task.

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
rdraskicTT: Claude response: No — an aborted sequence cannot be scheduled for decode. abortRequest() explicitly removes the sequence from decode_queue_ (lines 223-226) and erases it from sequences_ (line 228) before tryScheduleDecode ever runs.

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
rdraskicTT: I agree, adapted this version in latest commit

knovokmetTT (approved):

PR #2693, opened 2026-04-01
Use uint32 for TaskId

ljovanovicTT:
## Issue
[#2680 Convert task id to uint32 format](https://github.com/tenstorrent/tt-inference-server/issues/2680)

## Problem
In order to be compatible with new `Pipeline Manager`, we need to ensure taskId is of type `uint_32`.
This way, generating new taskId is faster, and taskId takes up less memory.


tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -138,32 +138,24 @@ void LLMService::consumerLoopForWorker(size_t workerIdx) {
...
   // Mitigation: observe ITL once every kItlSampleStride consecutive tokens.
   // The reported quantiles remain statistically representative; the absolute
   // ITL values are accurate because the caller pre-computes elapsed time
   // between the two most recent CONSECUTIVE tokens rather than relying on the
   // spacing between sampled events.
-  //
-  // TODO: Remove once task IDs become uint32 — then per-event cost is
-  // negligible (no heap allocation) and onToken() can be called for every
-  // token unconditionally.
   static constexpr int kItlSampleStride = 5;
```
ljovanovicTT, line 152: Should we remove it now?

idjuricTT (approved):

tt-media-server/cpp_server/include/utils/id_generator.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * Utility class for generating unique task IDs and serialization.
+ *
+ * Task IDs are uint32_t values generated via atomic counter.
+ * This provides:
+ * - Thread-safe generation
+ * - Deterministic sequential IDs (1, 2, 3, ...)
+ * - Compact 4-byte serialization for IPC
+ * - Simple debugging (sequential IDs are easy to track)
+ */
+class TaskIDGenerator {
```
knovokmetTT, line 20: nice !

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -161,7 +161,6 @@ void Scheduler::preempt(Sequence& seq) {
   seq.status = SequenceStatus::WAITING;
   block_manager_.deallocate(seq);
   prefill_queue_->push(seq);
-  sequences_.erase(seq.taskId);
```
knovokmetTT, line 164: why is this removed?

tt-media-server/cpp_server/include/domain/manage_memory.hpp
```diff
@@ -76,12 +77,12 @@ enum class ManageMemoryStatus : std::uint8_t {
 };
 
 struct ManageMemoryResult {
-  TaskID taskId;
+  uint32_t taskId;
   ManageMemoryStatus status{ManageMemoryStatus::FAILURE};
   std::vector<std::uint32_t> slotIds;
 
   void serialize(std::ostream& os) const {
```
knovokmetTT, line 78: do we ever actually use these serialize/deserialize methods?

ljovanovicTT: We are using it here:
https://github.com/tenstorrent/tt-inference-server/blob/dev/tt-media-server/cpp_server/include/ipc/boost_ipc_memory_queue.hpp#L68

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
ljovanovicTT: This was a previously undiscovered bug that was discovered after reducing the taskID size (from a UUID to uint32) - the test `Preempt_MovesSequenceBackToWaiting` started failing

The issue was in the test itself where we were checking:
```
  EXPECT_EQ(batch[0]->status, SequenceStatus::WAITING);
```
after calling preempt(), where batch[0] was deleted and left as a dangling pointer.

One option was to update the test, but after reviewing the `scheduler.cpp` implementation, safer fix seemed to be to avoid deleting the sequence from `sequences_`. This is because the sequence may still be referenced elsewhere like it was in tests. 
There is no risk of memory leaks with this approach, as the sequence is eventually destroyed (on completion, reschedule or abort) and `sequences_` remains consistent even after preempting.

PR #2714, opened 2026-04-02
For continuation or small ISL do prefill on decode

idjuricTT:
Do prefill on decode if:

- Valid session is found
- ISL is below 100 tokens (move to 1000 when done with testing)

### DIFF
--- a/tt-media-server/cpp_server/include/api/llm_controller.hpp
+++ b/tt-media-server/cpp_server/include/api/llm_controller.hpp
@@ -100,9 +100,8 @@ class LLMController : public drogon::HttpController<LLMController> {
   /**
    * Determine if disaggregated prefill should be used for this request.
    */
-  bool shouldDoPrefillOnDecode(
-      const domain::CompletionRequest& request,
-      bool validSessionFound) const;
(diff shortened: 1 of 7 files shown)


tt-media-server/cpp_server/include/api/llm_controller.hpp
```diff
@@ -85,6 +96,13 @@ class LLMController : public drogon::HttpController<LLMController> {
                                const std::string& type,
                                const Json::Value& param = Json::nullValue,
                                const Json::Value& code = Json::nullValue);
+
+  /**
+   * Determine if disaggregated prefill should be used for this request.
+   */
+  bool shouldDoPrefillOnDecode(
+      const domain::CompletionRequest& request,
```
ljovanovicTT: Unrelated to these changes, but we should be able to remove `CompletionRequest` from the codebase now

ljovanovicTT (approved):

tt-media-server/cpp_server/include/api/llm_controller.hpp
ztorlakTT: @ljovanovicTT  good catch, I'll address this

ztorlakTT (approved):

PR #2716, opened 2026-04-02
Enable virtual warmup dispatch and non-blocking server startup

### LINKED ISSUE #2710: Warmup for SP runner
Check / develop warmup for sp runner

ztorlakTT:
- Make `IRunner::warmup()` a virtual method so subclass overrides (`SpPipelineRunner`, `SpPrefillRunner`, `EmbeddingRunner`) are properly dispatched through the base interface. 
- Removes the blocking `waitForFirstWarmup()` call from `WorkerManager::start()` so the server can begin accepting requests while workers warm up in the background. Removing the blocking wait is safe because the server already handles unready workers gracefully — the `isModelReady()` check on each request returns 503 until at least one worker has completed warmup, and the `/tt-liveness` endpoint reports `model_ready: false` in the meantime, so no requests are silently dropped.
- This also consolidates the `INVALID_SLOT_ID` constant into a single definition in `manage_memory.hpp` (replacing scattered `std::numeric_limits<uint32_t>::max()` usages across `session.hpp`, `session_manager.hpp`, and `socket_messages.hpp`).

### DIFF
--- a/tt-media-server/cpp_server/src/worker/worker_manager.cpp
+++ b/tt-media-server/cpp_server/src/worker/worker_manager.cpp
@@ -61,7 +61,6 @@ WorkerManager::~WorkerManager() { stop(); }
 void WorkerManager::start() {
   startWarmupListener();
   startWorkers();
-  waitForFirstWarmup();
 }
 
 void WorkerManager::stop() {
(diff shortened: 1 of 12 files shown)


tt-media-server/cpp_server/src/worker/worker_manager.cpp
```diff
@@ -209,12 +208,6 @@ void WorkerManager::startWarmupListener() {
   });
 }
 
-void WorkerManager::waitForFirstWarmup() {
```
idjuricTT, line 212: Wouldn't this let requests pass even before the warmup?

idjuricTT (approved):

tt-media-server/cpp_server/src/worker/worker_manager.cpp
ztorlakTT: No, LLMController checks isModelReady() before processing any request and returns 503 until warmup completes

Issue #2720, opened 2026-04-02
Implement Text-Level Stop String Matching in Chat Completions API


ljovanovicTT:
We need to stop generating text when model output matches `stop` string from `ChatCompletionRequest`.

Api Reference: [stop: optional string or array of string](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20stop%20%3E%20(schema))

ljovanovicTT:
Closing since we now rely on Dynamo

PR #2747, opened 2026-04-03
Implement Text Level Stop String Matching

ljovanovicTT:
## Issue
[#2720 Implement Text-Level Stop String Matching in Chat Completions API](https://github.com/tenstorrent/tt-inference-server/issues/2720)

## Problem
We need to stop generating text when model output matches stop string from ChatCompletionRequest.

## TODO
Verify cancelation request for `SPPipelineRunner` is propagated when stop string is detected.


idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -71,6 +71,22 @@ std::vector<tt::worker::WorkerInfo> LLMService::getWorkerInfo() const {
 void LLMService::preProcess(domain::LLMRequest& request) const {
   BaseService::preProcess(request);
 
+  if (!request.stop.empty()) {
+    static constexpr size_t kMaxStopSequences = 4;
```
knovokmetTT, line 75: magic number?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -71,6 +71,22 @@ std::vector<tt::worker::WorkerInfo> LLMService::getWorkerInfo() const {
 void LLMService::preProcess(domain::LLMRequest& request) const {
   BaseService::preProcess(request);
 
+  if (!request.stop.empty()) {
```
knovokmetTT, line 74: maybe wrap this logic that checks for stop seq len in an inline func? can be a private namespace above the class in the file even

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -202,6 +223,41 @@ domain::LLMStreamChunk buildStreamChunk(
...
+    if (text.size() >= stopStr.size() &&
+        text.compare(text.size() - stopStr.size(), stopStr.size(), stopStr) ==
+            0) {
+      // Found a stop string match
+      state.stop_detected = true;
+      // Truncate stop string from output (MVP: always exclude)
+      state.accumulated_text.resize(text.size() - stopStr.size());
+
+      // Signal scheduler to abort this sequence
+      for (auto& cq : queueManager->cancel_queues) {
```
knovokmetTT, line 247: is this actually tested? Awesome if this works :O

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -233,6 +289,7 @@ void LLMService::consumerLoopForWorker(size_t workerIdx) {
   std::unordered_map<uint32_t,
                      std::unique_ptr<tt::utils::Tokenizer::StreamDecoder>>
       streamDecoders;
+  std::unordered_map<uint32_t, TaskStreamState> streamStates;
```
knovokmetTT, line 292: this map is not thread safe? Will multiple drogon threads access it?

ljovanovicTT: nice catch! but this map is local to each consumer thread in llm service

tt-media-server/cpp_server/src/services/llm_service.cpp
ljovanovicTT: thank you for you concern! 
this does work!

PR #2759, opened 2026-04-03
Standardize signedness of count/size types

### LINKED ISSUE #2740: Standardize signedness of count/size/index types in llm_engine
Throughout the LLM engine, `int` is used for quantities that are never negative: `numSeqs`, `numBatchedTokens`, `seqLimit`, `block_size_`, `max_num_batched_tokens_`, `blockId\`, `blockSize`. Meanwhile `max_in_flight_count_` is already `size_t`, creating signed/unsigned mismatches.
Standardize on `size_t` for counts and sizes. Keep `int` where negative values are meaningful (e.g. `hash == -1` sentinel). Note: `blockTable` (`vector<int>`) is serialized to shared memory via IPC, so changing its element type requires updating the binary wire format.

ztorlakTT:
Replace int with size_t for internal counters and sizes in the scheduler and block manager that are never negative and interact with STL .size() returns. This eliminates signed/unsigned comparison warnings and removes 7 unnecessary static_cast calls.

Serialized fields (blockTable, Sequence::blockSize, block IDs) remain int to preserve the IPC wire format.

### DIFF
--- a/tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
+++ b/tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
@@ -32,10 +32,9 @@ std::unique_ptr<Scheduler> makeScheduler(const Config& config,
 
 Scheduler::Scheduler(const Config& config, ITaskQueue* taskQueue,
                      size_t maxInFlightCount)
-    : block_size_(static_cast<size_t>(config.kvcache_block_size)),
+    : block_size_(config.kvcache_block_size),
       max_in_flight_count_(maxInFlightCount),
-      max_num_batched_tokens_(
-          static_cast<size_t>(config.max_num_batched_tokens)),
+      max_num_batched_tokens_(config.max_num_batched_tokens),
       stop_token_ids_(config.stop_token_ids.begin(),
                       config.stop_token_ids.end()),
       block_manager_(config.num_kvcache_blocks, config.kvcache_block_size),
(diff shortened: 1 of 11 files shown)


idjuricTT (approved):

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -32,9 +32,10 @@ std::unique_ptr<Scheduler> makeScheduler(const Config& config,
 
 Scheduler::Scheduler(const Config& config, ITaskQueue* taskQueue,
                      size_t maxInFlightCount)
-    : block_size_(config.kvcache_block_size),
+    : block_size_(static_cast<size_t>(config.kvcache_block_size)),
```
ljovanovicTT: Maybe we can switch `kvcache_block_size` and `max_num_batched_tokens` in `runner_config.hpp` to `size_t` as well?

ljovanovicTT (approved):

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
```diff
@@ -46,7 +47,8 @@ bool Scheduler::isFinished() const {
 
 Sequence& Scheduler::addRequest(uint32_t taskId, std::vector<int64_t> prompt,
                                 const SamplingParams& params) {
-  auto seq = std::make_unique<Sequence>(std::move(taskId), block_size_,
+  auto seq = std::make_unique<Sequence>(std::move(taskId),
+                                        static_cast<int>(block_size_),
```
ljovanovicTT, line 50: Same for `block_size_` in `scheduler.hpp`

knovokmetTT (approved):

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
ztorlakTT: `block_size_` in `scheduler.hpp` is already `size_t` 😄

tt-media-server/cpp_server/src/runners/llm_runner/scheduler.cpp
ztorlakTT: Updated, thank you!

Issue #1767, opened 2026-01-15
Unify tt-media-server and tt vLLM Docker images


tstescoTT:
Use separate uv venvs for different inference servers and build into one image.

PR #2765, opened 2026-04-04
feat: introduce tt-config-override

fivanovicTT:


bgoelTT (commented):
Not blocking this PR by any means, but eventually I want to address https://github.com/tenstorrent/tt-inference-server/issues/1767. This would require the vLLM & media-server entrypoints to have the same docker interface.

Right now, the vLLM entrypoint takes `sys.argv` and forward it to the main vLLM run API server function. I think it would make sense for media-server to also support this, where optional configuration is passed as arguments instead of environment variables. I.e, `docker run ... <image> --arg1 --arg2 ...`

Issue #659, opened 2025-09-15
Use OOP for eval / benchmark scripts if possible


idjuricTT:
Investigate using OOP for eval / benchmark scripts

PR #2787, opened 2026-04-06
fix(wan22): use mounted local weights and cache TT weights across restarts

kevinmi-TT:
## Problem

`TTWan22Runner` had three bugs that compounded to cause model warmup timeouts on every multi-chip runner, resulting in `SubDeviceManagerTracker is not initialized on MeshDevice 0` and `500 Internal Server Error` on `/tt-liveness`.

**1. HF weights re-downloaded on every container start**

`create_pipeline()` called `WanPipeline.create_pipeline()` without a `checkpoint_name`, so it fell back to the hardcoded default `"Wan-AI/Wan2.2-T2V-A14B-Diffusers"` and triggered a full HuggingFace download on every container start — even though `setup_host.py` had already downloaded the weights to the host and mounted the snapshot directory into the container via `MODEL_WEIGHTS_DIR`.

`TTFlux1Runner` and `TTQwenImageRunner` already solve this correctly by passing `checkpoint_name=self.settings.model_weights_path`. Wan2.2 was the only runner that didn't.

**2. TT-converted weights rebuilt on every container start**

`TTWan22Runner` was the only DiT runner that did not set `TT_DIT_CACHE_DIR`, so `cache.load_model()` rebuilt TT-format weights from the PyTorch state dict on every container restart and every new worker. Logs show repeated:

```
Loading transformer weights from PyTorch state dict. To use caching, set the TT_DIT_CACHE_DIR environment variable.
```

**3. `FabricTensixConfig.MUX` crashes `MeshDevice` on P300x2 (BHQBGE)**

On P300x2 (QuietBox GE), setting `fabric_tensix_config = FabricTensixConfig.MUX` caused `MeshDevice` to appear remote-only, which trips the `SubDeviceManagerTracker` assertion before any inference begins. Other Blackhole SKUs are unaffected.

A `load_weights()` override returning `False` also blocked the standard weight loading path.

---

## Fix

Single file changed: `tt-media-server/tt_model_runners/dit_runners.py`

1. **`create_pipeline()`** — pass `checkpoint_name = os.environ.get("MODEL_WEIGHTS_DIR") or self.settings.model_weights_path`. `from_pretrained()` resolves the already-mounted local snapshot and skips all network access.
2. **`__init__()`** — set `TT_DIT_CACHE_DIR=/tmp/TT_DIT_CACHE` so TT weights are cached within the run but wiped between runs (intentional — avoids stale compiled cache across image updates).
3. **`get_pipeline_device_params()`** — remove `FabricTensixConfig.MUX` and `dispatch_core_axis=ROW` on Blackhole. Neither is needed; removing them aligns Wan2.2 with all other DiT runners.
4. Remove spurious `load_weights()` override.

**Note:** Activating `--host-hf-cache` in CI so that `MODEL_WEIGHTS_DIR` is set inside the container is handled in a companion tt-shield change (PR tenstorrent/tt-shield#659). Both changes are required for the HF weight fix.

---

## Impact by runner

| Runner | Root cause | Fixed? |
|---|---|---|
| BHQBGE (p300x2) | MUX crash → remote-only mesh | ✅ |
| BHQBGE (p300x2) | HF re-download on every start (companion: tt-shield #659) | ✅ |
| Galaxy (6u) | HF re-download (49 files) → 38-min liveness window exhausted | ✅ |
| BH-LB (p150x8) | TT re-conversion per request → liveness window exhausted | ✅ |
| BH-QB-AE (p150x4) | Same as BHQBGE | ✅ |
| T3K | TT re-conversion overhead + WH inference speed (separate perf issue) | ⚠️ Partial |

---

## First-run note

TT weight cache will be cold on the first run after this lands; the runner will rebuild and cache to `/tmp/TT_DIT_CACHE`. Subsequent runs within the same container will be fast. The cache is intentionally ephemeral — it resets on container restart to avoid stale compiled weights.

---

Fixes: tenstorrent/tt-metal#39307

### DIFF
--- a/tt-media-server/tt_model_runners/dit_runners.py
+++ b/tt-media-server/tt_model_runners/dit_runners.py
@@ -298,10 +298,17 @@ class TTMochi1Runner(TTDiTRunner):
 class TTWan22Runner(TTDiTRunner):
     def __init__(self, device_id: str):
         super().__init__(device_id)
+        os.environ["TT_DIT_CACHE_DIR"] = "/tmp/TT_DIT_CACHE"
 
     def create_pipeline(self):
         try:
-            return WanPipeline.create_pipeline(mesh_device=self.ttnn_device)
+            checkpoint_name = (
+                os.environ.get("MODEL_WEIGHTS_DIR") or self.settings.model_weights_path
+            )
+            return WanPipeline.create_pipeline(
+                checkpoint_name=checkpoint_name,
+                mesh_device=self.ttnn_device,
+            )
         except Exception as e:
             log_exception_chain(
                 self.logger,
@@ -310,9 +317,6 @@ class TTWan22Runner(TTDiTRunner):
                 e,
             )
             raise
-
-    def load_weights(self):
-        return False
 
     @log_execution_time(f"{dit_runner_log_map[get_settings().model_runner]} inference")
     def run(self, requests: list[VideoGenerateRequest]):
@@ -344,9 +348,6 @@ class TTWan22Runner(TTDiTRunner):
         device_params = {
             "fabric_config": ttnn.FabricConfig.FABRIC_1D,
         }
-        if ttnn.device.is_blackhole():
-            device_params["fabric_tensix_config"] = ttnn.FabricTensixConfig.MUX
-            device_params["dispatch_core_axis"] = ttnn.device.DispatchCoreAxis.ROW
-        elif tuple(self.settings.device_mesh_shape) == (4, 8):
+        if tuple(self.settings.device_mesh_shape) == (4, 8):
             device_params["fabric_config"] = ttnn.FabricConfig.FABRIC_1D_RING
         return device_params


kevinmi-TT:
@fvranicTT could you please take a look. should unblock the wan2.2 failing on BHGE2

kevinmi-TT:
passing CI run for video gen part: https://github.com/tenstorrent/tt-shield/actions/runs/24053190150/job/70153473818

error due to path name being wrong in dev branch, now fixed.

fvranicTT:
> @fvranicTT could you please take a look. should unblock the wan2.2 failing on BHGE2

Wrong Filip here 😅

@fivanovicTT PTAL

fivanovicTT:
Hey @kevinmi-TT let's not merge this PR yet, I have increased the timeout for weight download, it seems like that fixes the issue:

- PR: https://github.com/tenstorrent/tt-inference-server/pull/2840
- CI run: https://github.com/tenstorrent/tt-shield/actions/runs/24197266953

fivanovicTT:
Hey @kevinmi-TT, I think we can close this PR - I have increased the timeout for downloading the weights so we don't need to mount any HF cache

Issue #2518, opened 2026-03-19
Clarify model-hardware support for ModelSpecTemplate defintions that are locked to older release_versions


tstescoTT:
The re-building of old tt-metal commit images is due to breaking changes in tt-metal build process (e.g. adding uv, which is a good thing). Supported infinitely backwards compatible build process in tt-inference-server patching each breaking change is a high-effort/low-reward task I believe we can avoid by simply directing users to the previous release version in the case of breaking changes occuring. 

The previous release versions will remain complete with code + images that have been tested so no functionality will ever be regressed across the set of releases and this minimizes the team time and machine capacity required for future releases using Models CI.

## Proposed solution:

<img width="2363" height="1878" alt="Image" src="https://github.com/user-attachments/assets/dc448f6f-038b-457b-971c-f9c1b1cfc885" />

bgoelTT:
@acvejicTT now that #2790 has been resolved is this issue fully unblocked?

bgoelTT:
@acvejicTT is it possible to address this before the v0.14.0 release? We're seeing lots of new tt-inference-server users hit this and blow their foot off

bgoelTT:
Daily:
* to be merged before v0.14.0 release

bgoelTT:
Daily:
* PR approved https://github.com/tenstorrent/tt-inference-server/pull/3543/
* After merging to `main` it will need to be cherry-picked to `stable` by @bgoelTT

bgoelTT:
Closing this issue as #3543 was merged and will be included in the v0.14.0 release

PR #2790, opened 2026-04-06
Introduce version per model in model_spec

vmaksimovicTT:


bgoelTT (approved):

bgoelTT:
@vmaksimovicTT is this good to merge so that work can start on #2518 ?

PR #2802, opened 2026-04-07
Fix shared memory bug

vpetrovicTT:
## Issue

**Python 3.10's `SharedMemory` Resource Tracker Bug**

---

### Problem

Python 3.10's `SharedMemory` automatically registers every opener with an internal **resource tracker daemon** — even processes that only attach (`create=False`) and don't own the segment.

### Root Cause

When the `SPRunner` process exits, the resource tracker sees unregistered segments and calls `shm_unlink()`, destroying:

- `/dev/shm/tt_video_in`
- `/dev/shm/tt_video_out`

...while `video_runner.py` is still using them.

### DIFF
--- a/tt-media-server/ipc/video_shm.py
+++ b/tt-media-server/ipc/video_shm.py
@@ -205,7 +205,7 @@ class VideoShm:
                 f"/{self._name}", "shared_memory"
             )
         self._buf = self._shm.buf
-
+ 
     def close(self) -> None:
         if self._shm:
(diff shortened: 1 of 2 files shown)


idjuricTT (approved):

fivanovicTT (approved):
Nice! 🚀

knovokmetTT (approved):

ztorlakTT (approved):

PR #2807, opened 2026-04-07
Support structured outputs

### LINKED ISSUE #2225: Support structured outputs
Special case of sampling where server dynamically takes output logits and prevents some tokens from being returned

ztorlakTT:
Adds support for OpenAI-compatible `response_format` parameter in chat completions, enabling `json_object` and `json_schema` output modes.
Uses xgrammar for grammar-guided token masking to constrain LLM output to valid JSON / JSON Schema during decoding. Integrated into the LLMRunner loop with per-request grammar state tracked via GuidedDecoderManager. SP Pipeline path logs a warning since per-step masking isn't supported there yet.

Test example:
```
curl -s http://localhost:8000/v1/chat/completions -H "Authorization: Bearer your-secret-key" -H "Content-Type: application/json" -d '{"messages": [{"role": "user", "content": "Give me a person"}], "max_tokens": 100, "response_format": {"type": "json_schema", "json_schema": {"name": "person", "strict": true, "schema": {"type": "object", "properties": {"name": {"type": "string"}, "age": {"type": "integer"}}, "required": ["name", "age"], "additionalProperties": false}}}}' | python3 -m json.tool
{
    "choices": [
        {
            "finish_reason": "length",
            "index": 0,
            "logprobs": null,
            "message": {
                "content": " {\"age\":-123456789,\"name\":\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\\"#$%&'()*+,-./0",
                "role": "assistant"
            }
        }
    ],
    "created": 1775570911,
    "id": "chatcmpl-15",
    "model": "default",
    "object": "chat.completion",
    "usage": {
        "completion_tokens": 100,
        "prompt_tokens": 7,
        "sessionId": "424f0c96-d700-4364-b06f-67b5feecaaaa",
        "total_tokens": 107,
        "tps": 6250.0,
        "ttft_ms": 16.0
    }
}
```


idjuricTT:
@bgoelTT  and @rreece could you please check if this PR makes sense for you?

bgoelTT (commented):
Please correct me if I'm wrong, but my concern with this is that it appears to require model to be returning the raw logits from the model but that is not how our models will operate in practice. Sampling will be performed on-device (else we pay massive host computation penalty). In order to sample, the model must receive this logit array over the total model vocab. Guided decoding takes the mask from xgrammar and applies it to the logit array **before sampling is performed**. If we wanted to apply the mask on-host then we'd have to short-circuit the model to return the logit array to host, apply mask, then move it back to device to apply sampling.

I'm not sure if this PR is just to pipe clean the inclusion of xgrammar and mask handling, but E2E guided decoding at perf needs to have the mask application on-device.

tt-media-server/cpp_server/include/runners/llm_runner.hpp
```diff
@@ -15,7 +15,8 @@
 
 namespace tt::services {
 class MemoryManager;
-}
+class GuidedDecoderManager;
```
knovokmetTT, line 18: not sure if we need to predclare these classes here btw... this pattern is used when there is a circular depenency , but I dont see it here... just polluting the code :(

tt-media-server/cpp_server/src/runners/llm_runner.cpp
```diff
@@ -27,27 +30,54 @@ LLMRunner::LLMRunner(const Config& config,
...
 
     if (result.isError) {
+      if (guidedDecoder) guidedDecoder->removeRequest(result.taskId);
       scheduler_->removeSequence(result.taskId);
       ipc::pushErrorToken(*result_queue_, result.taskId);
       return;
     }
 
+    if (guidedDecoder && guidedDecoder->hasGuidedDecoding(result.taskId)) {
+      guidedDecoder->acceptToken(result.taskId,
```
knovokmetTT: we call acceptToken but dont check the value? true/false?

ztorlakTT: fixed

tt-media-server/cpp_server/include/runners/llm_runner.hpp
ztorlakTT: I would avoid pulling guided_decoder_manager.hpp which transitively includes sampling_params.hpp into every file that includes llm_runner.hpp

idjuricTT (approved):

PR #2809, opened 2026-04-07
cpp server async session manager

### LINKED ISSUE #2748: Make session slot allocation async
createSession() blocks the Drogon IO thread for up to 1 second waiting for IPC slot allocation. This can stall the entire IO event loop. Make async or move to a dedicated thread.

knovokmetTT:


idjuricTT (approved):

ztorlakTT (commented):
I don't see we now have timeouts so `PendingAllocation` can stay in the map forever. This could lead to a memory leak

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -305,14 +204,113 @@ void SessionManager::sendDeallocRequest(const std::string& sessionId,
...
+                 sessionId, slotId);
+  } else {
+    TT_LOG_WARN(
+        "[SessionManager] Dealloc queue full, deferring session {} slot {}",
+        sessionId, slotId);
+    deferredDeallocQueue.push({sessionId, slotId});
+  }
+}
+
+void SessionManager::createSession(
```
ztorlakTT, line 217: Now we don't parse a `slot_id` from the request - is this change intentional?

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -305,14 +204,113 @@ void SessionManager::sendDeallocRequest(const std::string& sessionId,
...
+    callerEventLoop->queueInLoop([onError = std::move(onError)]() {
+      onError("Memory management IPC not available");
+    });
+    return;
+  }
+
+  domain::Session session = domain::Session(INVALID_SLOT_ID);
+  auto pendingAllocation =
+      PendingAllocation(std::move(session), std::move(onCompletion),
+                        std::move(onError), callerEventLoop, 10);
```
ztorlakTT: Can we avoid having magic numbers? 

```
maxAllocationAttempts = 10;
queueFullRetryDelay = std::chrono::milliseconds(50);
allocationFailureRetryDelay = std::chrono::milliseconds(500);
```

tt-media-server/cpp_server/src/services/session_manager.cpp
knovokmetTT: yes, but now I see we need it for testing, so tests dont have to init another proc for memory manager, so adding it back!

PR #2816, opened 2026-04-07
SDXL core_grid fix

jmitrovicTT:
## Problem description
SDXL is failing on Galaxy for some time: [example_run_link](https://github.com/tenstorrent/tt-shield/actions/runs/24053679049/job/70160606666)

```
KeyError: 'TT_METAL_CORE_GRID_OVERRIDE_TODEPRECATE'
```

## What's changed
- `TT_METAL_CORE_GRID_OVERRIDE_TODEPRECATE` is set for SDXL runners when device is galaxy

## Checklist
- [x] [n150 on dispach sdxl](https://github.com/tenstorrent/tt-shield/actions/runs/24100744347) passed✅
- [x] [6u galaxy on dispach sdxl](https://github.com/tenstorrent/tt-shield/actions/runs/24100730280) passed✅

### DIFF
--- a/tt-media-server/utils/runner_utils.py
+++ b/tt-media-server/utils/runner_utils.py
@@ -4,7 +4,7 @@
 
 import os
 
-from config.constants import ModelRunners
+from config.constants import ModelRunners, DeviceTypes
 from config.settings import settings
 from telemetry.telemetry_client import get_telemetry_client
 
@@ -60,6 +60,14 @@ def setup_runner_environment(
             )
             _setup_blackhole_mesh_config(tt_metal_home)
 
+    _RUNNERS_REQUIRING_GRID_OVERRIDE = {
+        ModelRunners.TT_SDXL_TRACE.value,
+        ModelRunners.TT_SDXL_EDIT.value,
+        ModelRunners.TT_SDXL_IMAGE_TO_IMAGE.value,
+    }
+    if settings.is_galaxy and settings.model_runner in _RUNNERS_REQUIRING_GRID_OVERRIDE:
+        _setup_grid_override(settings.device)
+
 
 def setup_cpu_threading_limits(cpu_threads: str, num_torch_threads: int = 1):
     """Set up CPU threading limits for PyTorch to prevent CPU oversubscription"""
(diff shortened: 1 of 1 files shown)


tt-media-server/utils/runner_utils.py
```diff
@@ -60,6 +60,14 @@ def setup_runner_environment(
...
             _setup_blackhole_mesh_config(tt_metal_home)
 
+    _RUNNERS_REQUIRING_GRID_OVERRIDE = {
+        ModelRunners.TT_SDXL_TRACE.value,
+        ModelRunners.TT_SDXL_EDIT.value,
+        ModelRunners.TT_SDXL_IMAGE_TO_IMAGE.value,
+    }
+    if settings.is_galaxy and settings.model_runner in _RUNNERS_REQUIRING_GRID_OVERRIDE:
+        _setup_grid_override(settings.device)
+
```
fivanovicTT: `settings.is_galaxy` actually does not mean that in fact it is galaxy machine :( Sorry, it is misleading

You can check if ti is galaxy: 
```
    device: Optional[str] = "n150"
```

like this:
```
settings.device == DeviceTypes.GALAXY.value  # "galaxy"
```

fivanovicTT (changes requested):

fivanovicTT:
If I can propose another solution to this:

In `Settings.py`place it between `model_config` and `__init__`, like:
```
  model_config = SettingsConfigDict(env_file=".env")
  
  @property
  def is_galaxy_device(self) -> bool:
      return self.device == DeviceTypes.GALAXY.value
      
  def __init__(self, **kwargs):
```

jmitrovicTT:
> If I can propose another solution to this:
> 
> In `Settings.py`place it between `model_config` and `__init__`, like:
> 
> ```
>   model_config = SettingsConfigDict(env_file=".env")
>   
>   @property
>   def is_galaxy_device(self) -> bool:
>       return self.device == DeviceTypes.GALAXY.value
>       
>   def __init__(self, **kwargs):
> ```

I wouldn't go with `is_galaxy_device` property because it can again be misleading: does it mean galaxy or just WH galaxy...

and also I believe `if settings.is_galaxy` was wrong - so I think just removing it is fine

fivanovicTT:
> > If I can propose another solution to this:
> > In `Settings.py`place it between `model_config` and `__init__`, like:
> > ```
> >   model_config = SettingsConfigDict(env_file=".env")
> >   
> >   @property
> >   def is_galaxy_device(self) -> bool:
> >       return self.device == DeviceTypes.GALAXY.value
> >       
> >   def __init__(self, **kwargs):
> > ```
> 
> I wouldn't go with `is_galaxy_device` property because it can again be misleading: does it mean galaxy or just WH galaxy...
> 
> and also I believe `if settings.is_galaxy` was wrong - so I think just removing it is fine

So, that means it will run on other devices other than Galaxy? Is that behavior okay for you?

jmitrovicTT:
> So, that means it will run on other devices other than Galaxy? Is that behavior okay for you?

It will be set only on galaxy (both BH and WH):

```
def _setup_grid_override(device: str):
    _logger = TTLogger()
    if (
        device == DeviceTypes.BLACKHOLE_GALAXY.value
        or device == DeviceTypes.GALAXY.value
    ):
```

this condition was already inside `_setup_grid_override` so `if settings.is_galaxy` was redundant (by assumption it works)

fivanovicTT (approved):

Issue #2780, opened 2026-04-06
Investigate Kafka overheads


idjuricTT:
Try the following scenario:

- Producer uses SessionManager and when session count is over 80% it sends offload request to consumers via Kafka
- There is 1 or more consumers which are separate Drogon instances that have only Kafka listener and migraitonWorker
- For now MigrationWorker just prints a message when signal is received

Goal of the task: Measure overhead between sending a signal and receiving.

PR #2835, opened 2026-04-08
feat: kafka producer and consumer

lmuravljovTT:
#2780 

  ## Summary                                                                                                                            
                                                                                                                                        
  Implements Kafka-based messaging infrastructure to measure overhead between sending session offload signals (producer) and receiving them (consumers). This is Phase 1 of distributed session load balancing - establishes the messaging backbone. (Kafka Producer and Consumer)
  
  ## What was added 

### Producer Side 
   **KafkaProducer** (kafka_client.hpp) exposing _sendCopy_ method for sending request messages 
   
   
   TODO: Use it in SessionManager when session count exceeds threshold ( 80% )
   
### Consumer side 
   **KafkaConsumer** (kafka_client.hpp) exposing _pollPayload_ method for  polling messages
   **tt_consumer** - separate Drogon instance that runs independently from the main server
   **MigrationWorker** - Polls Kafka topic (currently hardcoded for local dev) 
   
### General 
   - updated install_dependencies.sh - install needed kafka libs if flag enabled 
   - updated build.sh - flag added to build with kafka
   - updated Cmake - linking up sources based on KAFKA_ENABLED flag


tt-media-server/cpp_server/include/messaging/kafka_consumer.hpp
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: Apache-2.0
```
knovokmetTT, line 1: maybe rename the file kafka_producer?

tt-media-server/cpp_server/include/messaging/kafka_producer.hpp
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @param payload Message content to send
+   * @param errorMessage Optional output parameter for error details. If
+   * provided and send fails, contains human-readable error description.
+   *
+   * @return true if message was successfully queued for sending, false
+   * otherwise
+   *
+   * @note This is asynchronous - the message may not have been delivered when
+   *       this function returns. Use flush() or destructor for guaranteed
```
knovokmetTT, line 39: where is the flush ?

tt-media-server/cpp_server/include/messaging/kafka_producer.hpp
```diff
@@ -0,0 +1,49 @@
...
+   * delivery.
+   */
+  bool sendCopy(std::string_view payload, std::string* errorMessage = nullptr);
```
knovokmetTT: is the name an overkill? just send maybe? wdyt?

tt-media-server/cpp_server/include/worker/migration_worker.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+  ~MigrationWorker();
+
+  MigrationWorker(const MigrationWorker&) = delete;
+  MigrationWorker& operator=(const MigrationWorker&) = delete;
+
+  /**
+   * Starts the worker thread that polls Kafka for messages.
+   *
+   * Creates a background thread that continuously polls the Kafka topic.
```
knovokmetTT: redundant comments? saying the same thing no?

tt-media-server/cpp_server/include/messaging/kafka_producer.hpp
lmuravljovTT: good catch
 
Actually I'd like your opinion here, 
rd_kafka_flush is called in the destructor currently if needed, does it make sense to also expose it separately? Not sure if there are some specific situations where we'd like to flush the requests?

tt-media-server/cpp_server/include/messaging/kafka_producer.hpp
lmuravljovTT: Yeah I'd change them up in both producer and consumer actually 

send / write for the Producer 
receive / poll / read for Consumer ? 

Or maybe keep the "payload" to follow kafka naming? 
with 
writePayload and pollPayload? 

Any preference?

tt-media-server/cpp_server/src/consumer_main.cpp
```diff
@@ -0,0 +1,113 @@
...
+          .topic = tt::config::kafkaTopic(),
+          .group_id = tt::config::kafkaGroupId()});
+
+  TT_LOG_INFO("[Consumer] Starting MigrationWorker...");
+  worker->start();
+  TT_LOG_INFO("[Consumer] MigrationWorker started");
+
+  (void)std::system("mkdir -p ./consumer_logs");
+
+  drogon::app()
```
knovokmetTT, line 86: is drogon needed here just for health checks?  just curious

knovokmetTT (approved):

tt-media-server/cpp_server/include/messaging/kafka_producer.hpp
knovokmetTT: I cant think of anything on top of my head, but I dont see any harm in exposing one more func via the iface, who know if / when we will need it 🤷

tt-media-server/cpp_server/include/messaging/kafka_producer.hpp
knovokmetTT: > send for the Producer
> receive for Consumer 

seems the most reasonable to me, no need to overcomplicate IMO 😄 
but any of the ideas are fine by me, up to you

tt-media-server/cpp_server/src/consumer_main.cpp
lmuravljovTT: It was a task note to have consumers which are separate Drogon instances:

> There is 1 or more consumers which are separate Drogon instances that have only Kafka listener and migraitonWorker

Guess it's for future development

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -32,12 +32,11 @@ constexpr size_t MAX_SESSIONS_COUNT = 64;
 constexpr unsigned SESSION_EVICTION_RATE = 90;
 constexpr size_t SESSION_EVICTION_COUNT = 10;
 constexpr size_t MAX_TOKENS_TO_PREFILL_ON_DECODE = 100;
-constexpr unsigned SESSION_ALLOCATION_MAX_RETRIES = 10;
+constexpr const char* KAFKA_BROKERS = "localhost:9092";
+constexpr const char* KAFKA_TOPIC = "session-offload";
```
idjuricTT: Nitpick: Maybe a better name would be OFFLOAD_TOPIC_NAME?
we'll have load back event too so I would like to avoid redoing this next week 😄

tt-media-server/cpp_server/src/messaging/kafka_consumer.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+KafkaConsumer::KafkaConsumer(KafkaConsumerConfig config)
+    : impl_(std::make_unique<Impl>()) {
+  rd_kafka_conf_t* conf = rd_kafka_conf_new();
+  if (!conf) {
+    TT_LOG_ERROR("[Kafka] rd_kafka_conf_new failed");
+    return;
+  }
+
+  if (!kafka_utils::setConfigOrLog(conf, "bootstrap.servers",
```
idjuricTT, line 35: This part is quite similar between consumer and producer. Would it make sense that we have a KafkaClient that is inherited by consumer and producer?

tt-media-server/cpp_server/src/worker/migration_worker.cpp
```diff
@@ -0,0 +1,115 @@
...
+      root.get("action", Json::Value("unknown")).asString();
+  const std::string sessionId =
+      root.get("session_id", Json::Value("unknown")).asString();
+  const int currentCount =
+      root.get("current_session_count", Json::Value(0)).asInt();
+  const int maxSessions = root.get("max_sessions", Json::Value(0)).asInt();
+
+  TT_LOG_WARN("[MigrationWorker] ✅ OFFLOAD REQUEST RECEIVED");
+  TT_LOG_WARN("[MigrationWorker]   Action:      {}", action);
+  TT_LOG_WARN("[MigrationWorker]   Session ID:  {}", sessionId);
```
idjuricTT: TT_LOG_DEBUG? Sorry for the nitpick 😄

idjuricTT (approved):

tt-media-server/cpp_server/src/messaging/kafka_consumer.cpp
lmuravljovTT: Hmm I'm not convinced that enough is similar to have it extracted

We will want different config for both , only part that I see we could extract is the kafkaHandle definition? Do you think we will have more common code?

tt-media-server/cpp_server/include/config/defaults.hpp
lmuravljovTT: Changed to KAFKA_OFFLOAD_TOPIC_NAME :D so they all have KAFKA_ prefix

PR #2838, opened 2026-04-08
Update MD format

idjuricTT:


ljovanovicTT (approved):

ztorlakTT (approved):

knovokmetTT (changes requested):
this should be maybe in settings.cpp and have a default value

PR #2845, opened 2026-04-08
Correct product names

kfranzTT:
Changed Quietbox GE to Quietbox 2, and Added Quietbox in front of 4xp150


fivanovicTT:
Hey @kfranzTT, can you checkout from latest `dev` and changes should be merged to `dev`?

Currently, our `dev` is like the `main` in tt-metal repo

bgoelTT (changes requested):
100% on board with the proposed change but it needs to be edit in the actual doc generation code and not the output .md artefacts

PR #2846, opened 2026-04-09
Patches for Training Runners

mmilosevicTT:
This PR introduces several changes:
- Check that the training request is asking for the right device, based on the runner and how server was spun up.
- Addition of max_steps field to training request.
- Change of defaults for training requests.
- Updates to llama runner to support both singlechip and multichip setup.


mcupacTT (approved):
LGTM, thanks!

I only did not go into the multichip related stuff as I do not have much experience there. Everything else looks good.

tt-media-server/utils/job_manager.py
```diff
@@ -382,21 +385,36 @@ def _cleanup_old_jobs(self):
...
+                if is_old_terminal or is_stuck:
                     jobs_to_remove.append(job)
 
         if not jobs_to_remove:
             return
 
         for job in jobs_to_remove:
-            if job.is_in_progress():
-                job.mark_cancelling()
+            if job.is_in_progress() or job.status == JobStatus.CANCELLING:
```
```suggestion
            if job.is_in_progress() or job.is_cancelling():
```

ztorlakTT (approved):

PR #2882, opened 2026-04-09
Implement LoRA Inference Runner to Perform Inference on Chosen Fine-Tuned Model (Base Model + Adapter)

mcupacTT:
## Major Changes
- Introduces the `LoraInferenceRunner`, which enables inference using any fine-tuned model stored on the server. The specific model is selected per request.
- For each request, a user can run inference with the same adapter or switch to a different one. If a new adapter is selected, the model is recompiled accordingly.
- To support this behavior, an optional `adapter` field has been added to both `CompletionRequest` and `ChatCompletionRequest`.
- The runner implementation is based on and adapted from [`tt-xla/examples/pytorch/llama.py`](https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/llama.py).

## Minor Updates
- Resolved a potential vulnerability in training runners related to evaluating the `dtype` value provided via requests.
- Changes applied to:
  - `training_gemma_lora_runner.py`
  - `training_llama_lora_runner.py`


tt-media-server/config/constants.py
```diff
@@ -105,6 +105,7 @@ class ModelRunners(Enum):
     TT_XLA_VIT = "tt-xla-vit"
     TRAINING_LLAMA_LORA = "training-llama-lora"
     TRAINING_GEMMA_LORA = "training-gemma-lora"
+    LORA_INFERENCE = "lora-inference"
```
mmilosevicTT: Maybe let's make it clear that this is only for single chip use for now.

tt-media-server/tt_model_runners/forge_runners/lora_single_chip_runner.py
```diff
@@ -0,0 +1,265 @@
...
+        self.device = torch_xla.device()
+
+    @log_execution_time("Lora Inference")
+    def run(self, requests: list[CompletionRequest]):
+        request = requests[0]
+        self._validate_request(request)
+
+        if request.adapter:
+            adapter_info = resolve_adapter(request.adapter)
+            self._load_adapter(adapter_info)
```
mmilosevicTT, line 49: Does this need to go under try catch? If we try to load adapter from wrong model? As in do we need a check for jobID to make sure that we are actually running adapters from the right model.

tt-media-server/domain/chat_completion_request.py
```diff
@@ -34,3 +34,5 @@ class ChatCompletionRequest(BaseModel):
     n: int = 1
     seed: int | None = None
     user: str | None = None
+
+    adapter: str | None = None
```
mmilosevicTT, line 38: Is this how openAI sets it?

tt-media-server/tt_model_runners/forge_runners/lora_single_chip_runner.py
```diff
@@ -0,0 +1,265 @@
...
+            )
+        ]
+
+    def _validate_request(self, request: CompletionRequest):
+        if isinstance(request.prompt, str) and not request.prompt.strip():
+            raise ValueError("Prompt must not be empty")
+        if isinstance(request.prompt, list) and len(request.prompt) == 0:
+            raise ValueError("Prompt token list must not be empty")
+
+    def _load_adapter(self, adapter_info: AdapterInfo):
```
mmilosevicTT, line 91: We should definitely discuss this, I am not sure this is what we need for deploying models. I would expect deployed model to be run on a single tt-inference-server and then can generate responses. Once we deploy, I would not expect adapters to be sent in following requests, so it might be better to have adapter value in env variables?

mmilosevicTT (commented):
Left one comment for discussion, but otherwise it looks ok to me.

tt-media-server/tt_model_runners/forge_runners/lora_single_chip_runner.py
mcupacTT: Agree, we can discuss this offline and decide what we prefer

tt-media-server/config/constants.py
```diff
@@ -142,6 +143,7 @@ class ModelServices(Enum):
         ModelRunners.VLLMForge_LLAMA_70B,
         ModelRunners.LLM_TEST,
         ModelRunners.LLAMA_RUNNER,
+        ModelRunners.LORA_INFERENCE,
```
ztorlakTT:
```suggestion
        ModelRunners.LORA,
```
I'd suggest renaming this, because the "Inference" part is implied by the runner pattern, and it distinguishes itself from the training runners by not having "Training" in the name. The same goes for LoraInferenceRunner

tt-media-server/domain/chat_completion_request.py
mcupacTT: OpenAI does not expose LoRA adapters through the api, but among OpenAI-compatible inference servers, the most common convention is to encode the adapter through `model` field, as far as i was able to see.

On the other hand, we agreed offline to set up a separate adapter field for this purpose.

Anyway it won't break OpenAI API compliance since it is an optional field.

We can discuss this.

tt-media-server/tt_model_runners/forge_runners/lora_single_chip_runner.py
ztorlakTT, line 1: The current location for this runner is misleading as it is not training runner. Maybe you can move it to the `forge_runners`?

tt-media-server/tt_model_runners/forge_runners/lora_single_chip_runner.py
mcupacTT: You specify the adapter in the form `{job_id}/ckpt-step-{step}` (and this is also how we store it on the server)

And then you read the base model from the adapter config - it is not the user who set the base model, so we always take the base model which corresponds with the corresponding job_id. The `model` field in this case is not used in the adapter inference.

Regarding the try catch, no need for that at this place since the whole `run` function is wrapped up inside try catch in `device_worker.py`

tt-media-server/tt_model_runners/forge_training_runners/training_gemma_lora_runner.py
```diff
@@ -31,6 +31,11 @@
     TrainingOptimizers.ADAMW.value: torch.optim.AdamW,
 }
 
+DTYPE_MAP = {
+    "torch.bfloat16": torch.bfloat16,
+    "torch.float32": torch.float32,
+}
```
ztorlakTT: Can we extract this map to a shared util, because we have identical maps defined in `training_gemma_lora_runner.py` and `training_llama_lora_runner.py`?

ztorlakTT (approved):

tt-media-server/config/constants.py
mcupacTT: Considering also @mmilosevicTT comment from above, I would rename it to `LORA_SINGLE_CHIP`

tt-media-server/tt_model_runners/forge_training_runners/training_gemma_lora_runner.py
mcupacTT: Both dtype_map and optimizer_map are now moved to a separate utils file, thanks for the feedback

mcupacTT:
Most of the comments are resolved, apart from the marked discussions from above, thanks for the feedback :)

vmilosevic (approved):

PR #2912, opened 2026-04-14
Add Org Scoping for Jobs

mmilosevicTT:
Currently this PR only introduces passing orgID in training Jobs, but for other features it defaults to null.

### DIFF
--- a/tt-media-server/tests/test_job_manager.py
+++ b/tt-media-server/tests/test_job_manager.py
@@ -1632,7 +1632,7 @@ class TestJobManager:
             await asyncio.sleep(0.1)
             return "result.pt"
 
-        job_data = await job_manager.create_job(
+        await job_manager.create_job(
             job_id="org-job-1",
             job_type=JobTypes.TRAINING,
             model="test-model",
--- a/tt-media-server/tests/test_job_manager.py
+++ b/tt-media-server/tests/test_job_manager.py
@@ -178,26 +178,6 @@ class TestJob:
         }
         assert result["completed_at"] is not None
 
-    def test_to_public_dict_with_org_id(self):
-        """Test converting job with org_id includes it in public dict"""
-        job = Job(
-            id="test-123",
-            job_type="training",
-            model="test-model",
-            created_at=1000,
-            org_id="org-abc",
-        )
-
-        result = job.to_public_dict()
-        assert result["org_id"] == "org-abc"
-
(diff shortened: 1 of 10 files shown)


tt-media-server/utils/job_manager.py
```diff
@@ -95,6 +96,8 @@ def to_public_dict(self) -> dict:
             "model": self.model,
             "request_parameters": self.request_parameters,
         }
+        if self.org_id is not None:
+            data["org_id"] = self.org_id
```
mcupacTT, line 100: What would be the use case of exposing org_id related data to public dict?

tt-media-server/tests/test_job_manager.py
```diff
@@ -1621,3 +1641,140 @@ async def task_func(req):
...
+        )
+
+        assert job_data["org_id"] == "org-abc"
+
+        if job_manager.db:
+            db_job = job_manager.db.get_job_by_id("org-job-1")
+            assert db_job["org_id"] == "org-abc"
+
+    @pytest.mark.asyncio
+    async def test_create_job_without_org_id_omits_from_public_dict(
```
mcupacTT: seems to be unnecessary as we already have the test above when the job is created without org_id

mcupacTT (commented):
Left minor comments, overall LGTM!

@ztorlakTT if you can also review and say if you agree with the changes.

tt-media-server/utils/job_manager.py
mmilosevicTT: You are right, we only need to keep it internally in the database. Will update.

tt-media-server/utils/job_manager.py
```diff
@@ -187,59 +191,81 @@ async def create_job(
...
-        """Get metadata for all jobs, optionally filtered by job type."""
+    def _get_job_if_authorized(
+        self, job_id: str, org_id: Optional[str] = None
+    ) -> Optional[Job]:
+        """Retrieve a job by ID, returning None if org_id is set and doesn't match."""
+        job = self._jobs.get(job_id)
+        if job is None:
+            return None
+        if org_id is not None and job.org_id != org_id:
+            return None
```
ztorlakTT: nit:
```suggestion
        if job is None or (org_id is not None and job.org_id != org_id):
            return None
```

ztorlakTT (approved):

ztorlakTT (approved):

PR #2915, opened 2026-04-14
Request continuation

idjuricTT:
Flag continuation when request contains a valid session_id


tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -224,23 +225,14 @@ void SpPipelineRunner::handleRequest(
   assert(slotId != tt::domain::INVALID_SLOT_ID);
 
   auto it = running.find(slotId);
```
knovokmetTT: we dont need this variable now?
I'd maybe do a defensive programming here tho and do
assert(it != running.end())

ztorlakTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -52,6 +52,7 @@ void LLMController::resolveSession(
     auto slotId = sessionManager->acquireSessionSlot(req->sessionId.value());
```
ljovanovicTT, line 52: We should handle the error case where session is already in flight and a new request comes in, preferably within `SessionManager`'s `acquireSessionSlot`

From an API perspective, it would be great to have a clear error code

idjuricTT: yep, I'm adding this in a separate PR 😄

ljovanovicTT (approved):

Issue #2799, opened 2026-04-07
[Helm] Automatic configuration of host resource requests


bgoelTT:
Define minimum guarantees and maximum constraints for standard compute resources.

Tasks:
* Implement configurable `resources.requests` and `resources.limits` for CPU and Memory.
* Ensure default requests are inferred from the selected ModelSpec.

Issue #2786, opened 2026-04-06
[Helm] Automated values.yaml Catalogue Generation


bgoelTT:
Develop the Python tooling required to generate the monolithic `values.yaml` catalogue from the existing model_spec.py logic.

Tasks:
* Implement `workflows/generate_helm_values.py` to expand ModelSpecTemplate into a YAML catalogue.
* Structure the models block in `values.yaml` as a two-level map keyed by <model-name> and <device-name>.
* Ensure each leaf node contains full container specifications (image, env, deviceResource, etc.)

bgoelTT:
Clarifying a point of confusion that was not explicitly mentioned in the original Helm design doc: **we should not support any vLLM images <= v0.10.1 as they require an unwieldly docker interface**

PR #2977, opened 2026-04-16
Helm charts, initial version

### LINKED ISSUE #2784: [Helm] Implement core template trinity
Initialize the standard Helm chart directory structure and implement the base resource templates required for the MVP v0.1.0 tt-inference-server chart. 

Specifically, these steps should be perfomed in order:
1. Create Chart.yaml with initial metadata.
2. Implement `templates/service.yaml` supporting ClusterIP by default.
3. Implement `templates/deployment.yaml` for the main inference container.
4. Implement `templates/configmap.yaml` to store the resolved `ModelSpec`'s environment configuration.
  a. Update `templates/deployment.yaml` to use `envFrom.configMapRef` instead of hard-coded environment blocks.
5. Add checksum/spec annotations to the Pod specification in `templates/deployment.yaml` to trigger rolling restarts on config changes. 
6. Define basic `values.yaml` structure to support replicaCount and image parameters.
7. After #2786 is addressed, implement the following (BLOCKED UNTIL ADDRESSED)
  b. Perform ModelSpec lookup via `templates/_helpers.tpl` see example helper below
  c. Template `configmap.yaml`, `service.yaml`, `deployment.yaml` from ModelSpec selection in step 2.

Example `templates/_helpers.tpl`:
```
{{/*
================================================================================
  tt-inference-server helpers
================================================================================
*/}}

{{- define "tt-inference-server.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
{{- end }}

{{- define "tt-inference-server.fullname" -}}
{{- if .Values.fullnameOverride }}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
{{- else }}
{{- $name := default .Chart.Name .Values.nameOverride }}
{{- if contains $name .Release.Name }}
{{- .Release.Name | trunc 63 | trimSuffix "-" }}
{{- else }}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
{{- end }}
{{- end }}
{{- end }}

{{- define "tt-inference-server.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
{{- end }}

{{- define "tt-inference-server.labels" -}}
helm.sh/chart: {{ include "tt-inference-server.chart" . }}
{{ include "tt-inference-server.selectorLabels" . }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{- define "tt-inference-server.selectorLabels" -}}
app.kubernetes.io/name: {{ include "tt-inference-server.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}

{{- define "tt-inference-server.serviceAccountName" -}}
{{- if .Values.serviceAccount.name }}
{{- .Values.serviceAccount.name }}
{{- else }}
{{- include "tt-inference-server.fullname" . }}
{{- end }}
{{- end }}

{{/*
================================================================================
  Model spec lookup  — the heart of the indirection layer
  
  Usage:  {{ include "tt-inference-server.modelSpec" . }}
  Returns the spec dict for (.Values.model, .Values.device).
  Fails loudly with a clear error if the combination doesn't exist.
================================================================================
*/}}
{{- define "tt-inference-server.modelSpec" -}}

{{- /* 1. Validate that --set model=X was provided */ -}}
{{- if not .Values.model }}
{{- fail (printf "\n\n[tt-inference-server] ERROR: --set model=<name> is required.\nAvailable models:\n%s"
    (join "\n" (keys .Values.models | sortAlpha))) }}
{{- end }}

{{- /* 2. Validate that --set device=Y was provided */ -}}
{{- if not .Values.device }}
{{- fail (printf "\n\n[tt-inference-server] ERROR: --set device=<name> is required.\nAvailable devices for model %q:\n%s"
    .Values.model
    (join "\n" (keys (index .Values.models .Values.model) | sortAlpha))) }}
{{- end }}

{{- /* 3. Look up the model family */ -}}
{{- if not (hasKey .Values.models .Values.model) }}
{{- fail (printf "\n\n[tt-inference-server] ERROR: model %q not found in catalogue.\nAvailable models:\n%s"
    .Values.model
    (join "\n" (keys .Values.models | sortAlpha))) }}
{{- end }}

{{- $modelFamily := index .Values.models .Values.model -}}

{{- /* 4. Look up the device variant */ -}}
{{- if not (hasKey $modelFamily .Values.device) }}
{{- fail (printf "\n\n[tt-inference-server] ERROR: device %q is not valid for model %q.\nAvailable devices:\n%s"
    .Values.device
    .Values.model
    (join "\n" (keys $modelFamily | sortAlpha))) }}
{{- end }}

{{- /* 5. Emit the resolved spec as JSON so callers can fromJson it */ -}}
{{- index $modelFamily .Values.device | toJson -}}

{{- end }}
```

vmaksimovicTT:
Initial version of helm charts. Several models (Llama-3.1-8B, Llama-3.1-70B, whisper-large, Qwen3-32B) have been tested on the single node k8s cluster (UF machine).

What is there yet do to:
[ ] - More comprahensive testing of these charts for every releas, for example
[ ] - A scallable way of figuring lout, automatically, what is startup time for each model, and their peak resource usage during warmup/setup


acvejicTT:
@vmaksimovicTT would it make sense to have a Readme.md in chart dir, to explain how it is generated, how should be used etc.

.github/workflows/test-gate.yml
```diff
@@ -1230,6 +1233,35 @@ jobs:
...
+          curl -sSL "https://github.com/yannh/kubeconform/releases/download/${KUBECONFORM_VERSION}/kubeconform-linux-amd64.tar.gz" \
+            | tar xz
+          sudo mv kubeconform /usr/local/bin/
+
+      - name: Helm template validation
+        run: |
+          helm template ci-test charts/tt-inference-server \
+            --set model=Llama-3.1-8B-Instruct \
+            --set device=galaxy \
+            | kubeconform -strict -summary
```
mjeremicTT: You are just validating against specific model/device combination here, what if some required prop is missed for other model/device combination? This is not expensive operation, I would template all the possible combinations and validate them with kubeconform.

charts/tt-inference-server/templates/deployment.yaml
```diff
@@ -0,0 +1,140 @@
+{{- include "tt-inference-server.validateValues" . }}
```
mjeremicTT, line 1: I think there is no need to validateValues in every manifest, you could just keep it in `deployment.yaml` as it is most important one

charts/tt-inference-server/values.yaml
```diff
@@ -0,0 +1,373 @@
...
+fullnameOverride: ""
+
+# Optional: override the cache host path (defaults to /opt/cache/<model>-<device>)
+cache:
+  hostPath: ""
+
+# ---------------------------------------------------------------------------
+# Defaults — merged with (and overridden by) per-model config below.
+# ---------------------------------------------------------------------------
+defaults:
```
mjeremicTT, line 28: There are some default values which are pretty static and not meant to be changed frequently, such as replicaCount, progressDeadlineSeconds, hostIPC, etc. and they are also not being overwritten by model/device combinations. From my point of view, I would probably add their values directly to manifest without need for templating, but I leave decision up to you

charts/tt-inference-server/values.yaml
```diff
@@ -0,0 +1,373 @@
...
+      periodSeconds: 30
+      timeoutSeconds: 5
+      failureThreshold: 3
+    readiness:
+      enabled: true
+      path: /health
+      initialDelaySeconds: 2400
+      periodSeconds: 10
+      timeoutSeconds: 5
+      failureThreshold: 3
```
mjeremicTT: NOTE: Be careful with readiness and liveness probes as we are using only one replica. I would probably give slightly more initialDelaySeconds for readiness.
[https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/](https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/)

charts/tt-inference-server/values.yaml
```diff
@@ -0,0 +1,373 @@
...
+    serverType: media
+    galaxy:
+      image:
+        repository: ghcr.io/tenstorrent/tt-media-inference-server
+        tag: 0.10.0-555f240
+      resources:
+        limits:
+          memory: 64Gi
+        requests:
+          memory: 32Gi
```
mjeremicTT: For sake of curiosity, how you determined requests and limits for memory?

vmaksimovicTT: For this small batch of models - trial and error. I started the model (usually with large enough memory limit), monitored the usage of the pod, and based on the peak set these values. Per the PR description, this approach is far from ideal, and not scalable at all. It should be revisited and improved in one of the next iterations

charts/tt-inference-server/values.yaml
vmaksimovicTT: These initial delays are also quite more then it actually took for these model to warmup. Again, the values were set manually, and the actual startup time for the pod on one galaxy was considerably shorter. This approach is also just for v0.1, and should be improved so that it can easily scale and support adding new models without manually testing the inital delay

vmaksimovicTT: BTW, interesting blog post :)

.github/workflows/test-gate.yml
vmaksimovicTT: Done

charts/tt-inference-server/values.yaml
vmaksimovicTT: Good point, I moved several things to the manifests. I did not move replicaCount, though, as we want to support multiple instances of the server on the same machine, and it will be used then.

I also did not move progressDeadlineSeconds, as there are several big models for whom the value is overwritten to be greater than 3600

charts/tt-inference-server/values.yaml
```diff
@@ -0,0 +1,369 @@
...
+# Required at install time:
+#   helm install my-model ./charts/tt-inference-server-v2 \
+#     --set model="Llama-3.1-8B-Instruct" \
+#     --set device=galaxy \
+#     --set hfToken="hf_xxx"
+# ---------------------------------------------------------------------------
+
+model: ""
+device: ""
+hfToken: ""
```
sott0n, line 14: don’t see `hfCacheDir` defined anywhere. Is it required?

charts/tt-inference-server/templates/deployment.yaml
```diff
@@ -0,0 +1,140 @@
...
+  labels:
+    {{- include "tt-inference-server.labels" . | nindent 4 }}
+spec:
+  replicas: {{ $cfg.replicaCount }}
+  progressDeadlineSeconds: {{ $cfg.progressDeadlineSeconds }}
+  selector:
+    matchLabels:
+      {{- include "tt-inference-server.selectorLabels" . | nindent 6 }}
+  strategy:
+    type: RollingUpdate
```
sott0n: With `RollingUpdate`, the old Pod may still hold mmap'd hugepages while the new Pod starts. In that case, even if the cleanup, the actual memory might not be freed, and the new Pod may fail to allocate hugepages.
If we switch to `Recreate`, the new Pod would start only after the old Pod has fully terminated, so cleanup should work correctly, I think.
Did your tests work without issues using this strategy?

charts/tt-inference-server/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: tt-inference-server
+description: Tenstorrent inference server — vLLM and media backends on Galaxy hardware
+type: application
+version: 0.2.0
```
bgoelTT:
```suggestion
version: 0.1.0
```

charts/tt-inference-server/README.md
```diff
@@ -0,0 +1,235 @@
+# tt-inference-server Helm Chart
+
+Deploys vLLM and media inference backends on Tenstorrent Galaxy hardware.
+
+**Chart version:** 0.2.0 | **App version:** 0.12.0
```
bgoelTT:
```suggestion
**Chart version:** 0.1.0 | **App version:** 0.12.0
```

charts/tt-inference-server/README.md
```diff
@@ -0,0 +1,235 @@
...
+---
+
+## Overview
+
+This chart creates a single `Deployment`, `Service`, `ConfigMap`, and `Secret` for one inference model on one Tenstorrent device. The chart ships with pre-validated configurations for every supported model and device combination. You select a model and device at install time; the chart merges your selection against a layered config system to produce the final Kubernetes resources.
+
+Two server types are supported:
+
+- **vllm** — large language models served via the vLLM OpenAI-compatible API
+- **media** — image/audio/video models served via the tt-media-inference-server API
```
bgoelTT:
```suggestion
- **media** — image/audio/video/embedding models served via the tt-media-inference-server API
```

charts/tt-inference-server/README.md
```diff
@@ -0,0 +1,235 @@
...
+---
+
+## Values Reference
+
+| Value | Required | Default | Description |
+|---|---|---|---|
+| `model` | yes | `""` | Model name. Must match a key in `models`. |
+| `device` | yes | `""` | Device name. Must match a key under `models.<model>`. |
+| `hfToken` | yes* | `""` | HuggingFace token. Injected as `HF_TOKEN`. Required unless weights are pre-downloaded. |
+| `hfCacheDir` | no | `""` | Host path to a pre-downloaded HuggingFace weights directory. Skips download at startup. |
```
bgoelTT: Is this the path to the weights or the root of the entire HF cache directory?

charts/tt-inference-server/README.md
```diff
@@ -0,0 +1,235 @@
...
+| `cache.hostPath` | no | `""` | Override the host path used for the cache volume. Defaults to `/opt/cache/<model>-<device>`. |
```
bgoelTT:
```suggestion
| `cache.hostPath` | no | `""` | Override the host path used for the ttnn cache volume. Defaults to `/opt/cache/<model>-<device>`. |
```

charts/tt-inference-server/README.md
```diff
@@ -0,0 +1,235 @@
...
+| `defaults.service.protocol` | `TCP` | Service protocol. |
+| `defaults.service.annotations` | `{}` | Annotations applied to the Service. |
+| `defaults.resources.limits.cpu` | `8` | CPU limit. |
+| `defaults.resources.limits.memory` | `128Gi` | Memory limit (overridden per model). |
+| `defaults.resources.limits.hugepages-1Gi` | `32Gi` | Hugepage limit. |
+| `defaults.resources.requests.cpu` | `6` | CPU request. |
+| `defaults.resources.requests.memory` | `64Gi` | Memory request (overridden per model). |
+| `defaults.resources.requests.hugepages-1Gi` | `32Gi` | Hugepage request. |
+| `defaults.probes.liveness.enabled` | `true` | Enable liveness probe. |
+| `defaults.probes.liveness.path` | `/v1/models` | Liveness probe HTTP path. |
```
bgoelTT: We might have to make this conditional on the serverType. I don't know if media server always serves `/v1/models` for all models

charts/tt-inference-server/README.md
```diff
@@ -0,0 +1,235 @@
...
+| `whisper-large-v3` | galaxy |
+| `distil-large-v3` | galaxy |
+| `stable-diffusion-xl-base-1.0` | galaxy |
+| `stable-diffusion-xl-base-1.0-img-2-img` | galaxy |
+| `FLUX.1-dev` | galaxy |
+| `FLUX.1-schnell` | galaxy |
+| `Wan2.2-T2V-A14B-Diffusers` | galaxy |
+| `mochi-1-preview` | galaxy |
+
+To add a new model, add an entry under `models` in `values.yaml` with at least one device sub-key containing `image.repository` and `image.tag`.
```
bgoelTT, line 266: We should aim to automatically populated `values.yaml` using the model spec. This is okay for now

charts/tt-inference-server/templates/deployment.yaml
```diff
@@ -0,0 +1,140 @@
...
+        - name: inference-server
+          image: {{ include "tt-inference-server.image" . }}
+          imagePullPolicy: {{ $cfg.image.pullPolicy }}
+          {{- if eq $cfg.serverType "vllm" }}
+          args:
+            - "--model"
+            - {{ .Values.model | quote }}
+            - "--tt-device"
+            - {{ .Values.device | quote }}
+          {{- end }}
```
bgoelTT, line 70: Is the media server entrypoint compatible with these kwargs?

bgoelTT (changes requested):
Looks great! Just a few nitpicks

charts/tt-inference-server/values.yaml
acvejicTT: This should come from model_spec, as work under this ticket:
https://github.com/tenstorrent/tt-inference-server/issues/2799

charts/tt-inference-server/README.md
acvejicTT: @idjuricTT do you have an answer to this? This is to test liveness of server?

charts/tt-inference-server/README.md
acvejicTT: That's work under this [issue](https://github.com/tenstorrent/tt-inference-server/issues/2786), correct?

PR #2991, opened 2026-04-16
Fix: pass CACHE_ROOT env var to docker container for all engine types

### LINKED ISSUE #2990: CACHE_ROOT not passed to docker container for non-media engine types
## Summary

`CACHE_ROOT` is only set in `get_media_server_docker_env_vars`, which means it is only injected into the container for `FORGE` and `MEDIA` engine types. For all other engines (e.g. vLLM, SGLang), `CACHE_ROOT` is never passed as an environment variable even though the container entrypoint relies on it.

## Fix

Set `docker_env_vars["CACHE_ROOT"]` unconditionally whenever `setup_config` is present in `generate_docker_run_command`.

stisiTT:
## Summary

- `CACHE_ROOT` was only injected into the container for `FORGE`/`MEDIA` engines via `get_media_server_docker_env_vars`
- Non-media engines (vLLM, SGLang, etc.) never received `CACHE_ROOT`, breaking container entrypoints that depend on it
- Set `CACHE_ROOT` unconditionally from `setup_config.cache_root` whenever `setup_config` is present

Closes #2990

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -197,6 +197,7 @@ def generate_docker_run_command(
 
     docker_env_vars = {}
     if setup_config:
+        docker_env_vars["CACHE_ROOT"] = str(setup_config.cache_root)
         if (
             setup_config.container_model_weights_path
             and setup_config.host_model_weights_mount_dir


stisiTT:
Closing this PR after investigation. The fix is real but only necessary when a non-default `cache_root` path is used. For standard deployments, the Dockerfile already bakes in `ENV CACHE_ROOT=/home/container_app_user/cache_root` which matches `SetupConfig.cache_root`'s default — so this line would be a no-op in practice. The timing of this branch (created during the dev→main transition) was coincidental; neither `main` nor `dev` have this line. If custom host volume paths ever become a real use case, this fix would be needed to prevent the hard assert in `run_vllm_api_server.py` from failing due to a mismatch between the mounted path and the env var.

Issue #2692, opened 2026-04-01
Get pipeline manager hang signal and flag it in inference server


idjuricTT:
If hang signal is caught notify Kubernetes

PR #2996, opened 2026-04-17
Hang detection support: Report worker metrics 

dmadicTT:
Issue: http://github.com/tenstorrent/tt-inference-server/issues/2692
PR: https://github.com/tenstorrent/tt-inference-server/pull/2996

Reports
- heartbeat timestamp - each event loop step
- last output timestamp - each token
- active requests

Hang detection
- last heartbeat was long time ago, meaning event loop does not run 
- last output timestamp is long time ago while we have active requests, model does not output

Next steps
- Get feedback from Jacek and others
- New PR for worker to die once it detects a hang. This way /health monitoring can see somethig is wrong.

image

https://github.com/user-attachments/assets/25b7e8b0-52ca-4625-9e8e-108e35ca8227


tt-media-server/cpp_server/include/worker/single_process_worker_metrics.hpp
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+#pragma once
```
idjuricTT, line 4: I would call this single_process_worker_metrics - sounds odd but it would be more clear what it tracks

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -192,6 +194,7 @@ SpPipelineRunner::getMemoryRequest() {
 }
 
 void SpPipelineRunner::handleOutput(const pm::OutputMessage& output) {
+  tt::worker::WorkerMetrics::instance().updateOutputHeartbeat();
```
idjuricTT: Would this affect the perf?

idjuricTT: Nitpick but maybe better to do it in the end of the method, not the beginning?

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
```diff
@@ -224,6 +230,7 @@ inline void SpPipelineRunner::evictSlot(uint32_t slotId) {
 
 void SpPipelineRunner::handleRequest(
     std::unique_ptr<tt::runners::llm_engine::Sequence> request) {
+  tt::worker::WorkerMetrics::instance().incrementActiveRequests();
```
idjuricTT: Same here, I'd do this in the end of the method, otherwise it might get out of sync if i.e. slot id is invalid

idjuricTT (approved):

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
dmadicTT: > Would this affect the perf?

I don't think so. It should be the same as writing to a variable.

tt-media-server/cpp_server/include/services/llm_service.hpp
```diff
@@ -52,6 +52,10 @@ class LLMService
    */
   void abortRequest(uint32_t taskId);
 
+  /** Borrowed pointer to the worker manager, used by main to wire the
+   * worker metrics aggregator. Lifetime tied to this LLMService. */
+  tt::worker::WorkerManager* workerManager() { return worker_manager_.get(); }
```
knovokmetTT: func can be const

tt-media-server/cpp_server/include/worker/worker_metrics_shm.hpp
```diff
@@ -0,0 +1,90 @@
...
+constexpr uint32_t WORKER_METRICS_SHM_MAGIC = 0x54545744;  // "TTWD"
+constexpr size_t MAX_WORKER_SLOTS = 32;
+constexpr size_t WORKER_SCRATCH_U64_COUNT = 32;
+
+/**
+ * Names what a slot's scratch area means. Each worker writes its own value
+ * during initialize() so the main-side aggregator can dispatch to the
+ * correct renderer. Append-only; never renumber existing values.
+ */
+enum class MetricsLayout : uint32_t {
```
knovokmetTT: can be uint8?

tt-media-server/cpp_server/src/worker/worker_metrics_shm.cpp
```diff
@@ -0,0 +1,118 @@
...
+  region->num_workers.store(static_cast<uint32_t>(numWorkers),
+                            std::memory_order_release);
+
+  TT_LOG_INFO(
+      "[WorkerMetricsShm] Created shared region '{}' ({} bytes, {} slots)",
+      name, REGION_SIZE, numWorkers);
+  return region;
+}
+
+WorkerMetricsShmRegion* openSharedRegion(const std::string& name) {
```
knovokmetTT: create will behave like open if the file exists already
if we dont care which process creates the actual file, this method is useless

tt-media-server/cpp_server/src/worker/worker_metrics_shm.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace tt::worker {
+
+namespace {
+
+constexpr size_t REGION_SIZE = sizeof(WorkerMetricsShmRegion);
+
+}  // namespace
+
+WorkerMetricsShmRegion* createSharedRegion(const std::string& name,
```
knovokmetTT: id prefer we dont expose raw pointers to shm RAM, but create inline write/read methods, and getters/setters for metadata

knovokmetTT (approved):

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
dmadicTT: > Nitpick but maybe better to do it in the end of the method, not the beginning?

As we have few exist points from the method I would keep it here

tt-media-server/cpp_server/src/runners/sp_pipeline_runner/sp_pipeline_runner.cpp
dmadicTT: If slot ID is invalid debug run will abort, so does not matter, while production run will skip it, so does not really matter.
Since we have multiple exit points from the method I would prefer to keep it at the top

PR #3006, opened 2026-04-17
Tool call handling base

lmuravljovTT:
#3010 

 # Add Tool Calling Support (Base Implementation)

  ## Overview

  This PR adds foundational tool calling support to the C++ media server. This base implementation enables the server to:
  - Accept tool definitions in chat completion requests
  - Apply model-specific chat templates with tool instructions (only deepseek template available rn.)
  - Parse tool calls from model responses
  - Return tool calls

  ## Changes Made

  ### 1. Request Parsing
  - **Tool Definitions**: Parse `tools` array from `ChatCompletionRequest`
  - **Tool Choice**: Parse `tool_choice` parameter (currently supports `auto` only)
  - **Domain Objects**: Added `Tool`, `FunctionDefinition`, and `ToolChoice` structures

  ### 2. Chat Template Application
  - **Tokenizer Integration**: Extended `applyChatTemplate()` to accept optional tools parameter
  - **DeepSeek Implementation**: Added tool calling template format for DeepSeek-R1
    - Includes system instructions for tool usage
    - Formats tool definitions in model-specific syntax
    - Uses special markers: `<｜tool▁calls▁begin｜>`, `<｜tool▁call▁begin｜>`, etc.
  - **Enum Refactoring**: Replaced `bool addGenerationPrompt` with `GenerationPrompt` enum for
  better clarity

  ### 3. Response Parsing
  - **Parser Interface**: Created `IToolCallParser` interface for model-specific parsing
    - `parseComplete()`: Extract tool calls from model output
    - `stripMarkers()`: Remove tool call markers from content
  - **DeepSeek Implementation**: `DeepSeekToolCallParser` extracts function calls from DeepSeek
  format
  - **Factory Pattern**: `createToolCallParser()` selects parser based on `ModelType` enum
  - **Response Integration**:
    - Added `tool_calls` field to `ChatCompletionMessage`
    - `LLMService.postProcess()` parses and populates tool calls
    - Sets `finish_reason` to "tool_calls" when tools are detected

  ### 4. Testing

  **Unit Tests:**
  - `test_chat_template_tools.cpp`: Validates chat template application with tools
    - Configurable test pattern for multiple tokenizers (DeepSeek implemented)
    - Tests single tool, multiple tools, conversation history with tools
    - Exact template matching with detailed diff output

  - `test_tool_call_parser.cpp`: Validates tool call parsing
    - Configurable format pattern for multiple models (DeepSeek implemented)
    - Tests single/multiple tool calls, complex JSON arguments, edge cases
    - Validates OpenAI format compliance

  **Integration Tests:**
  - `test_tool_call_integration.cpp`: End-to-end data flow verification
    - `LLMResponse` → `ChatCompletionResponse` conversion with tool calls
    - Single/multiple tool calls in response
    - Responses without tool calls
    - Complex nested JSON arguments

  ## Architecture

  Request Flow:
  ChatCompletionRequest (with tools)
    → applyChatTemplate(messages, tools)
    → Model generates response with markers
    → IToolCallParser.parseComplete()
    → LLMChoice (with tool_calls)
    → ChatCompletionResponse (OpenAI format)

  **Extensibility:**
  - Interface-based design allows easy addition of new models
  - Tokenizer-specific configs separate format from logic
  - Parser factory enables model-specific implementations

  ## References

  - [vLLM Tool Calling Documentation](https://docs.vllm.ai/en/latest/features/tool_calling/)
  - [OpenAI Tool Calling API](https://platform.openai.com/docs/guides/function-calling)
  - [DeepSeek-R1 Model Format Specification](https://github.com/sgl-project/sglang/blob/main/examples/chat_template/tool_chat_template_deepseekr1.jinja)


tt-media-server/cpp_server/tests/test_tool_call_parser.cpp
bgoelTT, line 1: Looks great, could you also please add a test to verify the `parallel_tool_calls` API parameter? https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20parallel_tool_calls%20%3E%20(schema)

It should cover the two cases where `parallel_tool_calls=True` and `parallel_tool_calls=False`

bgoelTT (changes requested):
Just one request for an additional parameter and test

tt-media-server/cpp_server/tests/test_tool_call_parser.cpp
lmuravljovTT: Yep, 
I've added only cutting of the list in the parser for now, as its a quick addition and an okay place to start. 
But I'd guess we'll want to do it using constrained decoding soon enoguh?

tt-media-server/cpp_server/include/domain/tool_calls/function_definition.hpp
ztorlakTT, line 1: Typo - it should be `function_definition.hpp`

tt-media-server/cpp_server/include/domain/tool_calls.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <json/json.h>
+#include <json/value.h>
+
+#include <optional>
+#include <string>
+
+namespace tt::domain {
+
+struct FunctionDefinition {
```
ztorlakTT: We have FunctionDefinition defined in the function_definition.hpp too

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    // Pattern to match:
+    // function<｜tool▁sep｜>function_name\n```json\n{...}\n``` Using a simpler
+    // approach: find the markers and extract content between them
+    size_t pos = 0;
+    while (true) {
+      if (!parallelToolCalls && toolCallsArray.size() == 1) {
+        return toolCallsArray;
+      }
+      // Find next tool call
```
ztorlakTT, line 53: Can you define constexpr for all three and then use .size(), for example:
```
constexpr std::string_view kToolCallBegin  = "<｜tool▁call▁begin｜>";
constexpr std::string_view kToolSep        = "<｜tool▁sep｜>";
constexpr std::string_view kToolCallsEnd   = "<｜tool▁calls▁end｜>";
```

tt-media-server/cpp_server/include/domain/tool_calls.hpp
lmuravljovTT: good catch, this file should be removed

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
lmuravljovTT: Yep I should extract the tags .. Not sure where we should use .size() though?

ztorlakTT: I think you could use kToolCallBegin.size() instead of hardcoded 20

tt-media-server/cpp_server/include/services/tool_call_parser.hpp
```diff
@@ -0,0 +1,58 @@
...
+   *     "type": "function",
+   *     "function": {
+   *       "name": "get_weather",
+   *       "arguments": "{\"location\":\"San Francisco\"}"
+   *     }
+   *   }
+   * ]
+   */
+  virtual std::optional<Json::Value> parseComplete(
+      const std::string& text, const bool& parallelToolCalls = true) const = 0;
```
ztorlakTT: `parallelToolCalls` is a trivial type so you can pass it by value
```suggestion
      const std::string& text, const bool parallelToolCalls = true) const = 0;
```

ztorlakTT: Also, `parallelToolCalls` is never passed when calling `parseComplete`

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
lmuravljovTT: Got it, pushed :D

tt-media-server/cpp_server/include/services/tool_call_parser.hpp
lmuravljovTT: It's passed in one relevant test

ztorlakTT (approved):

tt-media-server/cpp_server/tests/test_tool_call_parser.cpp
bgoelTT: I think this is a great place to start 👍 

Yes, when guided decoding is implemented it would be worth writing an integration or e2e test that validates only a single tool call is generated.

bgoelTT (approved):
Nice!

PR #3028, opened 2026-04-17
Fix SessionManager TOCTOU race and orphaned workers

### LINKED ISSUE #2907: TOCTOU race in SessionManager::evictOldSessions can orphan KV cache slots
## Summary

`SessionManager::evictOldSessions()` has a Time-of-Check, Time-of-Use (TOCTOU) race between scanning sessions via `forEach` and removing them via `take`. A concurrent `acquireSessionSlot` call can mark a session as `in_flight` in the gap between these two operations, but the subsequent `take` removes the session unconditionally — without re-checking the `in_flight` flag.

This causes the slot to be deallocated while an active request still holds a reference to it, resulting in orphaned pipeline slots and loss of conversation context.

## Affected Code

`src/services/session_manager.cpp` — `evictOldSessions()` (lines 145–218)

```cpp
// Step 1: forEach holds the lock, checks isInFlight(), builds eviction heap
sessions.forEach([&heap, ...](const std::string& id, domain::Session& session) {
  if (session.isInFlight()) return;  // CHECK
  // ... add to heap
});
// ← Lock RELEASED here — acquireSessionSlot() can run on another thread

// Step 2: take re-acquires lock but does NOT re-check isInFlight()
for (const auto& [_, sessionId] : heap) {
  auto session = sessions.take(sessionId);  // USE — removes unconditionally
  sendDeallocRequest(sessionId, slotId);
}
```

`closeSession()` has a similar (but deterministic) variant — it calls `sessions.take()` without checking `in_flight` at all, so a `DELETE /v1/sessions/{id}` during an active request deallocates the slot out from under it.

## Race Scenario

| Step | Thread A (User A, turn 2 request) | Thread B (new session creation) | Pipeline Runner |
|------|-----------------------------------|--------------------------------|-----------------|
| T0 | | | slot 1 in `cached` (User A's prior turn) |
| T1 | | `createSession()` → `evictOldSessions()` | |
| T2 | | `forEach`: sees User A's session not in-flight, adds to eviction heap. **Lock released.** | |
| T3 | `acquireSessionSlot("S1")`: sets `in_flight=true`, returns slot 1 | | |
| T4 | Creates `Sequence(slot=1)`, pushes to task queue | | |
| T5 | | `sessions.take("S1")`: removes session **despite `in_flight=true`**. Sends `DEALLOCATE` for slot 1. | |
| T6 | | | Processes `DEALLOCATE`: `CANCEL(slot=1)`, `evictSlot(1)` clears `cached` |
| T7 | | | Processes User A's `Sequence(slot=1)`: `isNew=true` → fresh `SUBMIT` (not `CONTINUE`) |
| T8 | Response completes. `setSessionInFlight("S1", false)` → no-op (session gone). Slot 1 left in `cached`, orphaned. | | |

## Impact

- **Slot leak**: The deallocated-then-reused slot ends up in the runner's `cached` set with no owning session. It is never reclaimed.
- **Lost conversation context**: The `CANCEL` wipes the KV cache before the request's `Sequence` is processed. The runner does a fresh `SUBMIT` instead of a `CONTINUE`, so the user loses all multi-turn history.
- **Potential slot collision**: If pipeline manager reuses the slot number for a new allocation before User A's stale `SUBMIT` is processed, two users can race on the same hardware slot.

## Suggested Fix

Add a conditional take to `ConcurrentMap` that atomically checks a predicate:

```cpp
template <typename Pred>
std::optional<Value> takeIf(const Key& key, Pred&& pred) {
  std::lock_guard lock(mutex);
  auto it = map_.find(key);
  if (it == map_.end()) return std::nullopt;
  if (!pred(it->second)) return std::nullopt;
  auto value = std::move(it->second);
  map_.erase(it);
  return value;
}
```

Then in `evictOldSessions`:

```cpp
auto session = sessions.takeIf(sessionId, [](const domain::Session& s) {
  return !s.isInFlight();
});
if (!session.has_value()) continue;  // was acquired between forEach and now
```

For `closeSession`, either reject closing in-flight sessions or defer the deallocation until the in-flight request completes.

### LINKED ISSUE #3001: [cpp_server] Main process aborts with 'malloc(): unaligned tcache chunk detected' under streaming load; suspected SessionManager eviction race
## Summary

Under sustained streaming benchmark load (3 workers, `mock_pipeline` backend, stream=1, ~hundreds of in-flight `/v1/chat/completions` requests), the main `tt_media_server_cpp` process **aborts with glibc heap corruption** the first time the `SessionManager` evicts sessions:

```
malloc(): unaligned tcache chunk detected
```

The three worker subprocesses are not killed (they continue running orphaned under PID 1), and the POSIX shm region `/dev/shm/tt_worker_metrics` is left behind because the main process did not run its graceful shutdown path.

## Environment

- Branch: `dmadicTT/worker-metrics`, commit `b74ea74b9e8403ebc64eaeee3de3832915406fb2` (this PR only touches worker-metrics SHM; none of these code paths are on the allocation hot path).
- Build: `./build.sh --blaze` (Release, no sanitizers).
- Launch: `DEVICE_IDS="(0),(1),(2)" LLM_DEVICE_BACKEND=mock_pipeline ./build/tt_media_server_cpp -p 8000`.
- Workload: streaming `/v1/chat/completions`, max_tokens=128, stream=1, high concurrency (task_ids climbing through ~540 when the crash fired).

## Symptom / evidence

Final tail of `cpp_server/logs/server.log`:

```
[07:40:16.196445] [warning] [SessionManager] No pending allocation found for task ID: 511
[07:40:16.196448] [warning] [SessionManager] No pending allocation found for task ID: 512
... (many more, task ids 509..517) ...
[07:40:16.201409] [info]    [LLMController] /v1/chat/completions task_id=519 ...
...
[07:40:16.312378] [info]    [SessionManager] Evicted 10 oldest session(s) (active: 59/64, threshold: 90%)
[07:40:16.313008] [warning] [SessionManager] No pending allocation found for task ID: 540
[07:40:16.313017] [warning] [SessionManager] No pending allocation found for task ID: 541
... (more task ids 540..549) ...
malloc(): unaligned tcache chunk detected
```

`malloc(): unaligned tcache chunk detected` is glibc's tcache integrity check aborting the process (SIGABRT). It is almost always caused by one of:
1. double-free of a heap chunk,
2. use-after-free where the subsequent free sees corrupted tcache metadata,
3. a heap buffer overrun that corrupts an adjacent chunk's tcache header.

## Suspected root cause

A race between `SessionManager::evictOldSessions` and `SessionManager::handleMemoryResult` / the `/v1/chat/completions` request path.

Supporting signals:
- The burst of `No pending allocation found for task ID` warnings around and immediately after eviction is coming from `[SessionManager::handleMemoryResult](tt-media-server/cpp_server/src/services/session_manager.cpp:366)` when `pendingAllocationsMap.take(result.taskId)` returns empty. That means a task id's pending allocation got removed concurrently from somewhere other than the result handler.
- The abort happens in the same second as the eviction (`Evicted 10 oldest session(s) ... threshold: 90%`), under concurrent traffic.
- Eviction ([session_manager.cpp:210-226](tt-media-server/cpp_server/src/services/session_manager.cpp:210)) loops calling `sessions.take(sessionId)` + `sendDeallocRequest()` while request threads are concurrently creating/looking up sessions on the same container. Depending on the `sessions` / `pendingAllocationsMap` container thread-safety guarantees, this is a plausible site for double free / use-after-free of a session or pending allocation value.

I did NOT bisect further or run with sanitizers; this is the best hypothesis from log evidence alone.

## Repro

1. Build release: `./build.sh --blaze`.
2. Launch with 3 workers + mock backend (see Environment).
3. Drive `/v1/chat/completions` with stream=1 at high concurrency (64+ parallel streams, max_tokens=128, random prompts) until you cross the SessionManager eviction threshold (default 90% of `MAX_SESSIONS`, e.g. 58/64 active).
4. The main process aborts within a second or two of the first `Evicted ... session(s)` log line.

## How to collect more signal next time

Any single one of these will pinpoint the offending line:

### 1. ASan (fastest path to a named culprit)
```bash
./build.sh --asan
ASAN_OPTIONS='abort_on_error=1:halt_on_error=1:detect_leaks=0:print_stacktrace=1' \
  DEVICE_IDS='(0),(1),(2)' LLM_DEVICE_BACKEND=mock_pipeline \
  ./build/tt_media_server_cpp -p 8000 2>&1 | tee logs/server_asan.log
```
Then replay the benchmark. ASan will print:
- `heap-use-after-free` / `double-free` / `heap-buffer-overflow` (tells us which class of bug),
- allocation site, free site, and access site with symbolized stacks,
- the two threads involved.

### 2. TSan (for the race itself, not the heap symptom)
```bash
./build.sh --tsan
```
TSan names the two threads + code locations accessing the shared `sessions` / `pendingAllocationsMap` state without synchronization.

### 3. Glibc debug + core dump (keeps release build)
```bash
ulimit -c unlimited
MALLOC_CHECK_=3  ./build/tt_media_server_cpp ...
# On abort:
coredumpctl dump <pid> -o core.ttms     # or read /proc/sys/kernel/core_pattern
gdb ./build/tt_media_server_cpp core.ttms -ex 'thread apply all bt' -ex 'quit'
```
`MALLOC_CHECK_=3` makes glibc abort at the first inconsistency with a plain backtrace instead of waiting for tcache corruption symptoms.

### 4. Extra application-level logging
Consider promoting the following to `INFO` temporarily:
- `[SessionManager::evictOldSessions](tt-media-server/cpp_server/src/services/session_manager.cpp:193)` DEBUG lines.
- `[SessionManager::handleMemoryResult](tt-media-server/cpp_server/src/services/session_manager.cpp:366)` DEBUG lines.
- The DEALLOCATE path around `sendDeallocRequest` and the deferred dealloc queue.
So we can correlate which specific sessionId / taskId got freed twice.

## Impact

- High availability: the 3-worker HA setup loses the main process while workers keep running orphaned. Requests in flight fail, and externally the service is down until the parent is restarted by the orchestrator.
- Diagnosability: the corruption is pseudo-random; without sanitizers we won't catch it in CI.
- Cleanup: orphaned workers hold onto `/dev/shm/tt_worker_metrics`. Defensive `shm_unlink` on next main start recovers this, but users need to manually `kill` the orphaned worker PIDs.

## Proposed next steps

1. Reproduce with `--asan` (5 min) to confirm the class of bug and the offending line.
2. Depending on what ASan names, fix the SessionManager / pendingAllocations concurrency model — likely by extending the lock scope around eviction so eviction is atomic w.r.t. allocation result handling and request ingress.
3. Add a stress test that repeatedly drives SessionManager eviction under concurrent `/v1/chat/completions` to guard the fix in CI.
4. Orthogonal but related: consider changing `WorkerManager` so it terminates its worker subprocesses when the main process exits abnormally (via `prctl(PR_SET_PDEATHSIG, SIGTERM)` in each worker after `execv`), so an HA orchestrator's liveness probe drives a clean restart instead of leaving zombies.

## Notes

- This is unrelated to the worker-metrics PR: the SHM writer is a single relaxed atomic store per update with no allocation, and the main-side aggregator runs only in the scrape thread which is not on the path of this crash. The crash would reproduce on `dev` HEAD before this PR under the same load.

ztorlakTT:
Fixes two concurrency bugs in SessionManager:
- TOCTOU race: Added ConcurrentMap::takeIf to atomically check isInFlight and remove a session in one lock scope. evictOldSessions and closeSession now use it, preventing dealloc of a KV slot while an active request holds it. closeSession returns CloseSessionResult so the HTTP layer can distinguish:
  - 200 OK — closed immediately
  - 409 Conflict — session was in-flight; close is deferred and executes automatically when the request completes (via a pendingClose flag on the session, checked in setSessionInFlight)
  - 404 Not Found — session does not exist
- Orphaned workers: Added prctl(PR_SET_PDEATHSIG, SIGTERM) in the worker child so it receives SIGTERM if the main process dies unexpectedly.


tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -60,13 +60,25 @@ void SessionManager::readerLoop() {
   }
 }
 
-bool SessionManager::closeSession(const std::string& sessionId) {
+CloseSessionResult SessionManager::closeSession(const std::string& sessionId) {
```
knovokmetTT, line 63: so we dont allow us to kill the session mid flight? I dont think this is good
It means that someone disconnects , he closes the tab, while he had a very long decode loop going, we wont stop the decode even tho nobody is there to receive the tokens

tt-media-server/cpp_server/src/services/session_manager.cpp
knovokmetTT: how does this guard not protect us? if it truly doesnt , then why even have it in the first place here?

tt-media-server/cpp_server/src/services/session_manager.cpp
ztorlakTT: Updated

tt-media-server/cpp_server/src/services/session_manager.cpp
ztorlakTT: The guard is useful because it avoids putting actively-in-flight sessions into the heap at all, but it can't cover the window between this forEach and takeIf

idjuricTT (approved):

tt-media-server/cpp_server/include/domain/session.hpp
```diff
@@ -86,6 +97,7 @@ class Session {
   std::string session_id_;
   uint32_t slot_id_;
   bool in_flight_{false};
+  bool pending_close_{false};
```
dmadicTT, line 100: We should think about state machine(s) for the requests. Multiple booleans can make it easier to make programming mistakes, cause inconsistent states and make debugging harder.

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -60,24 +60,46 @@ void SessionManager::readerLoop() {
...
```
dmadicTT, line 63: Unrelated to this PR.
I'm kinda affraid of adding more logic to the session manager without any unit tests.
Can we prioritize that @idjuricTT ?

dmadicTT (approved):

tt-media-server/cpp_server/include/domain/session.hpp
ztorlakTT: Agreed, I was thinking the same — I'd leave it for a separate PR though

PR #3033, opened 2026-04-19
Prefill error handling

idjuricTT:
Changes:

- Update shared memory task_id format
- Add check for errors on prefill side
- Crash prefill runner if number of consecutive errors / timeouts > 5


tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
```diff
@@ -0,0 +1,156 @@
...
+      // Check if we should terminate
+      if (errorCount >= MAX_CONSECUTIVE_ERRORS) {
+        TT_LOG_CRITICAL(
+            "[SpPrefillRunner] Max consecutive errors ({}) reached. "
+            "Terminating worker process.",
+            MAX_CONSECUTIVE_ERRORS);
+        std::exit(1);  // Kill the worker process
+      }
+
+      continue;
```
madicd: I did a quick research:
- `continue` will cause the future destructor to be called.
- The destructor will wait for the future to complete.
- Which means the timeout has no effect.
- Next iteration will anyway run after `forward` completes.

Maybe moving the timeout logic to model runner, which responds with error token in case it timed out?
This level, the runner, can still keep track of consecutive errors and die in case it's more than threshold.
This way we avoid any new threads or futures.

madicd: Also the model itself is synchronous, so there is not much we can do in parallel while it's working.
We could overlap it with fetching the next sequence, but this is probably a separate motive for change, and not only related to this error case.

dmadicTT:
`mock_prefill_runner.py` and `test_shm_roundtrip.py` seems to still have references to old task ID format

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
idjuricTT: Makes sense, I'll make the change

idjuricTT:
> `mock_prefill_runner.py` and `test_shm_roundtrip.py` seems to still have references to old task ID format

@dmadicTT mock_prefill_runner uses shared_memory.py which is fixed so it should be applied.
For round_trip test good catch, I'll remove that file since no model developers started using it 😄

tt-media-server/cpp_server/src/runners/sp_prefill_runner/sp_prefill_runner.cpp
idjuricTT: Changes applied

dmadicTT (approved):

tt-media-server/cpp_server/src/runners/shared_memory.py
```diff
@@ -25,29 +25,30 @@
...
-        + task_id(36) + token_ids(max_token_ids × 8)
+    Layout per slot (must match C++ exactly):
+        state(4) + max_tokens(4) + num_token_ids(4) + task_id(4) + fast_mode(4)
+        + padding(4) + token_ids(max_token_ids × 8)
     """
 
     SLOTS = 64
     TOKEN_ID_SIZE = 8
-    TASK_ID_SIZE = 36
-    _HEADER_SIZE = 48
```
ljovanovicTT, line 44: Might be missing some context, but why are we deleting these fully and not aligning the new size?

ljovanovicTT (approved):

ztorlakTT (approved):

Issue #2946, opened 2026-04-15
Grab a hang flag from pipeline manager


idjuricTT:
Pipeline manager might throw a hang error - it should be propagated to the inference server.
Killing the process and flagging it is also fine

PR #3034, opened 2026-04-20
cpp server: self-terminate worker on model output hang

dmadicTT:
## Summary

- When the SP pipeline stops producing output while requests are in flight, the inference server can't recover on its own. This change detects that condition in `BlazeRunner` and `SpPipelineRunnerDemo` and `abort()`s the worker so the infrastructure monitoring stack (Prometheus `tt_worker_alive`, WorkerManager `CRITICAL` log, `/tt-liveness`) notices the crash and restarts the server instead of hanging silently.
- Per-runner timer tracks the steady-clock time of the last model output; reset on output arrival and on the 0→1 idle→active transition. If ``elapsed > threshold`` while at least one request is in flight, the runner logs `TT_LOG_CRITICAL` and calls `std::abort()`, which routes through the existing `fatalSignalHandler` / `WorkerManager` crash path.
- Threshold defaults to 10 s (`defaults::OUTPUT_HANG_TIMEOUT_MS`) and is overridable via the `OUTPUT_HANG_TIMEOUT_MS` env var, same pattern as `WARMUP_TIMEOUT_MS` / `PM_CONNECT_TIMEOUT_MS`.

Refs: https://github.com/tenstorrent/tt-inference-server/issues/2946

### DIFF
--- a/tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
+++ b/tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
@@ -4,6 +4,8 @@
 #include "runners/sp_pipeline_runner/blaze_runner.hpp"
 
 #include <cassert>
+#include <chrono>
+#include <cstdlib>
 #include <cstring>
 #include <pipeline_manager/pipeline_manager_types.hpp>
 #include <thread>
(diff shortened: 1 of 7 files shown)


knovokmetTT (approved):

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
```diff
@@ -211,6 +217,28 @@ void BlazeRunner::handleOutput(const pm::OutputMessage& output) {
   }
 }
 
+void BlazeRunner::checkOutputHang() {
+  if (running.empty()) {
+    lastOutputTime = std::chrono::steady_clock::now();
+    return;
+  }
+  auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
```
idjuricTT, line 225: Can we also add a check if i.e. we get an error 5 consecutive times that we stop the runner?

idjuricTT (approved):

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
dmadicTT: I'll see with @knovokmetTT what makes sense here

dmadicTT: He'll check it out. I will merge this for now

ztorlakTT (approved):

Issue #2927, opened 2026-04-15
Investigate prefix caching simulation on host and match with session id


idjuricTT:
Check how to simulate prefix caching to match session with request and isolate delta for sending to model

PR #3021, opened 2026-04-17
Add tool output support for DeepSeek

lmuravljovTT:
#3011 

## Add Tool Output Parsing and Handling                                                                                          
                                         
  Implements support for tool output messages (`role: "tool"`), completing the tool calling flow by allowing the model to receive function execution results.                                                                                                      
                                                                                                                                   
  ### Changes                                                                                                                                                                                                                                  
  **Domain Model** (`chat_message.hpp`)                                                                                            
  - Added `tool_call_id` field to `ChatMessage` with validation                                                                    
  - Tool messages must include `tool_call_id` or throw error
  - JSON parsing and serialization support                                                                                         
  
  **Chat Template** (`deepseek_tokenizer.cpp`)                                                                                     
  - Format tool outputs with DeepSeek markers (`<｜tool▁outputs▁begin｜>`, `<｜tool▁output▁begin｜>`)
  - Support single and multiple parallel tool outputs                                                                              
                                                                                                                                   
  **Tests** (`test_chat_template_tools.cpp`)                
  - Added tests for single and multiple tool outputs with exact matching                                                           
  - Created reusable test fixtures (`createWeatherTool()`, `createToolOutputMessage()`, etc.)                                                      
                                                                                                                                   
  ### Example Flow                                                                                                                 
                                                                                                                                   
  ```json                                                   
  1. Client → Model: [{"role": "user", "content": "Get SF weather"}]
  2. Model → Client: [{"role": "assistant", "tool_calls": [{"id": "call_1", ...}]}]                                                
  3. Client → Model: [{"role": "tool", "tool_call_id": "call_1", "content": "{\"temp\": 72}"}]                                     
  4. Template formats with markers → Model generates response

PR #3035, opened 2026-04-20
SessionManager Prefix Caching

ljovanovicTT:
## Issue
[#2927 Investigate prefix caching simulation on host and match with session id](https://github.com/tenstorrent/tt-inference-server/issues/2927)

## Problem
In order to remain compatible with OpenAI `/v1/chat/completions` API while staying efficient, we’re looking to implement prefix caching similar to vLLM.

The idea is to map the conversation history (the `messages` array received with each request [reference](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20messages%20%3E%20(schema))) to a specific session that tracks a `slotId`. By linking this ID to the "address" of the existing KV cache, we can pick up right where the conversation left off. This lets us maintain continuity without having to waste resources re-processing (prefillig) the entire prompt every time a new message is added.

## Implementation
Screenshot 2026-04-20 at 16 56 32

### Prompts:
USER1
```
curl http://localhost:8000/v1/chat/completions   -H "Content-Type: application/json"   -H "Authorization: Bearer your-secret-key"   -d '{
    "messages": [
      {"role": "user", "content": "Hello"} 
    ],                                                                  
    "stream": true,
    "max_tokens": 20
  }'
```

USER2
```
curl http://localhost:8000/v1/chat/completions   -H "Content-Type: application/json"   -H "Authorization: Bearer your-secret-key"   -d '{
    "messages": [
      {"role": "user", "content": "Hello"} 
    ],                                                                  
    "stream": true,
    "max_tokens": 20
  }'
```

USER1 Comes Back
```
curl http://localhost:8000/v1/chat/completions   -H "Content-Type: application/json"   -H "Authorization: Bearer your-secret-key"   -d '{
    "messages": [
      {"role": "user", "content": "Hello"},
      {"role": "assistant", "content": "Hi there! How can I help you?"},
      {"role": "user", "content": "How are you?"}
    ],
    "stream": true,
    "max_tokens": 20
  }'
```

### Logs:
Note: since this is on mock runner `slotId=123` is hardcoded, with real memory allocation slotIds would differ in first two logs.
```
USER1
[2026-04-20 15:43:39.125948] [tt-media-server] [info] [LLMController] /v1/chat/completions task_id=1 model=default stream=1 messages=1 last_msg=[user: "Hello"] max_tokens=20 temperature=none top_p=none top_k=none min_p=none presence_penalty=0 frequency_penalty=0 n=1 stop_count=0
[2026-04-20 15:43:39.127076] [tt-media-server] [info] [SessionManager] registerPrefixHash: registered slotId=123 under hash=11620829393098143408
[2026-04-20 15:43:39.127089] [tt-media-server] [info] [LLMController] New session: slotId=123, registered under hash=11620829393098143408

USER2
[2026-04-20 15:43:42.306496] [tt-media-server] [info] [LLMController] /v1/chat/completions task_id=3 model=default stream=1 messages=1 last_msg=[user: "Hello"] max_tokens=20 temperature=none top_p=none top_k=none min_p=none presence_penalty=0 frequency_penalty=0 n=1 stop_count=0
[2026-04-20 15:43:42.308044] [tt-media-server] [info] [SessionManager] registerPrefixHash: registered slotId=123 under hash=11620829393098143408
[2026-04-20 15:43:42.308055] [tt-media-server] [info] [LLMController] New session: slotId=123, registered under hash=11620829393098143408

USER1 Comes Back
[2026-04-20 15:43:46.797360] [tt-media-server] [info] [LLMController] /v1/chat/completions task_id=5 model=default stream=1 messages=3 last_msg=[user: "How are you?"] max_tokens=20 temperature=none top_p=none top_k=none min_p=none presence_penalty=0 frequency_penalty=0 n=1 stop_count=0
[2026-04-20 15:43:46.797383] [tt-media-server] [info] [SessionManager] tryAcquireByPrefixHash: acquired slotId=123 for hash=11620829393098143408
[2026-04-20 15:43:46.797386] [tt-media-server] [info] [LLMController] Prefix cache HIT: hash=11620829393098143408, slotId=123, sending delta
[2026-04-20 15:43:46.797389] [tt-media-server] [info] [SessionManager] registerPrefixHash: registered slotId=123 under hash=5374002104659326034
```

## TODO:
- Add `Cache Hits` metric to Grafana - vLLM style [ref](https://docs.vllm.ai/en/stable/design/metrics/#v1-metrics)
- Add Unit Tests
- Remove `sessionId` and `slotId` from request
- Investigate how to handle different roles like "tools" and "developer" [ref](https://developers.openai.com/api/reference/resources/chat#(resource)%20chat.completions%20%3E%20(model)%20chat_completion_message_param%20%3E%20(schema))


tt-media-server/cpp_server/include/utils/conversation_hasher.hpp
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @param messages Input chat messages
+ * @return Messages with system messages filtered out
+ */
+std::vector<domain::ChatMessage> stripSystemMessages(
+    const std::vector<domain::ChatMessage>& messages);
+
+/**
+ * Return the prefix used to LOOK UP a continuing session: non-system
+ * messages with the trailing [assistant, user] pair removed.
```
lmuravljovTT: with [tool output ](https://github.com/tenstorrent/tt-inference-server/pull/3021), we will sometimes have the last two trailing messages be [assistant, tool] roles, might be good to handle that right away?

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -55,12 +56,67 @@ LLMController::LLMController() {
...
 void LLMController::resolveSession(
-    std::shared_ptr<domain::LLMRequest> req, trantor::EventLoop* loop,
+    std::shared_ptr<domain::LLMRequest> req,
+    const std::vector<domain::ChatMessage>& messages, trantor::EventLoop* loop,
     std::function<void(SessionInfo)> onResolved,
     std::function<void(const SessionError&)> onError) const {
   SessionInfo info;
 
-  if (req->sessionId.has_value() && sessionManager) {
+  // Compute routing information from messages
```
idjuricTT: Can we keep both?
I mean:

1. If session id exists, use it
2. If not, try prefix caching

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -55,12 +56,67 @@ LLMController::LLMController() {
...
+      "registrationHash={}",
+      routingInfo.hasPriorTurn,
+      routingInfo.lookupHash.has_value()
+          ? std::to_string(*routingInfo.lookupHash)
+          : "none",
+      routingInfo.registrationHash);
+
+  // New hash-based routing: try to find a session by prefix hash
+  if (routingInfo.hasPriorTurn && routingInfo.lookupHash.has_value() &&
+      sessionManager) {
```
idjuricTT: Nitpick if no session manager we don't need routingInfo at all on controller level

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -55,12 +56,67 @@ LLMController::LLMController() {
...
+    try {
+      auto acquired =
```
idjuricTT: Maybe a stupid question, but why acquired is not a type of session?

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -55,12 +56,67 @@ LLMController::LLMController() {
...
+    } catch (const services::SessionInFlightException& e) {
+      // All sessions with this hash are busy
+      TT_LOG_WARN("[LLMController] All sessions busy for hash={}: {}",
+                  *routingInfo.lookupHash, e.what());
+      onError({SessionErrorType::RATE_LIMIT, e.what()});
+      return;
+    }
+  }
+
+  // Legacy path: explicit sessionId provided by client
```
idjuricTT: OK I see it now, I would keep this one first since it's faster and more accurate

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -496,6 +587,142 @@ void SessionManager::retryFailedDeallocs() {
...
+  bool hashExists =
+      sessions.modify(prefixHash, [&](std::list<domain::Session>& sessionList) {
+        found = true;
+
+        // Log all sessions in this bucket for debugging
+        TT_LOG_INFO(
+            "[SessionManager] tryAcquireByPrefixHash: found {} session(s) "
+            "under hash={}",
+            sessionList.size(), prefixHash);
+        for (const auto& s : sessionList) {
```
idjuricTT: 😄

tt-media-server/cpp_server/include/utils/conversation_hasher.hpp
ljovanovicTT: Great catch, thank you for noticing! On it ✅

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -0,0 +1,117 @@
...
+  // Should not happen if preconditions are met, but return empty as fallback
+  return "";
+}
+
+PrefixCachingInfo computePrefixCachingInfo(
+    const std::vector<domain::ChatMessage>& messages) {
+  PrefixCachingInfo info;
+
+  // Strip system messages
+  auto nonSystemMessages = stripSystemMessages(messages);
```
dmadicTT: Working with `guidellm` I learned that prefix-caching very connected to system prompts and that Claude Code, Cursor and other coding assistants heavily rely on it.
I believe we should keep them.

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -0,0 +1,117 @@
...
+uint64_t hashConversationPrefix(
+    const std::vector<domain::ChatMessage>& prefix) {
+  // Empty prefix should have a deterministic hash
+  if (prefix.empty()) {
+    return 0;
+  }
+
+  // Render the prefix with addGenerationPrompt=false
+  const auto& tokenizer = tokenizers::activeTokenizer();
+  std::string rendered = tokenizer.applyChatTemplate(prefix, false);
```
dmadicTT, line 74: We might have to think how to call `applyChatTemplate` less often

Image

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  // Compute stable 64-bit hash
+  std::hash<std::string> hasher;
+  return hasher(rendered);
```
dmadicTT: I just learned about a term "stable hash"
> A stable hash = a hash function whose output for a given input is fixed forever, across:
different processes / restarts (no random seed),
different machines / OSes / compilers / standard libraries,
different versions of the library itself.

> std::hash<std::string> does not guarantee any of those. Per the C++ standard it only has to be consistent within one program execution; implementations are free to (and do) salt or change algorithms.

The only reason I see we want to use it is once we persist sessions to survive server restarts with https://github.com/tenstorrent/tt-inference-server/issues/3040 
If it's easy to do, I would still do it now.

Suggestion is:
xxHash3 (XXH3_64bits) — fast, stable across versions, widely used.

tt-media-server/cpp_server/include/api/llm_controller.hpp
```diff
@@ -81,6 +83,7 @@ class LLMController : public drogon::HttpController<LLMController> {
    */
   void handleStreaming(
       std::shared_ptr<domain::LLMRequest> reqPtr,
```
knovokmetTT, line 83: what about embedding the messages array as an optional thing in LLMRequest?

tt-media-server/cpp_server/include/domain/session.hpp
```diff
@@ -89,12 +100,14 @@ class Session {
   Json::Value toJson() const {
     Json::Value json;
     json["session_id"] = session_id_;
+    json["hash"] = static_cast<Json::UInt64>(hash_);
```
knovokmetTT: should the client see the hash? is that a part of the API spec?

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -0,0 +1,117 @@
...
+    return std::nullopt;
+  }
+
+  // Remove the trailing [assistant, user] pair
+  std::vector<domain::ChatMessage> priorPrefix;
+  priorPrefix.reserve(nonSystemMessages.size() - 2);
+
+  for (size_t i = 0; i < nonSystemMessages.size() - 2; ++i) {
+    priorPrefix.push_back(nonSystemMessages[i]);
+  }
```
knovokmetTT, line 55: maybe
```
std::vector<domain::ChatMessage> priorPrefix(
    nonSystemMessages.begin(), 
    nonSystemMessages.begin() + (nonSystemMessages.size() - 2)
);
```

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
knovokmetTT: issue: tool calls for deepseek are embedded into system prompt

dmadicTT: Can you share examples what does this mean?
How do mulitple-turns of single sessions look like?

knovokmetTT: @lmuravljovTT

lmuravljovTT: You can take a look at the standard [chatTemplate for deepseek ](https://github.com/vllm-project/vllm/blob/main/examples/tool_chat_template_deepseekv31.jinja)

Specifically this section 
  {% set ns.system_prompt = ns.system_prompt + '\n\n' + tool_ns.text %}

If tool calls are sent through the request, they will be added to the system prompt. If the user sends different tool definitions in different requests ( or leaves them out  ), the system message will change resulting in a different hash

tt-media-server/cpp_server/src/api/llm_controller.cpp
knovokmetTT: we will never receive session id via API in prod?

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -0,0 +1,117 @@
...
+  // Render the prefix with addGenerationPrompt=false
+  const auto& tokenizer = tokenizers::activeTokenizer();
+  std::string rendered = tokenizer.applyChatTemplate(prefix, false);
+
+  // Compute stable 64-bit hash
+  std::hash<std::string> hasher;
+  return hasher(rendered);
+}
+
+std::string renderLastUserTurn(
```
dmadicTT, line 80: C++ has `find_if` API I recently found out about , so this piece would be:

```
auto it = std::find_if(messages.rbegin(), messages.rend(),
                       [](const auto& m) { return m.role == "user"; });
if (it == messages.rend()) {
  return "";
}
return tokenizers::activeTokenizer().applyChatTemplate({*it}, true);
```

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
lmuravljovTT: For example: 

Turn 1: No Tools (Baseline)                                                                                       
```
                                                                                                                    
  Request:                                                                                                          
  {                                                                                                                 
    "model": "deepseek-v3",                                                                                         
    "messages": [                                           
      {"role": "user", "content": "What's the capital of France?"}
    ],                                                            
    "session_id": "session_123"                                                                                     
  }
  
  ```
                                                                                                                    
  System Prompt Generated:         
  
  ```                         
  <|begin▁of▁sentence|><|System|>                                                                                   
  You are a helpful AI assistant.                           
  <|end▁of▁sentence|>            


```


Turn 2: Add Tools with tool_choice: "auto"                

```
  Request:
  {
    "model": "deepseek-v3",
    "messages": [                                                                                                   
      {"role": "user", "content": "What's the capital of France?"},
      {"role": "assistant", "content": "The capital of France is Paris."},                                          
      {"role": "user", "content": "What's the weather there?"}            
    ],                                                                                                              
    "tools": [
      {                                                                                                             
        "type": "function",                                 
        "function": {                                                                                               
          "name": "get_weather",
          "description": "Get current weather for a location",                                                      
          "parameters": {                                     
            "type": "object",
            "properties": {  
              "location": {"type": "string"},                                                                       
              "units": {"type": "string", "enum": ["celsius", "fahrenheit"]}
            },                                                                                                      
            "required": ["location"]                        
          }
        }                                                                                                           
      }
    ],                                                                                                              
    "tool_choice": "auto",                                  
    "session_id": "session_123"
  }

```
  System Prompt Generated:
  
  ```
  <|begin▁of▁sentence|><|System|>
  You are a helpful AI assistant with tool calling capabilities.
                                                                
  # Tools
         
  You have access to the following tools:
                                         
  ## get_weather
  Get current weather for a location
                                    
  Parameters:
  - location (string, required): The location to get weather for
  - units (string, optional): Temperature units. Options: celsius, fahrenheit
                                                                                                                    
  ## Tool Usage Instructions
                                                                                                                    
  When you need to use a tool, respond with JSON in this format:
  ```json                                                                                                           
  {                                                         
    "name": "function_name",                                                                                        
    "arguments": {                                          
      "param1": "value1"
    }                   
  }  
   
  Only use tools when necessary. If you can answer without a tool, respond naturally.
  <|end▁of▁sentence|>                                                                                               
                                                                                          
```

File: tt-media-server/cpp_server/src/api/llm_controller.cpp [hunk 14]
```diff
@@ -55,12 +56,67 @@ LLMController::LLMController() {
...
+      if (acquired.has_value()) {
+        // HIT: found matching session, send delta only
+        TT_LOG_INFO(
+            "[LLMController] Prefix cache HIT: hash={}, sessionId={}, "
+            "slotId={}, sending delta",
+            *routingInfo.lookupHash, acquired->sessionId, acquired->slotId);
+        req->slotId = acquired->slotId;
+        req->sessionId = acquired->sessionId;  // Use stable UUID from session
+        req->continuation = true;
+        req->prompt = routingInfo.deltaPrompt;  // Switch to delta prompt
```
dmadicTT: Alternative is sending the full prompt and `position_id` marking the size of the prefix
We should think pros/cons

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
dmadicTT: Thanks @lmuravljovTT !

We are filtering out "messages.system" JSON atributtes.
As I see tool calling does not really add any messages with `role: "system"`, so tool calling will work even if we stop filtering.

Adding tool calling to system prompt, and putting system prompt at the beginning seems to be the behavior of `applyChatTemplate` of the DeepSeek tokenizer.

tt-media-server/cpp_server/src/api/llm_controller.cpp
ljovanovicTT: It is first in the flow now, where each step fallbacks to the next one:

1. Check if `sessionId` is passed
2. Check if we can do prefix caching
3. Create session

tt-media-server/cpp_server/src/api/llm_controller.cpp
ljovanovicTT: Done!

tt-media-server/cpp_server/src/api/llm_controller.cpp
ljovanovicTT: Done!

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
ljovanovicTT: Great catch! I have created a follow-up Issue for this: https://github.com/tenstorrent/tt-inference-server/issues/3108

tt-media-server/cpp_server/include/api/llm_controller.hpp
ljovanovicTT: Great idea, done!

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
ljovanovicTT: Something like this is included with the latest changes!

tt-media-server/cpp_server/src/api/llm_controller.cpp
ljovanovicTT: Agreed, here is a follow-up Issue in which scope we can investigate this: https://github.com/tenstorrent/tt-inference-server/issues/3109

tt-media-server/cpp_server/include/domain/session.hpp
ljovanovicTT: Nice catch! Its really not needed for client, done with latest commit

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
ljovanovicTT: Agreed - we should definitely investigate what we want to perform hashing on. Since this is not the final version of this flow, I would keep it as is, and properly address it within https://github.com/tenstorrent/tt-inference-server/issues/3102, since this is convenient for the time being

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
ljovanovicTT: With the most recent changes, we filter out the changeable `tools` and `function` part of the `messages` field - this way system is kept as is

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -157,16 +162,22 @@ uint32_t SessionManager::acquireInFlight(const std::string& sessionId,
...
-                    ms.session.updateActivityTime();
-                    ms.session.markInFlight();
-                    ms.cancelFn = std::move(cancelFn);
-                    result = ms.session.getSlotId();
-                  });
+  bool found = sessions.modify(
+      sessionId, [&result, &wasInFlight,
+                  cancelFn = std::move(cancelFn)](ManagedSession& ms) mutable {
+        wasInFlight = !ms.session.isIdle();
+        if (wasInFlight) return;
```
idjuricTT, line 169: Not your change but I wonder how did we get this case 😄

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  // Check if second-to-last is assistant
+  if (turns[turns.size() - 2].role != "assistant") {
+    return std::nullopt;
+  }
+
+  // Remove the trailing [assistant, user] pair
+  std::vector<domain::ChatMessage> priorPrefix;
+  priorPrefix.reserve(turns.size() - 2);
```
idjuricTT, line 51: Nitpick: let's try to do a regex, I'll create a follow up task, it's going to be way fast

idjuricTT (approved):

dmadicTT (approved):

PR #3042, opened 2026-04-20
Replace boolean flags with SessionState state machine

### LINKED ISSUE #3036: Session manager unit tests
Add UTs to session manager

### LINKED ISSUE #3043: Session uses raw booleans instead of a state machine
Session tracks lifecycle with two independent booleans (in_flight_, pending_close_), allowing invalid combinations and making transitions implicit. Replace with a SessionState enum and guarded transition methods to enforce valid state changes and improve debuggability.

ztorlakTT:
Replaces the `in_flight_` / `pending_close_` boolean pair in Session with a `SessionState` enum and guarded transition methods. Adds an abort-callback mechanism so streaming requests are cancelled immediately when closeSession is called while a request is in flight.

State | Meaning
-- | --
IDLE | No active request; session is free to accept a new one
IN_FLIGHT | A request is actively being processed
CLOSE_REQUESTED | closeSession was called while in-flight; waiting for the current request to finish before deallocating the slot
CLOSING | Request has finished on a close-requested session; slot deallocation is pending

SessionManager now stores a per-session abort callback. LLMController registers one after resolving the session for a streaming request; it captures the taskId and calls LLMService::abortRequest. When closeSession is called on an in-flight session the callback is taken and invoked immediately, giving the client a fast response. Non-streaming requests continue to use the deferred-close path (no callback registered).

### Other changes
- setSessionInFlight(id, false) → releaseInFlight(id) — cleaner caller API
- Removed CloseSessionResult::IN_FLIGHT; closeSession now always returns SUCCESS for existing sessions
- Extracted finalizeSessionClose private method to eliminate duplicated dealloc/metric/log logic
- Warning logs on unexpected state transitions (returns false instead of asserting)
- concurrent_map_test — ConcurrentMap::takeIf unit tests + TOCTOU race regression
- session_test — Session state machine transitions + SessionManager abort callback behaviour

### DIFF
--- a/tt-media-server/cpp_server/src/domain/session.cpp
+++ b/tt-media-server/cpp_server/src/domain/session.cpp
@@ -23,7 +23,7 @@ bool Session::markInFlight() {
 
 bool Session::clearInFlight() {
   if (state_ != SessionState::IN_FLIGHT &&
-      state_ != SessionState::PENDING_CLOSE) {
+      state_ != SessionState::CLOSE_REQUESTED) {
     return false;
   }
   state_ = (state_ == SessionState::IN_FLIGHT) ? SessionState::IDLE
@@ -31,9 +31,9 @@ bool Session::clearInFlight() {
   return true;
 }
 
-bool Session::markPendingClose() {
+bool Session::markCloseRequested() {
   if (state_ != SessionState::IN_FLIGHT) return false;
-  state_ = SessionState::PENDING_CLOSE;
+  state_ = SessionState::CLOSE_REQUESTED;
   return true;
 }
 
--- a/tt-media-server/cpp_server/src/domain/session.cpp
+++ b/tt-media-server/cpp_server/src/domain/session.cpp
@@ -5,12 +5,36 @@
 
 #include <iomanip>
 #include <mutex>
+#include <random>
+#include <sstream>
 
 namespace tt::domain {
 
 Session::Session(uint32_t slotId)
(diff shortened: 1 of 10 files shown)


tt-media-server/cpp_server/src/domain/session.cpp
```diff
@@ -13,6 +15,28 @@ Session::Session(uint32_t slotId)
...
 }
 
+bool Session::markInFlight() {
+  if (state_ != SessionState::IDLE) return false;
+  state_ = SessionState::IN_FLIGHT;
+  return true;
+}
+
+bool Session::clearInFlight() {
+  if (state_ != SessionState::IN_FLIGHT &&
```
dmadicTT, line 25: If we ever end up in this situation is it because of developer bug?
If yes, we should know about it, I would log a warning.
Similar for above case.

dmadicTT (approved):

tt-media-server/cpp_server/src/domain/session.cpp
ztorlakTT: We have logging but it's in SessionManager. I'd leave it there because Session is a domain object and shouldn't carry infrastructure dependencies

dmadicTT: But in session manager we never care about `markInFlight` return value, so we will not log anything

knovokmetTT (approved):

PR #3056, opened 2026-04-20
cpp server - prefill: both sides can open SHM

### LINKED ISSUE #2997: Make shared memory opening from both sides
In prefill runner shared memory should be able to be open from both sides - inference and model so it's not important which one starts first

knovokmetTT:


tt-media-server/cpp_server/include/ipc/slot_ring_buffer.hpp
```diff
@@ -24,10 +24,11 @@ constexpr int PREFILL_MAX_TOKEN_IDS =
     131072;  // matches Config::MAX_INPUT_TOKENS (128k)
 constexpr int DECODE_MAX_TOKEN_IDS = 1;
 
-enum SlotState { FREE, TAKEN };
+enum SlotState { EMPTY, FILLED };
```
knovokmetTT, line 27: makes more sense in my head

tt-media-server/cpp_server/include/ipc/slot_ring_buffer.hpp
```diff
@@ -78,11 +79,15 @@ class SlotRingBuffer {
   SlotRingBuffer& operator=(const SlotRingBuffer&) = delete;
 
   void open() {
-    int fd = shm_open(name.c_str(), O_RDWR, 0);
+    int fd = shm_open(name.c_str(), O_CREAT | O_RDWR, 0666);
     if (fd < 0) {
       throw std::runtime_error(
           "SlotRingBuffer: unable to open shared memory: " + name);
     }
+    if (ftruncate(fd, Msg::K_TOTAL_SIZE) < 0) {
```
knovokmetTT, line 87: no op if file already exists and same size, so its safe

tt-media-server/cpp_server/include/ipc/slot_ring_buffer.hpp
```diff
@@ -130,53 +136,58 @@ class SlotRingBuffer {
                                std::to_string(MaxTokenIds));
     }
 
-    auto& msg = acquireMsg();
+    auto& slot = acquireWriterSlot();
```
knovokmetTT, line 139: also makes more sense to call this a slot, not a msg, just aesthetics

tt-media-server/cpp_server/src/runners/shared_memory.py
```diff
@@ -79,19 +83,10 @@ def open(self) -> None:
                 name=self._name, create=True, size=self._total_size
             )
         except FileExistsError:
-            temp_shm = _shm.SharedMemory(name=self._name, create=False)
-            temp_shm.unlink()  # delete the existing shared memory block
```
knovokmetTT, line 83: do not unlink and zero the mem if it exists anymore, just attach, because cpp can create it as well, they race to it

dmadicTT (approved):

idjuricTT (approved):

PR #3058, opened 2026-04-20
Sync video setup

idjuricTT:


idjuricTT:
@ztorlakTT could you please check does this still hit the video cleanup policy?

ljovanovicTT (approved):

tt-media-server/open_ai_api/video.py
```diff
@@ -38,9 +40,59 @@ async def submit_generate_video_request(
...
+                serve_path,
+                media_type="video/mp4",
+                filename=f"video_{request._task_id}.mp4",
+                headers={
+                    "Content-Disposition": f"attachment; filename=video_{request._task_id}.mp4"
+                },
+            )
+        
+        # Async mode: create job and return job metadata
         job_data = await service.create_job(JobTypes.VIDEO, request)
```
ayerofieiev-tt, line 94: wouldn't it be better to do

```py
        job_data = await service.create_job(JobTypes.VIDEO, request)
        if not async: # do extra here
            result = job_data.result()
            ....
```

tt-media-server/open_ai_api/video.py
```diff
@@ -38,9 +40,59 @@ async def submit_generate_video_request(
         service.scheduler.check_is_model_ready()
     except Exception:
         raise HTTPException(status_code=405, detail="Model is not ready")
+    
     try:
+        # Synchronous mode: process and return video directly
+        if not settings.use_async_video:
```
ayerofieiev-tt, line 46: when requests are sync, what happens with the server accepting next requests?
its basically clients responsibility to properly wait for the response?

idjuricTT: We have a multiprocessing queue so feel free to throw whatever you have on the service 😄 
Longer queue = longer waiting time since there’s only 1 video at the time on SP

tt-media-server/open_ai_api/video.py
idjuricTT: I don’t want to have job service engaged at all on sync creation. It’s just a noise in that case

tt-media-server/open_ai_api/video.py
ayerofieiev-tt: thank you

tt-media-server/open_ai_api/video.py
ayerofieiev-tt: just thinking it would be great to have a single flow, thats being forced to wait vs having 2 different flows
totally your call.

idjuricTT: I would keep it configurable on statrtup since probably you want it to be async on T3K where you wait 10ish minutes for the response.

for 2s SP it doesn’t make sense for it to be async.

but I’m happy to discuss further if needed 😄

vpetrovicTT (approved):

PR #3074, opened 2026-04-21
Add Qwen 3 8B Training And Unify LoRA Multichip Training

mmilosevicTT:
This PR adds and updates following things:
- Support for Qwen 3 8B Training
- Unification of LoRA Training (for qwen and llama) to single script
- Rename constant mapping for runners to only include inference runners
- Addition of need for MODEL parameter in training

There are 2 things left for clean-up:
- Gemma 1.1 should be unified into the same runner
- Fix for single hardcoded sharding patterns


tt-media-server/utils/build_catalog.py
```diff
@@ -26,20 +26,36 @@
...
+def _build_models_catalog(model_runner: str, model_weights_path: str = ""):
     try:
         runner_enum = ModelRunners(model_runner)
     except ValueError:
         return []
     models = []
-    for model_name in MODEL_RUNNER_TO_MODEL_NAMES_MAP.get(runner_enum, set()):
+    model_names_set = MODEL_RUNNER_TO_MODEL_NAMES_MAP.get(runner_enum, set())
+    for model_name in model_names_set:
+        model_config = SupportedModels[model_name.name].value
```
mcupacTT: nit: i would rename this to hugging face path or repo id to be more clear

tt-media-server/utils/build_catalog.py
```diff
@@ -26,20 +26,36 @@
...
+def _weights_path_matches(model_weights_path: str, repo_id: str) -> bool:
+    """Check if model_weights_path matches the HF repo ID.
+
+    model_weights_path may be the repo ID itself ("Qwen/Qwen3-8B") or a
+    resolved local snapshot path that embeds it
+    ("…/models--Qwen--Qwen3-8B/snapshots/…").
+    """
+    if model_weights_path == repo_id:
+        return True
+    cache_fragment = f"models--{repo_id.replace('/', '--')}"
```
mcupacTT: this is the format that hugging face saves the cached models, right? 

This part will be likely extended with loading adapters from local storage and path we also pass through model_weights_path

But I will tackle it in the corresponding PR about it

ztorlakTT (approved):

tt-media-server/model_services/training_service.py
```diff
@@ -20,9 +19,17 @@ class TrainingService(BaseJobService):
...
+        super().__init__()
+
+    def _resolve_model_name(self) -> str:
+        """Resolve the short model name for the configured runner and model.
+
+        Runners with a single supported model return that name directly.
+        Runners with multiple supported models require the configured
+        `model` to match one of the supported model configs.
+        """
+        self._model_name = ModelNames(self.settings.model).value
```
mcupacTT: where and when is `settings.model` variable set? I can not find it

Apart from that LGTM!

mmilosevicTT: Good catch, I will add a field to settings to actually populate this as well.

ztorlakTT (approved):

Issue #3015, opened 2026-04-17
Add none tool_choice support


lmuravljovTT:
Prevent tool calling when tool_choice is set to "none". The chat template must instruct the model to respond directly without using any tools, even when tools are defined in the request. Validate that parsed responses 
  contain no tool calls and set appropriate finish_reason.
                                                                                                                                                                                                                            
When tools are available but tool_choice: "none" is set, the model may still output tool-like syntax by habit or confusion. Structured output constraints ensure the response conforms to a       
  text-only schema, preventing accidental tool call formatting.                                                                                                                                                             
                                                                                                                                                                                                                            
  Where to look:                                                                                                                                                                                                            
  - applyChatTemplate() in utils/tokenizers/tokenizer.hpp - add template logic for none mode
  - LLMService::postProcess() in src/services/llm_service.cpp - validate no tool_calls parsed                                                                                                                               
  - Add tests in tests/test_tool_output_parsing.cpp

PR #3084, opened 2026-04-21
Add tool_choice "none" support 

lmuravljovTT:
#3015 
### Overview

  Implements support for `tool_choice: "none"` in chat completion requests. When set, this parameter instructs the server to skip tool call parsing from the models response.

  ### What is `tool_choice: "none"`?

  `tool_choice: "none"` is an OpenAI-compatible parameter that controls tool calling behavior:
  - **`"auto"`** (default): Model decides whether to call tools
  - **`"required"`**: Model must call at least one tool
  - **`"none"`**: Model must NOT call any tools (respond with text only)
  - **`{"type": "function", "function": {"name": "..."}}`**: Model must call the specified function

  ### Implementation

  When `tool_choice: "none"` is specified:
  1. The request parsing validates the `tool_choice` parameter
  2. The `tools` list is cleared (`req.tools = std::nullopt`) before being passed to the model
  3. No tools are included in the chat template/prompt
  4. Model generates natural language response without tool call syntax

  This approach ensures the model has no context about tools and cannot accidentally generate tool calls.

  ### Changes

  - **`chat_completion_request.hpp`**: Added logic to clear tools when `tool_choice: "none"`
  - **`test_chat_completion_request.cpp`**: Added unit tests for tool_choice parsing and validation
  
  
  
    References

  -[ vLLM Chat API - Tool Choice](https://docs.vllm.ai/en/latest/serving/openai_compatible_server/#tool-calling-in-chat-completions-api)

  
  
  Example Usage

```
  curl -X POST http://localhost:8000/v1/chat/completions \                                                          
    -d '{                                                                                                           
      "model": "deepseek-v3",                               
      "messages": [{"role": "user", "content": "What'\''s the weather in Paris?"}],                                 
      "tools": [{                                                                                                   
        "type": "function",                                                                                         
        "function": {                                                                                               
          "name": "get_weather",                            
          "parameters": {"type": "object", "properties": {"location": {"type": "string"}}}
        }                                                                                                           
      }],
      "tool_choice": "none"                                                                                         
    }' 
```                                                     

  Response: Natural language text (no tool calls)                                                                   
```
  {
    "choices": [{                                                                                                   
      "message": {                                          
        "role": "assistant",
        "content": "I don't have access to real-time weather data..."                                               
      },                                                             
      "finish_reason": "stop"                                                                                       
    }]                                                      
  }
```


bgoelTT (commented):
Was checking with Gemini to see how the `tools` array is treated when `tool_choice="none"` in vLLM, it yielded that the tools are still added to the input context via chat templating, but tool parsing is skipped over.

Source: https://docs.nvidia.com/nim/vision-language-models/latest/release-notes.html

Image

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -354,6 +354,10 @@ void LLMService::processStreamingRequest(
   StreamCallbackEntry entry{std::move(callback), request.skip_special_tokens};
   streamCallbacks.insert(taskId, std::move(entry));
 
+  if (request.tool_choice_type.has_value()) {
+    toolChoiceMap.insert(taskId, request.tool_choice_type.value());
```
idjuricTT: How is this tracked between multiple requests? Would it case bias of behavior?

idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
lmuravljovTT: Not sure I understand what you mean? toolChoiceMap is a ConcurrentMap, and the tool_choice is mapped to the task_id, and removed from the map once handled. Can there be overlap if done this way?

ztorlakTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -73,6 +73,16 @@ std::vector<tt::worker::WorkerInfo> LLMService::getWorkerInfo() const {
 void LLMService::preProcess(domain::LLMRequest& request) const {
   BaseService::preProcess(request);
 
+  if (request.tool_choice.has_value()) {
+    const auto& type = request.tool_choice->type;
+    if (type != "auto" && type != "none") {
+      throw std::invalid_argument(
```
idjuricTT, line 79: Minor: Please check would this return an error or crash the main process

idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
lmuravljovTT: good catch, it's getting handled now in the llm_controller

PR #3087, opened 2026-04-21
Fix runner selection when MODEL and MODEL_RUNNER env vars are both set

### LINKED ISSUE #3086: MODEL env var ignores MODEL_RUNNER and picks runner non-deterministically
When both MODEL and MODEL_RUNNER are set, the server ignores MODEL_RUNNER and instead selects the runner by iterating MODEL_RUNNER_TO_MODEL_NAMES_MAP and picking the first match. This causes two problems:
- MODEL_RUNNER is silently overridden, making it impossible to select a training runner for a model that also has an inference runner (e.g. Qwen3-8B with vllm_forge as default)
- The iteration order depends on dict/set insertion order, so adding a new runner to the map can change which runner gets selected for existing models

ztorlakTT:
Fixes the runner selection logic when MODEL env var is used for config overrides.

Changes:
- Add MODEL_NAME_TO_RUNNER_MAP as the single source of truth for model→runner mapping, replacing MODEL_RUNNER_TO_MODEL_NAMES_MAP
- MODEL_RUNNER env var now correctly takes precedence over the canonical runner, enabling use of training runners for models that also have inference runners
- Add models_for_runner() helper (O(1) lookup, precomputed at import time) to derive runner→models direction
- Fix non-deterministic set iteration in model_weights_path fallback and TrainingService — multi-model runners now log a warning and require MODEL to be set explicitly

### DIFF
--- a/tt-media-server/config/settings.py
+++ b/tt-media-server/config/settings.py
@@ -8,8 +8,9 @@ from typing import Optional
 
 from config.constants import (
     MODEL_NAME_OVERRIDES,
-    MODEL_RUNNER_TO_MODEL_NAMES_MAP,
+    MODEL_NAME_TO_RUNNER_MAP,
     MODEL_SERVICE_RUNNER_MAP,
+    models_for_runner,
     SDXL_VALID_IMAGE_RESOLUTIONS,
     AudioTasks,
     DeviceTypes,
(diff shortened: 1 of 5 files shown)


tt-media-server/config/settings.py
```diff
@@ -289,46 +280,50 @@ def _calculate_audio_chunk_duration(self):
...
+        config_key = (model_runner_enum, device_type_enum)
+        matching_config = ModelConfigs.get(config_key)
+        logger.info(
+            f"Config lookup: runner={model_runner_enum}, device_type={device_type_enum}, "
+            f"key_exists={config_key in ModelConfigs}, "
+            f"matching_config={matching_config}"
+        )
+
+        if not matching_config:
+            return
```
vpetrovicTT, line 304: Is this early return going to prevent vllm.model initialization? Do we want that?

PR #3090, opened 2026-04-22
Fix Checkpointing in Training Runners

mmilosevicTT:
This PR changes:
- Update default for `save_interval` 100 -> 25
- Fix Gemma Lora Training runner to increase `global_step` after `optimizer_step`


mcupacTT (approved):
LGTM!
nit: Just change in the PR description that save_interval is reduced to 25 instead of 20

idjuricTT (approved):

PR #3096, opened 2026-04-22
Add eval_configs for Llama-3.2-3B and Qwen3-4B based on similar models

vmaksimovicTT:
Placeholder evals config for two new forge models (so that the release workflow does not crash)
https://github.com/tenstorrent/tt-shield/actions/runs/24769606296/job/72479060797
https://github.com/tenstorrent/tt-shield/actions/runs/24778863273/job/72504283805

### DIFF
--- a/evals/eval_config.py
+++ b/evals/eval_config.py
@@ -2082,49 +2082,6 @@ _eval_config_list = [
         ],
     ),
     EvalConfig(
-        hf_model_repo="meta-llama/Llama-3.2-3B",
-        tasks=[
-            EvalTask(
-                task_name="mmlu_pro",
-                num_fewshot=5,
-                score=EvalTaskScore(
-                    published_score=None,
-                    published_score_ref=None,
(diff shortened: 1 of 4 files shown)


evals/eval_config.py
bgoelTT, line 1: We should only be using the `-instruct` tuned version of models for these evals, so for `Llama-3.2-3B-Instruct` this already exists. Therefore the solution is just to update the Forge model spec to have include Llama-3.2-3B-Instruct in the weights list **and** update the `model-ci-config.json` to use the instruct tuned weights too

bgoelTT (changes requested):

evals/eval_config.py
vmaksimovicTT: Should be done. Had do change the schema of the models-ci-config.json to allow for multiple implementations of the same model

AleksKnezevic (approved):

bgoelTT (approved):

PR #3097, opened 2026-04-22
cpp server - make ContiguousMemoryManager allocate from a slot pool

dmadicTT:
Adapts ContiguousMemoryManager to own a pool of N slot IDs instead of returning slot ID 123 for all requests.
This enables testing of upcoming prefix caching.

### DIFF
--- a/tt-media-server/cpp_server/src/services/memory_services/contiguous_memory_manager.cpp
+++ b/tt-media-server/cpp_server/src/services/memory_services/contiguous_memory_manager.cpp
@@ -4,22 +4,84 @@
 #include "services/memory_services/contiguous_memory_manager.hpp"
 
 #include "domain/manage_memory.hpp"
+#include "utils/logger.hpp"
 
 namespace tt::services {
 
 using tt::domain::ManageMemoryResult;
 using tt::domain::ManageMemoryStatus;
 using tt::domain::ManageMemoryTask;
+using tt::domain::MemoryManagementAction;
+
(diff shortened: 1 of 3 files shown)


tt-media-server/cpp_server/src/services/memory_services/contiguous_memory_manager.cpp
```diff
@@ -4,22 +4,84 @@
...
+        ManageMemoryResult result{};
+        result.taskId = task.taskId;
+        result.status = ManageMemoryStatus::FAILURE;
+        resultQueue->push(result);
+        return;
+      }
+
+      uint32_t slotId = *freeSlots.begin();
+      freeSlots.erase(freeSlots.begin());
+      allocatedSlots.insert(slotId);
```
ljovanovicTT, line 42: Are we okay with this not being atomic?

ljovanovicTT (approved):

tt-media-server/cpp_server/src/services/memory_services/contiguous_memory_manager.cpp
dmadicTT: Just checked. I think we are safe since we access it from single thread.
```
void SpPipelineRunnerDemo::memoryLoop() {
  while (!stopped.load(std::memory_order_relaxed)) {
    auto task = memoryManager->getRequest();
    if (task) {
      memoryManager->handleRequest(*task);
    } else {
      std::this_thread::sleep_for(std::chrono::milliseconds(1));
    }
  }
}
```

idjuricTT (approved):

PR #3106, opened 2026-04-22
Add single user multi-turn test to CI

dmadicTT:
Enables prefix caching testing
- Runs single user conversation with 29 turns up to 128k context length
- Verifies there is 29 active sessions reported by /metrics
- Once we merge prefix caching this will be changed to 1 active session


.github/workflows/test-gate.yml
```diff
@@ -965,6 +965,140 @@ jobs:
...
+        run: cpp_server/install_dependencies.sh
+
+      - name: Download C++ build
+        uses: actions/download-artifact@v4
+        with:
+          name: cpp-server-build
+          path: tt-media-server/cpp_server
+
+      - name: Make binaries executable
+        run: chmod +x cpp_server/build/tt_media_server_cpp
```
ljovanovicTT: Is this needed?

.github/workflows/test-gate.yml
```diff
@@ -965,6 +965,140 @@ jobs:
           printf '%s\n' "$BENCH_ERRORS"
           exit 1
 
+  cpp-server-single-user-multi-turn:
```
ljovanovicTT: Can we reuse `cpp-server-benchmarks` and have `Multi-Turn Smoke Test` as a step?
That way we can install dependencies, fetch build files and start cpp server in mock mode only once

ljovanovicTT (approved):

ztorlakTT (approved):

PR #3111, opened 2026-04-22
cpp server - blaze: handle PM return vals

knovokmetTT:


tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
```diff
@@ -185,6 +188,12 @@ inline void BlazeRunner::handleMemoryRequest(
 }
 
 inline void BlazeRunner::handleResponse(const pm::PMResponse& response) {
+  if (response.error_code != 0) {
```
dmadicTT: Just to share that I see that `error_code=1` means no free slots
I did not find any other error codes

No changes needed in the code

dmadicTT (approved):

PR #3114, opened 2026-04-22
fix: use float timestamps for sub-second generation time precision

anatarajan-tt:
## Summary

`created_at` and `completed_at` were stored as `int`, using `int(time.time())` which truncates to whole seconds. This caused generation times to display as `2.0s` instead of the actual `2.3s`.

## Changes

**`tt-media-server/utils/job_manager.py`**
- `created_at: int` → `float`
- `completed_at: Optional[int]` → `Optional[float]`
- `int(time.time())` → `time.time()` in `__post_init__`, `mark_completed`, `mark_cancelled`, `mark_failed`

**`tt-media-server/utils/job_database.py`**
- `created_at INTEGER` → `REAL` in SQLite schema
- `completed_at INTEGER` → `REAL` in SQLite schema
- Type annotations `int` → `float` for `created_at`/`completed_at` parameters

## Effect

`completed_at` is now returned as a float (e.g. `1776881712.648`) instead of an integer (`1776881712`). The `flexTime` parser in consumers that handle both Unix integers and floats will preserve the sub-second precision.

### DIFF
--- a/tt-media-server/utils/job_database.py
+++ b/tt-media-server/utils/job_database.py
@@ -50,8 +50,8 @@ class JobDatabase:
                     model TEXT NOT NULL,
                     request_parameters TEXT,
                     status TEXT NOT NULL,
-                    created_at INTEGER NOT NULL,
-                    completed_at INTEGER,
+                    created_at REAL NOT NULL,
+                    completed_at REAL,
                     error_message TEXT,
                     result_path TEXT,
                     org_id TEXT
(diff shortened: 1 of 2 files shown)


idjuricTT (commented):
@mmilosevicTT please check would your team benefit from more precise decimal points

mmilosevicTT:
> @mmilosevicTT please check would your team benefit from more precise decimal points

@idjuricTT I would say for training the stamps for `created_at` and `completed_at` will not be that close, training is expected to run for minutes or more. Only thing that we might benefit from is `started_at` - to represent when the job actually started.

PR #3125, opened 2026-04-23
feat: use model runner passed as env

fivanovicTT:


tt-media-server/config/settings.py
```diff
@@ -289,12 +289,17 @@ def _calculate_audio_chunk_duration(self):
...
+        model_runner_enum = ModelRunners(explicit_runner) if explicit_runner else None
+        logger.info(
+            f"Explicit MODEL_RUNNER={explicit_runner!r} for MODEL={model_to_run!r}"
+        )
+
         for runner, model_names in MODEL_RUNNER_TO_MODEL_NAMES_MAP.items():
             if model_name_enum in model_names:
-                model_runner_enum = runner
-                break
+                if not model_runner_enum:
```
mmilosevicTT, line 306: We can move this if outside of the loop, to not bother with traversing the map.

tt-media-server/config/settings.py
```diff
@@ -289,12 +289,17 @@ def _calculate_audio_chunk_duration(self):
...
 
-        # Find the appropriate model runner for this model name
-        model_runner_enum = None
+        explicit_runner = os.getenv("MODEL_RUNNER")
+        model_runner_enum = ModelRunners(explicit_runner) if explicit_runner else None
+        logger.info(
+            f"Explicit MODEL_RUNNER={explicit_runner!r} for MODEL={model_to_run!r}"
+        )
+
         for runner, model_names in MODEL_RUNNER_TO_MODEL_NAMES_MAP.items():
```
mmilosevicTT, line 304: I am still worried adding same model name to different runners in the map will be a breaking change without warning if MODEL_RUNNER is not set. This is because we will default to the first one extracted, so if someone adds an entry at the start, we have a breaking change.

mmilosevicTT (approved):
LGTM from training perspective, but default case (no MODEL_RUNNER set) still remains faulty.

mmilosevicTT (approved):
LGTM, for training purposes

tt-media-server/config/settings.py
```diff
@@ -289,12 +289,23 @@ def _calculate_audio_chunk_duration(self):
...
+                f"Explicit MODEL_RUNNER={explicit_runner!r} for MODEL={model_to_run!r}"
+            )
+
         for runner, model_names in MODEL_RUNNER_TO_MODEL_NAMES_MAP.items():
             if model_name_enum in model_names:
-                model_runner_enum = runner
-                break
+                if not model_runner_enum:
+                    model_runner_enum = runner
+                    break
```
mmilosevicTT, line 308: nit: Could you move this into the if on line 294? As we do not need to run this when we have model_runner_enum set.

mjeremicTT (approved):

vpetrovicTT (approved):

vmaksimovicTT (approved):

tt-media-server/config/settings.py
```diff
@@ -289,12 +289,23 @@ def _calculate_audio_chunk_duration(self):
     def _set_config_overrides(self, model_to_run: str, device: str):
         model_name_enum = ModelNames(model_to_run)
 
-        # Find the appropriate model runner for this model name
-        model_runner_enum = None
+        explicit_runner = os.getenv("MODEL_RUNNER")
```
idjuricTT, line 292: We already support this if MODEL_RUNNER is sent but not MODEL

acvejicTT (approved):
Lgtm

idjuricTT (approved):

PR #3146, opened 2026-04-23
cpp server fix perm issue boost queue exabox

knovokmetTT:


idjuricTT (approved):

tt-media-server/cpp_server/include/ipc/boost_ipc_queue.hpp
dmadicTT, line 1: As you have learned this stuff the hard way, is it possible to write automated tests for this kind of issues?

PR #3156, opened 2026-04-24
Validate docekr images exist in the GHCR

vmaksimovicTT:
Script that validates all of the images in the `release_model_spec.json` actually exist. Used as the part of the release process


bgoelTT (approved):
Just a small note that it appears the same file is duplicated

Issue #621, opened 2025-09-04
YOLOv4 Object Detection Benchmarks Implementation


arobergeTT:
## Summary
Extend the performance benchmarking system to support YOLOv4 object detection models alongside existing LLM benchmarks using the --workflow benchmarks command.

## Current State
YOLOv4 model runner: `/tt-metal-sdxl/tt_model_runners/yolov4_runner.py`
Object detection inference: Batch processing, timeout handling
Missing: Integration with `--workflow benchmarks`
Missing: Performance metrics collection (FPS, latency, throughput)

#### Performance Metrics
* FPS (Frames Per Second) for sustained inference
* Latency metrics (TTFI - Time to First Inference, per-frame latency)
* Throughput (images/second) under various batch sizes
* Memory utilization and device efficiency

#### Benchmark Configuration
* Batch size scaling (1, 4, 8, 16 images)
* Concurrent request handling
* Different confidence thresholds impact on performance

## Framework Extension
* Add BenchmarkTaskCNN extension to `benchmarking/benchmark_config.py`
* Create YOLOv4-specific benchmark runner
* Add performance targets to `benchmark_targets/model_performance_reference.json`
* Implement object detection performance profiling

#### Core Infrastructure
- [ ] Extend BenchmarkTaskCNN for YOLOv4-specific metrics
- [ ] Create YOLOv4 benchmark runner compatible with run_benchmarks.py
- [ ] Add YOLOv4 model specs with performance targets

#### Performance Profiling
- [ ] FPS measurement under sustained load
- [ ] Latency profiling (preprocessing, inference, post-processing)
- [ ] Memory usage tracking during inference
- [ ] Device utilization metrics collection

#### Benchmark Configuration
- [ ] Define performance targets in model_performance_reference.json
- [ ] Multi-resolution benchmark scenarios
- [ ] Batch size scaling analysis
- [ ] Integration with existing reporting system

Expected Usage
```
python3 run.py --model yolov4 --device n150 --workflow benchmarks
```

## Success Criteria
- [ ] YOLOv4 benchmarks run using standard workflow command
- [ ] Performance metrics collection and reporting
- [ ] Integration with existing benchmark infrastructure
- [ ] Automated performance regression detection

PR #3188, opened 2026-04-27
model_spec: set gpt-oss-120b trace_region_size to 128MB

sraizada-tt:
## Why

`gpt-oss-120b` does not set `trace_region_size` and falls back to vLLM's default 50 MB, which is too small for prefill traces (~85 MB observed at SL=131072 in tt-shield runs). Without this, the server crashes during trace capture on long-context cells:

```
Creating trace buffers of size 84877312B on MeshDevice 0,
but only 50000000B is allocated for trace region.
```

The benchmark JSONs come back with `output_lens=[0,0,...]`, and the downstream report-gen step crashes with `ZeroDivisionError: float division by zero` while computing throughput. (Seen in tt-shield run [#621](https://github.com/tenstorrent/tt-shield/actions/runs/24960859686/job/73089333672) and [later](https://github.com/tenstorrent/tt-shield/actions/runs/24983710463/job/73157371874).)

Other 120B-class models in this file already set explicit `trace_region_size` for the same reason — e.g. Llama-3.1-70B Galaxy uses 184 MB.

## What

Set `trace_region_size = 134217728` (128 MB) in the GALAXY and T3K device specs for `openai/gpt-oss-120b` in `workflows/model_spec.py` (the source of truth). `release_model_spec.json` will be regenerated by the release tooling.

Per [@bgoelTT review](https://github.com/tenstorrent/tt-inference-server/pull/3188#pullrequestreview-4184569088): change goes in `workflows/model_spec.py`, not `release_model_spec.json`.

128 MB picked over the previous 100 MB because it was empirically validated against the full benchmarks workflow on Galaxy 4x8 — all 19 cells passed including the previously-failing `isl=65536 conc=1` (was DRAM OOM in attention all-reduce) and `isl=16384 conc=31 n=62` (was hung at near-OOM KV cache state). Headroom over Ben's observed 85 MB high-water mark.

## Test plan
- [x] Full `--workflow benchmarks` on gpt-oss-120b galaxy: all 19 cells PASS
- [ ] Re-run tt-shield gpt-oss-120b release-galaxy benchmark; verify no trace OOM and `output_lens > 0`

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -1005,6 +1005,9 @@ llm_templates = [
                 max_context=16 * 1024,
                 default_impl=True,
                 tensor_cache_timeout=5400.0,
+                override_tt_config={
+                    "trace_region_size": 134217728,  # 128 MB; vLLM default 50 MB is too small for long prefill
+                },
             ),
             DeviceModelSpec(
                 device=DeviceTypes.GALAXY,
@@ -1023,6 +1026,7 @@ llm_templates = [
                 },
                 override_tt_config={
                     "sample_on_device_mode": "all",
+                    "trace_region_size": 134217728,  # 128 MB; vLLM default 50 MB is too small for long prefill
                 },
             ),
         ],


bgoelTT (changes requested):
Nice find - especially so since I don't see any complaints about trace region size in the docker server log in Models CI. Did you diagnose this by running it locally?

The change is good but it should not be applied to the release_model_spec.json and instead here https://github.com/tenstorrent/tt-inference-server/blob/3ec30649259f2fa6ab0ecb041b25609750f70f55/workflows/model_spec.py#L1025

sraizada-tt:
yeah, needed with prefill optimizations branch we are merging soon https://github.com/tenstorrent/tt-metal/pull/43147

bgoelTT (approved):

bgoelTT:
Running dispatch here too https://github.com/tenstorrent/tt-shield/actions/runs/25068826410

PR #3200, opened 2026-04-27
Qwen3-Embedding-4B/n150 - Modifications for running the server 

### LINKED ISSUE #3190: failing "Models CI" pipeline for "Qwen3-Embedding-4B | n150"
models-ci job for Forge engine and model "**Qwen3-Embedding-4B | n150**" is failing constantly in the last 2 months.
I have faced several errors that were preventing the server to start.
I have done couple of required modifications which led that now server is started and only _spec_tests_ are failing.

Please double check the corrections which were introduced in several commits.

I  will list in here errors and workarounds that I did on my branch so that testings phase for this model could progress.

**Error1:** 
`ERROR - [warmup] async failed after 2.6761 seconds. Error: EngineArgs.__init__() got an unexpected keyword argument 'task'`

Suggestion solution:
```
You need to update tt-forge to a version that bundles vLLM ≥ 0.6.0. 
If no newer tt-forge exists yet with that vLLM version, 
an alternative is to add an explicit vLLM pin below it (pip will upgrade the transitive dependency):                                           
  --extra-index-url https://pypi.eng.aws.tenstorrent.com/                                                                                                                                                                                                                     
  tt-forge==1.0.0                                                                                                                                                                                                                                                            
  vllm>=0.6.0                                                                                                                                                                                                                                                                
                                                         
  Note: Adding vllm>=0.6.0 directly alongside tt-forge==1.0.0 risks a 
conflict if tt-forge pins an incompatible vLLM internally. 
The safest path is to identify a newer tt-forge version on 
pypi.eng.aws.tenstorrent.com that already requires vLLM >= 0.6.0 and update the  
  pin.
```
Commit1:
`vllm>=0.6.0` in `[tt-media-server/tt_model_runners/forge_runners/requirements.txt‎]` 
https://github.com/tenstorrent/tt-inference-server/commit/8ee1e58a73e80bcd7483e309d70ab392269a1251

**Error2:**
```
ImportError: cannot import name 'ImageChunk' from 'mistral_common'
  The server failed to start due to a mistral_common package version mismatch in the Docker container's Python environment.                                                                                                                      
  
 vLLM loads TTQwen3ForCausalLM
    → generator_vllm.py (qwen3_embedding_8b/demo/) line 25                                                                                                                                                                                       
      → models/tt_transformers/tt/generator_vllm.py line 18                                                                                                                                                                                      
        → vllm.model_executor.models.mistral3 (line 38 of mistral3.py)                                                                                                                                                                           
          → vllm.model_executor.models.pixtral (pixtral.py line 13)                                                                                                                                                                              
            → from mistral_common.protocol.instruct.messages import (ImageChunk, TextChunk, ...)                                                                                                                                                 
              ❌ ImportError: ImageChunk does not exist in installed mistral_common  
```

Commit2:

https://github.com/tenstorrent/tt-inference-server/commit/d364c5d26b6113b9d8b7c5138a598e8849606bc7

```
"mistral-common>=1.0.0",
in [‎tt-vllm-plugin/pyproject.toml‎]
```

**Error3:** 

```
  Root Cause                                             
                                                                                                                                                                                                                                                 
  The server started with the wrong model — tt-xla-resnet instead of Qwen3-Embedding-4B.
                                                                                                                                                                                                                                                 
  Evidence                                               

  The liveness check at attempt 18/230 (~14:16:54) finally passes and reveals the actual state:                                                                                                                                                  
  
  Liveness check response: {                                                                                                                                                                                                                     
    'status': 'alive',                                                                                                                                                                                                                           
    'model_ready': True,
    'runner_in_use': 'tt-xla-resnet',   ← WRONG MODEL                                                                                                                                                                                            
    ...                                                  
  }                                                                                                                                                                                                                                              
  
  The forge media inference server started and loaded its default ResNet image classification runner instead of the Qwen3-Embedding-4B model. ResNet has no /v1/embeddings endpoint, so every embedding request returned:                        
                                                         
  POST http://localhost:8000/v1/embeddings → HTTP 404 Not Found                                                                                                                                                                                  
  {'detail': 'Not Found'}                                
                                                                                                                                                                                                                                                 
  Failure Chain
                                                                                                                                                                                                                                                 
  1. Forge server starts (tt-media-inference-server-forge image)
  2. ~3 min of model compilation: UMD IOMMU + Conv2D ops for ResNet                                                                                                                                                                              
  3. Health check passes: model_ready=True, runner_in_use='tt-xla-resnet'                                                                                                                                                                        
  4. Test calls POST /v1/embeddings  →  404 Not Found                                                                                                                                                                                            
  5. Eval reports: "EMBEDDING MediaTaskType.EVALUATION failed: Error code: 404"                                                                                                                                                                  
  6. Job concludes: failure                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                 
  Why ResNet Instead of Qwen3-Embedding-4B?              
                                                                                                                                                                                                                                                 
  The docker run command passed model: Qwen/Qwen3-Embedding-4B in the runtime spec, but the forge media server ignored it and loaded its default runner. This means the Qwen3-Embedding-4B vllm-plugin integration is not registered/wired into  
  the forge media server's router — the server doesn't know how to route this model to the vLLM embedding backend and falls back to its default (ResNet).
```


Solution3: 

 ```
 ---                                                                                                                                                                                                                                            
  The registration infrastructure already exists — the issue is env var routing                                                                                                                                                                  
                                                                                                                                                                                                                                                 
  All the pieces are in place in tt-inference-server:                                                                                                                                                                                            
  - ModelRunners.VLLMForge_QWEN_EMBEDDING is defined                                                                                                                                                                                             
  - MODEL_RUNNER_TO_MODEL_NAMES_MAP maps it to ModelNames.QWEN_3_EMBEDDING_4B                                                                                                                                                                    
  - VLLMForgeEmbeddingQwenRunner class exists in tt_model_runners/vllm_forge_qwen_embedding_runner.py                                                                                                                                            
  - runner_fabric.py has the VLLMForge_QWEN_EMBEDDING → VLLMForgeEmbeddingQwenRunner entry in AVAILABLE_RUNNERS                                                                                                                                  
                                                                                                                                                                                                                                                 
  The failure is that the server starts with runner_in_use: 'tt-xla-resnet' — meaning settings.model_runner resolved to tt-xla-resnet at startup instead of vllmforge_qwen_embedding.                                                            
                                                                                                                                                                                                                                                 
  The fix is in one place:                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                 
  tt-media-server/config/settings.py — method _set_config_overrides()                                                                                                                                                                            
                                                         
  This method reads the MODEL environment variable and uses MODEL_RUNNER_TO_MODEL_NAMES_MAP to resolve it to the correct runner. The problem is that the MODEL env var passed to the Docker container is either:                                 
  
  1. Not set — server falls back to default (ResNet), OR                                                                                                                                                                                         
  2. Set to Qwen3-Embedding-4B but settings.py is not finding a match and silently falling back to the default runner
                                                                                                                                                                                                                                                 
  Check settings.py's _set_config_overrides() — specifically the fallback/default logic when the model name lookup fails. The MODEL_RUNNER_TO_MODEL_NAMES_MAP lookup must match the exact string being passed (i.e., "Qwen3-Embedding-4B" vs     
  "Qwen/Qwen3-Embedding-4B") — a string mismatch here would silently route to the default ResNet runner.      

Root cause found. Dockerfile.forge has:                                                                                                                                                                                                        
                                                         
  ARG MODEL_RUNNER=tt-xla-resnet                                                                                                                                                                                                                 
  ENV MODEL_RUNNER=${MODEL_RUNNER}                                                                                                                                                                                                               
                                                                                                                                                                                                                                                 
  The MODEL_RUNNER=tt-xla-resnet env var is baked into the image. When the container starts, settings.py reads os.getenv("MODEL_RUNNER") → "tt-xla-resnet", and that explicit runner overrides the auto-detection from                           
  MODEL_RUNNER_TO_MODEL_NAMES_MAP. The pipeline passes -e MODEL=Qwen3-Embedding-4B but never passes -e MODEL_RUNNER=vllmforge_qwen_embedding to override the baked-in value.
                                                                                                                                                                                                                                                 
  Let me verify by checking if (TT_XLA_RESNET, N150) exists in ModelConfigs: 

Root cause confirmed. (TT_XLA_RESNET, N150) is in ModelConfigs (added via a loop at line 1203), so the resnet config is found and self.model_runner = "tt-xla-resnet" is set.                                                                  
                                                         
  Here is the complete trace for pipeline run #24957835658:                                                                                                                                                                                      
   
  The bug — Dockerfile.forge line 21-22:                                                                                                                                                                                                         
  ARG MODEL_RUNNER=tt-xla-resnet                         
  ENV MODEL_RUNNER=${MODEL_RUNNER}                                                                                                                                                                                                               
                                                         
  MODEL_RUNNER=tt-xla-resnet is baked into the image at build time. The pipeline starts the container with -e MODEL=Qwen3-Embedding-4B but never passes -e MODEL_RUNNER=vllmforge_qwen_embedding.                                                
                                                                                                                                                                                                                                                 
  Execution trace through settings.py:                                                                                                                                                                                                           
                                                                                                                                                                                                                                                 
  1. pydantic-settings reads MODEL_RUNNER=tt-xla-resnet from the process env → self.model_runner = "tt-xla-resnet" (before __init__ logic runs)                                                                                                  
  2. model_to_run = os.getenv("MODEL") → "Qwen3-Embedding-4B" ✓
  3. _set_config_overrides("Qwen3-Embedding-4B", "n150") is called                                                                                                                                                                               
  4. model_name_enum = ModelNames("Qwen3-Embedding-4B") → QWEN_3_EMBEDDING_4B ✓                                                                                                                                                                  
  5. explicit_runner = os.getenv("MODEL_RUNNER") → "tt-xla-resnet" (from baked-in ENV)                                                                                                                                                           
  6. model_runner_enum = ModelRunners("tt-xla-resnet") → TT_XLA_RESNET ← wrong runner selected here                                                                                                                                              
  7. Map loop runs but if not model_runner_enum: is False → auto-detection is skipped                                                                                                                                                            
  8. config_key = (TT_XLA_RESNET, DeviceTypes.N150) → found in ModelConfigs (the bulk loop at line 1203)                                                                                                                                         
  9. matching_config is non-empty → self.model_runner = "tt-xla-resnet" ✓ (matches log)                                                                                                                                                          
  10. Server starts with runner_in_use: tt-xla-resnet → /v1/embeddings route not registered → HTTP 404                                                                                                                                           
                                                                                                                                                                                                                                                 
  Fix — tt-media-server/Dockerfile.forge:                                                                                                                                                                                                        
                                                                                                                                                                                                                                                 
  Remove the ENV MODEL_RUNNER line so the baked-in value doesn't override runtime auto-detection:                                                                                                                                                
   
  # Remove these two lines (or change the default ARG to empty):                                                                                                                                                                                 
  ARG MODEL_RUNNER=tt-xla-resnet   # ← remove or change to: ARG MODEL_RUNNER=                                                                                                                                                                    
  ENV MODEL_RUNNER=${MODEL_RUNNER}  # ← remove entirely                                                                                                                                                                                          
                                                                                                                                                                                                                                                 
  With MODEL_RUNNER unset in the container, os.getenv("MODEL_RUNNER") returns None, the map lookup runs correctly, finds VLLMForge_QWEN_EMBEDDING for Qwen3-Embedding-4B, and the server starts with the correct embedding runner.

```
Commit3:

https://github.com/tenstorrent/tt-inference-server/commit/568f240b950933ae2e19c0697ff2842ad42d71bd

Perhaps, this needs to be resolved in a different way, so that proper model runner is selected? Please assist on this part

**Error4:**
`ERROR - [warmup] async failed after 2.7859 seconds. Error: EngineArgs.__init__() got an unexpected keyword argument 'task'`
https://github.com/tenstorrent/tt-shield/actions/runs/24978940829/job/73139118077

Solution4:

                                                                                                                                                                                                                                                                             
```
⏺ Root cause confirmed. Both standard vLLM 0.16.0 and the Tenstorrent fork removed task from EngineArgs — it no longer exists there. The task parameter was moved to a different level (per-method via pooling_task, or auto-detected from model config).                    
                                                                                                                                                                                                                                                                             
  The error: EngineArgs.__init__() got an unexpected keyword argument 'task'                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                             
  What happened: vllm_forge_qwen_embedding_runner.py passes "task": "embed" in llm_args → vllm.LLM(**llm_args) → internally forwarded to EngineArgs(**kwargs) → EngineArgs doesn't accept task in vLLM 0.16.0 → crash.                                                       
                                                                                                                                                                                                                                                                             
  Fix: Remove "task": "embed" from llm_args in vllm_forge_qwen_embedding_runner.py. The warmup() and run() methods already call self.llm.embed(...) which is the correct embedding API — no task specification is needed at init time.                                       
                                                                                                                                                                                                                                                                             
  The change is in tt-media-server/tt_model_runners/vllm_forge_qwen_embedding_runner.py:      
```                                                                                                                                                                               
 Commit4:
https://github.com/tenstorrent/tt-inference-server/commit/a4be99d20d5764f706dae7097e5149069df2a62f          


After all these changes we managed to run server correctly and evals and benchmarks workflows finally had a successful run. After that only spec_tests failed which should be further examined by the inference team.

Please double check the changes that are done through commits and specify if some of those changes must NOT be introduced (impacting other modes?), and suggest a different type of workaround.


Workflow with successful server run and successful benchmarks and evals workflows is the following:
https://github.com/tenstorrent/tt-shield/actions/runs/24995948902/job/73197542226

.........................................................
INFO: ✅ Completed workflow: evals
.........................................................
INFO: ✅ Completed workflow: benchmarks
.............................................................
ERROR: ⛔ workflow: spec_tests, failed with return code: 1


PR with changes:

### PR DESCRIPTION
**Pull request by vcankovicTT**
closes #3190

### DIFF
--- a/tt-media-server/Dockerfile.forge
+++ b/tt-media-server/Dockerfile.forge
@@ -13,14 +13,14 @@ LABEL org.opencontainers.image.source=https://github.com/tenstorrent/tt-inferenc
 ARG DEBIAN_FRONTEND=noninteractive
 
 # default model runner, override with --build-arg MODEL_RUNNER=<runner>
-ARG MODEL_RUNNER=tt-xla-resnet
+# ARG MODEL_RUNNER=tt-xla-resnet
 
 # CONTAINER_APP_UID is a random ID, change this and rebuild if it collides with host
 ARG CONTAINER_APP_UID=1000
 ARG DEBIAN_FRONTEND=noninteractive
 
 ENV ENVIRONMENT=development
-ENV MODEL_RUNNER=${MODEL_RUNNER}
+# ENV MODEL_RUNNER=${MODEL_RUNNER}
 
 ENV SHELL=/bin/bash
 ENV CONTAINER_APP_USERNAME=container_app_user
--- a/tt-media-server/Dockerfile.forge
+++ b/tt-media-server/Dockerfile.forge
@@ -12,15 +12,11 @@ LABEL org.opencontainers.image.source=https://github.com/tenstorrent/tt-inferenc
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-# default model runner, override with --build-arg MODEL_RUNNER=<runner>
-# ARG MODEL_RUNNER=tt-xla-resnet
-
 # CONTAINER_APP_UID is a random ID, change this and rebuild if it collides with host
 ARG CONTAINER_APP_UID=1000
 ARG DEBIAN_FRONTEND=noninteractive
 
 ENV ENVIRONMENT=development
-# ENV MODEL_RUNNER=${MODEL_RUNNER}
 
 ENV SHELL=/bin/bash
 ENV CONTAINER_APP_USERNAME=container_app_user
--- a/tt-media-server/tt_model_runners/forge_runners/requirements.txt
+++ b/tt-media-server/tt_model_runners/forge_runners/requirements.txt
@@ -1,6 +1,5 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
 tt-forge==1.0.0
-vllm>=0.6.0
 torchvision
 
 tabulate
--- a/tt-media-server/tt_model_runners/forge_runners/requirements.txt
+++ b/tt-media-server/tt_model_runners/forge_runners/requirements.txt
@@ -1,5 +1,6 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
 tt-forge==1.0.0
+vllm>=0.6.0
 torchvision
 
 tabulate
--- a/tt-media-server/tt_model_runners/vllm_forge_qwen_embedding_runner.py
+++ b/tt-media-server/tt_model_runners/vllm_forge_qwen_embedding_runner.py
@@ -30,7 +30,6 @@ class VLLMForgeEmbeddingQwenRunner(BaseDeviceRunner):
         ]
         llm_args = {
             "model": SupportedModels.QWEN_3_EMBEDDING_4B.value,
-            "task": "embed",
             "dtype": "bfloat16",
             "disable_sliding_window": True,
             "enable_prefix_caching": False,
--- a/tt-vllm-plugin/pyproject.toml
+++ b/tt-vllm-plugin/pyproject.toml
@@ -31,6 +31,7 @@ dependencies = [
     "numpy>=1.24.4,<2",
     # Pin transformers to match requirements.txt
     "transformers==4.55.0",
+    "mistral-common>=1.0.0",
 ]
 
 [project.optional-dependencies]

### CODE REVIEW

File: tt-media-server/tt_model_runners/forge_runners/requirements.txt [hunk 1]
```diff
@@ -1,5 +1,6 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
 tt-forge==1.0.0
+vllm>=0.6.0
```
ljovanovicTT: Are we sure this is needed?

ljovanovicTT (approved):

tt-media-server/tt_model_runners/forge_runners/requirements.txt
vcankovicTT: I'll do one more testing without explicitly referencing vllm version. As I was informed by Kyle "forge 1.0.0 IIRC is from ~Mar 27th and vllm v0.6.0 was used in tt-xla/forge since Mar 4", so this should be already embedded with the current forge version, although I did face some failures, prior all other changes that were introduced afterwards. 
I'll let you know!

vcankovicTT: Confirmed, this is not mandatory, this version is probably embedded within the forge 1.0.0

tt-media-server/Dockerfile.forge
```diff
@@ -12,15 +12,11 @@ LABEL org.opencontainers.image.source=https://github.com/tenstorrent/tt-inferenc
...
 
-# default model runner, override with --build-arg MODEL_RUNNER=<runner>
-ARG MODEL_RUNNER=tt-xla-resnet
-
 # CONTAINER_APP_UID is a random ID, change this and rebuild if it collides with host
 ARG CONTAINER_APP_UID=1000
 ARG DEBIAN_FRONTEND=noninteractive
 
 ENV ENVIRONMENT=development
-ENV MODEL_RUNNER=${MODEL_RUNNER}
```
ljovanovicTT, line 23: The deleted code is just setting the default runner - I dont think it was causing any issues, but I dont mind deleting it since it could be confusing when using the image

kmabeeTT: Just spotted this :) - this line was causing me issues too actually... when launching an LLM because the MODEL_RUNNER env-var is parsed via https://github.com/tenstorrent/tt-inference-server/pull/3125, it was picking up the version here from docker image with no way to override when using run.py

PR #3211, opened 2026-04-28
Media client refactor 

### LINKED ISSUE #3155: Refactor utils/media_clients/
https://github.com/tenstorrent/tt-inference-server/pull/3044#discussion_r3129528753

The goal is to replace per-media client files with unified eval and benchmark files. Client files go away

ddjukicTT:
Issue: https://github.com/tenstorrent/tt-inference-server/issues/3155

The goal is to replace per-media client files with unified eval and benchmark files. Client files go away


fivanovicTT (changes requested):
Can you break down both `media_benchmark_tests` and `media_eval_tests` into separate benchmark and eval tests?

For example: `audio_benchmark_tests` and `audio_eval_tests`etc

Reason for this is that we have granular files with specific only tests related to that model. You have also created a new folder structure for both benchmarks and evals, so it can support it. With this change, I know exactly what file I need to update/change and we don't have a single file which can become bloated

fivanovicTT (approved):

Issue #3174, opened 2026-04-24
[DS Server] Use Prefill on Decode When Prefill Server is not Connected


ljovanovicTT:
Currently, if we run DS setup with only Decode server running, server is unresponsive, but we should be able to provide inference with prefill on decode

PR #3213, opened 2026-04-28
Use Prefill on Decode When Prefill Server Not Connected

ljovanovicTT:
## Issue
[#3174 [DS Server] Use Prefill on Decode When Prefill Server is not Connected](https://github.com/tenstorrent/tt-inference-server/issues/3174)

## Problem
Currently, if we run DS setup with only Decode server running, server is unresponsive, but we should be able to provide inference with prefill on decode


idjuricTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -410,6 +412,18 @@ bool LLMController::shouldDoPrefillOnDecode(const domain::LLMRequest& request,
     return true;
   }
 
+  // In disaggregated decode mode, fall back to running prefill locally if the
+  // prefill server socket is unavailable — otherwise the request would be sent
+  // to a peer that cannot service it.
+  if (tt::config::llmMode() == tt::config::LLMMode::DECODE_ONLY &&
```
ztorlakTT: The `llmMode() == DECODE_ONLY` check is redundant because `shouldDoPrefillOnDecode` is only called from decode server

ztorlakTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
ljovanovicTT: You're right, thanks for noticing!
Removed in the latest commit

Issue #3013, opened 2026-04-17
Add named function tool_choice support


lmuravljovTT:
# Add Named Function Tool Choice Support                                                                                  
                                                                                                                            
  ## What to Implement                                                                                                      
                       
  **1. Parse Named Function** (`domain/tool_calls/tool_choice.hpp`)                                                         
  - Extend `ToolChoice::fromJson()` to handle:                                                                              
    ```json                                                                                                                 
    {                                                                                                                       
      "type": "function",                                                                                                   
      "function": {"name": "get_weather"}                   
    }
  - Current: Only supports "auto" (string)
  - New: Support object with type + function.name                                                                           
                                                 
  2. Validation (domain/chat_completion_request.hpp)                                                                        
  - If tool_choice specifies function name:                                                                                 
    - Verify function exists in tools array                                                                                 
    - Throw error if function not found                                                                                     
                                                            
  3. Chat Template Enforcement (utils/tokenizers/deepseek_tokenizer.cpp)                                                    
  - When tool_choice.function.name is set:
    - Add instruction: "You MUST use the function_name function"                                                            
    - Or filter tools list to only include that function        
  - Apply to both DeepSeek and Llama tokenizers  

4. Must use structured output 
 - When sending a request to the model, add structured output schema to the request, which will match the expected output format for the given function call                                                                            
                                                                                                                            
  5. Tests                                                                                                                  
  - Parse tool_choice with function name                                                                                    
  - Template includes enforcement instruction                                                                               
  - Validation rejects unknown function names
  - Multiple tool_choice modes: auto, required, none, named function                                                        
                                                                                                                            
  Request Example                                                                                                           
                                                                                                                            
  {                                                                                                                         
    "tools": [                                                                                                              
      {"type": "function", "function": {"name": "get_weather", ...}},
      {"type": "function", "function": {"name": "get_time", ...}}
    ],                                                                                                                      
    "tool_choice": {
      "type": "function",                                                                                                   
      "function": {"name": "get_weather"}  ← Force this one 
    }                                                                                                                       
  }
                                                                                                                            
  Acceptance                                                

  - ✅ Parse named function from tool_choice                                                                                
  - ✅ Validate function exists in tools
  - ✅ Chat template enforces specific function                                                                             
  - ✅ Tests cover all tool_choice modes                    
                                                                                                                            
  Current Support
                                                                                                                            
  - ✅ "auto" (model decides)                                                                                               
[...]

Issue #2224, opened 2026-02-26
Tool calling


idjuricTT:
Support tool calling for DeepSeek

PR #3214, opened 2026-04-28
Add named function calling

lmuravljovTT:
#3013
# Add Named Function Call Support via tool_choice                                                                          
                                         
  ## Summary                                                                                                                 
   
  Adds support for `tool_choice` with `type: "function"` to enforce structured output for specific tool calls. When a named  function is specified via `tool_choice`, the system uses that function's parameter schema for guided decoding, ensuring generated arguments strictly conform to the tool's specification.
                                                                                                                             
  ## Feature                                                

  This PR implements named function calling, allowing clients to force the model to call a specific tool 
  with schema-validated arguments.                                                                                           
   
                                                                                                                                                                             
  ## Examples
                                                                                                                             
  ### Example 1: Named tool choice enforces tool parameter schema

  **Request:**
  ```json
  {
    "messages": [
      {
        "role": "user",
        "content": "What is the weather in San Francisco?"
      }                                                                                                                      
    ],
    "tools": [                                                                                                               
      {                                                     
        "type": "function",
        "function": {
          "name": "get_weather",
          "parameters": {
            "type": "object",
            "properties": {
              "location": {"type": "string"},
              "units": {"type": "string"}                                                                                    
            },
            "required": ["location"]                                                                                         
          }                                                 
        }
      }
    ],
    "tool_choice": {
      "type": "function",
      "function": {"name": "get_weather"}
    },                                                                                                                       
    "max_tokens": 300
  }                                                                                                                          
                                                            
  Response:
  {
    "choices": [
      {         
        "finish_reason": "tool_calls",
        "message": {                  
          "role": "assistant",
          "tool_calls": [     
            {                                                                                                                
              "type": "function",
              "function": {                                                                                                  
                "name": "get_weather",                      
                "arguments": "{\"location\":\"San Francisco\",\"units\":\"celsius\"}"
              }
            }                                                                                                                
          ]
        }                                                                                                                    
      }                                                     
    ]
  }
```
  ✅ Result: Tool call arguments conform to get_weather parameter schema

  ---
  Example 2: Explicit response_format takes precedence when tool_choice is "auto"
```json                                                                                                                  
  Request:
  {                                                                                                                          
    "messages": [                                           
      {          
        "role": "user",
        "content": "Give me a person"
      }                              
    ], 
    "tools": [
      {       
        "type": "function",
        "function": {      
          "name": "get_weather",
          "parameters": {                                                                                                    
            "type": "object",
            "properties": {                                                                                                  
              "location": {"type": "string"}                
            },                              
            "required": ["location"]
          }                         
        }  
      }  
    ], 
    "tool_choice": {"type": "auto"},                                                                                         
    "response_format": {            
      "type": "json_schema",                                                                                                 
      "json_schema": {                                      
        "schema": {   
          "type": "object",
          "properties": {  
            "person_name": {"type": "string"}
          },                                                                                                                 
          "required": ["person_name"],
          "additionalProperties": false                                                                                      
        }                                                   
      }
    },
    "max_tokens": 300
  }

  Response:
  {
    "choices": [                                                                                                             
      {         
        "finish_reason": "length",                                                                                           
        "message": {                                        
          "role": "assistant",
          "content": "{\"person_name\":\"John Smith\"}"
        }                                              
      }  
    ]  
  }  
   ```
  ✅ Result: Explicit response_format schema is used (not tool schema)
                                                                                                                             
  ---
  Priority Logic                                                                                                             
                                                            
  The system determines which schema to use for guided decoding in the following order:
                                                                                                                             
  1. If tool_choice.type == "function" with named function → Use tool's parameter schema
  2. Else if response_format is set → Use explicit response_format schema                                                    
  3. Else → No guided decoding (plain text generation)

### CODE REVIEW

File: tt-media-server/cpp_server/include/domain/chat_completion_request.hpp [hunk 1]
```diff
@@ -285,13 +287,63 @@ struct ChatCompletionRequest : BaseRequest {
...
+    if (lastAssistantIdx == req.messages.size()) return;
+
+    const auto& lastAssistant = req.messages[lastAssistantIdx];
+
+    if (!lastAssistant.tool_calls.has_value() ||
+        lastAssistant.tool_calls->empty()) {
+      return;
+    }
+
+    const std::string& expectedCallId = lastAssistant.tool_calls->at(0).id;
```
ztorlakTT, line 321: Should we also validate the IDs of other tool calls, not just the first?

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
```diff
@@ -285,13 +287,63 @@ struct ChatCompletionRequest : BaseRequest {
...
+
+    if (lastAssistantIdx + 1 >= req.messages.size()) {
```
ztorlakTT, line 323: Here we accept a tool call without following tool message, right? It would be good to comment this

tt-media-server/cpp_server/include/domain/sampling_params.hpp
```diff
@@ -47,8 +49,18 @@ struct SamplingParams {
...
+  std::optional<std::vector<tool_calls::Tool>> tools;
+  std::optional<tool_calls::ToolChoice> tool_choice;
+
   bool hasGuidedDecoding() const {
-    return response_format_type != ResponseFormatType::TEXT;
+    if (response_format_type != ResponseFormatType::TEXT) {
+      return true;
+    }
+    if (tool_choice.has_value() && tool_choice->type == "function" &&
+        tool_choice->function.has_value() && tools.has_value()) {
```
ztorlakTT, line 60: Should we check whether function refers to an existing tool?

tt-media-server/cpp_server/src/runners/guided_decoder_manager.cpp
```diff
@@ -39,19 +39,46 @@ GuidedDecoderManager::~GuidedDecoderManager() = default;
...
+        Json::Value wrappedSchema;
+        wrappedSchema["type"] = "object";
+
+        wrappedSchema["properties"]["name"]["const"] =
+            tool.functionDefinition.name;
+        wrappedSchema["properties"]["arguments"] =
+            tool.functionDefinition.parameters;
+        wrappedSchema["required"].append("name");
+        wrappedSchema["required"].append("arguments");
+
```
ztorlakTT, line 65: Is omitting `additionalProperties` intentional here?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -405,30 +437,72 @@ void LLMService::postProcess(domain::LLMResponse& response) const {
...
+      choice.text = toolCallParser->stripMarkers(choice.text);
+      continue;
+    }
+
+    if (toolChoice.type == "function") {
+      // Generate unique tool call ID in OpenAI format (call_1, call_2, ...)
+      std::string toolCallId = tt::utils::ToolCallIDGenerator::generate();
+
+      Json::Value decodedOutput;
+      Json::Reader reader;
```
ztorlakTT: I'd suggest using `Json::CharReaderBuilder`

tt-media-server/cpp_server/include/utils/tool_call_id_generator.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Utility class for generating unique tool call IDs.
+ *
+ * Tool call IDs are strings in format "call_<N>" where N is a sequential
+ * number. This provides:
+ * - Thread-safe generation via atomic counter
+ * - Deterministic sequential IDs (call_1, call_2, call_3, ...)
+ * - Simple debugging (sequential IDs are easy to track in logs)
+ * - OpenAI-compatible format
```
ztorlakTT, line 16: The tool call ID format here differs slightly from OpenAI's. They use random alphanumeric strings, while this generates sequential integers. I opened an [issue](https://github.com/tenstorrent/tt-inference-server/issues/3226) for it, but it's not a blocker for this PR — we can follow up separately and see if it's actually needed

lmuravljovTT: Thanks!

tt-media-server/cpp_server/src/runners/guided_decoder_manager.cpp
lmuravljovTT: nope, will add it

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
lmuravljovTT: This is a good point, valid when parallel_tool_calls is set to true, 

I think there's more loop holes throughout the code for that case as it wasn't taken into consideration all that much when creating the base PR. 

For now, I think this would be okay, and I'd come back to implement the proper parallel calling logic with this [task ](https://github.com/tenstorrent/tt-inference-server/issues/2224?issue=tenstorrent%7Ctt-inference-server%7C3236)

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
lmuravljovTT: Hmm yeah this was added just to handle the edgecase but then again not sure it could happen?
Only if the client sends a request without appending their message / tool result, guess we should throw some exception then?

tt-media-server/cpp_server/include/domain/sampling_params.hpp
lmuravljovTT: That's already validated in llm_service, so it wouldn't get to here if it's not found

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
ztorlakTT: Agreed, I think it would be better to throw an exception

ztorlakTT (approved):

Issue #3225, opened 2026-04-28
Add ModelSpec metadata fields to the release_model_spec.json during release


bgoelTT:
In `workflows/model_spec.py` the `ModelSpec` objects have a `metadata` attribute that should be propogated to the model specification objects in the `release_model_spec.json` that is updated during releases.

PR #3227, opened 2026-04-28
Add metadata to release_model_spec.json

vmaksimovicTT:
Addresses https://github.com/tenstorrent/tt-inference-server/issues/3225

### DIFF
--- a/release_model_spec.json
+++ b/release_model_spec.json
@@ -116,7 +116,11 @@
             "subdevice_type": null,
             "uses_tensor_model_cache": true,
             "has_builtin_warmup": true,
-            "cli_args": {}
+            "cli_args": {},
+            "metadata": {
+              "reasoning_parser_name": "openai_gptoss",
+              "tool_call_parser_name": "openai"
+            }
           }
         }
       },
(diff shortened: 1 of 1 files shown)


bgoelTT (commented):
Looks good in terms of what the final json will look like, but could you also add the changes required to perform this in an automated way?

vmaksimovicTT:
> Looks good in terms of what the final json will look like, but could you also add the changes required to perform this in an automated way?

Those changes are already present, this was created automtically. The reason why the metadata was not present in the release_model_spec.json after the last releas - I removed it manually, as I was not sure how the new field would affect current deployment to console

bgoelTT:
> > Looks good in terms of what the final json will look like, but could you also add the changes required to perform this in an automated way?
> 
> Those changes are already present, this was created automtically. The reason why the metadata was not present in the release_model_spec.json after the last releas - I removed it manually, as I was not sure how the new field would affect current deployment to console

Ahh okay, yeah we actually added that metadata field for Console specifically

bgoelTT (approved):

PR #3245, opened 2026-04-29
Tool Call ID Generation fix

lmuravljovTT:
OpenAI Tool Call IDs are typically 29 characters long (including the prefix). They follow a standard pattern:

    Prefix: call_ (5 characters)

    Suffix: 24 random alphanumeric characters (case-sensitive).


tt-media-server/cpp_server/include/utils/tool_call_id_generator.hpp
```diff
@@ -1,29 +1,34 @@
...
-    return "call_" + std::to_string(++counter);
+    static constexpr const char kChars[] =
+        "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+
+    thread_local std::mt19937 gen(std::random_device{}());
+    std::string result = "call_";
+    result.reserve(29);
+
+    for (int i = 0; i < 24; ++i) {
+      result += kChars[gen() % 62];
```
ztorlakTT: I'm just thinking out loud - would it make any difference if we used std::uniform_int_distribution?

tt-media-server/cpp_server/include/utils/tool_call_id_generator.hpp
```diff
@@ -1,29 +1,34 @@
...
-   * Generate a new unique tool call ID using atomic counter.
-   * Thread-safe, returns "call_1", "call_2", "call_3", etc.
-   */
   static std::string generate() {
-    static std::atomic<uint64_t> counter{0};
-    return "call_" + std::to_string(++counter);
+    static constexpr const char kChars[] =
+        "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+
+    thread_local std::mt19937 gen(std::random_device{}());
```
ztorlakTT: I'm wondering whether we need a bigger seed, e.g. using std::seed_seq with multiple random_device values. What do you think?

ztorlakTT (approved):

tt-media-server/cpp_server/include/utils/tool_call_id_generator.hpp
lmuravljovTT: Sounds like a good idea!

tt-media-server/cpp_server/include/utils/tool_call_id_generator.hpp
lmuravljovTT: Yep! Updated it to use unfirom distribution instead of just modding.

PR #3246, opened 2026-04-29
New DiT runner for i2v

### LINKED ISSUE #3242: Runner logic for new model

vpetrovicTT:


vpetrovicTT:
Random CI run in order to validate nothing is broken: https://github.com/tenstorrent/tt-shield/actions/runs/25306434479

Dedicated PR runs for the new functionality comming after adding the model to Shield

server_tests/test_cases/video_generation_i2v_test.py
```diff
@@ -0,0 +1,155 @@
...
+from pathlib import Path
+
+import aiohttp
+
+from server_tests.base_test import BaseTest
+
+logger = logging.getLogger(__name__)
+
+# Relative to repo root; same dir as other server_tests datasets.
+FIXTURE_IMAGE_PATH = Path(__file__).parent.parent / "datasets" / "i2v_fixture.png"
```
fivanovicTT: I don't see that you have commited this `png` file? I suppose it is under `.gitignore`, so locally you will have to locate the file, uncomment this folder from gitignore and push it

server_tests/test_cases/video_generation_i2v_test.py
```diff
@@ -0,0 +1,155 @@
...
+POLL_INTERVAL_SECONDS = 5
+DEFAULT_JOB_TIMEOUT_SECONDS = 1200
+DEFAULT_NUM_INFERENCE_STEPS = 40
```
fivanovicTT: Can we utilize `Test Categorization System` and just add a new JSON file (since this is a new model) with these configurations? 

Example: `image.json`

server_tests/test_cases/video_generation_i2v_test.py
```diff
@@ -0,0 +1,155 @@
...
+    raw_bytes = FIXTURE_IMAGE_PATH.read_bytes()
+    return base64.b64encode(raw_bytes).decode("ascii")
+
+
+class VideoGenerationI2VTest(BaseTest):
+    """Happy-path test for the I2V endpoint end-to-end."""
+
+    async def _run_specific_test_async(self):
+        self.base_url = f"http://localhost:{self.service_port}"
+        submit_url = f"{self.base_url}/v1/videos/generations/i2v"
```
fivanovicTT: This can be a constant in the file

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -387,26 +399,117 @@ def get_pipeline_device_params(self):
...
+
+    @log_execution_time(
+        f"{dit_runner_log_map[get_settings().model_runner]} inference",
+        TelemetryEvent.MODEL_INFERENCE,
+        os.environ.get("TT_VISIBLE_DEVICES"),
+    )
+    def run(self, requests: list[VideoI2VGenerateRequest]):
+        self.logger.debug(f"Device {self.device_id}: Running inference")
+        request = requests[0]
+        # TODO: Move parameterization outside of runner class.
```
fivanovicTT: Unnecessary comment or still work left to do?

kevinmi-TT:
Hey @vpetrovicTT, looks great i think this extends a lot from what i have on branch, two things worth picking up from kevinmi/wan22-i2v-support before this merges:

MPI image broadcasting, the MPI video_runner.py on main only broadcasts the text VideoRequest to all ranks. For I2V to work on quad/galaxy multi-host setups the conditioning image bytes need to be distributed to all ranks before inference. That branch has the implementation for this.

Base64 padding fix in ImageManager.base64_to_pil_image, base64 strings transmitted over HTTP/JSON often have trailing = padding stripped. Without restoring it (+= "=" * (-len(base64_string) % 4)) before calling b64decode, clients sending stripped strings will get a binascii.Error. That branch also adds HTTP/HTTPS URL support to the same method if the caller passes a URL instead of a base64 string, if case we want to support that

vpetrovicTT:
### Tested locally on galaxy

---


Screenshot 2026-05-05 at 10 41 48
Screenshot 2026-05-05 at 10 41 48

vpetrovicTT:
Follow up incoming for multi host i2v solution

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -141,20 +154,22 @@ def distribute_block():
...
-                    )
-                ],
-            )
+            self.run([self._build_warmup_video_request()])
 
         self.logger.info(f"Device {self.device_id}: Model warmup completed")
 
         return True
 
+    def _build_warmup_video_request(self) -> VideoGenerateRequest:
```
idjuricTT, line 166: Please double check will this work, @sadesoyeTT  mentioned this used to throw na error since num_inference_steps 2 Is invalid

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -314,13 +329,64 @@ def get_pipeline_device_params(self):
...
+    """Build the kwargs dict shared by Wan2.2 T2V and I2V ``__call__`` sites."""
+    pipeline_args = {
+        "prompt": request.prompt,
+        "height": resolution.height,
+        "width": resolution.width,
+        "num_frames": WAN22_NUM_FRAMES,
+        "num_inference_steps": request.num_inference_steps,
+        "guidance_scale": 4.0,
+        "guidance_scale_2": 3.0,
+        "seed": int(request.seed or 0),
```
idjuricTT: I don't think we should reset seed to 0 always if this is not the intial seed

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -340,73 +406,99 @@ def load_weights(self):
...
+
+    @log_execution_time(
+        f"{dit_runner_log_map[get_settings().model_runner]} inference",
+        TelemetryEvent.MODEL_INFERENCE,
+        os.environ.get("TT_VISIBLE_DEVICES"),
+    )
+    def run(self, requests: list[VideoI2VGenerateRequest]):
+        self.logger.debug(f"Device {self.device_id}: Running inference")
+        request = requests[0]
+        resolution = wan22_target_resolution(self.settings.device_mesh_shape)
```
idjuricTT: This part is static for all the requests, I wouldn't load it dynamically

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -340,73 +406,99 @@ def load_weights(self):
...
+        The I2V pipeline resizes the image to the target (height, width)
+        before VAE encoding, so the input resolution is irrelevant — a
+        small black frame exercises the same kernels as a real photo.
+        """
+        import base64
+        import io
+
+        import PIL.Image
+
+        from domain.video_i2v_generate_request import ImagePromptEntry
```
idjuricTT: Can be loaded in the beginning since it's always used

tt-media-server/utils/image_manager.py
```diff
@@ -12,7 +12,7 @@
 
 
 class ImageManager:
-    def __init__(self, storage_dir: str):
+    def __init__(self, storage_dir: str = ""):
```
idjuricTT, line 15: Will this change the behavior?

vpetrovicTT: It should not

tt-media-server/tt_model_runners/dit_runners.py
vpetrovicTT: Fix incoming

tt-media-server/tt_model_runners/dit_runners.py
vpetrovicTT: It should work now, cause we are using it just for the warmup.

That same value 2 is hardcoded inside the pipeline's own warmup buffer initialization:
```
    def warmup_buffers(self, height, width, image_prompt=None):
        self.run_single_prompt(
            prompt="warmup",
            image_prompt=image_prompt,
            height=height,
            width=width,
            num_frames=81,
            num_inference_steps=2,
        )
```

idjuricTT (approved):

PR #3251, opened 2026-04-29
Start spliting num_of_devices to two different vars

### LINKED ISSUE #2122: Split num_of_devices and num_of_requests
In order to be more specific with the spec tests, when running a test for i.e. whisper on Galaxy, be more specific on number of devices and requests, currently it's all "num_of_devices" which might be confusing to the end user of the tests

vpetrovicTT:
## Overview 
The PR splits the overloaded num_of_devices targets key into two distinct concepts: a new num_concurrent_requests for client-side HTTP concurrency in load tests (audio transcription, TTS, CNN, embedding, image-gen, video-gen), while num_of_devices stays reserved for physical chip count used by DeviceLivenessTest / DeviceStabilityTest.

## Use case 

In your test-suite JSON, inside a load test's targets: block, use num_concurrent_requests instead of num_of_devices — for example:
```
{
    "template": "AudioTranscriptionLoadTest",
    "targets": { "num_concurrent_requests": 8, "dataset": "60s" }
}
```
Existing configs using num_of_devices still work but will log a deprecation warning (will be migrated in the next PR) . No code changes are needed in test templates themselves — they already read the new key through the base class.


vpetrovicTT:
CI runs:
[SDXL | galaxy ](https://github.com/tenstorrent/tt-shield/actions/runs/25306457468)
[Whisper | t3k](https://github.com/tenstorrent/tt-shield/actions/runs/25306479988)

fivanovicTT (approved):

PR #3267, opened 2026-04-30
Unique test for health check in spec_tests

### LINKED ISSUE #1835: Unique test for health check in spec_tests
Currently, every test has it's own implementation of calling `DeviceHealthCheck` test, we need a unified place for this.

fivanovicTT:
### Github Issue
[Unique test for health check in spec_tests](https://github.com/tenstorrent/tt-inference-server/issues/1835)

### Summary

- Currently, every test has it's own implementation of calling DeviceHealthCheck test, we need a unified place for this.
- DeviceLivenessTest is registered once as a prerequisite test in server_tests_config.json and is automatically prepended to every suite by the categorization system. You don't need to add it to each test case — it just runs.
- For specific eval test there is a code duplication for _wait_for_server_ready. It should be centralized in single class - base_test.py. I will add fix in relation to this issue

### DIFF
--- a/server_tests/base_test.py
+++ b/server_tests/base_test.py
@@ -8,12 +8,27 @@ import os
 import time
 import traceback
 from abc import ABC, abstractmethod
-from typing import Any, Dict
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+import requests
 
 from server_tests.test_classes import TestConfig
 
 # Set up logging
 logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class HealthCheckConfig:
+    """Health check configuration."""
+
+    MAX_ATTEMPTS: int = 230
+    RETRY_DELAY: int = 10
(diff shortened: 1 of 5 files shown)


server_tests/base_test.py
```diff
@@ -133,6 +148,67 @@ def get_logs(self):
...
+                            "Server is ready after %s attempt(s)",
+                            attempt,
+                        )
+                        return True
+                logger.info(
+                    "Server not ready (attempt %s/%s), retrying in %ss...",
+                    attempt,
+                    max_attempts,
+                    retry_delay,
+                )
```
idjuricTT, line 198: For the future, ideally this should be compatible with c++ server too

idjuricTT (approved):

server_tests/base_test.py
fivanovicTT: Yes, good call out! Once we start working on C++ server, we should follow with changes in Test Framework

Issue #2903, opened 2026-04-13
fix: SERVICE_PORT not passed to media/forge Docker containers


stisiTT:
## Problem

`get_media_server_docker_env_vars()` in `workflows/run_docker_server.py` does not include `SERVICE_PORT` in the environment variables passed to media-engine and forge-engine Docker containers. The container's `run_uvicorn.sh` reads `${SERVICE_PORT:-8000}`, so without the env var it always defaults to port 8000 internally.

Meanwhile, `generate_docker_run_command()` publishes `--publish bind_host:service_port:service_port`, using the same value for both host and container ports. This means `--service-port 7001` publishes `7001:7001` on the host side, but the container still listens on 8000 — traffic never reaches the server.

The `--service-port` CLI option is effectively broken for all media and forge engine models (whisper, speecht5, SDXL, etc.).

## Expected behavior

`--service-port 7001` should make the container listen on 7001 (via `SERVICE_PORT=7001` env var) and publish `7001:7001`, so the port mapping works end-to-end.

## Fix

Pass `runtime_config` to `get_media_server_docker_env_vars()` and add `SERVICE_PORT` to the returned dict. This is consistent with `run_local_server.py` which already sets `SERVICE_PORT` in its environment (line 176).

The default `RuntimeConfig.service_port` is `"8000"`, matching the `run_uvicorn.sh` fallback, so there is no behavioral change when `--service-port` is not specified.

## Relevant files

- `workflows/run_docker_server.py` — `get_media_server_docker_env_vars()` (line 46) and call site (line 219)
- `tt-media-server/run_uvicorn.sh` — reads `${SERVICE_PORT:-8000}`
- `workflows/runtime_config.py` — `service_port` defaults to `"8000"`

PR #3284, opened 2026-05-01
Anirud/fixes-for-media-server-models

anirudTT:
Still WIP , 
Need to re test speecht5 and whisper on the qb2


stisiTT (commented):
Hey @anirudTT — the fix works functionally (media/forge containers will now get `SERVICE_PORT` and the publish mapping lines up), but it doesn't match what the issue lays out, and the placement has some side effects worth flagging:

**1. Wrong function.** The issue (#2903) specifies the change should go in `get_media_server_docker_env_vars()` at `workflows/run_docker_server.py:46`, with the call site at line 219 updated to pass `runtime_config` in. Putting `SERVICE_PORT` there keeps the helper as the source of truth for media/forge env vars — otherwise a future reader looking at that helper won't realize `SERVICE_PORT` is part of the contract.

**2. Scope leaks into vLLM.** Because the line sits at line 265 — after the `if MEDIA or FORGE` guard at 215-219 — `SERVICE_PORT` now gets injected as an env var into every container, including vLLM. vLLM already receives `--service-port` as a CLI flag at lines 283-284, so this is redundant for that path and broader than the issue asks for.

Could you move the assignment into `get_media_server_docker_env_vars()` and thread `runtime_config` through? It's a one-line shift that keeps the change scoped to the actual bug.

**3. Drop the `model_spec.py` change.** The `dev_mode` image-name fallback (appending `-dev` when `-release-` isn't present) is unrelated to #2903 and silently changes behavior for any image whose tag doesn't contain `-release-` — `dev_mode` would now resolve to a different image entirely instead of leaving the tag untouched. That's a foot-gun, and not something this PR should be carrying.

PR #3285, opened 2026-05-01
fix: add runtime_config parameter to get_media_server_docker_env_vars…

anirudTT:
### fix: pass SERVICE_PORT env var to media/forge containers

  Description:

  Summary

  - Media and forge server containers were launched with --publish host_port:host_port but uvicorn inside the container always defaulted to port 8000 (via ${SERVICE_PORT:-8000}), causing connection
   resets when hitting the host port from outside
  - Fix: inject SERVICE_PORT into the container via get_media_server_docker_env_vars() so uvicorn binds to the correct port and the publish mapping lines up

Addresses issues cited [here](https://github.com/tenstorrent/tt-inference-server/pull/3284#pullrequestreview-4211600139)

### DIFF
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -43,12 +43,13 @@ def short_uuid():
     return str(uuid.uuid4())[:8]
 
 
-def get_media_server_docker_env_vars(model_spec):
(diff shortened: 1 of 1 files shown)


bgoelTT:
Closing as this will be addressed by #3325

Issue #2853, opened 2026-04-09
Test Module Overhaul


fivanovicTT:

fivanovicTT:
TODO: investigate how to incorporate vllm test stuff here (separate folder structure, using vllm bench serve)

fivanovicTT:
All sub-issues completed

PR #3304, opened 2026-05-04
v2 stress tests

### LINKED ISSUE #3209: Stress tests refactor

ddjukicTT:
- Copy v1 stress tests into the v2 test_module and wire into the report renderer
 - Move sweep params from result filenames into the JSON body                                                                                     
 - Null-guard prefill throughput on failed runs and remove the dead image branch


fivanovicTT (approved):
I don't want to block this PR merge by any means, but would like the following things to be addressed in the upcoming PRs:

1. `run_stress_tests.py` -> make it OOP, not a script call
2. Don't use `print` methods, we should use `logger`
3. `stress_tests_record_builder.py` -> make it OOP, not a script call

I would create a separate issue and inside of test module so we can clean this up later

ddjukicTT:
> I don't want to block this PR merge by any means, but would like the following things to be addressed in the upcoming PRs:
> 
> 1. `run_stress_tests.py` -> make it OOP, not a script call
> 2. Don't use `print` methods, we should use `logger`
> 3. `stress_tests_record_builder.py` -> make it OOP, not a script call
> 
> I would create a separate issue and inside of test module so we can clean this up later

Thanks for the feedback! I've opened an [issue](https://github.com/tenstorrent/tt-inference-server/issues/2853?issue=tenstorrent%7Ctt-inference-server%7C3309) to capture this so I can address it in a follow-up PR

fivanovicTT (approved):

PR #3306, opened 2026-05-04
Parallel tool call validation

lmuravljovTT:
## Parallel Tool Call Support                                                                                                                                 

  This PR adds support for parallel tool calls in the C++ server, matching OpenAI's API specification.                                                          
   
When an assistant message contains multiple `tool_calls`, the server now validates that the conversation includes corresponding tool responses before processing continues. This validation ensures the integrity of multi-turn tool call conversations, which is essential for parallel tool calling where the model can request multiple tool invocations simultaneously.

### DIFF
--- a/tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
+++ b/tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
@@ -5,12 +5,12 @@
 
 #include <json/json.h>
 
+#include <algorithm>
 #include <optional>
 #include <sstream>
 #include <stdexcept>
 #include <string>
 #include <vector>
-#include <algorithm>
 
 #include "domain/base_request.hpp"
 #include "domain/chat_message.hpp"
@@ -396,7 +396,7 @@ struct ChatCompletionRequest : BaseRequest {
     }
     for (const auto& expectedId : expectedToolCallIds) {
       size_t count = std::count(actualToolCallIds.begin(),
-                                 actualToolCallIds.end(), expectedId);
+                                actualToolCallIds.end(), expectedId);
 
       if (count == 0) {
         throw std::invalid_argument("Missing tool response for tool_call_id '" +
--- a/tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
+++ b/tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
@@ -408,21 +408,6 @@ struct ChatCompletionRequest : BaseRequest {
             "Duplicate tool response for tool_call_id '" + expectedId +
             "': found " + std::to_string(count) + " responses");
       }
-
-      if (std::find(expectedToolCallIds.begin(), expectedToolCallIds.end(),
-                    expectedId) == expectedToolCallIds.end()) {
-        throw std::invalid_argument("Unknown tool_call_id '" + expectedId +
-                                    "' not found in assistant's tool_calls");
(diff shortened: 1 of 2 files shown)


idjuricTT (approved):

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
```diff
@@ -327,32 +328,101 @@ struct ChatCompletionRequest : BaseRequest {
...
+      }
+
+      if (count > 1) {
+        throw std::invalid_argument(
+            "Duplicate tool response for tool_call_id '" + expectedId +
+            "': found " + std::to_string(count) + " responses");
+      }
+
+      if (std::find(expectedToolCallIds.begin(), expectedToolCallIds.end(),
+                    expectedId) == expectedToolCallIds.end()) {
```
ztorlakTT: `expectedId` is taken from `expectedToolCallIds`, so you can remove this check

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
```diff
@@ -327,32 +328,101 @@ struct ChatCompletionRequest : BaseRequest {
...
+        throw std::invalid_argument("Unknown tool_call_id '" + expectedId +
+                                    "' not found in assistant's tool_calls");
+      }
+    }
+
+    for (const auto& actualId : actualToolCallIds) {
+      if (std::find(expectedToolCallIds.begin(), expectedToolCallIds.end(),
+                    actualId) == expectedToolCallIds.end()) {
+        throw std::invalid_argument("Unknown tool_call_id '" + actualId +
```
ztorlakTT: Can you double-check this condition - I'm not sure if we'll ever hit this throw given the earlier checks?

ztorlakTT (approved):

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
lmuravljovTT: yep, this was redundant, thanks

tt-media-server/cpp_server/include/domain/chat_completion_request.hpp
```diff
@@ -327,32 +328,86 @@ struct ChatCompletionRequest : BaseRequest {
...
+    std::vector<std::string> expectedToolCallIds;
+    for (const auto& toolCall : toolCalls) {
+      expectedToolCallIds.push_back(toolCall.id);
     }
 
-    if (!nextMessage.tool_call_id.has_value()) {
-      throw std::invalid_argument(
-          "Message with role='tool' must include 'tool_call_id' field");
+    std::vector<std::string> actualToolCallIds;
+    for (size_t i = lastAssistantIdx + 1; i < req.messages.size(); ++i) {
```
dmadicTT, line 345: Does it make sense to validate the whole conversation instead of just the last messages?

lmuravljovTT: I was curious about this as well, I'd guess it's a valid situation that the client could send an improper message history over, and that could mess with the models output 
On the other hand, with bigger conversations and assuming it will usually be the happy scenario - always the same messages passed through ( with new ones added) , then we would check the same messages on every request, which seems like a waste...

dmadicTT (approved):

PR #3321, opened 2026-05-04
Add non-streaming mock backend bench to the test gate

### LINKED ISSUE #3317: Add non-streaming test in the CI

ztorlakTT:
- Adds a `Bench mock backend (non-streaming)` step to the `cpp-server-benchmarks` job so the test gate exercises `/v1/chat/completions` with `stream=false` alongside the existing streaming bench.
- Uses an inline `aiohttp` client (128/128 tokens, 1000 prompts, 64 concurrency)
- Fails on any non-200 or malformed response; emits a `$GITHUB_STEP_SUMMARY` table (throughput, e2e mean/p50/p99) and uploads `vllm-bench-mock-non-streaming.json` with the other bench artifacts.


idjuricTT (approved):

.github/workflows/test-gate.yml
```diff
@@ -586,6 +588,116 @@ jobs:
             --result-filename "vllm-bench-mock.json" \
             2>&1 | tee bench_mock_output.txt
 
+      # `vllm bench serve` hard-codes stream=True, so drive stream=false coverage via inline aiohttp.
```
dmadicTT: From my perspective, it's great to have non-streaming test to make sure we do not break the endpoint.
Since the goal is not to break it I think the tests should reflect it.
I don't think we are interested in benchmarking the non-streaming behavior.

Also, if we want to have a custom python script for this, it should live in a separate file.
The test gate grew too much and we should think about shrinking it/making it easier to maintain than growing futher.

ztorlakTT: @dmadicTT,  I updated the PR, please take a look

PR #3372, opened 2026-05-06
Default max context 64k

dmadicTT:
- Allows max context length to be set by environment variable
- Sets default to 64k
- Fixes a bug where we reported only delta token count in "usage" instead of the full prompt tokens count


tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -226,6 +229,32 @@ void LLMController::chatCompletions(
...
+      static_cast<size_t>(std::max(0, request->prompt_tokens_count));
+
+  bool exceedsContext = false;
+  if (request->max_tokens.has_value()) {
+    const size_t maxTok =
+        static_cast<size_t>(std::max(0, request->max_tokens.value()));
+    exceedsContext = promptTokens + maxTok >= maxContextLength;
+  } else {
+    exceedsContext = promptTokens >= maxContextLength;
+  }
```
ztorlakTT: Should this be > instead of >=?
Also you can rephrase this to smth like:
```suggestion
  const size_t requested = promptTokens +
      (request->max_tokens.has_value()
           ? static_cast<size_t>(std::max(0, *request->max_tokens))
           : 1);
  exceedsContext = requested > maxContextLength;
```

tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
```diff
@@ -254,6 +254,9 @@ struct ChatCompletionRequest : BaseRequest {
     out.messages = messages;
     out.prompt = tt::utils::tokenizers::activeTokenizer().applyChatTemplate(
         messages, true, tools, enable_reasoning);
+    out.prompt_tokens_count = tt::utils::tokenizers::activeTokenizer()
+                                  .encode(std::get<std::string>(out.prompt))
+                                  .size();
```
ztorlakTT, line 282: Does every request now encode the full prompt twice? I think we already encode in the preProcess method too

ztorlakTT (approved):

tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
dmadicTT: You are right, we do. I also don't like it and consider this temporary change.
All the trouble comes from the preprocess receiving DELTA in case of continuation, so we lose the information how big the original prompt was.

We need to make all of this cleaner. Many of the components have more than one responsibility.
We should have one place/layer where we do tokenization, one place/layer where we apply chat template etc.

idjuricTT (approved):

tt-media-server/cpp_server/include/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -25,7 +25,7 @@ inline pm::GenerationParams makeGenerationParams(
   return {
       .max_new_tokens =
           static_cast<uint32_t>(seq.getSamplingParams().max_tokens.value_or(
-              static_cast<int>(config::LLMConfig::MAX_INPUT_TOKENS))),
+              static_cast<int>(tt::config::defaults::MAX_CONTEXT_LENGTH))),
```
ljovanovicTT: Shouldn't we use `tt::config::maxContextLength()`?

ljovanovicTT (approved):

tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
knovokmetTT: We can expand LLMRreq to hold orignal prompt field as well as delta field, then assert this in preprocess

dmadicTT: This seems just like another workaround tbh

PR #3394, opened 2026-05-06
Get rid of redundant in flight method

idjuricTT:
Changes

- Stream writer is no longer aware of the whole sessionManager
- Usage statistics does not include sessionId
- Removed removeInFlight() method that was in 3 classes for no reason
- Removed sessionId where not needed

### DIFF
--- a/tt-media-server/cpp_server/src/api/llm_controller.cpp
+++ b/tt-media-server/cpp_server/src/api/llm_controller.cpp
@@ -113,9 +113,9 @@ void LLMController::resolveSession(
       [req, routingInfo, onResolved, cancelFn = std::move(cancelFn),
        mgr = sessionManager](const domain::Session& session) mutable {
         req->sessionId = session.getSessionId();
-        req->session = std::make_shared<domain::Session>(session);
         req->slotId =
             mgr->acquireInFlight(session.getSessionId(), std::move(cancelFn));
+        req->session = mgr->getSession(session.getSessionId());
         req->continuation = false;
         mgr->registerPrefixHash(session.getSessionId(),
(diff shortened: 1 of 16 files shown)


dmadicTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -72,45 +72,6 @@ void LLMController::resolveSession(
...
-    } catch (const std::exception& e) {
-      TT_LOG_WARN(
-          "[LLMController] Legacy session acquisition failed for {}: {}; "
-          "falling back to prefix cache",
-          sessionId, e.what());
-      req->sessionId.reset();
-    }
-  }
-
   // Layer 2: Prefix-cache routing. Requires a prior [assistant, user] pair
```
ljovanovicTT, line 75: nit: Maybe update the comments

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -281,12 +242,12 @@ void LLMController::handleStreaming(
         try {
           service->preProcess(*reqPtr);
         } catch (const services::QueueFullException& e) {
-          releaseSessionInFlight(reqPtr->sessionId);
+          reqPtr->session->clearInFlight();
```
ljovanovicTT, line 245: We might get a `nullptr` crash if there if session is not found (!sessionManager==nullptr or getSession() returns nullptr)

Lets maybe add a simple check to make sure we have graceful error handling?

ljovanovicTT (approved):

PR #3396, opened 2026-05-06
Move tool-calling validation to API layer

dmadicTT:
Moves the tool_choice validation block out of LLMService::preProcess into ChatCompletionRequest::fromJson (via validateToolFields), so malformed requests are rejected at JSON parse time before any session or slot is acquired.

### DIFF
--- a/tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
+++ b/tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
@@ -319,6 +319,26 @@ struct ChatCompletionRequest : BaseRequest {
       throw std::invalid_argument("tool_choice='" + type +
                                   "' requires non-empty 'tools'");
     }
+
+    if (type == "function") {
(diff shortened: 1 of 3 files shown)


tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
```diff
@@ -319,6 +319,26 @@ struct ChatCompletionRequest : BaseRequest {
       throw std::invalid_argument("tool_choice='" + type +
                                   "' requires non-empty 'tools'");
     }
+
```
lmuravljovTT, line 322: > I think we should also keep other checks? 
> - tool_choice validness 
> - tools must be provided when tool_choice is function / required / auto 

Ignore, all checks are in place

lmuravljovTT (approved):

idjuricTT (approved):

ztorlakTT (approved):

PR #3403, opened 2026-05-07
Tool call streaming support

lmuravljovTT:
## Summary                                        
                                                                                                                                                                                          
  - Added streaming support for function calling — tool calls are now parsed and emitted incrementally as tokens are generated, enabling real-time tool call streaming in SSE responses   
  - Unified non-streaming with streaming path — non-streaming requests now use the streaming infrastructure internally, with results accumulated by `NonStreamResponseWriter`, eliminating
   duplicate code paths and ensuring consistent behavior                                                                                                                                  
  - Dual parser implementations (JSON vs DeepSeek) — two tool call parsers exist because different tool_choice values force models to output tool calls differently:                                                            
    - `JsonToolCallParser`: For structured JSON tool calls directly                                                                                                    
    - `DeepSeekToolCallParser`: For special markers (`<｜tool▁call▁begin｜>`, etc.) to delimit tool calls in the output                                 
                                                                                                                                                                                          
  ## Streaming chunk types                                                                                                                                                                
                                                                                                                                                                                          
  The streaming response follows OpenAI conventions with three chunk types:                                                                                                               
  
  1. **Initial chunk** — role-only chunk sent before content generation starts (`role: "assistant"`, empty content)                                                                       
  2. **Content chunk** — contains delta content (tool call args)
  3. **Usage chunk** — final chunk after args have been sent through -> finish_reason = "tool_calls"


idjuricTT (approved):

idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -449,77 +570,8 @@ void LLMService::processStreamingRequest(
...
-      choice.tool_calls = std::move(toolCalls);
-      choice.text = toolCallParser->stripMarkers(choice.text);
-      choice.finish_reason = "tool_calls";
-    }
-  }
+  // Clean up tool choice map entry if present
+  toolChoiceMap.take(response.task_id);
 }
 
 void LLMService::abortRequest(uint32_t taskId) {
```
madicd, line 563: Since the parsers hold lots of state, should we cleanup something on request abort?

dmadicTT (approved):

tt-media-server/cpp_server/src/services/llm_service.cpp
lmuravljovTT: Yep makes sense! calling finalizeTask on abort now

PR #3428, opened 2026-05-09
cpp server - propagate cancel to blaze + handle all cases

knovokmetTT:
* pipeline manager side still unhandled


idjuricTT (approved):

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
```diff
@@ -226,20 +240,62 @@ std::unique_ptr<tt::domain::llm::Sequence> BlazeRunner::getRequest() {
...
 
-inline void BlazeRunner::handleMemoryRequest(
+void BlazeRunner::handleMemoryRequest(
     const tt::domain::ManageMemoryTask& request) {
+  if (cancelTombstones.consumeCancelTombstone(request.taskId)) {
+    TT_LOG_DEBUG(
+        "[BlazeRunner] handleMemoryRequest: dropping cancelled taskId={}",
+        request.taskId);
+    // we ignore the allocation request and unblock the session
+    memoryManager->handleResponse(request.taskId, tt::domain::INVALID_SLOT_ID);
```
dmadicTT: As this task is not in `allocating` or `cancelling` in `memoryManager` it will log
```
  TT_LOG_WARN(
      "[BlazeMemoryManager] handleResponse: unknown taskId={}, slotId={}",
      taskId, slotId);
```
Could be confusing

tt-media-server/cpp_server/include/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -71,4 +74,30 @@ struct SlotContext {
   uint32_t tokensGenerated = 0;
 };
 
+struct CancelTombstones {
+  std::deque<uint32_t> cancelTombstoneOrder;
+  std::unordered_set<uint32_t> cancelTombstoneSet;
+  static constexpr size_t MAX_CANCEL_TOMBSTONES = 256;
+  bool consumeCancelTombstone(uint32_t taskId) {
```
dmadicTT, line 142: Consuming should also remove from the `cancelTombstoneSet`.

In case we never get the set size to `MAX_CANCEL_TOMBSTONES`, it will keep growing.

step | set | deque
-- | -- | --
remember A | {A} | [A]
consume A | {} | [A] (lazy)
remember B | {B} | [A, B]
consume B | {} | [A, B]
... 10⁶ more cycles ... | {} | [A, B, ..., 10⁶]

tt-media-server/cpp_server/src/services/memory_services/blaze_memory_manager.cpp
dmadicTT: We should rename this to `evicting`

tt-media-server/cpp_server/src/services/memory_services/blaze_memory_manager.cpp
```diff
@@ -43,7 +43,7 @@ void BlazeMemoryManager::handleRequest(
     case domain::MemoryManagementAction::DEALLOCATE: {
```
dmadicTT: We should rename this to `EVICT`

tt-media-server/cpp_server/include/runners/blaze_runner/blaze_runner.hpp
```diff
@@ -42,22 +44,31 @@ class BlazeRunner : public IRunner {
...
   ipc::IResultQueue* resultQueue;
   tt::ipc::ITaskQueue* taskQueue;
+  tt::ipc::ICancelQueue* cancelQueue;
   std::unique_ptr<tt::domain::llm::Sequence> requestToRetry;
   std::unique_ptr<pm::PipelineManager> pipelineManager;
   std::unordered_map<uint32_t, blaze_utils::SlotContext> slotContexts;
+  // Reverse index of slotContexts: taskId -> slotId. Kept in sync with
+  // slotContexts on every insert/erase so isTaskRunning is O(1).
+  blaze_utils::CancelTombstones cancelTombstones;
+  std::unordered_map<uint32_t, uint32_t> taskIdToSlotId;
```
dmadicTT: Having 2 data structures that must remain in sync creates a risk of developer error.
To reduce the risk, we should encapsulate this in a single data structure.
This will co-locate the logic and make it unit testable.

Example
```
class SlotIndex {
 public:
  void assign(uint32_t taskId, uint32_t slotId, SlotContext ctx);
  void eraseBySlot(uint32_t slotId);
  bool isTaskRunning(uint32_t taskId) const;
  SlotContext* contextBySlot(uint32_t slotId);
  size_t size() const;
  bool empty() const;

 private:
  std::unordered_map<uint32_t, SlotContext> bySlot;
  std::unordered_map<uint32_t, uint32_t> byTask;
};
```

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
```diff
@@ -346,6 +415,14 @@ void BlazeRunner::handleRequest(
   if (slotContexts.empty()) {
     lastOutputTime = std::chrono::steady_clock::now();
   }
+  auto existingSlot = slotContexts.find(slotId);
+  if (existingSlot != slotContexts.end() &&
```
dmadicTT: How can this happen?

tt-media-server/cpp_server/include/runners/blaze_runner/blaze_runner.hpp
dmadicTT, line 1: We should prioritize unit tests for blaze runner as the logic is getting more complex
Few use cases to test only for cancellation

- Cancel before ALLOCATE request is seen (ManageMemoryTask still in-flight).
- Cancel between ALLOCATE request and its PM response.
- Cancel between PM ALLOCATE response and handleRequest (Sequence in taskQueue).
- Cancel after handleRequest while running (current no-op).
- Cancel of requestToRetry.
- Cancel for a taskId that already finished (must be a harmless no-op + tombstone GC).

tt-media-server/cpp_server/include/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -71,4 +74,30 @@ struct SlotContext {
...
+    // Lazy removal from the deque: leave a stale entry behind. It will be
+    // skipped when popped during eviction. This keeps the common path O(1).
+    return true;
+  }
+  void rememberCancelTombstone(uint32_t taskId) {
+    if (!cancelTombstoneSet.insert(taskId).second) {
+      return;
+    }
+    cancelTombstoneOrder.push_back(taskId);
+    while (cancelTombstoneSet.size() > MAX_CANCEL_TOMBSTONES &&
```
dmadicTT: I would add a log statement when this happens

tt-media-server/cpp_server/include/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -71,4 +74,30 @@ struct SlotContext {
   uint32_t tokensGenerated = 0;
 };
 
+struct CancelTombstones {
+  std::deque<uint32_t> cancelTombstoneOrder;
+  std::unordered_set<uint32_t> cancelTombstoneSet;
+  static constexpr size_t MAX_CANCEL_TOMBSTONES = 256;
```
dmadicTT: Is there a story behind this number or it's just a solid starting point?

dmadicTT (approved):

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
```diff
@@ -226,20 +240,62 @@ std::unique_ptr<tt::domain::llm::Sequence> BlazeRunner::getRequest() {
...
   memoryManager->handleRequest(request);
 }
 
-inline void BlazeRunner::handleMemoryResponse(const pm::PMResponse& response) {
+void BlazeRunner::handleCancelRequest(uint32_t taskId) {
+  if (requestToRetry && requestToRetry->taskId == taskId) {
+    TT_LOG_DEBUG(
+        "[BlazeRunner] handleCancelRequest: dropping retry for taskId={}",
+        taskId);
+    requestToRetry = nullptr;
```
dmadicTT, line 272: Before saving the `requestToRetry` we did an `ALLOCATE` for it. 
Since we cancel it, we have to do `EVICT` in this case to release the slot.

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
dmadicTT: Actually, since it's not in `allocating` it will never send the failure back. It will just log.
This is a bigger issue.

```
  if (allocating.erase(taskId)) {           // false — we never inserted
    ...
    if (slotId == ...INVALID_SLOT) {
      // FAILURE branch — the one we want
      result.status = domain::ManageMemoryStatus::FAILURE;
      resultQueue->push(result);
      return;
    }
    ...
  }
```

tt-media-server/cpp_server/src/runners/blaze_runner/blaze_runner.cpp
```diff
@@ -226,20 +240,62 @@ std::unique_ptr<tt::domain::llm::Sequence> BlazeRunner::getRequest() {
...
+    return;
+  }
+  if (isTaskRunning(taskId)) {
```
dmadicTT, line 275: Do we need an early return in this `if`?
I assume we don't want to remember the tombstone yet

PR #3433, opened 2026-05-10
fix(model_spec): add max_num_batched_tokens field to DeviceModelSpec

### LINKED ISSUE #2491: Handle vLLM --max-num-batched-tokens in model_spec.py against device memory
The vllm parameter `--max-num-batched-tokens`: https://docs.vllm.ai/en/stable/cli/serve/?h=vllm+serve#-max-num-batched-tokens
Defines:
> Maximum number of tokens that can be processed in a single iteration.

This gives how many concurrent users can run at a given context, i.e. the constraint:
$$\sum_{u \in U} \mathrm{seqlen_u} <= \texttt{max-num-batched-tokens}$$

This is currently hardcoded in https://github.com/tenstorrent/vllm/blob/dev/vllm/v1/worker/tt_worker.py#L427, as `max_tokens_all_users`.
Most models default to 128k, some model-hardware combinations have reduced total context. This is limitation is to be addressed in vLLM fork tracked here: https://github.com/tenstorrent/vllm/issues/315

The handling in tt-inference-server should be as follows:
Where a model has reduced `max_num_batched_tokens` it should be in `model_spec.py` so that the client side scripts and user documentation can reflect that limitation. A similar pattern can be used for other model config parameters that are non-default.

Dev-X25874:
## Summary

Adds an optional `max_num_batched_tokens` field to `DeviceModelSpec` so
that models with a hardware memory budget different from their context
window can express that independently, without corrupting the semantics
of `max_context`.

Closes #2491

## Problem

`max_num_batched_tokens` was hardcoded to `str(self.max_context)` in
`DeviceModelSpec._infer_data()`. A TODO comment acknowledged this was
wrong. As a workaround, at least one model spec (Qwen3-32B on Galaxy)
inflates `max_context` just to get the right `max_num_batched_tokens`
value, with a NOTE comment explaining the hack. This also incorrectly
sets `--max-model-len` in the vllm args to a value the hardware does not
actually support.

## Fix

- Added `max_num_batched_tokens: Optional[int] = None` to `DeviceModelSpec`.
- When set, `_infer_data` uses it for `--max-num-batched-tokens` and as
  the base for `max_tokens_all_users` (used by client-side benchmark
  concurrency math).
- When `None` (all existing specs), falls back to `max_context` —
  identical to current behavior. No existing spec is changed.
- Removes the TODO comment.

## Testing

All 19 existing model spec unit tests pass unchanged.

## Notes

The Qwen3-32B Galaxy spec NOTE comment is left in place intentionally.
The new field unblocks cleaning that up correctly, but the right values
require hardware-side verification by the model owner.

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -289,6 +289,7 @@ class DeviceModelSpec:
     device: DeviceTypes
     max_concurrency: int
     max_context: int
+    max_num_batched_tokens: Optional[int] = None
     perf_targets_map: Dict[str, float] = field(default_factory=dict)
     default_impl: bool = False
     perf_reference: List[BenchmarkTaskParams] = field(default_factory=list)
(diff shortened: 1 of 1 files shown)


Dev-X25874:
Hey @bgoelTT @idjuricTT @tstescoTT — just wanted to gently ping on this. 
The PR has been open since last week. Would really appreciate a review 
when you get a chance! 🙏

Dev-X25874:
Hey @bgoelTT @idjuricTT @tstescoTT @ddjukicTT @dmadicTT @fivanovicTT @knovokmetTT @ljovanovicTT @vpetrovicTT @ztorlakTT — reopening this for visibility. The bug is still present in main — `max_num_batched_tokens` is still hardcoded to `str(self.max_context)` with a TODO comment acknowledging it's wrong. This PR fixes it properly with an optional field that falls back to `max_context` when not set, so all existing specs are completely unaffected. All 19 unit tests pass. Would really appreciate a review! Thanks!

idjuricTT (approved):

ztorlakTT (approved):

fivanovicTT (approved):

PR #3449, opened 2026-05-11
Organize IPC folder and cleanup dead code

lmuravljovTT:
# Clean up IPC layer

## Summary

Reorganize `tt-media-server/cpp_server/include/ipc/`
subdirectories and drop the `BoostIpc*` prefix from class names so the
transport is encoded in the namespace instead of the type name.

### Layout changes

```
ipc/
  interface/   # transport-agnostic queue interfaces (was ipc/*.hpp)
  boost/       # Boost.Interprocess implementations (was ipc/boost_ipc_*.hpp)
  posix/       # POSIX shm slot ring buffer (was ipc/slot_ring_buffer.hpp)
  helpers/     # small inline helpers, e.g. token_push (was ipc/token_push.hpp)
  queue_manager.hpp
```

### Renames

| Before                         | After                            |
| ------------------------------ | -------------------------------- |
| `tt::ipc::BoostIpcMemoryQueue` | `tt::ipc::boost::MemoryQueue`    |
| `tt::ipc::BoostIpcTaskQueue`   | `tt::ipc::boost::TaskQueue`      |
| `tt::ipc::BoostIpcResultQueue` | `tt::ipc::boost::ResultQueue`    |
| `tt::ipc::BoostIpcCancelQueue` | `tt::ipc::boost::CancelQueue`    |
| `tt::ipc::BoostIpcWarmupSignalQueue` | `tt::ipc::boost::WarmupSignalQueue` |
| `tt::ipc::SlotRingBuffer`      | `tt::ipc::posix::SlotRingBuffer` |
| `tt::ipc::pushToken`           | `tt::ipc::helpers::pushToken`    |

### Other cleanups

- Drop unused `IResultQueue::empty()` and `IResultQueue::isShutdown()`
  virtuals; the only consumer (`BoostIpcResultQueue`) tracked shutdown
  state that nothing read.
- Inside `tt::ipc::boost::MemoryQueue`, alias the global Boost namespace
  as `bi_ipc = ::boost::interprocess` so the surrounding `tt::ipc::boost`
  namespace doesn't shadow `::boost`.
  
  
  
image


dmadicTT:
Even though they are in different namespaces, not sure if we should have same file names for the interface and implementations

image

idjuricTT (approved):

tt-media-server/cpp_server/include/ipc/boost/boost_result_queue.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <memory>
+#include <string>
+
+#include "ipc/boost/boost_memory_queue.hpp"
+#include "ipc/interface/result_queue.hpp"
+
+namespace tt::ipc::boost {
+
+constexpr int RESULT_QUEUE_CAPACITY = 65536;
```
nanicicTT: Maybe better like this, not to have hardcoded values?

```suggestion
#include <stdint.h>
constexpr int  RESULT_QUEUE_CAPACITY = UINT16_MAX + 1;
```

knovokmetTT (approved):

tt-media-server/cpp_server/include/ipc/boost/boost_result_queue.hpp
lmuravljovTT: I've extracted all the capacities and constants to defaults.hpp 

We've got a bunch of hardcoded values here, does it make sense to re-write them all in a similar way?

PR #3475, opened 2026-05-12
Send prompt hash to prefill

idjuricTT:


tt-media-server/cpp_server/src/services/disaggregation_service.cpp
```diff
@@ -130,13 +130,9 @@ void DisaggregationService::setupSocketHandlers() {
...
           auto maxTokens = message.max_tokens;
-          using PromptVariant = std::variant<std::string, std::vector<int>>;
 
-          request.prompt =
-              message.token_ids.empty()
-                  ? PromptVariant(message.prompt)
-                  : PromptVariant(std::vector<int>(message.token_ids.begin(),
-                                                   message.token_ids.end()));
+          request.prompt = std::vector<int>(message.token_ids.begin(),
+                                            message.token_ids.end());
```
nanicicTT, line 135: First comment :) 
Cool, you can also use assign, might be a bit less verbose :) If we already use cpp 20

https://wandbox.org/permlink/9mvnqOVI8LaTCriG

lmuravljovTT (approved):

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
idjuricTT: Thanks!! I missed this, so I'll add a follow up

nanicicTT: you do not have to, it is just nice c++20 feature :) 
This way with std::vector is fully legitimate way to do so, too.

Issue #3480, opened 2026-05-12
Forge SDXL: implement LoRA (load/fuse/unload) to unlock Blackhole LoRA tests


ctr-lelanchelianTT:
## Context
`SDXLForgeRunner` has no LoRA code path. Diffusers' standard `pipeline.load_lora_weights` → `fuse_lora` → `unfuse_lora` mutates UNet weights in place; after that mutation, the device-resident compiled graph holds stale weights.

The trace runner already satisfies this contract via `BaseSDXLRunner._ensure_lora_state` (`tt-media-server/tt_model_runners/base_sdxl_runner.py:213-246`) calling `self.tt_sdxl.load_lora_weights` / `fuse_lora` / `unload_lora_weights`. Forge needs to expose the same API surface.

Blocks #3476 from enabling LoRA test entries on `sdxl-p150x8` / `sdxl-p300x2`.

## Strategy
**V1: recompile-on-change.** On LoRA load/unload: pull UNet (and TE/VAE under full-on-device) back to CPU, fuse/unfuse via diffusers, re-`compile(backend="tt")`, push back to device. Expensive per switch, but correctness-first and matches the trace contract. The LoRA load/rollback tests gate on batch wall-clock (120s target), not per-switch latency.

**V2 (follow-up if recompile cost is unacceptable):** prototype PEFT adapter swap (`set_adapters()`) on TT-XLA.

## Scope
- Add `load_lora_weights(local_path)`, `fuse_lora(scale)`, `unload_lora_weights()` to `SDXLForgeRunner`.
- Plumb `_ensure_lora_state` / trigger word injection into the Forge run path (or extract to a shared helper).
- On `sdxl-p150x8` / `sdxl-p300x2` in `server_tests/test_suites/image.json`: add `ImageGenerationEvalsTest` (LoRA pixel-art-xl), `ImageGenerationLoraLoadTest`, `ImageGenerationLoraRollbackTest`. Remove the `_comment` documenting the gap.

## Acceptance
- All three LoRA test templates pass on Blackhole.
- Behavior unchanged when no `lora_path` is set on the request.

## Dependencies
- #3476 (test classes already in place).

PR #3488, opened 2026-05-12
Forge SDXL: implement LoRA (load/fuse/unload) via recompile-on-change

ctr-lelanchelianTT:
## Summary

Adds LoRA support to `SDXLForgeRunner` so adapter-based requests work on Blackhole. **Strategy A (V1):** every LoRA path/scale change triggers a full component rebuild — reload UNet + both CLIP text encoders from HF, optionally fuse the adapter via a transient `StableDiffusionXLPipeline`, recompile UNet with `backend="tt"`, move back to the XLA device. Costs one UNet compile per switch (~30s+); correctness-first, matches the trace runner's contract from `base_sdxl_runner.py:213-246`.

Concretely:
- `_ensure_lora_state(request)`: no-op when path+scale match cached state; otherwise calls `_rebuild_components` and updates the cache.
- `_rebuild_components(lora_path, lora_scale)`: tears down old UNet/TE refs, reloads fresh from HF, applies LoRA, recompiles UNet, repins to device.
- `run()` calls `_ensure_lora_state` before encoding and routes the prompt through `prepare_prompt_with_lora` for trigger-word injection (matches trace path behavior).
- `_load_pipeline` stashes `_model_id` / `_variant` so the rebuild path knows what to reload from HF.

Enables `ImageGenerationEvalsTest` (LoRA pixel-art-xl, 100 prompts) on `sdxl-p150x8` and `sdxl-p300x2`. `ImageGenerationLoraLoadTest` stays disabled on BH — its concurrent-mixed-LoRA pattern triggers a recompile per request and won't fit the 120s target; needs Strategy B (PEFT adapter swap) as a follow-up.

## ⚠️ Known risks — please validate during review

**This PR has not been verified on Blackhole hardware.** Two specific concerns to confirm:

1. **torch_xla program cache** (`TT_RUNTIME_ENABLE_PROGRAM_CACHE=1`): may key by graph shape rather than parameter identity. If so, the cached compiled graph could execute with stale (non-LoRA'd) weights even though the `nn.Module` was rebuilt and `.compile(backend="tt")` re-called. **Symptom**: LoRA requests produce baseline images. **Fix path**: cache flush on rebuild, or move to Strategy B.
2. **Memory headroom on rebuild**: `_rebuild_components` allocates a fresh UNet on CPU before the old one is freed from device. Peak memory could spike by ~5GB on the worker.

## Verification status
- ✅ Python AST parses cleanly; JSON validates.
- ✅ Standalone `SDXLForgeRunner` import works in the worker venv with default settings.
- ❌ End-to-end Blackhole verification: my dev box has Wormhole, not Blackhole; cannot exercise the actual TT compile path locally.
- ❌ CPU verification: not run (40-70 min cost; would only validate the diffusers/state-machine slice, not the torch_xla concern that is the real risk).

## Test plan
- [ ] On Blackhole: smoke `ImageGenerationEvalsTest` (LoRA) entry — verify output reflects LoRA style.
- [ ] On Blackhole: baseline → LoRA → baseline image-hash sequence; confirm `L != B1` and `B1 == B2`.
- [ ] Inspect log: each LoRA switch logs `Rebuilding components ...` exactly once.
- [ ] Confirm peak worker RSS during LoRA switch doesn't OOM the host.

Refs #3480


vmilosevic (approved):

ctr-lelanchelianTT:
Closing as superseded by #4814.

Same work — Forge SDXL LoRA via recompile-on-change, same `_ensure_lora_state` design — but #4814 is written against the current tree after #4678 collapsed `tt-inference-server-v2/` into the repo root, whereas this branch still targets the old paths (`server_tests/test_suites/image.json`, `tests/test_matrix_expansion.py`) and is conflicting against `main` after ~3 months.

#4814 is also a superset: it adds `tt-media-server/tests/test_forge_lora_state_management.py`, covering first-LoRA rebuild, dropping LoRA back to a plain UNet, fuse-then-unload graph stability, and build-failure state clearing — none of which exist here.

`main` currently has no LoRA support in `sdxl_forge_runner.py`, so nothing shipped is lost by closing this; the work continues in #4814.

PR #3516, opened 2026-05-13
add QuietBox p300x2 support in models_by_hardware.md

### LINKED ISSUE #3474: missing BH 2xP300 in Models by Hardware page
BH 2xP300 QuietBox and supported models on this server are missing from Models by Hardware page

vcankovicTT:
closes #3474


bgoelTT (changes requested):
Looks great, but we need one change that I remember was requested. We should call it `BH QuietBox 2`. This requires changing this enum method https://github.com/tenstorrent/tt-inference-server/blob/5a731d2b2f50920253744ab37be61369c5aaf81d/workflows/workflow_types.py#L125

bgoelTT (approved):

PR #3519, opened 2026-05-13
PrefillGateway — 1:N decode-to-prefill routing

### LINKED ISSUE #3468: Gateway service for prefill
Create a gateway service for prefill that would distribute the tasks by:

- Longest sequence match and 
- Round robin

### LINKED ISSUE #3469: Inform decode which prefill node has the request
Just inform the decode about the request for now.
It can be reused for KV transfer

### LINKED ISSUE #3508: SocketManager: support multiple peer connections (1:N)
The current `SocketManager` holds a single `peerSocket`, which is fine for a 1:1 decode↔prefill setup but blocks the PrefillGateway from connecting to N prefill servers simultaneously.

Refactor `SocketManager` (or add a thin layer above it) to maintain a map of `server_id → connected socket`, with thread-safe send/receive per peer. Existing 1:1 callers should keep working unchanged.

Scope:
- Per-peer send queue and receive thread.
- Identity is `server_id` (not host:port) so addresses can change across reconnects.
- Surface clean callbacks for peer-up / peer-down so the gateway can update its registry.

### LINKED ISSUE #3509: SocketManager: reconnect with exponential backoff
On a peer drop, `SocketManager` should auto-reconnect with capped exponential backoff (e.g., 100ms → 200ms → 400ms → ... up to 5s), not block indefinitely or busy-loop.

Triggered by:
- Initial connection attempt at startup before the peer is up.
- Mid-life connection loss (e.g., gateway restart, prefill restart).

Required for the PrefillGateway and useful for the existing decode↔prefill path. Emit peer-state events so callers can react (drain in-flight, evict from routing tables, etc.).

### LINKED ISSUE #3510: Prefill → Gateway: notify cache block additions and evictions
To support longest-prefix-match routing in the PrefillGateway, the gateway needs to know which block hashes each prefill currently has cached.

Add two wire messages from prefill to gateway:
- `PrefillCacheBlocksAdded { server_id, block_hashes[] }` — emitted after a request adds new blocks to prefill's cache.
- `PrefillCacheBlocksEvicted { server_id, block_hashes[] }` — emitted on LRU eviction.

Prefill should batch/coalesce these to avoid one message per block. On gateway reconnect, prefill re-announces its full block set so the gateway can rebuild its view.

Depends on the chained-hash work (#3448, #3466, #3467).

ztorlakTT:
Adds a new `PrefillGateway` C++ service that sits between one decode server and N prefill servers, enabling disaggregated LLM serving at scale.

**`prefill_gateway/` (new service)**
- `Dispatcher` — routes `PrefillRequest` to a selected prefill, matches results back, fails in-flight tasks on prefill crash
- `PrefillRegistry` — tracks live prefill nodes; populated via `PrefillRegistrationMessage` sent on each (re)connect
- `AffinityCache` — caches `registrationHash` → prefill ID for sticky routing
- `PrefillSelector` — routing policy: sticky (hash match) → least-inflight → round-robin
- `Dockerfile.gateway` — minimal image (~30 MB), no TT hardware/Python dependencies;

**`cpp_server` changes**
- `USE_PREFILL_GATEWAY=1` inverts socket roles: decode = TCP client (→ gateway), prefill = TCP server (← gateway)
- New wire messages: `PrefillRegistrationMessage`, `PrefillAssignmentMessage`, `PrefillCacheBlocksAdded/EvictedMessage`
- `SocketTransport` extracted from `SocketManager` with reliability fixes: `atomic<int> peerSocket_` (TSan), `shutdownPeer()`, `connectionEstablishedCallback`, exponential backoff, correct `stop()` join order

- `prefill-gateway-format-check` — clang-format-20
- `prefill-gateway-build` — cmake + ctest + artifact upload
- `cpp-server-gateway-split` — E2E: gateway + 2 mock prefills + decode server + curl
- `prefill-gateway-tsan` — ThreadSanitizer (0 races required)

## New environment variables
| Variable | Default | Purpose |
|---|---|---|
| `USE_PREFILL_GATEWAY` | `0` | Enable gateway mode |
| `PREFILL_SERVER_ID` | `<hostname>:<port>` | Stable ID sent in registration |
| `PREFILL_MAX_IN_FLIGHT` | `0` (unlimited) | Capacity hint to gateway |
| `MAX_TOKENS_TO_PREFILL_ON_DECODE` | `1000` | Set to `0` to always offload |

Docker image build: https://github.com/tenstorrent/tt-shield/actions/runs/26042927787

Screenshot 2026-05-18 at 12 11 41


tt-media-server/cpp_server/include/sockets/socket_messages.hpp
```diff
@@ -199,4 +199,87 @@ struct LoadBalanceMessage {
...
+  void write(Archive& ar) const {
+    ar(server_id, block_hashes);
+  }
+
+  template <class Archive>
+  static PrefillCacheBlocksEvictedMessage read(Archive& ar) {
+    PrefillCacheBlocksEvictedMessage msg;
+    ar(msg.server_id, msg.block_hashes);
+    return msg;
+  }
```
nanicicTT: This feels repetitive, so many of the same methods. 
Could be defined once and then reused. Not so easy with templates, but there is something that is called [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) and it might be useful here. 

You have repeated code to read/write in every message. Then, the logic is always the same

```
write:
  call archive with all fields

read:
  create message
  call archive with all fields
  return message
```

and only the fields differ. 

You can put the common logic into this code using CRTP:

```cpp
template <class Derived>
struct SerializableMessage {
  template <class Archive>
  void write(Archive& ar) const {
    auto const& self = static_cast<Derived const&>(*this);

    self.fields([&](auto const&... xs) {
      ar(xs...);
    });
  }

  template <class Archive>
  static Derived read(Archive& ar) {
    Derived msg;

    // you take parameters here from outside and pass to this lambda that calls ar
    msg.fields([&](auto&... xs) {
      ar(xs...);
    });

    return msg;
  }
};
```

and now `SerializableMessage` says:
> I know how to serialize and deserialize any message type, as long as that type tells me what its fields are.

and then to use:

```cpp
struct PrefillRegistrationMessage
    : SerializableMessage<PrefillRegistrationMessage> {
  std::string server_id;
  uint32_t max_in_flight = 0;

  template <class F>
  void fields(F&& f) {
    f(server_id, max_in_flight);
  }

  template <class F>
  void fields(F&& f) const {
    f(server_id, max_in_flight);
  }
};
```

and if something changes it only changes in one place. You see how you pass arguments in fields function by universal reference, or a forwarding reference and it can bind to lvalue or rvalue reference.

tt-media-server/cpp_server/src/config/settings.cpp
```diff
@@ -443,6 +446,29 @@ uint16_t socketPort() {
...
+
+std::string prefillServerId() {
+  static const std::string cached = [] {
+    std::string v = envString("PREFILL_SERVER_ID", defaults::PREFILL_SERVER_ID);
+    if (!v.empty()) return v;
+    char host[256] = {0};
+    if (::gethostname(host, sizeof(host) - 1) != 0) {
+      std::snprintf(host, sizeof(host), "unknown-host");
+    }
+    return std::string(host) + ":" + std::to_string(socketPort());
```
nanicicTT: or why not using string and separate function:


```cpp
std::string getHostname() {
    std::string host(256, '\0');
    if (::gethostname(host.data(), host.size()) == 0) {
        host.resize(strlen(host.c_str()));
        return host;
    }
    return "unknown-host";
}

// And then use it like this:
return getHostname() + ":" + std::to_string(socketPort());
```

Somehow `char host[256] = {0};` is a C-style. Works, but there are modern CPP ways. 
Of course, code is correct, this was just modernization hint, disregard it if you will

tt-media-server/cpp_server/src/sockets/socket_transport.cpp
```diff
@@ -136,18 +137,29 @@ void SocketTransport::stop() {
   running_ = false;
   connected_ = false;
 
-  peerSocket_ = -1;
+  peerSocket_.store(-1);
```
nanicicTT: Do we need some memory order, like

```cpp
peerSocket_.store(-1, std::memory_order_release);
```

Just asking, default is `std::memory_order_seq_cst`. It is incredibly slow, and I wonder should it be default one.

ztorlakTT: Thanks for the review, @nanicicTT! All three comments addressed

idjuricTT (approved):

nanicicTT (approved):
Well done!

PR #3528, opened 2026-05-14
fix(forge): honor --no-auth by treating missing API_KEY as auth-disabled

### LINKED ISSUE #3527: Forge media server: `--no-auth` is silently ignored; inference endpoints always require auth
### Problem

`run.py --docker-server --no-auth ...` is documented as disabling vLLM API key authorization (`workflows/README.md:207`), and it works correctly for `vllm-tt-metal` servers. On the **forge media server**, however, the flag is silently dropped — `/v1/completions` and `/v1/chat/completions` always require `Authorization: Bearer <API_KEY>` regardless of `--no-auth`.

Mechanically: `workflows/run_docker_server.py:294-295` appends `--no-auth` to the container CLI, but the container's entrypoint (`run_uvicorn.sh`) only recognizes `--skip-venv` and discards the rest. The FastAPI auth check in `tt-media-server/server/security/api_key_checker.py:10` then enforces auth unconditionally because it falls back to a hardcoded default:

```python
API_KEY = os.getenv("API_KEY", "your-secret-key")
```

### Repro

```bash
SRV="http://127.0.0.1:8103"

# WITHOUT auth header — inference endpoint rejected with 401
curl -s -o /dev/null -w "%{http_code}\n" -X POST $SRV/v1/completions \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → 401   (expected with --no-auth: 200)

# Metadata endpoints are open (consistent with intent):
curl -s -o /dev/null -w "%{http_code}\n" $SRV/health      # → 200
curl -s -o /dev/null -w "%{http_code}\n" $SRV/v1/models   # → 200
```

This applies to any forge model. Server was started with `--no-auth` and the inference endpoint still 401s.

### Existing precedent in this repo

`vllm-tt-metal/src/run_vllm_api_server.py:416-447` already implements `--no-auth` for the vLLM/tt-metal server using the convention:

> **"No API key configured → no auth required."**

When `--no-auth` is set, it `del os.environ["VLLM_API_KEY"]`. The downstream auth path then sees no key and skips enforcement. The same logic applies when `VLLM_API_KEY` and `JWT_SECRET` are both unset — it logs a warning and continues without auth.

The forge server doesn't follow this convention because of the hardcoded `"your-secret-key"` fallback in `api_key_checker.py:10`, which means the env-unset path still has a key to gate on.

### Proposed fix

Align the forge server with the existing `vllm-tt-metal` convention. Two files, ~5 lines total:

#### `tt-media-server/server/security/api_key_checker.py`

```diff
-API_KEY = os.getenv("API_KEY", "your-secret-key")
-api_key_header = APIKeyHeader(name="Authorization")
+API_KEY = os.getenv("API_KEY")
+api_key_header = APIKeyHeader(name="Authorization", auto_error=False)


-def get_api_key(api_key: str = Security(api_key_header)):
+def get_api_key(api_key: str | None = Security(api_key_header)):
+    if not API_KEY:
+        # Match vllm-tt-metal convention: no API_KEY configured → no auth.
+        return None
     if api_key != f"Bearer {API_KEY}":
         raise HTTPException(...)
```

#### `workflows/run_docker_server.py`

```diff
 api_key = os.getenv("API_KEY")
-if api_key:
+if api_key and not runtime_config.no_auth:
     docker_env_vars["API_KEY"] = api_key
```

(Optional cleanup: remove the now-no-op `docker_command.append("--no-auth")` at line 294-295.)

### Backward compatibility

- Production deployments that explicitly set `API_KEY` continue to enforce auth — no change.
- Users who didn't set `API_KEY` and were relying on the implicit `"your-secret-key"` default lose that fallback. Argued as a security improvement (a public default key provides no real protection); flagging in case reviewers prefer to keep an explicit hardcoded fallback.
- `--no-auth` now works as documented — and as it already works for vllm-tt-metal.

### Verification once merged

```bash
python3 run.py --workflow server --docker-server --no-auth --model <forge-llm> ...
# wait for ready
curl -s -o /dev/null -w "%{http_code}\n" -X POST http://127.0.0.1:<port>/v1/completions \
  -d '{"model":"...","prompt":"hi","max_tokens":4}'
# → 200
```

kmabeeTT:
### Issue

Closes #3527

### Problem

- `--no-auth` was silently dropped by the forge container entrypoint, and `api_key_checker.py`'s hardcoded `"your-secret-key"` default kept auth enforced regardless. Inference endpoints (`/v1/completions`, `/v1/chat/completions`) always required `Authorization: Bearer your-secret-key`.

### What's changed

- Low risk, aligns with the existing convention used by `vllm-tt-metal/src/run_vllm_api_server.py:handle_secrets()`: no `API_KEY` configured → no auth required.
  - `tt-media-server/security/api_key_checker.py` — drop the `"your-secret-key"` default; short-circuit `get_api_key()` when `API_KEY` is unset; `auto_error=False` on the header so missing-header surfaces as None.
  - `workflows/run_docker_server.py` — omit `API_KEY` from container env when `runtime_config.no_auth` is set.

### Testing

- Tested locally on a forge container via `docker cp` + `docker restart`. With no `API_KEY` in env and no `Authorization` header, `/v1/completions` and `/v1/chat/completions` now return 200 (were 401). `/health` and `/v1/models` unaffected.

### DIFF
--- a/tt-media-server/security/api_key_checker.py
+++ b/tt-media-server/security/api_key_checker.py
@@ -8,14 +8,14 @@ from fastapi import HTTPException, Security
 from fastapi.security import APIKeyHeader
 from starlette.status import HTTP_401_UNAUTHORIZED
 
-# No API_KEY → no auth required, matching vllm-tt-metal's handle_secrets() convention.
-API_KEY = os.getenv("API_KEY")
-# auto_error=False so missing header is None (handled below), not 403.
-api_key_header = APIKeyHeader(name="Authorization", auto_error=False)
+API_KEY = os.getenv("API_KEY", "your-secret-key")
+NO_AUTH = os.getenv("NO_AUTH", "").lower() in ("1", "true", "yes")
+# auto_error=False when NO_AUTH so missing header is None, not 403.
+api_key_header = APIKeyHeader(name="Authorization", auto_error=not NO_AUTH)
 
 
 def get_api_key(api_key: str | None = Security(api_key_header)):
-    if not API_KEY:
+    if NO_AUTH:
         return None
     if api_key != f"Bearer {API_KEY}":
         raise HTTPException(
--- a/tt-media-server/security/api_key_checker.py
+++ b/tt-media-server/security/api_key_checker.py
@@ -8,11 +8,15 @@ from fastapi import HTTPException, Security
 from fastapi.security import APIKeyHeader
 from starlette.status import HTTP_401_UNAUTHORIZED
 
-API_KEY = os.getenv("API_KEY", "your-secret-key")  # Or use os.getenv("API_KEY")
-api_key_header = APIKeyHeader(name="Authorization")
+# No API_KEY → no auth required, matching vllm-tt-metal's handle_secrets() convention.
+API_KEY = os.getenv("API_KEY")
+# auto_error=False so missing header is None (handled below), not 403.
+api_key_header = APIKeyHeader(name="Authorization", auto_error=False)
 
 
-def get_api_key(api_key: str = Security(api_key_header)):
+def get_api_key(api_key: str | None = Security(api_key_header)):
+    if not API_KEY:
+        return None
     if api_key != f"Bearer {API_KEY}":
         raise HTTPException(
             status_code=HTTP_401_UNAUTHORIZED,
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -328,7 +328,8 @@ def generate_docker_run_command(
     ):
         docker_env_vars.update(get_media_server_docker_env_vars(model_spec))
         api_key = os.getenv("API_KEY")
-        if api_key:
+        # Skip API_KEY when --no-auth: container sees no key → auth disabled.
+        if api_key and not runtime_config.no_auth:
             docker_env_vars["API_KEY"] = api_key
         if _is_cpp_media_spec(model_spec):
             openai_api_key = os.getenv("OPENAI_API_KEY") or api_key
--- a/workflows/run_docker_server.py
+++ b/workflows/run_docker_server.py
@@ -328,8 +328,9 @@ def generate_docker_run_command(
     ):
         docker_env_vars.update(get_media_server_docker_env_vars(model_spec))
         api_key = os.getenv("API_KEY")
-        # Skip API_KEY when --no-auth: container sees no key → auth disabled.
-        if api_key and not runtime_config.no_auth:
+        if runtime_config.no_auth:
+            docker_env_vars["NO_AUTH"] = "1"
+        elif api_key:
             docker_env_vars["API_KEY"] = api_key
         if _is_cpp_media_spec(model_spec):
             openai_api_key = os.getenv("OPENAI_API_KEY") or api_key


kmabeeTT:
## Extra details for review about testing, etc

**Test results per endpoint** (with no `API_KEY` env var, no `Authorization` header):

| Endpoint | Before | After |
|---|---|---|
| `GET /health` | 200 | 200 |
| `GET /v1/models` | 200 | 200 |
| `POST /v1/completions` | 401 | **200** |
| `POST /v1/chat/completions` | 401 | **200** |

**Backward compatibility:**

- Deployments that explicitly set `API_KEY` continue to enforce auth — no change.
- Deployments that never set `API_KEY` and were implicitly relying on the `"your-secret-key"` default lose that fallback. Argued as a security improvement — the default key is publicly known and provided no real protection — but flagging in case reviewers prefer to keep an explicit hardcoded fallback.

**Precedent reference:**

`vllm-tt-metal/src/run_vllm_api_server.py:416-447` (`handle_secrets()`) already implements `--no-auth` using the "no key configured → no auth" convention. That function `del os.environ["VLLM_API_KEY"]` when `--no-auth` is set, and also short-circuits when neither `VLLM_API_KEY` nor `JWT_SECRET` are set. This PR brings the forge media server in line with that pattern.

**Server-side check is per-route:**

`/health` and `/v1/models` aren't auth-gated even today — they don't `Depends(get_api_key)`. This PR doesn't change that. Only routes that already require auth (the inference endpoints) are affected by the new "no key → no auth" branch.

tt-media-server/security/api_key_checker.py
```diff
@@ -8,11 +8,15 @@
 from fastapi.security import APIKeyHeader
 from starlette.status import HTTP_401_UNAUTHORIZED
 
-API_KEY = os.getenv("API_KEY", "your-secret-key")  # Or use os.getenv("API_KEY")
-api_key_header = APIKeyHeader(name="Authorization")
+# No API_KEY → no auth required, matching vllm-tt-metal's handle_secrets() convention.
+API_KEY = os.getenv("API_KEY")
```
idjuricTT: We use default one as a safety net. I'm not against this change but it would require changing all the other test flows which is tricky

kmabeeTT: Whoops thx, good catch - the behavior change for the "no host `API_KEY`, no `--no-auth`" case (server returning 200 instead of enforcing the default key) was unintentional on my end. 

I reworked the patch w/ follow up commit: kept the `"your-secret-key"` default in `api_key_checker.py` so existing test flows are unchanged, and switched to an explicit `NO_AUTH`env var that `run_docker_server.py` propagates only when `--no-auth` is passed to fix the original issue here only. The flag now opts into auth-disabled behavior instead of relying on absence-of-config. 

New commit pushed. Validated all 4 cases locally on Llama-3.2-3B-Instruct — see follow-up comment for the matrix.

kmabeeTT:
Pushed a change based on feedback above, here is some extra info about testing done w/ it - I think we're good now.

## Test matrix — validated all 4 cases on Llama-3.2-3B-Instruct (forge)

Local validation against the patched branch. Two containers spun up side-by-side on devices 0 / 1, ports 8100 / 8101, with `tt-media-server/security/api_key_checker.py` bind-mounted from the patched host file. Container image: `tt-media-inference-server-forge:709c6a767e...`.

| Case | Host `API_KEY` | `--no-auth` | Expected server behavior | Observed |
|---|---|---|---|---|
| 1 | `foo` | (no) | Enforce `Bearer foo`; reject anything else | ✅ |
| 2 | (unset) | (no) | Enforce `Bearer your-secret-key` (default-key safety net — unchanged from main) | ✅ |
| 3 | `foo` | yes | `--no-auth` strips `API_KEY` from container env, sets `NO_AUTH=1`; server returns 200 for any/no Authorization header | ✅ |
| 4 | (unset) | yes | Server returns 200 for any/no Authorization header | ✅ |

### Case 2 — default-key safety net (unchanged from main)

Server launched via `./run_server_llama_3_2_3b_authon.sh` with no host `API_KEY` set.

```bash
curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8100/v1/completions \
  -H 'Authorization: Bearer your-secret-key' \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 200, {"id":"cmpl-...","choices":[{"text":" there's a precaution",...}], ...}

curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8100/v1/completions \
  -H 'Authorization: Bearer wrong-key' \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 401, {"detail":"Invalid or missing API Key"}
```

Confirms that test flows / deployments that don't set `API_KEY` continue to get auth-on against the literal `"your-secret-key"` default — identical to main.

### Case 4 — `--no-auth` with no host `API_KEY`

Server launched via `./run_server_llama_3_2_3b_authoff.sh`.

```bash
# (a) no Authorization header at all
curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8101/v1/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 200, real completion text returned

# (b) junk Authorization header (proves auth is genuinely off,
#     not just defaulting-to-allow on missing header)
curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8101/v1/completions \
  -H 'Authorization: Bearer total-garbage' \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 200, real completion text returned
```

Container env propagation sanity-check:

```bash
docker exec $(docker ps -q --filter "publish=8101") env | grep -E 'API_KEY|NO_AUTH'
# → NO_AUTH=1
#   (no API_KEY line — confirms run_docker_server.py omits API_KEY when --no-auth)
```

### Case 1 — host `API_KEY=foo` overrides default (auth still on)

Server relaunched with `API_KEY=foo ./run_server_llama_3_2_3b_authon.sh`.

```bash
curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8100/v1/completions \
  -H 'Authorization: Bearer foo' \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 200, {"id":"cmpl-...","choices":[{"text":" there's a precaution",...}], ...}

curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8100/v1/completions \
  -H 'Authorization: Bearer your-secret-key' \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 401, {"detail":"Invalid or missing API Key"}
# (default key no longer valid once host overrode with API_KEY=foo)
```

### Case 3 — `--no-auth` with host `API_KEY=foo` set (flag wins)

Server relaunched with `API_KEY=foo ./run_server_llama_3_2_3b_authoff.sh`.

```bash
curl -sw "\nHTTP %{http_code}\n" -X POST http://localhost:8101/v1/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"meta-llama/Llama-3.2-3B-Instruct","prompt":"hi","max_tokens":4}'
# → HTTP 200, real completion text returned
# (no auth header; --no-auth flag wins regardless of host API_KEY)

docker exec $(docker ps -q --filter "publish=8101") env | grep -E 'API_KEY|NO_AUTH'
# → NO_AUTH=1
#   (no API_KEY line — confirms --no-auth strips host API_KEY from container env;
#    the flag is authoritative even when API_KEY=foo was set on the host)
```

---

### Summary

- **Default behavior (Case 2) is unchanged from main**: the `"your-secret-key"` convenience default is preserved exactly as the reviewer noted — no test-flow migration needed.
- **`--no-auth` flag is now honored end-to-end** (Cases 3, 4): host-side `run_docker_server.py` propagates `NO_AUTH=1` to the container; `api_key_checker.py` short-circuits auth when that env var is set.
- **Explicit host `API_KEY` still works** (Case 1): no change for deployments that set `API_KEY` explicitly.

Implementation uses an explicit `NO_AUTH` env var rather than treating "missing `API_KEY`" as auth-disabled, which preserves the existing default-key convention.

idjuricTT (approved):

Issue #3529, opened 2026-05-14
evals: run_evals.py picks wrong auth path for FORGE/MEDIA LLMs (overwrites API key with JWT)


kmabeeTT:
### Problem

When `run.py --workflow evals` runs against a **forge or media LLM**, every inference request fails with HTTP 401 `{"detail":"Invalid or missing API Key"}`. The lm-eval-harness subprocess sends `Authorization: Bearer <jwt>` to a server that only validates `Authorization: Bearer your-secret-key`.

Mechanically, `evals/run_evals.py:380-392` picks an auth path based on `model_spec.model_type`:

```python
if model_spec.model_type in EVAL_TASK_TYPES:    # CNN/IMAGE/AUDIO/etc.
    _setup_openai_api_key(args, logger)          # uses API key
elif args.jwt_secret:                            # ANY LLM, regardless of engine
    encoded_jwt = jwt.encode(...)
    os.environ["OPENAI_API_KEY"] = encoded_jwt   # ← overwrites with JWT
```

For FORGE/MEDIA LLM, the `elif` branch fires (since `JWT_SECRET` is in `.env` from CI) and overwrites the `OPENAI_API_KEY=your-secret-key` that CI just put in `.env` (`.github/workflows/workflow_run-tests-with-inference-server.yml:277`). The forge/media server validates the literal API key, not JWTs → 401 on every request.

### Existing precedent

`benchmarking/run_benchmarks.py:345-351` has the engine-aware auth branch, and `run_benchmarks.py:439` does an explicit `env_config.vllm_api_key = os.getenv("VLLM_API_KEY")` re-read so the in-process `PromptClient` picks up env updates the dataclass default (`utils/prompt_configs.py:51`) missed. Both are missing in `run_evals.py`.

### Repro

```bash
python3 run.py \
  --model Llama-3.2-3B-Instruct --tt-device n150 \
  --engine forge --impl forge-vllm-plugin \
  --workflow evals --service-port 8103
```

**In-process `PromptClient` trace capture** silently fails — `EnvironmentConfig.vllm_api_key` default was captured at module load, so `PromptClient` sends no auth header, server 401s silently, harness logs zero tokens and a garbage TPOT (`time.time()` returned because no token-arrival was ever recorded):

```
2026-05-14 03:24:17,060 - utils.prompt_client - INFO - Text trace completed: tokens_generated=0, TTFT=0.000ms, TPOT=1387303443.598ms
[...]

PR #3532, opened 2026-05-14
fix(forge): accept batched prompts on /v1/completions to fix 422 Errors

### LINKED ISSUE #3531: Forge /v1/completions schema rejects List[str] and List[List[int]] prompts; blocks all forge-LLM evaluations
### TL;DR

The `/v1/completions` pydantic schema in the forge media server accepts `prompt: str | list[int]` only. OpenAI's spec (and what `lm-evaluation-harness` actually sends) is `prompt: str | list[str] | list[int] | list[list[int]]`. As a result, **every forge-LLM eval fails with HTTP 422** before the model is invoked.

### Problem

Run `--workflow evals` against any forge-engine LLM (verified across Qwen3-4B, Llama-3.2-3B-Instruct, and the upcoming Falcon3-7B-Instruct). Every eval request fails with the same response:

```
openai.UnprocessableEntityError: Error code: 422 - {'detail': [
  {'type': 'string_type', 'loc': ['body', 'prompt', 'str'],
   'msg': 'Input should be a valid string',
   'input': [[128000, 128006, 882, 128007, 271, ...]]}
]}
```

(Captured from a live `--workflow evals` run against Llama-3.2-3B-Instruct on forge; the inner list is the chat-templated, tokenized prompt that lm-eval-harness wraps in an outer list for batched submission.)

In the eval sample JSONL, every prompt is logged as `__INFERENCE_ERROR__`. lm-evaluation-harness then surfaces a "failure noise" baseline score (~9% on IFEval) that is **not a real evaluation result** — it's the harness's default for "all responses empty."

#### The schema

`tt-media-server/domain/completion_request.py:31`:

```python
# prompt can be a string or a list of token ids
prompt: str | list[int]
```

#### What clients actually send

`lm-evaluation-harness` (and llama-cookbook, which wraps it) always wraps prompts in a list when calling `/v1/completions`. Depending on `tokenizer_backend`:

| Setting | What gets sent | OpenAI spec? | Forge schema? |
|---|---|---|---|
| `tokenizer_backend=huggingface` (default) | `prompt: [[101, 23, 4521, …]]` | ✅ valid | ❌ rejected (422) |
| `tokenizer_backend=none` + `apply_chat_template=True` | `prompt: ["<chat-templated string>"]` | ✅ valid | ❌ rejected (422) |
| `tokenizer_backend=none` + `apply_chat_template=False` | `prompt: ["raw text"]` | ✅ valid | ❌ rejected (422) |

OpenAI's official `/v1/completions` schema accepts `Union[str, List[str], List[int], List[List[int]]]`. The forge server implements only the first and third.

There is **no `lm-evaluation-harness` setting** that produces an unwrapped single string for `/v1/completions` — the harness always batches (even at `batch_size=1`).

### Minimal reproduction

Against any running forge LLM server on `/v1/completions`:

```bash
# prompt as nested token-id list — what lm-eval sends with tokenizer_backend=huggingface
curl -s -i -X POST <server>/v1/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"<model>","prompt":[[1,2,3,4]],"max_tokens":4}' | head -10
# → 422 Unprocessable Entity
```

The `list[str]` form (`"prompt":["hello"]`) — what lm-eval sends with `tokenizer_backend=none` — fails the same way. Both shapes are inside OpenAI's documented schema.

Verified across Qwen3-4B, Llama-3.2-3B-Instruct, and Falcon3-7B-Instruct — same 422 regardless of task, eval venv, or `apply_chat_template` setting. The only common factor is the client list-wrapping + the schema rejecting it.

### Why this is showing up now

This bug is latent — it has existed since the forge server's `/v1/completions` endpoint was first written, but no client has hit it because:

- **Other forge models** (CNN / IMAGE / AUDIO / EMBEDDING / VIDEO / TTS) don't go through `/v1/completions` — they have their own endpoints.
- **LLM evals on the `tt-transformers` server** work because that server implements the full OpenAI `Union[str, List[str], List[int], List[List[int]]]` schema and accepts the list-wrapped prompts.
- **Forge LLM evals** are a new combination — no team has run `python3 run.py --workflow evals --engine forge --impl forge-vllm-plugin --model <any-LLM>` until now.

Once any team enables a forge-engine LLM for evals in CI, this bug fires.

### Fix direction

Expand the `prompt` union in `tt-media-server/domain/completion_request.py` to the full OpenAI shape (`str | list[str] | list[int] | list[list[int]]`). The downstream runner takes a single prompt per request, so the handler in `open_ai_api/llm.py` also needs to expand the batched forms into N sub-requests and aggregate into N `choices`. PR follows.

Out of scope: the same payload-shape concern on `/v1/chat/completions` — evals run against `/v1/completions` only.

kmabeeTT:
### Issue

Closes #3531 

### Problem

`tt-media-server/domain/completion_request.py` types `prompt` as `str | list[int]`, but the OpenAI `/v1/completions` spec accepts the full union `str | list[str] | list[int] | list[list[int]]`. lm-evaluation-harness always wraps prompts in a list (even at `batch_size=1`), so every forge-LLM eval request fails with HTTP 422 before the model is invoked.

### What's changed

- `tt-media-server/domain/completion_request.py`: expand `prompt` union to match the OpenAI spec.
- `tt-media-server/open_ai_api/llm.py`:
  - Add `_split_batched_prompts(req)` helper. For `list[str]` / `list[list[int]]`, it produces N single-prompt sub-requests via `req.model_copy(update={"prompt": item})`; for `str` / `list[int]` it returns `[req]` unchanged.
  - Non-streaming path fans out via `asyncio.gather(*(service.process_request(r) for r in sub_requests))` and aggregates results into N indexed `choices`. Order-preserving gather keeps the choice index stable.
  - Token-length validation now loops over all sub-prompts (not just the first), so an oversized batch entry fails fast at the API layer instead of wasting a runner slot.
  - Streaming + batched returns HTTP `501 Not Implemented`. No current client (lm-eval-harness, vllm bench) sends that combination; multiplexing N async generators into one SSE stream is significant work for zero current benefit.

### Testing

Depends on, and was tested with fix for #3529 in place otherwise can't see these errors.

Stock forge container running Llama-3.2-3B-Instruct, `--workflow evals` (lm-eval-harness via the `EVALS_META` venv, `tokenizer_backend=huggingface` so prompts arrive as `list[list[int]]`):

- **Baseline (no fix):** every request 422s; `samples_*.jsonl` is all `__INFERENCE_ERROR__`; lm-eval reports failure-noise scores.
- **With fix:** 0 × 422, 0 × 401 across the run; `meta_ifeval` generated real completions end-to-end, `meta_gpqa` in progress. lm-eval-harness's batched payload is now accepted and fanned out into N choices that map back to its expected response shape.

### DIFF
--- a/tt-media-server/domain/completion_request.py
+++ b/tt-media-server/domain/completion_request.py
@@ -27,8 +27,9 @@ class CompletionRequest(BaseRequest):
     # Model identifier
     model: str | None = None
 
-    # prompt can be a string or a list of token ids
-    prompt: str | list[int]
+    # Per OpenAI /v1/completions spec, prompt can be a string, list of strings,
+    # flat list of token ids, or list of token-id lists (batched).
+    prompt: str | list[str] | list[int] | list[list[int]]
 
     # Response configuration
     echo: bool | None = False
--- a/tt-media-server/open_ai_api/llm.py
+++ b/tt-media-server/open_ai_api/llm.py
@@ -2,6 +2,7 @@
 #
 # SPDX-FileCopyrightText: © 2025 Tenstorrent USA, Inc.
 
+import asyncio
 import json
 import time
 import uuid
@@ -19,6 +20,19 @@ from open_ai_api.chat import _count_tokens
 
 logger = TTLogger()
 router = APIRouter()
+
+
+def _split_batched_prompts(req: CompletionRequest) -> list[CompletionRequest]:
+    """Split a batched-prompt request into N single-prompt requests.
+
+    OpenAI /v1/completions accepts prompt as List[str] or List[List[int]] for
+    batched submission. The downstream runner accepts only a single prompt
+    per request, so split here and aggregate results into N choices.
+    """
+    p = req.prompt
+    if isinstance(p, list) and p and isinstance(p[0], (list, str)):
+        return [req.model_copy(update={"prompt": item}) for item in p]
+    return [req]
 
 
 @router.post("/completions")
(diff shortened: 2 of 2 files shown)


kmabeeTT:
## Extra detail for review context

**Concurrency note:** the downstream `service.process_request` is serialized by a single task queue inside the scheduler, so `asyncio.gather` doesn't introduce parallel device execution — it pipelines enqueueing so subsequent sub-requests don't wait for the previous one's full round-trip before being submitted. Order-preserving gather is also required to keep `choices[i].index == i`.

**Partial-failure semantics:** `asyncio.gather` without `return_exceptions=True` fails the whole batch on any sub-request error (matches vLLM / OpenAI server precedent — the spec leaves this undefined).

idjuricTT (approved):

PR #3536, opened 2026-05-14
Use ZMQ socket for decode to prefill communication

idjuricTT:


tt-media-server/cpp_server/CMakeLists.txt
```diff
@@ -149,6 +149,45 @@ FetchContent_MakeAvailable(spdlog)
...
+    set(WITH_DOCS OFF CACHE BOOL "" FORCE)
+    set(WITH_PERF_TOOL OFF CACHE BOOL "" FORCE)
+    FetchContent_MakeAvailable(libzmq)
+    set(ZMQ_FROM_SOURCE TRUE)
+endif()
+# cppzmq (header-only C++ bindings for libzmq)
+FetchContent_Declare(
+    cppzmq
+    GIT_REPOSITORY https://github.com/zeromq/cppzmq.git
+    GIT_TAG        v4.10.0
```
nanicicTT: 4.11.0 is the newest tag, from June this year :) 
We can use the latest one.

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
```diff
@@ -0,0 +1,185 @@
...
+  }
+}
+
+void ZmqSocketTransport::start() {
+  if (running_) return;
+  running_ = true;
+
+  if (mode_ == Mode::CLIENT) {
+    // DEALER: send an empty "hello" frame so the ROUTER learns our identity.
+    std::lock_guard<std::mutex> lock(sendMutex_);
```
nanicicTT: Design should be, IMHO, one dedicated thread that owns the ZeroMQ socket. 
Here in `send()` and in `sendRawData()` (https://github.com/tenstorrent/tt-inference-server/pull/3536/changes#diff-2d9bc729b7f620391cb8d52cbdf2c7783bc64b367bf9bd24977abc97655d2649R114)
this mutex protects send, not the socket. 

But `receiveRawData()` does this without the same lock: https://github.com/tenstorrent/tt-inference-server/pull/3536/changes#diff-2d9bc729b7f620391cb8d52cbdf2c7783bc64b367bf9bd24977abc97655d2649R136

So, one thread calls:
Thread A: sendRawData()
Thread B: receiveRawData()
and they, theoretically, can access the same zmq::socket_t at the same time. 

For ZeroMQ, I believe normal design is that `context_t` can be shared across threads and
`socket_t` should be owned by one thread. 
What is not ideal here, I believe, is that for many threads -> same `socket_` is protected only around `send()`. 
I think what is better is that one thread owns `socket_` and other threads communicate with that thread through queue/inproc/channel, whatever.

tt-media-server/cpp_server/src/sockets/socket_transport_factory.cpp
```diff
@@ -0,0 +1,22 @@
...
+namespace tt::sockets {
+
+std::unique_ptr<ISocketTransport> createSocketTransport() {
+  auto type = tt::config::socketTransport();
+  if (type == "zmq") {
+    TT_LOG_INFO("[SocketTransport] Using ZMQ transport");
+    return std::make_unique<ZmqSocketTransport>();
+  }
+  TT_LOG_INFO("[SocketTransport] Using TCP transport");
+  return std::make_unique<TcpSocketTransport>();
```
lmuravljovTT, line 25: Maybe we should check whether type is actually "zmq" or "tcp" , since a random value would still make TcpSocketTransport

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
```diff
@@ -0,0 +1,185 @@
...
+  running_ = false;
+  connected_ = false;
+
+  if (socket_) {
+    socket_->close();
+    socket_.reset();
+  }
+  if (context_) {
+    context_->close();
+  }
```
nanicicTT: This might be risky since one thread might be inside `socket_->send(<something>)` or recv at the same time as this. 
So you can get use-after-reset, maybe undefined behavior, or some random ZeroMQ exceptions. 
I think at minimum socket access should be protected by one mutex and every place that touches `socket_` should use it. 
Imagine like:

```
start()
sendRawData()
receiveRawData()
stop()
```
But the cleaner ZeroMQ design is still (from [the book](https://zguide.zeromq.org/)): **one socket owner thread.**

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
```diff
@@ -0,0 +1,185 @@
...
+std::string ZmqSocketTransport::getStatus() const {
+  if (!running_) return "stopped";
+  if (connected_) {
+    return mode_ == Mode::SERVER ? "server:connected" : "client:connected";
+  }
+  return mode_ == Mode::SERVER ? "server:waiting" : "client:connecting";
+}
+
+bool ZmqSocketTransport::sendRawData(const std::vector<uint8_t>& data) {
+  if (!running_ || !socket_) return false;
```
nanicicTT: If different threads call sendRawData(), receiveRawData(), start(), or stop(), these should be maybe
```cpp
std::atomic<bool> running_{false};
std::atomic<bool> connected_{false};
```

since these are read/written from multiple functions. One example is here. 
Right now, if this transport is used from multiple threads, those booleans are data races.

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
```diff
@@ -0,0 +1,185 @@
...
+    TT_LOG_ERROR("[ZmqSocketTransport] Send failed: {}", e.what());
+    return false;
+  }
+}
+
+std::vector<uint8_t> ZmqSocketTransport::receiveRawData() {
+  if (!running_ || !socket_) return {};
+
+  try {
+    if (mode_ == Mode::SERVER) {
```
lmuravljovTT: These seem to be a bit more different when Mode is Server vs Client than the tcp implementaton was.. 
Might be better to separate them?

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
```diff
@@ -0,0 +1,185 @@
...
+      if (!idResult.has_value()) return {};
+
+      // Store peer identity (update on every receive to stay current).
+      {
+        std::lock_guard<std::mutex> idLock(peerIdMutex_);
+        peerId_.assign(
+            static_cast<uint8_t*>(identity.data()),
+            static_cast<uint8_t*>(identity.data()) + identity.size());
+        if (!connected_) {
+          connected_ = true;
```
nanicicTT: I guess this does not mean

- server received hello
- server is alive
- router knows this client
 - network path is valid

just that dealer socket was created and `connect()` was called. Right?
With ZeroMQ, `connect()` does not necessarily mean the TCP connection is fully established. Messages may be queued locally. 
Connection issues are hard. We can, maybe, set `connected_ = true` only after receiving an **ACK**, heartbeat response, or valid message from the peer.

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
lmuravljovTT: should whole methods be guarded by mutex? 
`  if (!running_ || !socket_) return false;`
We could still get a race condition between two atomic value loads here right?

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+  if (mode_ == Mode::CLIENT) {
+    // DEALER: send an empty "hello" frame so the ROUTER learns our identity.
+    std::lock_guard<std::mutex> lock(sendMutex_);
+    try {
+      zmq::message_t hello(0);
+      socket_->send(hello, zmq::send_flags::dontwait);
+    } catch (...) {
+    }
+    connected_ = true;  // Client considers itself connected once socket is up.
```
nanicicTT: If send fails, the code ignores it and still marks the client connected. It is outside of the catch block. 
Maybe we can see result value what it is and dies it have it, also to catch specific `zmq::error_t& e` errors?

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
nanicicTT: in general, 
> But the cleaner ZeroMQ design is still (from the book): one socket owner thread.

I would say this is the right design altogether.

tt-media-server/cpp_server/CMakeLists.txt
idjuricTT: Thanks, done!

tt-media-server/cpp_server/src/sockets/zmq_socket_transport.cpp
idjuricTT: Fixed, thanks!!

ztorlakTT (approved):

nanicicTT (approved):

PR #3542, opened 2026-05-14
add: Terminal-Bench-2 and SWEBench-Verified agentic evaluations

ipastalTT:
## Summary

- Introduces a unified evals/agentic/ package that dispatches between two agentic eval backends, Harbor (terminal-bench) and mini-swe-agent (SWEBench) , from a single EVALS_AGENTIC venv
- Adds eval config for Qwen3.6-27B with two agentic tasks: terminal_bench_2 (published 59.3%) and swe_bench_verified (published 77.2%)
- Integrates agentic result.json processing into run_reports.py via a unified Harbor-shaped format shared by both backends
- Adds tests/agentic/ test suite covering terminal-bench config writing, SWEbench command building, and report parsing

## Structure

- evals/agentic/terminal_bench.py: Harbor invocation and config generation
- evals/agentic/swebench.py: mini-swe-agent + SWE-bench harness runner
- evals/agentic/run_agentic_eval.py: subprocess entry point with terminal-bench / swebench subcommands
- evals/agentic/report.py: result parsing for both backends

## Work in Progress

- Select specific tasks for the CLI nightly runs to reduce time limit
- Test with TT Qwen-3.6 implementation (?)
- Remove code that is not needed in normal mode.

## Test plan

-  Unit tests pass: pytest tests/agentic/ tests/test_run_evals.py 
-  Report generation verified against real result files from a vLLM-backed run 
-  Dedicated script executes end-to-end with smoke-test mode ( temporary code changes to smoke tests,both tasks skipped at 0 samples)

### Metadata: Qwen3.6-27B on gpu
```json
{
    "report_id": "id_nvidia_vllm_Qwen3.6-27B_gpu_2026-05-14_04-39-21",
    "model_name": "Qwen3.6-27B",
    "model_id": "id_nvidia_vllm_Qwen3.6-27B_gpu",
    "runtime_model_spec_json": "evals/dev_model_specs/qwen36_runtime_model_spec.json",
    "model_repo": "Qwen/Qwen3.6-27B",
    "model_impl": "nvidia-vllm",
    "inference_engine": "vLLM",
    "device": "gpu",
    "server_mode": "API",
    "tt_metal_commit": "n/a",
    "vllm_commit": "n/a",
}
```
### Acceptance Criteria

- Acceptance status: `FAIL`
- `benchmarks_summary.0.target_checks`: Missing target_checks in benchmark summary.
- `evals.terminal_bench_2`: Accuracy check failed; actual score=46.0674; reference published_score=59.3000; ratio=0.7769.
- `evals.swe_bench_verified`: Accuracy check failed; actual score=62.0000; reference published_score=77.2000; ratio=0.8031.

### Accuracy Evaluations for Qwen3.6-27B on gpu

| model       | device | task_name          | task_type | accuracy_check | score | ratio_to_reference | gpu_reference_score | ratio_to_published | published_score                                  |
|-------------|--------|--------------------|-----------|----------------|-------|--------------------|---------------------|--------------------|--------------------------------------------------|
| Qwen3.6-27B | gpu    | terminal_bench_2   | text      | FAIL ⛔         | 46.07 | N/A                | [None](TBD)         | 0.78               | [59.30](https://huggingface.co/Qwen/Qwen3.6-27B) |
| Qwen3.6-27B | gpu    | swe_bench_verified | text      | FAIL ⛔         | 62.00 | N/A                | [None](TBD)         | 0.80               | [77.20](https://huggingface.co/Qwen/Qwen3.6-27B) |

Note: The ratio to published scores defines if eval ran roughly correctly, as the exact methodology of the model publisher cannot always be reproduced. For this reason the accuracy check is based first on being equivalent to the GPU reference within a +/- tolerance. If a value GPU reference is not available, the accuracy check is based on the direct ratio to the published score.


workflows/run_reports.py
```diff
@@ -2640,6 +2651,16 @@ def evals_generate_report(args, server_mode, model_spec, report_id, metadata={})
             f"{get_default_workflow_root_log_dir()}/evals_output/{file_name_pattern}"
         )
         files = glob(file_path_pattern)
+        agentic_file_name_pattern = generate_agentic_report_data(
```
bgoelTT: I think this function should be renamed to reflect the fact it's really generating a file path pattern for agentic evals results

evals/agentic/report.py
```diff
@@ -0,0 +1,181 @@
...
+from typing import Any, Optional
+
+logger = logging.getLogger(__name__)
+
+
+def generate_agentic_report_data(model_spec: Any, eval_run_id: str) -> str:
+    """Generate a file path pattern to locate agentic eval result files for the
+    given evaluation run ID."""
+    # TODO: discuss this -> should we enforce similar structure to other evals
+    #  from llms-eval for example?
```
bgoelTT, line 19: Yes I think this is a good idea in general. I don't think we should address this now as @ddjukicTT @fivanovicTT are massively refactoring the generation of reports

requirements/evals-agentic.txt
```diff
@@ -0,0 +1,14 @@
...
+# Dependencies for the EVALS_AGENTIC workflow venv.
+# Used by: setup_evals_agentic() in workflows/workflow_venvs.py
+#
+# SWE-agent is NOT listed here — it is cloned from GitHub and installed as an
+# editable package by the setup_evals_agentic() hook, so that its config
+# directory is available at a known path on disk.
+
+harbor==0.6.5
+mini-swe-agent
+git+https://github.com/epoch-research/SWE-bench.git
```
bgoelTT: Can you also please pin the version of `mini-swe-agent` that you've been using? This will help preventing version drift which introduce breaking CLI changes, etc

Could the same also be done for the SWE-bench package?

evals/run_evals.py
```diff
@@ -79,12 +76,31 @@ def _select_eval_config(
     if limit_mode != EvalLimitMode.SMOKE_TEST or not eval_config.tasks:
         return eval_config
 
-    selected_task = eval_config.tasks[0]
-    logger.info(
-        "Smoke-test mode enabled; running only first eval task: %s",
-        selected_task.task_name,
-    )
-    return EvalConfig(hf_model_repo=eval_config.hf_model_repo, tasks=[selected_task])
+    # TODO: revert this logic later
```
bgoelTT: Does this logic still need to be reverted in this PR? If so, please do and if not, then please file a follow-up issue and paste the link here in this comment

evals/eval_config.py
```diff
@@ -26,6 +26,52 @@ class EvalTaskScore:
     tolerance: float = 0.05
 
 
+@dataclass(frozen=True)
+class AgenticEvalConfig:
```
bgoelTT: This config appears to be specific to terminal-bench-2, is there a reason not to name it so just like you did for the `SWEbenchEvalConfig` below?

evals/run_evals.py
```diff
@@ -16,9 +16,6 @@
 project_root = Path(__file__).parent.parent
 sys.path.insert(0, str(project_root))
 
-from utils.media_clients.base_strategy_interface import BaseMediaStrategy
-from utils.media_clients.media_client_factory import MediaClientFactory, MediaTaskType
-
```
bgoelTT, line 21: Why do these imports need to be moved into local imports?

evals/eval_config.py
```diff
@@ -91,6 +139,116 @@ class EvalConfig:
 
 
 _eval_config_list = [
+    # Leaving these here for now
+    # TODO: Move to a more appropriate place?
```
bgoelTT: I think these are in the correct place for now

evals/eval_config.py
```diff
@@ -3144,3 +3302,8 @@ class EvalConfig:
     for _, model_spec in MODEL_SPECS.items()
     if model_spec.hf_model_repo in _eval_config_map
 }
+# TODO: remove this when we have eval configs for all model specs, but in the meantime,
+# for dummy testing, we want to include all eval configs
+# even if they don't have a corresponding MODEL_SPEC
+for eval_config in _eval_config_list:
+    EVAL_CONFIGS.setdefault(eval_config.hf_model_repo.split("/")[-1], eval_config)
```
bgoelTT: I believe this was just used for testing, can this be removed now?

evals/agentic/report.py
```diff
@@ -0,0 +1,181 @@
...
+    return f"eval_{eval_run_id}/agentic/*/result.json"
+
+
+def is_harbor_result(data: Any) -> bool:
+    ## is this future proof though?
```
bgoelTT: Hmm probably not, a pattern @lukezTT implemented in the structured outputs benchmarking was to actually read back in the result `.json` after the eval completed and write a custom metadata field that was then read in a similar place as this https://github.com/tenstorrent/tt-inference-server/blob/18ec1547cfdaf970fbd33b4250c39447862daab7/benchmarking/run_benchmarks.py#L255

evals/eval_config.py
```diff
@@ -91,6 +139,116 @@ class EvalConfig:
...
+                    },
+                ),
+                swebench_eval_config=SWEbenchEvalConfig(
+                    dataset_name="SWE-bench/SWE-bench_Verified",
+                    sweagent_subset="verified",
+                    # we will need to specify specific tasks
+                    # for CI runs to keep runtime reasonable
+                    dataset_split="test",
+                    # that is the preferred agent for evaluation,
+                    # but we can also run with the default sweagent
```
bgoelTT: Can you please explain this preference? Also, should we keep the codepaths alive in `evals/agentic/` to allow running with the `sweagent` backend?

bgoelTT (changes requested):
A few minor comments after my first pass, looking really good

evals/eval_config.py
ipastalTT: SWEBench states mini agent is the preferred agent, sweagent is stale as well. We can remove support for sweagent if we want to

evals/eval_config.py
ipastalTT: I think we can, but we will have to wait for model spec updates to run agentic tasks

requirements/evals-agentic.txt
ipastalTT: Sure.
Just to note we are using a SWE-bench repo that has common docker layers, reducing size from about 200GB to 30GB

evals/eval_config.py
ipastalTT: Yeap, I should change that. The idea was to have a common config for both tasks initially, but their fields differ quite a lot

evals/eval_config.py
ipastalTT: from swe-bench-agent [link](https://github.com/SWE-agent/mini-swe-agent)

Should I use SWE-agent or mini-SWE-agent?
You should consider mini-swe-agent your default choice. In particular, you should use mini-swe-agent if

You want a quick command line tool that works locally
You want an agent with a very simple control flow
You want even faster, simpler & more stable sandboxing & benchmark evaluations
You are doing FT or RL and don't want to overfit to a specific agent scaffold
You should use swe-agent if

You want to experiment with different sets of tools, each with their own interface
You want to experiment with different history processors
What you get with both

Excellent performance on SWE-Bench
A trajectory browser

evals/run_evals.py
ipastalTT: LLM hallucinations, I was skeptical myself why this was needed...it imports correctly with the correct venv. Reverted

evals/agentic/report.py
ipastalTT: added "_result_format" : "harbor" in agentic results.

evals/eval_config.py
ipastalTT: removed

evals/eval_config.py
bgoelTT: Okay makes sense. Let's just keep the code for swe-agent but leave the default agent to be mini

evals/agentic/terminal_bench.py
```diff
@@ -100,6 +100,18 @@ def _write_harbor_config(config: TerminalBenchRunConfig) -> Path:
...
+def _annotate_result_file(result_file: Path) -> None:
+    try:
+        with result_file.open("r", encoding="utf-8") as f:
+            data = json.load(f)
+        if isinstance(data, dict) and "_result_format" not in data:
+            data["_result_format"] = "harbor"
+            with result_file.open("w", encoding="utf-8") as f:
+                json.dump(data, f, indent=2)
+    except (json.JSONDecodeError, IOError) as e:
+        logger.warning("Could not annotate result file %s: %s", result_file, e)
```
bgoelTT: Instead of warning I think we should raise an exception as the presence of this field is required by `evals/agentic/report.py` for proper report processing

bgoelTT (approved):
Great work. Approving now so that you can see whether my last comment makes sense. If it does, please address and then merge at will

evals/agentic/terminal_bench.py
ipastalTT: I changed this to raise a RuntimeError instead.

Issue #3105, opened 2026-04-22
[Model Readiness Support]: gpt-oss-120b on QB2


bgoelTT:
### Model Name

openai/gpt-oss-120b

### Model URL

https://huggingface.co/openai/gpt-oss-120b

### Model Tenstorrent code implementation

https://github.com/tenstorrent/tt-metal/tree/main/models/demos/gpt_oss

### What accuracy evals are needed?

Same evals from #1322 

### What hardware is supported?

BH QB2

### Model Readiness Support Preconditions

- [x] Tenstorrent code implementation has all runtime preconditions and configurations documented.
- [x] vLLM or other inference server integration complete.
- [ ] [Non-blocking] Performance targets (Top Perf) set in benchmarking/benchmark_targets/model_performance_reference.json.
- [ ] [Non-blocking] Required evals are implemented and integrated for model in evals/eval_config.py.
- [ ] [Non-blocking] GPU accuracy reference scores collected and set for model in evals/eval_config.py.

### Model Readiness Support Checklist

- [ ] Inference server can serve model via run.py server workflow.
- [ ] Inference server can run release workflow. Release markdown report for target hardware pasted in comment below.
- [ ] Specific inference server docs cover how to serve model and run release workflow.
- [ ] Model is added to top level README table.
- [ ] Model is added to tt-shield Model Readiness CI nightly run producing release report data.

arobergeTT:
Models CI dispatch run failed: https://github.com/tenstorrent/tt-shield/actions/runs/24865001505/job/72803162555#step:21:324

ERROR: 
```
2026-04-24 06:59:54.469 | critical |          Always | TT_FATAL: Number of blocks exceeds number of cores available: 23 blocks > 12 cores (assert.hpp:104)
```

----------

              MODEL CONFIG                  MESH CONFIG                  HARDWARE
              ─────────────                 ───────────                  ────────
              intermediate_size              MESH_DEVICE                Tile = 32×32
                = 2880                      = "P300x2"
                  │                           │
                  │                  vLLM worker maps to:
                  │                  mesh_device.shape = (1, 4)
                  │                           │
                  │                  GPT-OSS code does:
                  │                  TP = mesh_device.shape[1] = 4
                  │                           │
                  ▼                           │
              ┌─────────────────────┐         │
              │ Each chip's slice   │ ◀───────┘
              │ of intermediate dim │         intermediate_per_device
              │   = 2880 / TP       │           = 2880 / 4 = 720
[...]

PR #3548, opened 2026-05-14
Enable gpt-oss-120b on P300x2 (BH QB2)

arobergeTT:
## Summary

Enables running `openai/gpt-oss-120b` on a Tenstorrent P300x2 (Blackhole QB2) device via the local vLLM server + `lm-eval-harness` flow. Ref: https://github.com/tenstorrent/tt-inference-server/issues/3105. 

- `workflows/model_spec.py`: appends a `ModelSpecTemplate` for `openai/gpt-oss-120b` on `DeviceTypes.P300X2`
  - `tt_metal_commit=0cc6a39` (`main`: `handrews/gpt-oss-p150` branch merged to `main`), `vllm_commit=1abfcfc`
  - `max_context=128*1024`, `max_concurrency=1`, `MESH_DEVICE=(1,4)`, `p300_x2_mesh_graph_descriptor.textproto`
  - Harmony parsing: `reasoning_parser_name=openai_gptoss`, `tool_call_parser_name=openai`
- `evals/eval_config.py`: tunes the existing `gpt-oss-120b` tasks for P300x2
  - `aime25`, `gpqa_diamond_cot_zeroshot`: `max_concurrent=1`, `timeout=14400`, `stream=false`, `max_gen_toks=120*1024`
  - `mmlu_generative`: `stream=false` (reasoning_effort stays `low`, `max_gen_toks=64*1024`)
  - Rationale (in inline comments): vLLM launches with `max_num_seqs=1` on P300x2 so parallel client requests time out in the queue; lm-eval's SSE consumer only parses `/v1/completions`, not `/v1/chat/completions`; `max_gen_toks` must stay strictly below `max_context=131072` to avoid 1-token prefills + empty Harmony responses.
- `benchmarking/benchmark_targets/model_performance_reference.json`: adds a `p300x2` theoretical perf target under `gpt-oss-120b` (TTFT 51 ms, Tput/user 27 t/s).

No changes to other models or shared infra. Sibling PRs scoped separately:
- generic CI auth / vLLM client+server fixes -> `adam/fix-broken-dotenv` WIP
- `--eval-samples` passthrough -> `adam/eval-samples-passthrough` (PR: https://github.com/tenstorrent/tt-inference-server/pull/3524)


arobergeTT:
**Blocking**: In Models CI dispatch run, evals and benchmarks both generate 0 output tokens.
- evals: https://github.com/tenstorrent/tt-shield/actions/runs/25833254468/job/75907825395#step:11:4901
- benchmarks: https://github.com/tenstorrent/tt-shield/actions/runs/25833254468/job/75907825395#step:11:11087 
 
Both client tools have the same anti-pattern: they treat the HTTP response as "successful" if status is 200 or if the body is parseable JSON, and they swallow the missing choices key.

#### Why this surfaced as 0% instead of as a loud crash

1. Evals

For lm-eval (`lm_eval/models/openai_completions.py::parse_generations`):
```
WARNING [models.openai_completions:224] Could not parse generations: 'choices'
```
That warning is from a `try/except KeyError`. The except substitutes an empty string "" for every model output. lm-eval then scores "" against the gold answer (which is never ""), and you get `exact_match = 0.0` on every doc. Six warnings → six zero-scored docs → aime25 reports 0.0 with the run looking otherwise healthy (same story for the other evals as well).

2. Benchmarks 

For the vLLM benchmark tool (vllm bench serve):
```
Successful requests:        2
Total generated tokens:     0
Benchmark duration (s):     0.00
Total token throughput (tok/s):  6171758.10
```
Same anti-pattern. An HTTP response with no choices is counted as a "successful request" but contributes 0 generated tokens. The benchmark divides total-tokens by duration to compute throughput, and because both the duration and the token count are absurdly small/zero, you get a nonsense Total token throughput like 6.1 M tok/s, 0.00 TTFT, and Total generated tokens: 0.

#### How /health checks gave a false green

The only auth-related gate for evals/benchmarks is `prompt_client.wait_for_healthy()`, which calls GET `/health`. vLLM's `/health` endpoint does not require authentication. So a server that rejects every authenticated POST with 401 still reports itself "healthy", and the workflow proceeds straight into the silent failure path. 

```
Stale GHA runner env  +  .env missing JWT_SECRET
            │
            ▼
Host signs bearer B          Container validates against bearer A
            │                              │
            └───────── 401 Unauthorized ◄──┘
                       (no `choices` in body)
                              │
              ┌───────────────┴───────────────┐
              ▼                               ▼
    lm-eval: KeyError swallowed     vLLM bench: counts as
    → "" output for every doc       "successful request" with
    → exact_match = 0.0             0 generated tokens
              │                               │
              │                               ▼
              │                     mean_tps written as "N/A"
              │                               │
              │                               ▼
              │                     run_reports.py:1928
              │                     "N/A" / float → TypeError
              ▼
     0% on every task
```

What we actually see: 
```
Server (vLLM 1abfcfc, harmony parser regression)
            │
   harmony Rust error bubbles up past
   chat_completion_full_generator's narrow except
            │
            ▼
   FastAPI route handler turns the exception
   into an ErrorResponse and returns it with
            ▼
        HTTP 200    body = {"error": "...", ...}    (no "choices")
            │
  ┌─────────┴─────────┐
  ▼                   ▼
lm-eval:             vllm bench serve:
parse_generations    counts the 200 as "successful",
hits KeyError(       reads response.usage which is
'choices') →         empty/missing → records 0 tokens →
catches it,          benchmark JSON has 0 generated tokens
substitutes "" →     and duration ≈ 0.00
exact_match=0.0
```

bgoelTT (changes requested):
Please also add an entry for `p300x2` into the `openai/gpt-oss-120b` object in `.github/models-ci-config.json`

arobergeTT:
@bgoelTT Added `P300x2` for `gpt-oss-120b` in `.github/models-ci-config.json`

workflows/model_spec.py
```diff
@@ -1111,6 +1111,42 @@ def expand_to_specs(self) -> List["ModelSpec"]:
             },
         },
     ),
+    ModelSpecTemplate(
+        weights=["openai/gpt-oss-120b"],
+        impl=gpt_oss_impl,
+        version="0.12.0",
```
bgoelTT: Do not set the `version` field yet as this model has not been officially released

evals/eval_config.py
bgoelTT, line 1: We shouldn't globally change the max_concurrent in this eval_config. Instead @kmabeeTT will be introducing this change to dynamically clamp the max_concurrent based on the ModelSpec https://github.com/tenstorrent/tt-inference-server/pull/3627

bgoelTT (changes requested):

evals/eval_config.py
kmabeeTT: (That PR is merged now, btw)

arobergeTT:
tt-metal issue created: [[gpt-oss-120b]: Decode hangs on first iteration after paged_fill_cache (isl ≥ 1024) on P300x2 #45052](https://github.com/tenstorrent/tt-metal/issues/45052)

arobergeTT:
@bgoelTT I resolved both comments and it is now ready to be merged.

evals/eval_config.py
bgoelTT, line 1: I am still seeing some modifications to the max_concurrent by the way

bgoelTT (changes requested):

evals/eval_config.py
arobergeTT: Fixed!

bgoelTT: Sorry I should have been specific, can you please leave the default `max_concurrency=16`? This is so that the galaxy evals do not suffer repeated KV cache pre-emption

arobergeTT: Reverted aime25 and gpqa to max_concurrency=16

bgoelTT (approved):

PR #3391, opened 2026-05-06
Migrate image models to cpp_server (SDXL)

ztorlakTT:
Migrates SDXL image models (text-to-image, image-to-image, edit/inpaint) from the Python `tt-media-server` into the C++ `cpp_server`, behind the same OpenAI-compatible endpoints. Selected at runtime via `MODEL_SERVICE=image` and `MODEL_RUNNER_TYPE=tt_sdxl_{generate,image_to_image,edit}`.

- **New endpoints** (controller `ImageController`): `POST /v1/images/generations`, `POST /v1/images/image-to-image`, `POST /v1/images/edits`. Only the route matching the active `MODEL_RUNNER_TYPE` is registered.
- **Runner family for in-process media work**: `IMediaRunner<Request, Response>` alongside the existing IPC-loop `IRunner`, sharing a common `IRunnerBase` (warmup/stop/runnerType). SDXL pipelines are driven from C++ via pybind11, with all Python access guarded by the GIL.
- **Single `RunnerRegistry`** that handles both families: `registerIpcRunner` / `createIpc` for worker-process runners (LLM, embedding) with MOCK fallback, and `registerMediaRunner` / `createMedia<Runner>()` for direct-call media runners with type-safe downcasting.
- **Shared media config** (`MediaRunnerConfigBase`): `max_batch_size`, `device_mesh_shape`, `is_galaxy`, `model_weights_path`, `weights_distribution_timeout_seconds`. Read once via `readMediaRunnerConfig`, ready for upcoming audio/TTS/video runners.
- **Generic sync dispatch helper** (`dispatchJsonRequest`) used by `ImageController` and reusable by future media controllers — handles JSON validation, model-readiness, request parsing, thread-pool offload, and OpenAI-style error mapping (400 / 429 / 500 / 503).
- **Plumbing cleanup**: `runner_factory` renamed to `ipc_runner_factory` to reflect its IPC-only role; `runner_interface.hpp` → `ipc_runner.hpp`; `IService` lifecycle hooks (`warmup`, `stop`, `runnerType`) standardized on `IRunnerBase`. `RouteRegistry` only carries routes for the active service.

CI runs:
- SDXL N150: https://github.com/tenstorrent/tt-shield/actions/runs/25909766416
- SDXL P300x2: https://github.com/tenstorrent/tt-shield/actions/runs/25909734330
- SDXL galaxy: https://github.com/tenstorrent/tt-shield/actions/runs/25909609318
[...]

PR #3551, opened 2026-05-14
fix: wire SDXL runners into BH mesh-descriptor set + uplift SDXL-base tt_metal to 6900b0c

### LINKED ISSUE #3550: SDXL runners missing from `_RUNNERS_REQUIRING_MESH_DESCRIPTOR` — silently breaks SDXL init on Blackhole
## Gap

`tt-media-server/utils/runner_utils.py` keeps a `_RUNNERS_REQUIRING_MESH_DESCRIPTOR` set that decides which runner types get `TT_MESH_GRAPH_DESC_PATH` auto-exported at container start. Whisper and SpeechT5 are in the set. **The three SDXL runners (`TT_SDXL_TRACE`, `TT_SDXL_IMAGE_TO_IMAGE`, `TT_SDXL_EDIT`) are not.**

## Symptom

The moment SDXL's pinned `tt_metal_commit` becomes one that asserts on the custom fabric mesh graph desc path, SDXL device init on every Blackhole device type (P150 / P300 / P150X4 / P150X8) fails with:

```
TT_FATAL @ tt_cluster.cpp:273: this->rtoptions_.is_custom_fabric_mesh_graph_desc_path_specified()
```

There is no per-device-spec workaround for this in `workflows/model_spec.py` that would have been correct — the canonical wiring point is the runner-type set.

## Ask

- Add the SDXL runners to `_RUNNERS_REQUIRING_MESH_DESCRIPTOR` so they pick up the same auto-export Whisper/SpeechT5 already get.
- Treat additions to that set as the canonical wiring point so future media-server runners don't relearn this lesson on a customer demo.

stisiTT:
## Summary

- Add `TT_SDXL_TRACE`, `TT_SDXL_IMAGE_TO_IMAGE`, and `TT_SDXL_EDIT` to `_RUNNERS_REQUIRING_MESH_DESCRIPTOR` in `tt-media-server/utils/runner_utils.py` so `TT_MESH_GRAPH_DESC_PATH` is auto-exported at container start on Blackhole devices — matching the existing wiring for Whisper and SpeechT5.
- Bump SDXL-base `tt_metal_commit` from `bac8b34` to `6900b0c` in `workflows/model_spec.py`.

## Why the uplift

`bac8b34` hits `18 is not a valid ClusterType` on single-chip P300 boards (the bh-qb-ge hardware) — the Python `ClusterType` enum at that commit doesn't include the CUSTOM type the UMD now assigns to single-chip P300 boards. `6900b0c` is well past those tt-cluster fixes and is the commit the rest of the BH QB2 stack is already on. The runner_utils.py change is what makes `6900b0c` usable for SDXL (it now requires `TT_MESH_GRAPH_DESC_PATH`).

Closes #3550

## Test plan

- [x] tt-shield `on-dispatch.yml` run, SDXL-base on bh-qb-ge / p150 with tt_metal `6900b0c`: https://github.com/tenstorrent/tt-shield/actions/runs/25888128635


workflows/model_spec.py
```diff
@@ -2913,7 +2913,7 @@ def expand_to_specs(self) -> List["ModelSpec"]:
             "stabilityai/stable-diffusion-xl-base-1.0-img-2-img",
         ],
         version="0.11.1",
-        tt_metal_commit="bac8b34",
+        tt_metal_commit="6900b0c",
```
bgoelTT, line 2916: We should not be manually adjusting the tt-metal versions pinned in the ModelSpec as they should be adjusted during releases or when new model integration is performed

bgoelTT (changes requested):
The changes to everything but the model_spec.py are good

tt-media-server/utils/runner_utils.py
```diff
@@ -52,6 +52,9 @@ def setup_runner_environment(
     _RUNNERS_REQUIRING_MESH_DESCRIPTOR = {
         ModelRunners.TT_WHISPER.value,
         ModelRunners.TT_SPEECHT5_TTS.value,
+        ModelRunners.TT_SDXL_TRACE.value,
```
ztorlakTT, line 55: fyi: SDXL models are migrated to the cpp server (https://github.com/tenstorrent/tt-inference-server/pull/3391)

PR #3594, opened 2026-05-18
vLLD Dockerfile change and new plugin-config vLLM wrapper for "override-tt-config"

### LINKED ISSUE #3535: Fix vLLM build in Dockerfile by setting VLLM_TARGET_DEVICE=empty and installing plugin[runtime]
## Errors:

All `tt-shield` jobs targeting the `dev` branch of the `vllm` repo will fail because of the recently introduced change in `vllm` repo:  **"Fully separate TT code to vllm_tt_plugin"**
https://github.com/tenstorrent/vllm/commit/ae0f0730bdc25d770071dca870a3876869a1a633

This causes the following errors in the vLLM  dockerfile build:

The container has `VLLM_TARGET_DEVICE=tt`(Dockerfile line 31), and vllm's setup.py no longer recognizes that value.

tt is not set as VLLM_TARGET_DEVICE anymore:  `raise ValueError("Unsupported platform, please use CUDA, ROCm, or CPU).`

  ## Description
  
  The **vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile** fails to build the vLLM + Tenstorrent plugin stack. Two related issues:

  1. vLLM build attempts device-specific compilation

  The current install command runs:
`  uv pip install --index-strategy unsafe-best-match -e . --extra-index-url https://download.pytorch.org/whl/cpu`
  
  Without VLLM_TARGET_DEVICE set, vLLM's setup.py defaults to building for CUDA (or auto-detects an accelerator), which is wrong inside this container — the actual device backend is provided out-of-tree by vllm-tt-plugin against tt-metal. This causes the build to either fail or to pull in/compile unwanted CUDA-specific code paths.

  2. vllm-tt-plugin is installed with --no-deps
  
  The plugin was previously installed as:
`  uv pip install --no-deps -e plugins/vllm-tt-plugin`
  
  `--no-deps` skips the plugin's runtime dependencies entirely, so the image builds but runtime imports fail (or silently fall back) when the plugin tries to use libraries it declares in its [runtime] extra. The plugin also needs the same `--index-strategy  unsafe-best-match` and pytorch CPU `--extra-index-url` as vLLM itself, otherwise dependency resolution picks the wrong wheels (e.g., a CUDA torch build).

 ## Tried fix
  
  I have updated the RUN step to:

  1. Set `VLLM_TARGET_DEVICE=empty` for the vLLM editable install so vLLM builds without a device backend and defers device support to the plugin.
  2. Install vllm-tt-plugin with its [runtime] extras, using the same `--index-strategy unsafe-best-match` and `--extra-index-url https://download.pytorch.org/whl/cpu` flags as the vLLM install, instead of `--no-deps`.

Change done on my `tt-inference-server` branch

`https://github.com/tenstorrent/tt-inference-server/compare/main...vcankovic/vllm-dockerfile-changes`

 ## Failure: unrecognized arguments: --override-tt-config

I tried on-dispatch job for Llama and build passed.

https://github.com/tenstorrent/tt-shield/actions/runs/25848363779/job/75953950781

  - I have managed to build dockerfile cleanly without attempting CUDA-target compilation.
  - vllm-tt-plugin's runtime dependencies are resolved against the CPU pytorch index, fixing missing-dependency errors at runtime.

However the failure occured in "Run tests" step:
`run_vllm_api_server.py: error: unrecognized arguments: --override-tt-config`

In this way we are not able to pass arguments, like `trace_region_size` any more

vcankovicTT:
closes #3535 

## Description

  This PR introduces a `plugin_config` wrapping around TT-specific overrides under "tt", replacing the TT-fork-specific
  `--override-tt-config` vLLM argument with the vLLM `--plugin-config` interface.

  The internal representation of TT overrides (`override_tt_config` fields in all `ModelSpecTemplate` declarations) is unchanged. Only the plugin reading layer are adapted to the new vLLM interface.

## Error
  
  CI job: Llama-3.1-8B-Instruct (n150) https://github.com/tenstorrent/tt-shield/actions/runs/25848363779/job/75953950781
  `run_vllm_api_server.py: error: unrecognized arguments: --override-tt-config`

## Root cause: 

`model_spec.py` reads the TT override config into `vllm_args` under the key `override_tt_config`. This argument existed only in the `tenstorrent/vllm fork`. The new plugin mode does not recognize `--override-tt-config`, causing the server to fail immediately on startup.

## Introduced Changes

The following scripts/files are changed.

### Dockerfile "vllm.tt-metal.src.dev.Dockerfile" 

changes  confirmed by Viktor Pus. All code changes are separating all TT-specific code to a plugin.

```
&& VLLM_TARGET_DEVICE=empty uv pip install --index-strategy unsafe-best-match -e . --extra-index-url https://download.pytorch.org/whl/cpu \
    && if [ -d plugins/vllm-tt-plugin ]; then uv pip install --index-strategy unsafe-best-match -e 'plugins/vllm-tt-plugin[runtime]' --extra-index-url https://download.pytorch.org/whl/cpu; fi \
```

### model_spec.py

 Replacing the `--override-tt-config` vLLM argument with the upstream vLLM `--plugin-config` argument, adding a "tt" key within it, so vLLM can identify it as plugin-specific configuration for Tenstorrent.

 We don't do simple rename as the shape of the variables changes: `--override-tt-config` expected a flat dict, `--plugin-config` expects a nested-dict, one level deeper.

  **Change 1 — DeviceModelSpec._infer_data():**
  Before
  `"override_tt_config": json.dumps(self.override_tt_config)`
  
  After
  `"plugin_config": json.dumps({"tt": self.override_tt_config})`
  
  **Change 2 — ModelSpec.apply_overrides():**
   Before
  `"override_tt_config": json.dumps(merged_override_config)`
  
  After
  `"plugin_config": json.dumps({"tt": merged_override_config})`

  ### workflows/multihost_orchestrator.py
 
 We update the read and write paths to use the new plugin_config key and {"tt": ...} instead of the old `override_tt_config` key.

```
  existing_plugin_str = vllm_args.get("plugin_config", '{"tt": {}}')

  existing_override = json.loads(existing_plugin_str).get("tt", {})

  spec_dict["device_model_spec"]["vllm_args"]["plugin_config"] = json.dumps({"tt": merged_override})
```

  ### tt-vllm-plugin/tt_vllm_plugin/platform.py -- to be double checked(?)
  
 This file is used for platform config initialization
  
  Purpose: `TTPlatform.check_and_update_config()` is called by vLLM early in startup before any worker is created. It configures scheduler, cache, worker class, and sampling behavior based on TT-specific settings. It was previously reading
  `override_tt_config` from a hardcoded empty dict.

  We need to check what we want to achieve: to read `model_config.plugin_config` under the "tt" key, so setting like `sample_on_device_mode` is actually applied.

 **Before** — always empty, config was never applied in plugin mode
  `override_tt_config = {}`
  
  **After** — reads actual TT config passed via --plugin-config
```
  override_tt_config = getattr(
      vllm_config.model_config, "plugin_config", {}
  ).get("tt", {})
```
  
  ### tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py -- to be double checked(?)
  
  This file is used for worker config initialization.

  Purpose: `TTWorker.__init__()` is called when vLLM creates the TT worker process. It reads TT-specific config to determine device parameters such as `trace region size`. All the device initialization parameters — were never being read from the model spec in this worker. The device was always being opened with defaults. 

  This fix is the first time tt_worker.py in this repo actually reads the config and passes it through to device initialization. Is this what we want to achieve?

  Change:
  **Before** — always empty dict, all device overrides silently ignored
  `self.model_config.override_tt_config = {}`
  
  **After** — populated from --plugin-config {"tt": {...}}
```
  self.model_config.override_tt_config = getattr(
      self.model_config, "plugin_config", {}
  ).get("tt", {})
```
  
  By setting self.model_config.override_tt_config here, all existing downstream code in tt_worker.py (utility functions get_dispatch_core_config, get_fabric_config, device_params_from_override_tt_config, etc.) and in tt_loader.py continues
  to work without modification — they all read from model_config.override_tt_config which is now correctly populated.
  
  
  ## Test
  
  https://github.com/tenstorrent/tt-shield/actions/runs/25983643325/job/76378926361
  
  The arguments were correct. Here's the proof from the log:
     
```
  vllm serve \
    --model meta-llama/Llama-3.1-8B-Instruct \
    --plugin_config '{"tt": {"sample_on_device_mode": "decode_only", "trace_region_size": 105000000}}' \
    --port 8000
```

  The job failed for a completely unrelated reason — the trace_region_size value in the model spec is too small for the actual hardware requirement:

```
  trace_region_size configured: 105,000,000 bytes
  trace_region_size required:   107,331,584 bytes
  shortage: ~2.3 MB
```

### release_model_spec.json - will not be updated manually

`release_model_spec.json` is re-generated using the script `scripts/release/update_model_spec.py` 

The new json file format will have the following structure after it is being regenerated (example for `Llama-3.1-8B-Instruct`):

```
              "vllm_args": {
                "model": "meta-llama/Llama-3.1-8B-Instruct",
                "block_size": "64",
                "max_model_len": "32768",
                "max_num_seqs": "32",
                "max_num_batched_tokens": "32768",
                "max-log-len": "32",
                "seed": "9472",
                "plugin_config": "{\"tt\": {\"trace_region_size\": 52000000}}"
              },
              "override_tt_config": {
                "trace_region_size": 52000000
              },

```

### DIFF
--- a/tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
+++ b/tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
@@ -51,7 +51,9 @@ class TTWorker(WorkerBase):
 
         # Initialized by init_device
         self.mesh_device = None
-        self.model_config.override_tt_config = {}
+        self.model_config.override_tt_config = getattr(
+            self.model_config, "plugin_config", {}
+        ).get("tt", {})
 
         # Whether to use ttnn tracing for model execution
         override_tt_config = self.model_config.override_tt_config
--- a/tt-vllm-plugin/tt_vllm_plugin/platform.py
+++ b/tt-vllm-plugin/tt_vllm_plugin/platform.py
@@ -100,9 +100,9 @@ class TTPlatform(Platform):
         # TODO move this to tt_model_runner when request validation
         # stops depending on vllm_config
 
-        override_tt_config = getattr(
-            vllm_config.model_config, "plugin_config", {}
-        ).get("tt", {})
+        override_tt_config = getattr(vllm_config.model_config, "plugin_config", {}).get(
+            "tt", {}
+        )
         if (
             override_tt_config is not None
             and "sample_on_device_mode" in override_tt_config
--- a/tt-vllm-plugin/tt_vllm_plugin/platform.py
+++ b/tt-vllm-plugin/tt_vllm_plugin/platform.py
@@ -100,7 +100,9 @@ class TTPlatform(Platform):
         # TODO move this to tt_model_runner when request validation
         # stops depending on vllm_config
 
-        override_tt_config = {}
+        override_tt_config = getattr(
+            vllm_config.model_config, "plugin_config", {}
+        ).get("tt", {})
         if (
             override_tt_config is not None
             and "sample_on_device_mode" in override_tt_config
--- a/vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile
+++ b/vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile
@@ -95,8 +95,8 @@ RUN /bin/bash -c "git clone https://github.com/tenstorrent/vllm.git ${vllm_dir} 
     && git checkout ${TT_VLLM_COMMIT_SHA_OR_TAG} \
     && source ${PYTHON_ENV_DIR}/bin/activate \
     && uv pip install --upgrade pip \
-    && uv pip install --index-strategy unsafe-best-match -e . --extra-index-url https://download.pytorch.org/whl/cpu \
-    && if [ -d plugins/vllm-tt-plugin ]; then uv pip install --no-deps -e plugins/vllm-tt-plugin; fi \  
+    && VLLM_TARGET_DEVICE=empty uv pip install --index-strategy unsafe-best-match -e . --extra-index-url https://download.pytorch.org/whl/cpu \
+    && if [ -d plugins/vllm-tt-plugin ]; then uv pip install --index-strategy unsafe-best-match -e 'plugins/vllm-tt-plugin[runtime]' --extra-index-url https://download.pytorch.org/whl/cpu; fi \
     && rm -rf ${vllm_dir}/.git"
 
 # Build tt-smi in separate venv to avoid conflicts with tt-metal venv
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -348,7 +348,7 @@ class DeviceModelSpec:
             "max_num_batched_tokens": str(self.max_context),
             "max-log-len": "32",
             "seed": "9472",
-            "override_tt_config": json.dumps(self.override_tt_config),
+            "plugin_config": json.dumps({"tt": self.override_tt_config}),
         }
         merged_vllm_args = {**default_vllm_args, **self.vllm_args}
         object.__setattr__(self, "vllm_args", merged_vllm_args)
@@ -776,7 +776,7 @@ class ModelSpec:
             )
             merged_vllm_args = {
                 **self.device_model_spec.vllm_args,
-                "override_tt_config": json.dumps(merged_override_config),
+                "plugin_config": json.dumps({"tt": merged_override_config}),
             }
             object.__setattr__(self.device_model_spec, "vllm_args", merged_vllm_args)
 
--- a/workflows/multihost_orchestrator.py
+++ b/workflows/multihost_orchestrator.py
@@ -602,10 +602,10 @@ class MultiHostOrchestrator:
         # Get serialized model spec dict
         spec_dict = self.model_spec.get_serialized_dict()
 
-        # Get the existing override_tt_config from vllm_args (it's a JSON string)
+        # Get the existing TT config from vllm_args plugin_config (it's a JSON string)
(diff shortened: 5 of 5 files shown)


tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
```diff
@@ -51,7 +51,9 @@ def __init__(
 
         # Initialized by init_device
         self.mesh_device = None
-        self.model_config.override_tt_config = {}
```
dmadicTT, line 54: `tt-vllm-plugin` is not really used, it was just a PoC.
I have to make this explicit in the readme.

@viktorpusTT is creating a proper plugin for vLLM.

dmadicTT (approved):
Looks good to me, but I barely worked with this code
I would also ask @viktorpusTT and @vpetrovicTT to review

viktorpusTT:
I'm not familiar with this code, all I can confirm is the new config format looks correct.

This change of config format was done in https://github.com/tenstorrent/vllm/tree/dev though. I haven't modified https://github.com/tenstorrent/tt-inference-server/tree/main/tt-vllm-plugin at all.

bgoelTT (approved):

PR #3604, opened 2026-05-18
Add Cancellation to Multiturn Chat Script

ljovanovicTT:
## Problem
We want to add cancellation support to the multiturn-chat script, with behavior aligned to standard HTTP client cancellation semantics

1. If a request is cancelled before prefill completes, the last user turn should be discarded from the messages object:
```
User: 'hi'
Assistant: ^C
[cancelled before first token — discarding last user turn]
```

2. If a request is cancelled during decoding, preserve the partial conversation history and stop processing any further tokens after the cancellation occurs:
```
User: 'hi'
Assistant:  latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter latter^C
[cancelled — saving partial reply to history]
```

Also, with this PR customizable number of max tokens was enabled per request, as follows:
```
User: -n 10 'hi'
[max_tokens = 10]
Assistant:  latter latter latter latter latter latter latter latter
```

## Testing

1. Start `cpp_server`
2. Start `cpp_server/multiturn-chat.sh` script
3. Use `Ctrl+C` as the cancellation mechanism - note that if it is triggered while waiting for user input it terminates the multi-turn session.

### DIFF
--- a/tt-media-server/cpp_server/multiturn-chat.sh
+++ b/tt-media-server/cpp_server/multiturn-chat.sh
@@ -41,10 +41,20 @@ URL="http://localhost:${PORT}/v1/chat/completions"
 MESSAGES_JSON=$(jq -n --arg prompt "$PROMPT" '[{role: "user", content: $prompt}]')
 
 echo "Starting chat session. Type 'exit' or 'quit' to end."
+echo "Press Ctrl+C during the assistant's reply to cancel that turn"
+echo "  (partial reply is still saved to history)."
+echo "Prefix a user line with '-n NUM' to override max_tokens for that turn"
+echo "  (sticks for subsequent turns; current: $MAX_TOKENS)."
 echo "--------------------------------------------------"
 
 # pipefail so a curl/jq failure inside the $(...) capture below shows up in $?.
 set -o pipefail
+
+# Ctrl+C during the streaming pipeline cancels the current turn instead of
+# killing the script. The handler just sets a flag; SIGINT also reaches the
+# subshell's curl/jq/tee, which die normally and end the pipeline.
+CANCELLED=0
+on_cancel() { CANCELLED=1; }
 
 while true; do
   BODY=$(jq -n \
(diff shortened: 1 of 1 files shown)


tt-media-server/cpp_server/multiturn-chat.sh
```diff
@@ -102,6 +133,17 @@ while true; do
     echo "Goodbye!"
     break
   fi
+
+  # Allow per-turn `-n NUM` override at the start of the line. This is a
+  # one-shot: it applies only to the very next request, then $MAX_TOKENS (the
+  # launch default) takes over again. The rest of the line, if any, is the
+  # actual user prompt.
+  if [[ "$NEXT_PROMPT" =~ ^-n[[:space:]]+([0-9]+)([[:space:]]+(.*))?$ ]]; then
```
nanicicTT, line 141: Nice :) 
Cool way to check this is from terminal (no need for regex101):

```sh
echo "-n 5 3 bla" | grep -E '^-n[[:space:]]+([0-9]+)([[:space:]]+(.*))?$'
```

and it should work. 

Image

nanicicTT: you can resolve the comment, it was just interesting for me to share.

idjuricTT (approved):

knovokmetTT (approved):

PR #3606, opened 2026-05-18
Tool calling integration test

lmuravljovTT:
## Summary

  Add tool calling integration tests with end-to-end verification of the tokenizer and tool_call_parser
  pipeline.

  ### Changes

  - **New integration test suite** (`tool_calling_integration_test.cpp`)
    - Tests single and multiple tool calls parsing
    - Tests text content before tool calls
    - Tests complex JSON arguments (nested objects, arrays)
    - Tests realistic agentic tools (edit, read, exec, write, web_search)

  - **Test support infrastructure**
    - `MockToolCallRunner` - Streams predefined DeepSeek-format tool call tokens
    - `ToolCallRequest` - Fluent builder for tool-enabled chat completion requests
    - `ToolCallStream` - SSE response parser for tool_calls deltas
    
  - **Code organization**
    - Moved integration tests to `tests/integration/` directory


idjuricTT (approved):

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+  const auto stream = tt::test::ToolCallStream::parse(response);
+
+  EXPECT_TRUE(stream.endedWithDone());
+
+  ASSERT_TRUE(stream.hasToolCalls());
+  EXPECT_EQ(stream.toolCallFunctionName(0), "edit");
+
+  // Verify the edit arguments contain expected structure
+  const auto& args = stream.toolCallArguments(0);
+  EXPECT_NE(args.find("src/main.cpp"), std::string::npos);
```
dmadicTT, line 418: Easier to read
```
EXPECT_TRUE(args.contains("src/main.cpp")); 
```

It's in many other places, too

tt-media-server/cpp_server/tests/support/tool_call_stream.hpp
```diff
@@ -0,0 +1,160 @@
...
+      // Capture id and type from first delta for this index
+      if (tc.isMember("id") && !tc["id"].asString().empty()) {
+        pending.id = tc["id"].asString();
+      }
+      if (tc.isMember("type") && !tc["type"].asString().empty()) {
+        pending.type = tc["type"].asString();
+      }
+
+      // Capture function name and accumulate arguments
+      if (tc.isMember("function")) {
```
dmadicTT, line 134: Benefit of having `tool_call_stream` and `mock_tool_call_runner` is having super-readable tests.
This comes with a lot of logic (if-else statements) in this utility code.

Downside of this approach is tests can fail when someone makes a bug in this code, even thought the actual server logic under test did not change.
Or, tests can pass when they should fail in case `tool_call_stream` has a bug.

To fight against this we usually hardcode the inputs and the outputs.
Unfortunately this negaitvely impacts readability, but positive to reliability of the tests.

I think the way you designed the tests is great! And it's a huge step forward to what we have now.
I would not rush in changing this, but I wanted to make you aware what are the consequences of this approach.

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+  const auto response = tt::test::HttpResponse::parse(rawResponse);
+  EXPECT_EQ(response.statusCode(), 200);
+
+  const auto stream = tt::test::ToolCallStream::parse(response);
+
+  EXPECT_TRUE(stream.endedWithDone());
+
+  // Should have both content and tool call
+  EXPECT_FALSE(stream.content().empty()) << "Expected content before tool call";
```
dmadicTT: I guess one of those 2 is enough.
contains "check the weather"

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+  // Verify complex JSON is preserved
+  const auto& args = stream.toolCallArguments(0);
+  EXPECT_NE(args.find("/src"), std::string::npos);
+  EXPECT_NE(args.find("*.cpp"), std::string::npos);
+  EXPECT_NE(args.find("true"), std::string::npos);
+  EXPECT_NE(args.find("10"), std::string::npos);
+}
+
+// Test with realistic agentic tools (edit, exec, read, write, web_search)
```
dmadicTT, line 243: Great!

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+                    "type": "string",
+                    "description": "Filter by time: day, week, month, or year"
+                  }
+                },
+                "required": ["query"]
+              }
+            }
+          })json")
+          .toolChoice("auto")
+          .maxTokens(256)
```
dmadicTT: Is `maxTokens` important for the test?

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
dmadicTT, line 130: I know it's not scope of this work, but how come we need a `mutex` here?

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
```diff
@@ -264,21 +280,57 @@ class DeepSeekToolCallParser : public IToolCallParser {
           return std::nullopt;
 
         } else {
+          // Skip "json" marker and leading whitespace if coming after "```"
+          if (state.skip_json_marker) {
```
dmadicTT: The method `processToken` is getting huge. 
If you are confident in the tests we have in `test_tool_call_parser`, we should be breaking it down into smaller methods for readability.

tt-media-server/cpp_server/tests/support/mock_tool_call_runner.hpp
```diff
@@ -0,0 +1,151 @@
...
+      tok.token_index = static_cast<uint32_t>(i);
+      tok.token_id = tokens_[i];
+      tok.flags = (i == tokens_.size() - 1) ? ipc::SharedToken::FLAG_FINAL : 0;
+      queue_.push(tok);
+    }
+    tokens_.clear();
+  }
+
+  // Debug: print tokens and their decoded text
+  void debugPrint() const {
```
dmadicTT: Do we still need this?

dmadicTT (approved):

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
lmuravljovTT: hmm string.contains doesn't seem to be added until C++23, is it okay if we give it another bump? 
cc: @nanicicTT

nanicicTT: Yep, you are right, we can use it for maps, unordered maps and ranges in general, but surprisingly it is only added for strings in CPP 23. 
As for a bump, if not an issue, I always prefer latest and greatest compiler :) 
Probably some docker files need to changed and infra team probably needs to do that.

nanicicTT: https://www.linkedin.com/pulse/c23-new-features-sona-badalyan-qw9se

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
lmuravljovTT: similar to the reasoning_parser, we have one shared parser in the LLMService

multiple worker threads can call `processToken` for different tasks at the same time, grabbing the current parser state from `taskStates` map, overriding the ToolCallTaskState for one or the other

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
dmadicTT: Thanks for educating me! :)

lmuravljovTT: maybe we can do this separately then, and clean up other tests once we bump cpp :D

tt-media-server/cpp_server/tests/support/tool_call_stream.hpp
lmuravljovTT: Discussed a bit on slack, but just for the sake of it I'll add some comments here 

- reliability and readability are both important and we should balance them out when possible 
- making bug in the test code is always a possibility, and for that we should pay more attention to tests while reviewing
- this is an okay start for tool testing, and will be adapted as needed in the future 
- we could consider other ways to improve readability ( ex. plain Given/ when/ then or similar commenting format)

tt-media-server/cpp_server/tests/integration/tool_calling_integration_test.cpp
nanicicTT: @knovokmetTT told me metal is stuck on c++20 and it might not be a small feat to bump the version up.

PR #3611, opened 2026-05-19
Helm set updateStrategy to Recreate and introduced podAntiAffinity

vcankovicTT:


charts/tt-inference-server/values.yaml
```diff
@@ -24,6 +24,8 @@ defaults:
   serverType: vllm
   replicaCount: 1
   progressDeadlineSeconds: 3600
+  updateStrategy:
+    type: Recreate
```
mjeremicTT: I would move those out of `values.yaml` and add it directly to `deployment.yaml`

charts/tt-inference-server/values.yaml
```diff
@@ -54,7 +56,13 @@ defaults:
       initialDelaySeconds: 2400
   nodeSelector: {}
   tolerations: []
+  # Escape hatch: when set, replaces the auto-generated podAntiAffinity entirely.
   affinity: {}
+  # Enforces 1:1 Pod-to-Node allocation so pods do not collide on /dev/tenstorrent.
+  # Disable for multi-pod-per-node once a device plugin or DRA driver is in place.
+  podAntiAffinity:
+    enabled: true
+    topologyKey: kubernetes.io/hostname
```
mjeremicTT: Same for this

charts/tt-inference-server/templates/deployment.yaml
```diff
@@ -130,9 +130,17 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with $cfg.affinity }}
+      {{- if not (empty $cfg.affinity) }}
```
mjeremicTT: Why not just like: if cfg.affinity ? No need for double negation

charts/tt-inference-server/templates/deployment.yaml
```diff
@@ -130,9 +130,17 @@ spec:
...
       affinity:
-        {{- toYaml . | nindent 8 }}
+        {{- toYaml $cfg.affinity | nindent 8 }}
+      {{- else if $cfg.podAntiAffinity.enabled }}
```
mjeremicTT: Can we go just without else like: if $cfg.podAntiAffinity.enabled

charts/tt-inference-server/values.yaml
vcankovicTT: moved into deployment.yaml

charts/tt-inference-server/values.yaml
vcankovicTT: moved into deployment.yaml

charts/tt-inference-server/templates/deployment.yaml
vcankovicTT: we need to check with `if not (empty..."` because `if {}` will return True in Go templates. So `if $aff` would be always true for any dict value, including `{}`

charts/tt-inference-server/templates/deployment.yaml
vcankovicTT: based on the initial request in this phase , I suppose its even more safer if we opt-out this setting

PR #3612, opened 2026-05-19
Dynamo Protocol Integration

ljovanovicTT:
## Problem
We want to use the Dynamo Frontend as a proxy layer in front of our server. This approach provides several advantages:
- Built-in API layer — Endpoints such as /responses are already supported out of the box, reducing the need for manual endpoint implementation and allowing faster adaptation to evolving API standards.
- Preprocessing support — Features like tokenization and tool-calling for various models are already integrated, simplifying server-side logic.
- Load balancing — In future deployments with multiple cpp_server instances, Dynamo Frontend can distribute requests to the least busy node, improving resource utilization and scalability.
- Easier maintenance — Centralizing API handling and preprocessing reduces complexity in the cpp server itself, making the overall system easier to maintain and evolve.
- Improved scalability — Makes horizontal scaling of inference backends significantly cleaner.
- Observability opportunities — Easier integration of centralized logging, metrics, tracing, rate limiting, and monitoring at the proxy layer.

## Testing
1. Create Dynamo venv - note that Python 3.12 is a requirement
```
uv venv --python 3.12 ~/.venvs/dynamo
source ~/.venvs/dynamo/bin/activate
uv pip install ai-dynamo uvloop
```

2. Start Dynamo Frontend + cpp server Backend on desired ports
```
cd tt-inference-server/tt-media-srver/cpp_server
DYN_VENV=~/.venvs/dynamo \
HTTP_PORT=8000 SERVER_PORT=8001 \
  ./scripts/start_dynamo.sh
```

3. Run tests
```
vllm bench serve   --base-url http://1localhost:8000   --backend openai-chat 
  --endpoint /v1/chat/completions   --model deepseek-ai/DeepSeek-R1-05
28   --dataset-name random   --random-input-len 32   --random-output-
len 1024   --num-prompts 1000   --max-concurrency 64   --save-result   --result-filename vllm-bench-mock.json   2>&1 | tee bench_mock_output.txt

```

```
./multiturn-chat.sh -n 20 'hello'
```

## TODO
Follow-up PRs will tackle:
[] Use etcd Protocol for Discovery
[] Make Plan for Deployment and Testing
[] Investigate Adding Integration Tests or CI Gates
[] Measure Performance

### DIFF
--- a/tt-media-server/cpp_server/include/dynamo/dynamo_protocol.hpp
+++ b/tt-media-server/cpp_server/include/dynamo/dynamo_protocol.hpp
@@ -1,0 +1,249 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+#pragma once
+
+/**
+ * Dynamo TCP Wire Protocol
+ *
+ * Implements NVIDIA Dynamo's request plane so this C++ server can register
+ * as a native backend worker without a Python bridge. Originally adapted
+ * from https://github.com/ai-dynamo/dynamo and the dynamo-mock-backend
+ * sibling repo.
+ *
+ * Wire formats implemented:
+ *   - TCP request frame (ingress from frontend)
+ *   - TCP response frame (ACK back to frontend)
+ *   - TwoPartCodec (header+body framing for control + data)
+ *   - Call-home response stream (streaming tokens back to the frontend)
+ */
+
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include <json/json.h>
+
+namespace tt::dynamo {
+
+// ---------------------------------------------------------------------------
+// Wire-format helpers
+// ---------------------------------------------------------------------------
+
+inline void put_u16_be(std::vector<uint8_t>& buf, uint16_t val) {
+  buf.push_back(static_cast<uint8_t>((val >> 8) & 0xFF));
+  buf.push_back(static_cast<uint8_t>(val & 0xFF));
+}
+
+inline void put_u32_be(std::vector<uint8_t>& buf, uint32_t val) {
+  buf.push_back(static_cast<uint8_t>((val >> 24) & 0xFF));
+  buf.push_back(static_cast<uint8_t>((val >> 16) & 0xFF));
+  buf.push_back(static_cast<uint8_t>((val >> 8) & 0xFF));
+  buf.push_back(static_cast<uint8_t>(val & 0xFF));
+}
+
+inline void put_u64_be(std::vector<uint8_t>& buf, uint64_t val) {
+  for (int i = 7; i >= 0; --i) {
+    buf.push_back(static_cast<uint8_t>((val >> (i * 8)) & 0xFF));
+  }
+}
+
+inline uint16_t get_u16_be(const uint8_t* p) {
+  return static_cast<uint16_t>((static_cast<uint16_t>(p[0]) << 8) |
+                               static_cast<uint16_t>(p[1]));
+}
+
+inline uint32_t get_u32_be(const uint8_t* p) {
+  return (static_cast<uint32_t>(p[0]) << 24) |
+         (static_cast<uint32_t>(p[1]) << 16) |
+         (static_cast<uint32_t>(p[2]) << 8) | static_cast<uint32_t>(p[3]);
+}
+
+inline uint64_t get_u64_be(const uint8_t* p) {
+  uint64_t val = 0;
+  for (int i = 0; i < 8; ++i) {
+    val = (val << 8) | static_cast<uint64_t>(p[i]);
+  }
+  return val;
+}
+
+// ---------------------------------------------------------------------------
+// TCP request frame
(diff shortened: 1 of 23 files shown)


tt-media-server/cpp_server/include/dynamo/dynamo_protocol.hpp
```diff
@@ -0,0 +1,247 @@
...
+  bool ignore_eos = false;
+
+  // SamplingOptions --------------------------------------------------------
+  std::optional<float> temperature;
+  std::optional<float> top_p;
+  std::optional<int> top_k;
+  std::optional<int> seed;
+  std::optional<float> frequency_penalty;
+  std::optional<float> presence_penalty;
+  std::optional<float> repetition_penalty;
```
nanicicTT, line 178: would also look cool if these fields are packed in separate struct and having optional to it?

tt-media-server/cpp_server/src/dynamo/discovery.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/// Dynamo's slug validator rejects anything outside [a-z0-9_-]. HuggingFace
+/// model ids have `/` and uppercase, so map them to `-` and lowercase.
+std::string sanitizeSlug(const std::string& s) {
+  std::string out;
+  out.reserve(s.size());
+  for (char c : s) {
+    if ((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '-' ||
+        c == '_') {
```
nanicicTT, line 51:
```cpp
#include <ctype.h>
// ...
if (islower((unsigned char)c) || isdigit((unsigned char)c) || c == '-' || c == '_') {
    // Valid character!
}
```

this one could work, too, but more readable. 
In C, for example, if `c` is a standard char, it might be signed. Passing a negative value to `<ctype.h>` functions results in undefined behavior. So, it is highly recommended to cast the character to an unsigned char when using these functions.

And below for uppercase `isupper` could be used for check.

tt-media-server/cpp_server/src/dynamo/discovery.cpp
```diff
@@ -0,0 +1,193 @@
...
+    instance["endpoint"] = config.endpoint;
+    instance["namespace"] = config.namespace_name;
+    instance["instance_id"] = static_cast<Json::UInt64>(config.instance_id);
+
+    Json::Value transport(Json::objectValue);
+    transport["tcp"] = config.tcp_address;
+    instance["transport"] = std::move(transport);
+    instance["device_type"] = "cuda";
+
+    const std::string filePath = dir + "/" + encodedKey;
```
nanicicTT, line 94: Maybe good to use https://en.cppreference.com/cpp/filesystem/path for paths?

tt-media-server/cpp_server/src/dynamo/dynamo_endpoint.cpp
```diff
@@ -0,0 +1,330 @@
...
+          }
+
+          auto cb = [req, sendChunk, signalDone](
+                        const tt::domain::llm::LLMStreamChunk& chunk,
+                        bool isFinal) {
+            sendChunk(toTokenChunk(chunk, isFinal));
+            if (isFinal) {
+              if (req->session) req->session->clearInFlight();
+              signalDone();
+            }
```
nanicicTT, line 207: maybe `sendChunk` and `signalDone` could be collected in one function/method not to be forgotten to be called together, like in above and below methods. 
Even here, we send a Chunk, then do some check for req->session and finally send `signalDone()`, this, too, might make use of this method, right?

tt-media-server/cpp_server/src/dynamo/dynamo_endpoint.cpp
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  TT_LOG_INFO("[DynamoEndpoint] Shutting down");
+  if (server_) {
+    server_->shutdown();
+  }
+  unregister_discovery(discovery_);
+
+  if (keepalive_thread_.joinable()) keepalive_thread_.join();
+  if (server_thread_.joinable()) server_thread_.join();
```
nanicicTT, line 323: maybe https://en.cppreference.com/cpp/thread/jthread could be utilized...

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -126,4 +127,104 @@ PrefixCachingInfo computePrefixCachingInfo(
...
+    bool match = true;
+    for (size_t j = 0; j < n; ++j) {
+      if (tokens[i + j] != needle[j]) {
+        match = false;
+        break;
+      }
+    }
+    if (match) out.push_back(i + n);
+  }
+  return out;
```
nanicicTT, line 157: Cool, but there is another way to do so if you fancy it

```cpp
std::vector<size_t> out;

if (needle.empty() || tokens.size() < needle.size())
    return out;

const size_t n = needle.size();

for (size_t i = 0; i + n <= tokens.size(); ++i) {
    if (std::equal(
            needle.begin(),
            needle.end(),
            tokens.begin() + i)) {
        out.push_back(i + n);
    }
}

return out;
```

idjuricTT (approved):

PR #3616, opened 2026-05-19
Dynamo Integration - Enable etcd for Backend Discovery

ljovanovicTT:
## Problem
In order for the Dynamo Frontend to operate in a fully distributed system and act as a gateway to multiple distributed LLM nodes deployed across different machines, we need to enable distributed service discovery. In this case, `etcd` was chosen due to its simplicity, reliability, and proven support for production-scale workloads.

### DIFF
--- a/tt-media-server/cpp_server/include/dynamo/etcd_client.hpp
+++ b/tt-media-server/cpp_server/include/dynamo/etcd_client.hpp
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+#pragma once
+
+/**
+ * Minimal etcd v3 client over the JSON HTTP gateway.
+ *
+ * etcd's gRPC-gateway exposes the v3 KV/Lease API as JSON HTTP endpoints on
+ * the same port as gRPC (default 2379). This client speaks HTTP/1.1 directly
+ * over a POSIX socket so we don't pull in libcurl or grpc.
+ *
+ * Only the surface used by Dynamo discovery is implemented:
+ *   - LeaseGrant: create a lease the backend will associate with its keys.
+ *   - LeaseKeepAlive: refresh the lease so etcd doesn't reap our entries.
+ *   - LeaseRevoke: explicit deregistration on shutdown.
+ *   - Put / DeleteRange: write / remove the instance + MDC documents.
+ *
+ * HTTPS endpoints are NOT supported (frontends in the same trust domain
+ * typically use plain HTTP; production deployments should terminate TLS at a
+ * sidecar proxy).
+ */
+
+#include <cstdint>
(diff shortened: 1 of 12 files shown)


tt-media-server/cpp_server/include/dynamo/etcd_client.hpp
```diff
@@ -0,0 +1,70 @@
...
+  /// Put a single key-value pair, optionally attached to a lease (0 = none).
+  void put(const std::string& key, const std::string& value,
+           int64_t lease_id = 0);
+
+  /// Delete a single key (range_end is left unset).
+  void deleteRange(const std::string& key);
+
+ private:
+  std::string host_;
+  int port_ = 2379;
```
nanicicTT, line 66: this might be configured from outside, from some config file, but ok for now, just not to be forgotten, since it will be harder to maintain later on.

tt-media-server/cpp_server/scripts/start_dynamo.sh
```diff
@@ -30,16 +31,39 @@ if [[ ! -d "${DYN_VENV}" ]]; then
...
 
+# Quick sanity check: when running with etcd, fail fast if the gateway isn't
+# reachable so we don't waste a minute booting cpp_server only to crash on
+# first lease grant.
+if [[ "${DYN_DISCOVERY_BACKEND}" == "etcd" ]]; then
+    ETCD_HOST_PORT="${ETCD_ENDPOINTS#http://}"
+    ETCD_HOST_PORT="${ETCD_HOST_PORT%%/*}"
+    if ! (echo > "/dev/tcp/${ETCD_HOST_PORT/:/\/}") 2>/dev/null; then
+        echo "etcd not reachable at ${ETCD_ENDPOINTS}." >&2
+        echo "Start one with:  docker run --rm -d -p 2379:2379 --name etcd quay.io/coreos/etcd:v3.5.13 \\" >&2
```
nanicicTT, line 61: this is where we would make use of the config, read from one place and 2379 changes on one place only. Also on line 62 of this file.

tt-media-server/cpp_server/src/dynamo/discovery.cpp
```diff
@@ -59,135 +63,263 @@ std::string sanitizeSlug(const std::string& s) {
...
   Json::StreamWriterBuilder b;
   b["indentation"] = "";
   b["emitUTF8"] = true;
   return Json::writeString(b, v);
 }
 
-}  // namespace
+/// Hierarchical key that Dynamo's KVStoreDiscovery uses for both file and
+/// etcd backends: <namespace>/<component>/<endpoint>/<instance_id_hex>.
+std::string instanceKey(const DiscoveryConfig& c) {
```
idjuricTT, line 75: Should we add come randomness here? 
I had a lot of issues with double services having the same instance id

idjuricTT (approved):

PR #3620, opened 2026-05-19
Run prefill standalone script and session manager cleanup

dmadicTT:


tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -291,30 +262,11 @@ void SessionManager::sendDeallocRequest(const std::string& sessionId,
...
-      slotId.has_value() ? std::to_string(slotId.value()) : "none",
-      getActiveSessionCount());
+    trantor::EventLoop* callerEventLoop, size_t initialHash) {
+  TT_LOG_DEBUG("[SessionManager] createSession called, activeSessions={}",
+               getActiveSessionCount());
   evictOldSessions();
 
-  // Fast path: caller supplied a pre-assigned slot. Skip IPC allocation and
-  // insert the session synchronously.
-  if (slotId.has_value()) {
```
dmadicTT, line 304: `slotId` was always passed as `null`, so this was never called. 
As I cleaned this up, it cleaned up some more code.

ztorlakTT (approved):

ljovanovicTT (approved):

tt-media-server/cpp_server/tests/session_manager_test.cpp
```diff
@@ -55,8 +59,39 @@ std::string createSessionWithSlot(tt::services::SessionManager& manager,
...
+        continue;
+      }
+      foundAllocation = true;
+      break;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds{1});
+  }
+  if (!foundAllocation) {
+    throw std::runtime_error("Timed out waiting for allocation request");
+  }
```
nanicicTT, line 84: These flags as stateful things could be harder to follow. In this case it is simple enough, but in general, what I like is more functional way to do these things. 


```suggestion
auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds{2};

[&] {
  while (std::chrono::steady_clock::now() < deadline) {
      if (requestQueue->tryPop(request)) {
        if (request.action == tt::domain::MemoryManagementAction::ALLOCATE) {
          return;
        }
        continue;
      }

      std::this_thread::sleep_for(std::chrono::milliseconds{1});
    }

    throw std::runtime_error("Timed out waiting for allocation request");
}();
```

you want to find the `request.action` when it is `tt::domain::MemoryManagementAction::ALLOCATE` and that means you do not want to throw. 
Then we can easily return and skip the throwing and sleeping step. It is easier to read and to follow. Also it is early return without checks if ALLOCATE happened.

nanicicTT: also, one note in case if many non-ALLOCATE messages arrive, the loop will process them as fast as possible without sleeping. That is usually faster, but it may burn more CPU.

```
queue has message, but not ALLOCATE -> continue immediately, no sleep
queue has ALLOCATE                  -> break/return
queue empty                         -> sleep 1ms
```

so the `sleep_for()` is executed only when `tryPop()` fails, right?

There is this tradeoff, so maybe your approach is good, not sure about the traffic and number of irrelevant messages you might get:

> sleep only when queue is empty:
  faster, more CPU if lots of irrelevant messages

> sleep after every failed match:
  lower CPU, but can be much slower

tt-media-server/cpp_server/run-prefill-mock-standalone.sh
```diff
@@ -0,0 +1,47 @@
...
+  fi
+  wait "$SERVER_PID" 2>/dev/null || true
+  wait "$RUNNER_PID" 2>/dev/null || true
+}
+trap cleanup EXIT INT TERM
+
+./build/tt_media_server_cpp > prefill-server.log 2>&1 &
+SERVER_PID=$!
+
+python3 src/runtime/runners/mock_prefill_runner.py > prefill-model.log 2>&1 &
```
nanicicTT, line 34: this might be dangerous to be called like that, since it depends on where the script is called from. Might be good to be called relative to where the script is located physically. 

```sh
# Better maybe to get the directory where this script is located
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"

# Now call the python script relative to the script's location
python3 "$SCRIPT_DIR/src/runtime/runners/mock_prefill_runner.py" > "$SCRIPT_DIR/prefill-model.log" 2>&1 &
```

where `${BASH_SOURCE[0]}` is the path to the current script. And we work relative to it.

nanicicTT: Image
Image


Here's the example.

tt-media-server/cpp_server/run-prefill-mock-standalone.sh
```diff
@@ -0,0 +1,47 @@
...
+  fi
+  if [[ -n "$RUNNER_PID" ]] && kill -0 "$RUNNER_PID" 2>/dev/null; then
+    kill "$RUNNER_PID" 2>/dev/null || true
+  fi
+  wait "$SERVER_PID" 2>/dev/null || true
+  wait "$RUNNER_PID" 2>/dev/null || true
+}
+trap cleanup EXIT INT TERM
+
+./build/tt_media_server_cpp > prefill-server.log 2>&1 &
```
nanicicTT, line 31: Also this might be dangerous, local paths, script might be invoked from many different locations in folder structure. I would tie it relative to script's path, see
https://github.com/tenstorrent/tt-inference-server/pull/3620/changes#r3267053251

idjuricTT (approved):

tt-media-server/cpp_server/run-prefill-mock-standalone.sh
dmadicTT: Good comments! The script is work in progress, I will apply the comments in the follow ups.

PR #3623, opened 2026-05-19
Add grpc service

lmuravljovTT:
https://github.com/tenstorrent/tt-inference-server/issues/3747

# gRPC Server for LLM Inference

## Summary

Added a gRPC server to the C++ media server that exposes LLM inference capabilities via a streaming RPC interface. This provides an alternative to the existing HTTP/REST API with better performance characteristics for token streaming.

## What was done

- Defined `inference.proto` with `Generate` (streaming) RPC
- Implemented `GrpcInferenceService` to handle requests and stream token responses
- Implemented `GrpcServerHandle` for server lifecycle management
- Integrated gRPC server startup in `main.cpp` (enabled via `GRPC_LISTEN` env var)
- Updated CMake to optionally build with gRPC support (`-DENABLE_GRPC=ON`)
- Added smoke test script for quick validation

## Usage

```bash
# Build
./build.sh --grpc

# Run (gRPC on port 50051, HTTP on port 8080)
GRPC_LISTEN="0.0.0.0:50051" ./build/tt_media_server_cpp -p 8080
```

## Next steps

- Link up with Dynamo for end-to-end inference pipeline


tt-media-server/cpp_server/include/utils/concurrent_queue.hpp
```diff
@@ -146,4 +149,60 @@ class LockFreeSPSCQueue {
...
+template <typename T>
+class BlockingQueue {
+ public:
+  void push(T item) {
+    {
+      std::lock_guard<std::mutex> lock(mu);
+      queue.push(std::move(item));
+    }
+    cv.notify_one();
+  }
```
nanicicTT, line 162: I think there is something similar on Kosta's branch:
[knovokmetTT/cpp-server-blaze-cleanup](https://github.com/tenstorrent/tt-inference-server/pull/36240)

but I am not quite sure if it could be reusedf, or this file here needs to be changed, like you did it.

tt-media-server/cpp_server/CMakeLists.txt
```diff
@@ -217,6 +217,67 @@ endif()
 # Option for Kafka support
 option(KAFKA_ENABLED "Enable Kafka messaging support" OFF)
 
+# Option for gRPC server
+option(ENABLE_GRPC "Build gRPC server" OFF)
+
+if(ENABLE_GRPC)
```
nanicicTT: Maybe it is more readable `if (NOT ENABLE_GRPC)` to just do 
`message(STATUS "gRPC server: DISABLED (use -DENABLE_GRPC=ON to enable)")` 
and to early `return()` and then below to do the rest. Easier to read once that condition is out of the way, to focus on the main logic. Less cognitive load when reading. :)

tt-media-server/cpp_server/install_dependencies.sh
```diff
@@ -109,3 +118,45 @@ if ! drogon_found; then
...
+    $SUDO apt-get remove -y --purge libprotobuf-dev protobuf-compiler libgrpc++-dev libgrpc-dev \
+        protobuf-compiler-grpc libabsl-dev 2>/dev/null || true
+    $SUDO apt-get autoremove -y 2>/dev/null || true
+
+    GRPC_VERSION="v1.62.0"
+    GRPC_TMP="/tmp/grpc_build"
+    rm -rf "${GRPC_TMP}"
+    echo "Building gRPC ${GRPC_VERSION} from source (includes abseil, protobuf, utf8_range)..."
+    git clone --depth 1 --branch "${GRPC_VERSION}" --recurse-submodules --shallow-submodules \
+        https://github.com/grpc/grpc.git "${GRPC_TMP}"
```
nanicicTT, line 140: Maybe it would be good to stick to some release tag (version), since this is `main` and it is everchanging and there might be some bugs introduced every time we pull the main code (when we invoke the script). 
Usually it is better to work with release tags: more tested and stable version, we choose when to change the tag in the script, etc. 

Image

tt-media-server/cpp_server/CMakeLists.txt
```diff
@@ -926,6 +1003,13 @@ set(SOURCES
     src/dynamo/dynamo_endpoint.cpp
 )
 
+if(ENABLE_GRPC)
+    list(APPEND SOURCES
+        src/api/grpc/grpc_server.cpp
+        src/api/grpc/grpc_inference_service.cpp
```
nanicicTT, line 1051: Maybe `src/api/grpc` could be some CMAKE variable, so when the more files are added we reuse variable that will be changed in one place if need be.

tt-media-server/cpp_server/src/dynamo/dynamo_protocol.cpp
```diff
@@ -35,8 +35,8 @@ bool txTraceEnabled() {
     const char* v = std::getenv("DYN_TX_TRACE");
     if (v == nullptr) return false;
     std::string s(v);
-    if (s.empty() || s == "0" || s == "false" || s == "FALSE" ||
-        s == "False" || s == "off" || s == "OFF") {
+    if (s.empty() || s == "0" || s == "false" || s == "FALSE" || s == "False" ||
+        s == "off" || s == "OFF") {
       return false;
     }
     return true;
```
nanicicTT: or just 

```cpp
return !(s.empty() || s == "0" || s == "false" || s == "FALSE" || s == "False" ||
        s == "off" || s == "OFF")
```

tt-media-server/cpp_server/cmake/PatchTokenizersForGrpc.cmake
```diff
@@ -0,0 +1,52 @@
+# SPDX-License-Identifier: Apache-2.0
+# Patches tokenizers-cpp so SentencePiece is not built (avoids protobuf clash with gRPC).
+# FetchContent runs this with the tokenizers-cpp source directory as the working directory.
+cmake_minimum_required(VERSION 3.19)
```
nanicicTT: this is pretty old one 

Image

Maybe we can do better than that, at least 3.25?

tt-media-server/cpp_server/protos/inference.proto
```diff
@@ -0,0 +1,51 @@
...
+  rpc Generate(GenerateRequest) returns (stream TokenChunk);
+  rpc Health(HealthRequest) returns (HealthResponse);
+}
+
+message StopConditions {
+  optional uint32 max_tokens = 1;
+  optional uint32 min_tokens = 2;
+  repeated uint32 stop_token_ids = 3;
+  repeated string stop = 4;
+  optional bool ignore_eos = 5;
```
idjuricTT, line 21: not sure how this maps 😄

tt-media-server/cpp_server/protos/inference.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+message SamplingOptions {
+  optional float temperature = 1;
```
idjuricTT, line 25: I would be careful about those defaults since it might affect model behavior

tt-media-server/cpp_server/src/api/grpc/grpc_inference_service.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    if (chunkOpt->has_finish_reason()) {
+      return ::grpc::Status::OK;
+    }
+  }
+
+  return ::grpc::Status::OK;
+}
+
+::grpc::Status GrpcInferenceService::Generate(
```
idjuricTT, line 149: Please try some higher load tests so we can see how this behaves compared to HTTP

tt-media-server/cpp_server/build.sh
```diff
@@ -50,6 +51,14 @@ while [[ $# -gt 0 ]]; do
             KAFKA_ENABLED="ON"
             shift
             ;;
+        --grpc)
+            ENABLE_GRPC="ON"
+            shift
+            ;;
+        --no-grpc)
```
idjuricTT: I would use one as default, no matter npc or no-grpc

tt-media-server/cpp_server/protos/inference.proto
lmuravljovTT: Those aren't default values, but rather a proto convention for unique field tags

tt-media-server/cpp_server/CMakeLists.txt
lmuravljovTT: hmm I'm not completely convinced, I don't really expect the grpc folder to grow, not sure if it's worth extracting for 2 usages? 

Although, is this something we should consider for the rest of the file? 
ex: 
src/runtime/runners/
src/utils/tokenizers/
etc.

This parts not really the scope of this PR, just curious

nanicicTT: Sure, was just curious and asking, but for two variables I agree it is not worth it.

tt-media-server/cpp_server/protos/inference.proto
lmuravljovTT: // StopConditions ---------------------------------------------------------
  int max_tokens = 128;
  std::optional<int> min_tokens;
  std::vector<int> stop_token_ids;
  std::vector<std::string> stop;
  bool ignore_eos = false;

These are the stop conditions defined in the dynamo protocol, and they are passed through GenerateRequest, do you think there's an issue?

tt-media-server/cpp_server/include/utils/concurrent_queue.hpp
lmuravljovTT: cleaned up now, will be clear once https://github.com/tenstorrent/tt-inference-server/pull/3733 is merged to main

tt-media-server/cpp_server/src/api/grpc/grpc_inference_service.cpp
lmuravljovTT: I'll open a separate PR with linking dynamo up with grpc, and add the tests there, since this PR only adds the service to keep it smaller

tt-media-server/cpp_server/src/api/grpc/grpc_inference_service.cpp
```diff
@@ -0,0 +1,175 @@
...
+void prepareLLMRequest(LLMRequest& req,
+                       const inference::GenerateRequest* grpc) {
+  std::vector<int> tokens;
+  tokens.reserve(grpc->token_ids_size());
+  for (int i = 0; i < grpc->token_ids_size(); ++i) {
+    tokens.push_back(static_cast<int>(grpc->token_ids(i)));
+  }
+  const int promptLen = static_cast<int>(tokens.size());
+
+  req.stream = true;
```
idjuricTT, line 27: Not sure should we hardcode this

PR #3558, opened 2026-05-15
fix(forge): make concurrency reachable via run.py and --vllm-override-args + set batch2 for Llama-3.2-3B-Instruct

kmabeeTT:
### Issue

Closes #3556

### Problem

Configuring `max_num_seqs > 1` for a forge LLM is silently a no-op via the standard `run.py` flow. Two parallel requests always serialize. Four gaps stack (per the linked issue):

- `tt-media-server/config/settings.py`: `use_dynamic_batcher` defaults False → scheduler picks serial worker even when `max_num_seqs > 1`.
- `workflows/run_docker_server.py`: `model_spec.env_vars` is never propagated to the container for forge/media engines.
- `workflows/model_spec.py`: vllm-tt-only defaults (`TORCHDYNAMO_DISABLE` etc.) injected unconditionally; leak into forge containers and crash warmup.
- `workflows/model_spec.py`: `--vllm-override-args` updates `vllm_args` but never reaches the container env on forge/media.

### What's changed

Four commits in this PR:

1. `settings.py` — auto-enable `use_dynamic_batcher` when `max_num_seqs > 1`.
2. `run_docker_server.py` + `model_spec.py` — propagate `model_spec.env_vars` to forge/media containers; gate vllm-tt-only `default_env_vars` to non-forge engines.
3. `model_spec.py` — bridge `--vllm-override-args` keys (`max_num_seqs` → `MAX_NUM_SEQS`, `max_model_len` → `MAX_MODEL_LENGTH`) into `env_vars`.
4. `model_spec.py` — set `MAX_NUM_SEQS=2` as the default for Llama-3.2-3B-Instruct (N150, N300).

After: `python3 run.py --model Llama-3.2-3B-Instruct ...` delivers 2-concurrent execution by default. Other forge LLMs (Qwen3-4B, Llama-3.1-8B-Instruct, Qwen3-8B, Falcon3-7B-Instruct) keep their `max_num_seqs=1` default and can opt-in per-invocation via `--vllm-override-args '{"max_num_seqs": N}'` pending per-model validation on real N150/N300.

### Testing

- Two parallel curls against Llama-3.2-3B-Instruct: **1.37s identical wall-clock** (was 2.1s serialized pre-fix).
- Single-request throughput at batch=2: **~28.4 tok/s** vs the batch=1 baseline of ~30 tok/s (~5% regression — acceptable for the 2x aggregate win on batched workloads).
[...]

Issue #3533, opened 2026-05-14
Forge LLM evals time out: tt-media-server worker doesn't drop orphan tasks; eval batch_size exceeds device max_num_seqs


kmabeeTT:
### TL;DR

With #3531 applied (so batched-prompt requests no longer fail at HTTP 422), running `--workflow evals` against a forge LLM with realistic settings still fails with client timeouts. Two issues compound — would appreciate input on whether the analysis below is right and which layer to tackle first.

1. **Server (tt-media-server):** when `process_request` is cancelled, the result queue is torn down but the worker keeps re-processing the orphaned task indefinitely, blocking other sub-requests behind it.
2. **Config (tt-inference-server):** `evals/eval_config.py` uses `batch_size=32` for several forge-eligible models. Fanning out 32 prompts onto a device with `max_num_seqs=2` serializes into ~16 chunks and exceeds the client's read timeout, which then trips issue (1).

Neither was reachable before #3531 (every batched request 422'd immediately, so device-side work never ran long enough to expose either bug).

### Problem 1 — orphan-task re-process loop

Server log from a live `--workflow evals` run, after #3531 applied:

```
ERROR - [process_request] async failed after 1000.0009 seconds.
WARNING - No result queue found for task 82b3fbc4-...
          Current queues: ['a3432321-...']
INFO - Worker 0 processing tasks: 1
INFO - Device 0: Starting non-streaming batch generation for 1 requests
INFO - [run] executed in 71.3360 seconds. Run VLLM inference
WARNING - No result queue found for task 82b3fbc4-...     ← same task, again
          Current queues: ['a3432321-...']
INFO - Worker 0 processing tasks: 1
... (repeats indefinitely, ~71s per iteration)
```

The cancelled request's task keeps being pulled off the worker queue and processed (~71s on device per iteration). The legitimate sibling sub-request (`a3432321`) is starved behind it since the worker processes one task at a time.

A direct `curl` against `/v1/completions` from a separate terminal in this state hangs until `--max-time` aborts — confirming the worker queue is jammed, not just the eval client.

### Problem 2 — eval `batch_size` vs device `max_num_seqs`

[...]

PR #3627, opened 2026-05-20
fix(evals): clamp client concurrency to device_model_spec.max_concurrency

kmabeeTT:
### Issue

Addresses Problem 2 of #3533

### Problem

`evals/run_evals.py` builds lm-eval / lmms-eval commands with `num_concurrent=32` (from `EvalTask.max_concurrent=32` default) and `batch_size=32` on the EVALS_META path (post_init reassigns `batch_size = max_concurrent`). This works for engines/models that support a 32-wide batch (typical ttnn / vllm-tt) but over-subscribes engines that don't — e.g. some forge LLMs currently in bring-up run with a smaller `max_num_seqs` (Llama-3.2-3B-Instruct on N150 is at 2 today). When that happens the eval client queues 32 requests onto a smaller-than-32-slot server, the queue serializes, the client's read timeout fires, and requests cancel mid-flight.

The data needed to do the right thing already exists: `DeviceModelSpec.max_concurrency` is set per (model, device, engine) in `workflows/model_spec.py`. The eval command builder just wasn't consulting it.

This is also the fix shape the team had in mind. From prior discussion on #3533:

> "We've seen this issue a couple of times ourselves and the solution we want to implement but have not gotten around to, is to source the correct concurrency factor from the ModelSpec object." — @bgoelTT

### What changed

`evals/run_evals.py` — `build_eval_command()` now reads `model_spec.device_model_spec.max_concurrency` and clamps both `task.max_concurrent` and `task.batch_size` to that ceiling before they go into the lm-eval / lmms-eval command line. Logs each clamp. Falls back to the task's value if ModelSpec doesn't expose the field (defensive — keeps behavior unchanged for older specs).

Net effect:
- Entries with `max_concurrency >= 32` (typical ttnn / vllm-tt): no change, clamp is a no-op.
- Entries with `max_concurrency < 32` (some forge LLMs currently in bring-up): `num_concurrent` and `batch_size` are clamped down to the declared ceiling.

Note: the clamp applies universally — no `engine == "forge"` branch. It's a no-op when `device_model_spec.max_concurrency >= task.max_concurrent`, which is the case for most ttnn / vllm-tt entries today. The functional effect lands on entries that declare smaller capacity, which today is mainly forge LLMs in bring-up but may apply to any future entry. The spec stays the source of truth; the eval runner just honors it.

~30 lines including comments.

### Relationship to other recent PRs

- **#3558 (forge concurrency)** — set the smaller `max_concurrency` values on currently-in-bring-up forge LLM `DeviceModelSpec` entries (Llama-3.2-3B-Instruct N150/N300 went to 2). That made the model spec *truthful* about each entry's current capacity. This PR makes the eval client *honor* it. Without #3558, those entries would have still reported the legacy `max_concurrency=32` and the clamp would be a no-op. The two PRs are complementary.
- **#3532 (`/v1/completions` batched-prompt schema + fan-out)** — also a precondition for evals: lm-eval with `batch_size=N` packs N prompts into one HTTP request as `prompt: [str1, ..., strN]`, which #3532's schema + fan-out accept. Required for evals to reach the server at all.

### Testing

Llama-3.2-3B-Instruct on forge (`max_num_seqs=2`, P150 host, fresh tt-shield-built image), smoke eval (`--limit-samples-mode smoke-test`).

**Clamp log fires correctly:**
```
INFO Clamping meta_gpqa batch_size: 32 -> 2 (device_model_spec.max_concurrency=2)
```

**Resulting lm-eval command uses the clamped value:**
```
lm_eval --tasks meta_gpqa --model local-completions ... --batch_size 2 ...
```

**Server processes the now-properly-sized batch with full parallelism (server log, single-millisecond resolution):**
```
02:34:39,671 - Device 0: Starting non-streaming batch generation for 1 requests  <- sub-req A
02:34:39,671 - Device 0: Starting non-streaming batch generation for 1 requests  <- sub-req B  (same ms)
02:34:39,671 - Device 0: Starting non-streaming generation
02:34:39,671 - Device 0: Starting non-streaming generation
02:34:40,022 - Device 0: Non-streaming generation completed                       <- both done
02:34:40,022 - Device 0: Non-streaming generation completed
02:34:40,022 - Device 0: Non-streaming batch generation completed for 1 requests
02:34:40,022 - Device 0: Non-streaming batch generation completed for 1 requests
02:34:40,023 - [process_request] async executed in 0.3525 seconds. Base processing request
```

Both sub-requests started in the *same millisecond* (02:34:39.671) and finished in the *same millisecond* (02:34:40.022). They ran concurrently — that's `max_num_seqs=2` doing its job. Total `process_request` time: 0.35s, well under any read-timeout.

Before the fix, with `batch_size=32` against this 2-slot configuration, the queue would have been ~16-deep, taking ~70s per HTTP request — over the openai-python default 60s read-timeout, causing requests to cancel mid-flight (the original failure mode of #3533). The same dynamic applies to any other entry whose `max_concurrency` is below 32 — current or future.

**Validation on ttnn entries (no-op path):** `DeviceModelSpec.max_concurrency=32` on Llama-3.2-3B-Instruct ttnn entries (`workflows/model_spec.py:2089`) → clamp condition `min(32, 32) == 32` → no-op, command unchanged. Behavior preserved.

### Out of scope (intentional)

- **Problem 1 of #3533 (server-side worker orphan-task drain):** even after the clamp, evals can still hang because cancelled tasks remain on the worker queue and get re-processed (`No result queue found for task ...` warnings in the server log). The clamp drastically reduces the *amplification factor* (32× → 2×) and makes the bug far less reachable on real workloads, but doesn't fix the underlying server-side issue. A separate PR will address the worker-side drain in `tt-media-server/model_services/`. Both PRs together resolve the forge LLM eval timeout end-to-end.

### Files changed

- `evals/run_evals.py` — clamp + logging in `build_eval_command()`.

### DIFF
--- a/evals/run_evals.py
+++ b/evals/run_evals.py
@@ -334,9 +334,39 @@ def build_eval_command(
     else:
         api_url = f"{base_url}/completions"
 
+    # Clamp client concurrency/batch_size to the server's device_model_spec.max_concurrency.
+    # lm-eval / lmms-eval default to num_concurrent=32, which can exceed the server's
+    # max_num_seqs when the (model, device, engine) tuple supports a smaller batch —
+    # e.g. some forge LLMs currently in bring-up run with a smaller max_num_seqs. When
+    # that happens, the client over-subscribes the server, the queue serializes, and
+    # the client's read timeout fires. Source the correct ceiling from ModelSpec.
+    device_max_concurrency = getattr(
+        getattr(model_spec, "device_model_spec", None), "max_concurrency", None
+    )
+
+    effective_max_concurrent = task.max_concurrent
+    if effective_max_concurrent and device_max_concurrency:
+        effective_max_concurrent = min(effective_max_concurrent, device_max_concurrency)
+        if effective_max_concurrent != task.max_concurrent:
+            logger.info(
+                f"Clamping {task.task_name} num_concurrent: "
+                f"{task.max_concurrent} -> {effective_max_concurrent} "
+                f"(device_model_spec.max_concurrency={device_max_concurrency})"
+            )
+
+    effective_batch_size = task.batch_size
+    if effective_batch_size and device_max_concurrency:
+        effective_batch_size = min(effective_batch_size, device_max_concurrency)
+        if effective_batch_size != task.batch_size:
+            logger.info(
+                f"Clamping {task.task_name} batch_size: "
+                f"{task.batch_size} -> {effective_batch_size} "
+                f"(device_model_spec.max_concurrency={device_max_concurrency})"
+            )
+
     optional_model_args = []
-    if task.max_concurrent:
-        optional_model_args.append(f"num_concurrent={task.max_concurrent}")
+    if effective_max_concurrent:
+        optional_model_args.append(f"num_concurrent={effective_max_concurrent}")
 
     # lm-eval (text) expects full completions api route in base_url
     # lmms-eval (vision) expects base_url WITHOUT the endpoint path
@@ -391,7 +421,7 @@ def build_eval_command(
             "--output_path", output_dir_path,
             "--seed", task.seed,
             "--num_fewshot", task.num_fewshot,
-            "--batch_size", task.batch_size,
+            "--batch_size", effective_batch_size,
             "--log_samples",
             "--show_config",
         ]
@@ -405,7 +435,7 @@ def build_eval_command(
                 f"{model_kwargs_str}"
             ),
             "--tasks", task.task_name,
-            "--batch_size", str(task.batch_size),
+            "--batch_size", str(effective_batch_size),
             "--output_path", str(output_dir_path),
             "--log_samples",
         ]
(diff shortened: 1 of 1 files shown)


evals/run_evals.py
```diff
@@ -334,9 +334,39 @@ def build_eval_command(
...
+    effective_batch_size = task.batch_size
+    if effective_batch_size and device_max_concurrency:
+        effective_batch_size = min(effective_batch_size, device_max_concurrency)
+        if effective_batch_size != task.batch_size:
+            logger.info(
+                f"Clamping {task.task_name} batch_size: "
+                f"{task.batch_size} -> {effective_batch_size} "
+                f"(device_model_spec.max_concurrency={device_max_concurrency})"
+            )
+
```
idjuricTT, line 366: @vpetrovicTT  and @ddjukicTT could you please check this change?

bgoelTT (approved):

evals/run_evals.py
vpetrovicTT: This should work

vpetrovicTT (approved):

PR #3636, opened 2026-05-20
Refactor shared socket transport state

ztorlakTT:
- Extract common TCP/ZMQ transport state into `SocketTransportState`
- Share connection status, reconnect backoff, and connection-lost callback handling
- Keep TCP framing and ZMQ owner-thread send/receive logic transport-specific


idjuricTT (approved):

nanicicTT (approved):
Not sure if I missed something, but no remarks from my side, I'd say it looks pretty good and covered 👌🏻 💪

Issue #679, opened 2025-09-17
Process whisperX chunks in parallel


idjuricTT:
In order to improve performance  of Whisper processing, when WhisperX chunks are created submit all to the result queue and then pick those up in order.

In case multiple devices are available processing should be significantly faster

PR #3674, opened 2026-05-21
[WIP] Add centerpoint model

ddilbazTT:


tt-media-server/requirements.txt
```diff
@@ -18,8 +18,8 @@ silero_vad
 tqdm
 uvicorn
 whisperx==3.4.3
-torch==2.7.1+cpu
-torchaudio==2.7.1+cpu
+torch==2.10.0+cpu
+torchaudio
```
kmabeeTT, line 22: Is this an unrelated fix? Should go in as a seperate PR if it's fixing something, I think.

tt-media-server/tt_model_runners/forge_runners/fetch_models.sh
```diff
@@ -7,7 +7,7 @@
 
 set -e  # Exit on any error
 
-FORGE_MODELS_SHA="4703597c007ca78ec1c83a51a4433e9fc028b12b"
+FORGE_MODELS_SHA="89f187df4a5e86d6432dc33b4ce493e57ace1871"
```
kmabeeTT, line 10: Just a note to other reviewers, this is temporary hash from branch, will be replaced with hash from tt-forge-models main.

tt-media-server/open_ai_api/lidar.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Run CenterPoint 3D object detection on a LiDAR point cloud.
+
+    Accepts a nuScenes-format .bin file (float32 array, 5 channels per point:
+    x, y, z, intensity, ring).
+
+    Returns per-task 3D bounding box detections.
+    """
+    try:
+        file_content = await file.read()
```
kmabeeTT, line 34: Claude thought of this, and I think it makes sense...

**Nit / follow-up, not a blocker:** `await file.read()` here pulls the entire upload into memory with no cap. A typical nuScenes scan is ~1 MiB but the endpoint will accept multi-GB payloads, which an authenticated misbehaving client could use to OOM the worker. Other file-upload routes (`cnn.py`,
   `audio.py`) have the same gap, so this isn't unique to this PR — but LiDAR scans are larger and less self-bounded than images, so it's a good place to start.

  Minimal in-route fix:

  ```python
  MAX_LIDAR_UPLOAD_BYTES = 64 * 1024 * 1024  # 64 MiB

  chunks: list[bytes] = []
  total = 0
  while chunk := await file.read(1024 * 1024):
      total += len(chunk)
      if total > MAX_LIDAR_UPLOAD_BYTES:
          raise HTTPException(status_code=413, detail="LiDAR payload too large")
      chunks.append(chunk)
  file_content = b"".join(chunks)
  ```

  Longer-term, enforcing this once via ASGI middleware would cover every file-upload route instead of patching them individually. Happy to defer to a follow-up.

tt-media-server/tt_model_runners/forge_runners/centerpoint_runner.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self._voxelize = cp["voxelize"]
+        self._postprocess = cp["postprocess"]
+        self._grid_x = cp["GRID_X"]
+        self._grid_y = cp["GRID_Y"]
+
+        self.logger.info("Loading PillarFeatureNet weights on CPU ...")
+        self.pfn = cp["PillarFeatureNetCPU"]()
+        sd = cp["load_checkpoint"]()
+        self.pfn.load_state_dict(cp["remap_pfn_keys"](sd), strict=False)
```
kmabeeTT, line 134: This can load garbage... If _remap_pfn_keys is wrong, missing keys leave layers at random init; unexpected keys are silently dropped. Either capture and log missing_keys / unexpected_keys from the return value, or use strict=True. As-is, an upstream loader rename produces a silent accuracy regression.

tt-media-server/tt_model_runners/forge_runners/centerpoint_runner.py
```diff
@@ -0,0 +1,204 @@
...
+        voxels_np, coords_np, num_pts_np = self._voxelize(points)
+        voxels = torch.from_numpy(voxels_np)
+        coords = torch.from_numpy(coords_np)
+        num_pts = torch.from_numpy(num_pts_np)
+
+        with torch.no_grad():
+            pillar_features = self.pfn(voxels, num_pts, coords)
+            bev = self.scatter(pillar_features, coords, 1, self._grid_y, self._grid_x)
+
+        bev = bev.to(self.dtype).to(self.device)
```
kmabeeTT, line 193: bev.to(self.dtype).to(self.device) allocates twice. Use bev.to(self.device, dtype=self.dtype).

tt-media-server/domain/lidar_detection_request.py
```diff
@@ -0,0 +1,18 @@
...
+class LidarDetectionRequest(BaseRequest):
+    """Request for 3D object detection from a raw LiDAR point cloud.
+
+    The prompt field carries the raw binary content of a nuScenes-format
+    LiDAR scan (.bin file), interpreted as a flat array of float32 values
+    with 5 channels per point: x, y, z, intensity, ring.
+    """
+
+    prompt: bytes
+    score_threshold: float = Field(default=0.1, ge=0.0, le=1.0)
```
kmabeeTT, line 18: This score_threshold default of 0.1 gets duplicated elsewhere in lidar.py:21 too, can pull it out to variable in this file and re-use it both places to prevent drift/duplication:

```
  DEFAULT_SCORE_THRESHOLD = 0.1


  class LidarDetectionRequest(BaseRequest):
      prompt: bytes
      score_threshold: float = Field(
          default=DEFAULT_SCORE_THRESHOLD, ge=0.0, le=1.0
      )
      
and..

  @router.post("/detect", response_model=LidarDetectionResponse)
  async def detect(
      file: UploadFile = File(...),
      score_threshold: float = Form(default=DEFAULT_SCORE_THRESHOLD),
      ...
  ):      
     
 ```

File: tt-media-server/tt_model_runners/forge_runners/centerpoint_runner.py [hunk 7]
```diff
@@ -0,0 +1,204 @@
...
+        GRID_Y,
+        PillarFeatureNetCPU,
+        _PointPillarsScatterCPU,
+        _load_checkpoint,
+        _remap_pfn_keys,
+        get_single_input,
+        load_model_with_weights,
+        postprocess,
+        voxelize,
+    )
```
kmabeeTT, line 41: **Re: `_import_centerpoint_model()` reaching into `centerpoint.pytorch.src.model`**

  The loader exists in tt-forge-models PR #679 and `centerpoint/pytorch/src/__init__.py` already re-exports a clean public surface (`voxelize`, `postprocess`, `PillarFeatureNetCPU`, `get_single_input`, `load_model_with_weights`). But this runner imports from `...src.model` directly and grabs three
  underscore-prefixed symbols — `_PointPillarsScatterCPU`, `_load_checkpoint`, `_remap_pfn_keys` — that the upstream author explicitly marked private. A rename in `src/model.py` silently breaks warmup here.

  CenterPoint legitimately doesn't fit the canonical ViT-style loader (it's a CPU → TT → CPU pipeline with two models that share a checkpoint), so I'm not suggesting a full refactor to the `loader.ModelLoader()` pattern in this PR. But the private-symbol reach is fixable cheaply:

  **Suggestion:** promote those three names upstream (drop the leading underscore on `PointPillarsScatterCPU`, `load_checkpoint`, `remap_pfn_keys`) and add them to `centerpoint/pytorch/src/__init__.py`'s re-export list. Then this file's imports collapse to:

  ```python
  from tt_model_runners.forge_runners.model_loaders.centerpoint.pytorch.src import (
      GRID_X,
      GRID_Y,
      PillarFeatureNetCPU,
      PointPillarsScatterCPU,
      get_single_input,
      load_checkpoint,
      load_model_with_weights,
      postprocess,
      remap_pfn_keys,
      voxelize,
  )
  ```

  `_import_centerpoint_model()` and the dict-of-callables indirection go away — these become normal module-level imports after `ensure_model_loaders()` runs once. Should be a small additive commit on the still-open tt-forge-models PR with no semantic change.

  (Longer-term — not for this PR — these CPU stages could move behind `loader.input_preprocess()` / `loader.output_postprocess()` methods so the runner shrinks further, but that's a bigger interface change and worth deferring.)

kmabeeTT (commented):
Took an initial look, overall looks great - thank you! But some things to improve on, left comments.  Inference server folks have more experience, so I trust them with their review too :)

tt-media-server/requirements.txt
kmabeeTT: This is no longer needed, I encountered similar thing last week, and it is resolved now in latest main with a fix here: https://github.com/tenstorrent/tt-inference-server/pull/3947

PR #3678, opened 2026-05-22
Fix the bug in diarization pipeline

vpetrovicTT:


idjuricTT (approved):

tt-media-server/utils/audio_manager.py
```diff
@@ -20,6 +20,29 @@
...
+    # The audio worker only loads checkpoints from trusted sources (gated HF
+    # repos under HF_TOKEN and bundled weights); no user-supplied .pt ever
+    # reaches torch.load in this process, so the strict default offers no real
+    # safety here. Restore the pre-2.6 default for this process only. Callers
+    # that explicitly pass `weights_only=True` are still respected via setdefault.
+    _ORIGINAL_TORCH_LOAD = torch.load
+
+    def _torchLoadLegacyDefault(*args, **kwargs):
+        if kwargs.get("weights_only") is None:
+            kwargs["weights_only"] = False
```
nanicicTT, line 51: Cool, correct, but there is more Pythonic way to do so, I believe, but please test it:

If `weights_only` is not in kwargs, set it to `False`. 
If it is in kwargs, leave it alone.

```suggestion
    kwargs.setdefault("weights_only", False)
```

vpetrovicTT: gas

vpetrovicTT: It is prettier, but small concern is that _setdefault()_ is a **no-op** when the key is present but **'None'** .. :face_with_peeking_eye:

nanicicTT (approved):

ztorlakTT (approved):

PR #3679, opened 2026-05-22
Fix DeepSeek R1 tool calling: json prefix leak, tool_choice=none, and max_tokens floor

ksriv001:
## Summary

Fixes three issues that prevent DeepSeek R1 tool calling from working reliably on the C++ server, discovered via live testing against `console.tenstorrent.com`:

- **`json\n` prefix leaks into `function.arguments`**: When ` ``` ` and `json` arrive as separate tokens during streaming, the DeepSeek tool call parser enters the json block without stripping the "json" header. Adds a `json_header_stripped` flag to handle split-token scenarios, plus a safety strip in `finalizeSingleToolCall`.
- **`tool_choice=none` leaks raw special tokens into content**: Tools were still injected into the prompt via `applyChatTemplate` even when `tool_choice=none`, causing the model to generate `<|tool_calls_begin|>` etc. that appeared verbatim in the response content. Fix: pass `std::nullopt` for tools when `tool_choice` is `"none"`.
- **R1 exhausts token budget on reasoning before emitting tool calls**: DeepSeek R1 reasoning tokens consume the `max_tokens` budget. Clients often send low values (e.g. 8192) that get fully consumed by reasoning. Adds a DeepSeek-specific `max_tokens` floor of 16384 when tools are present.

## Test plan

- [x] Added `testSplitTokenJsonHeader()` with 3 test cases covering split-token scenarios
- [ ] Existing `test_tool_call_parser` tests still pass
- [ ] Verify `tool_choice=none` no longer leaks special tokens in response content
- [ ] Verify tool calls with `tool_choice=auto` return clean JSON arguments (no `json\n` prefix)
- [ ] Verify low `max_tokens` requests with tools get bumped to 16384 on DeepSeek R1

### DIFF
--- a/tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
+++ b/tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
@@ -272,8 +272,13 @@ struct ChatCompletionRequest : BaseRequest {
     out.messages = messages;
     out.skip_apply_chat_template = skip_apply_chat_template;
     const auto& tokenizer = tt::utils::tokenizers::activeTokenizer();
+    std::optional<std::vector<tool_calls::Tool>> effectiveTools =
+        (tool_choice.has_value() && tool_choice->type == "none")
+            ? std::nullopt
+            : tools;
     auto promptStr = tokenizer.applyChatTemplate(
-        messages, true, tools, enable_reasoning, skip_apply_chat_template);
+        messages, true, effectiveTools, enable_reasoning,
+        skip_apply_chat_template);
     out.full_prompt_tokens_count =
         static_cast<int>(tokenizer.encode(promptStr).size());
     out.prompt_tokens_count = out.full_prompt_tokens_count;
--- a/tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
+++ b/tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
@@ -85,6 +85,8 @@ struct ToolCallTaskState {
   Json::Value tool_calls;         // Array of completed tool calls
   int call_index = 0;             // Tool call counter
   bool in_json_block = false;     // Inside ```json...``` block
+  bool json_header_stripped =
+      false;  // Whether "json" header after ``` has been checked/stripped
   bool sent_tool_call_start =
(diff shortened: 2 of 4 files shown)


tt-media-server/cpp_server/include/domain/llm/chat_completion_request.hpp
```diff
@@ -272,8 +272,13 @@ struct ChatCompletionRequest : BaseRequest {
     out.messages = messages;
     out.skip_apply_chat_template = skip_apply_chat_template;
     const auto& tokenizer = tt::utils::tokenizers::activeTokenizer();
+    std::optional<std::vector<tool_calls::Tool>> effectiveTools =
```
lmuravljovTT, line 275: Even if tool_choice is none, tools should still be injected into the prompt if they are set. 
What should happen in that case is that the tool call-related model output should just be skipped. If they are leaking into the output, then the issue is likely somewhere in the deepseek_tool_call_parser 

You can see this as well [here](https://platform.claude.com/docs/en/agents-and-tools/tool-use/overview) 


> When you use tools, the API also automatically includes a special system prompt for the model which enables tool use. The number of tool use tokens required for each model are listed below (excluding the additional tokens listed above). Note that the table assumes at least 1 tool is provided. If no tools are provided, then a tool choice of none uses 0 additional system prompt tokens.

and the table below defines that even when tool_choice is `none`, the tool use system token prompt is >0

tt-media-server/cpp_server/src/services/deepseek_tool_call_parser.cpp
```diff
@@ -264,6 +270,35 @@ class DeepSeekToolCallParser : public IToolCallParser {
           return std::nullopt;
 
         } else {
+          // Strip "json" header if ``` and "json" arrived in separate tokens
```
lmuravljovTT, line 273: this should be handled now on main as well, might cause some conflicts

ksriv001: Confirmed — #4159 deleted `deepseek_tool_call_parser.cpp` (and `test_tool_call_parser.cpp`) on `main` and moved parsing to the Dynamo frontend, so these changes now conflict / are obsolete. Resolving.

PR #3713, opened 2026-05-25
Add first and block size, add a method to calculate blocks

idjuricTT:
- Add settings for first block and block size
- Add a method to calculate hashes by block
- Allow storing multiple hashes


tt-media-server/cpp_server/include/services/session_manager.hpp
```diff
@@ -137,9 +139,19 @@ class SessionManager {
   void removeFromPrefixIndex(const std::string& sessionId, uint64_t prefixHash);
 
   mutable utils::ConcurrentMap<std::string, domain::Session> sessions;
-  // Secondary index: prefix hash -> sessionIds registered under that hash.
+
+  // An entry in the prefix index: a group of sessions sharing the same prefix
+  // path, together with the remaining block hashes that follow (used for deeper
+  // prefix matching / numberOfMatchedTokens calculation).
+  struct PrefixIndexEntry {
+    std::list<std::string> sessionIds;    // sessions registered here
```
nanicicTT, line 147: We have several options here to use for ID:

```
using SessionId = boost::uuids::uuid;
// or
using SessionId = std::array<std::byte, 16>;
// or at least
using SessionId = std::string;
```

and string has some cons:
1. it has a larger memory
2. it has slower comparison than 16-byte UUID
3. It's easy to accidentally pass wrong string type

Maybe we can use

```cpp
struct SessionId {
    std::array<std::uint8_t, 16> bytes;
};
```

but string is also fine if this is not performance critical component. 

---

Also, I would not use std lists here:
This means every element is allocated separately on the heap. That is usually bad for cache locality and memory overhead.
I would go with `std::vector` if it works here. 

`std::vector<uint64_t>` is especially better for `remainingHashes`, because matching prefix tails is sequential scanning:

```
for (auto h : remainingHashes) {
    ...
}
```

That is exactly what vector is good at.


So I would make it all like:

```cpp
using SessionId = std::string; // or boost::uuids::uuid

struct PrefixIndexEntry {
    std::vector<SessionId> sessionIds;
    std::vector<uint64_t> remainingHashes;
};
```

tt-media-server/cpp_server/src/dynamo/dynamo_protocol.cpp
```diff
@@ -35,8 +35,8 @@ bool txTraceEnabled() {
     const char* v = std::getenv("DYN_TX_TRACE");
     if (v == nullptr) return false;
     std::string s(v);
-    if (s.empty() || s == "0" || s == "false" || s == "FALSE" ||
-        s == "False" || s == "off" || s == "OFF") {
+    if (s.empty() || s == "0" || s == "false" || s == "FALSE" || s == "False" ||
+        s == "off" || s == "OFF") {
       return false;
     }
     return true;
```
nanicicTT, line 42: Or


```suggestion
    return !(s.empty() || s == "0" || s == "false" || s == "FALSE" || s == "False" ||
        s == "off" || s == "OFF");
```

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -612,22 +615,38 @@ void SessionManager::addToPrefixIndex(const std::string& sessionId,
...
+    for (auto& entry : entries) {
+      auto& ids = entry.sessionIds;
+      ids.erase(std::remove(ids.begin(), ids.end(), sessionId), ids.end());
+    }
+    // Remove entries with no sessions left
+    entries.erase(std::remove_if(entries.begin(), entries.end(),
+                                 [](const PrefixIndexEntry& e) {
+                                   return e.sessionIds.empty();
+                                 }),
+                  entries.end());
```
nanicicTT, line 647: Or CPP20:

```cpp
for (auto& entry : entries) {
    std::erase(entry.sessionIds, sessionId);
}
```


So it removes all occurrences of `sessionId` from every `entry.sessionIds`.

PR #3728, opened 2026-05-25
cpp server expose pos + handle disaggregation

knovokmetTT:


idjuricTT (approved):

tt-media-server/cpp_server/include/domain/llm/sequence.hpp
```diff
@@ -98,10 +98,14 @@ class Sequence {
   bool isDisaggregated() const { return disaggregated; }
   void setDisaggregated(bool d) { disaggregated = d; }
 
+  std::optional<uint32_t> getKVCacheOffset() const { return kvCacheOffset; }
```
dmadicTT: let's call it position ID everywhere?

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -51,12 +51,17 @@ inline ds::GenerationParams makeGenerationParams(
...
       .top_k = static_cast<int32_t>(seq.getSamplingParams().top_k.value_or(-1)),
-      .disaggregated_decode = seq.isDisaggregated()};
+      .disaggregated_decode = seq.isDisaggregated(),
+      .position_id = seq.getKVCacheOffset()};
 }
 
 inline void fillSequenceFields(ds::ISRequest& req,
                                const tt::domain::llm::Sequence& seq) {
-  req.tokens.assign(seq.getTokenIds().begin(), seq.getTokenIds().end());
+  if (seq.isDisaggregated()) {
```
dmadicTT: How did disaggregation work before this change?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -633,6 +633,9 @@ void LLMService::processStreamingRequest(
       request.disaggregated,
       std::make_unique<tt::domain::llm::SamplingParams>(
           tt::utils::mapper::mapSamplingParams(request)));
+  if (request.disaggregated) {
+    sequence->setKVCacheOffset(sequence->getNumPromptTokens());
```
dmadicTT: What about adding it next to other disaggregation logic? https://github.com/tenstorrent/tt-inference-server/blob/dmadicTT/profiling-tools-may-2026/tt-media-server/cpp_server/src/services/disaggregation_service.cpp#L74

tt-media-server/cpp_server/include/domain/llm/llm_request.hpp
```diff
@@ -124,6 +124,13 @@ struct LLMRequest : BaseRequest {
   bool fast_mode = false;
   bool disaggregated = false;  // True if this is a disaggregated request
 
+  // For disaggregated decode: position in KV Cache of the migrated token (the first token
+  // produced by the prefill server) in the per-user KV cache. The decode
+  // scheduler uses this as `position_id` so the migrated token lands at the
+  // correct KV slot and `current_position` advances to it + 1. Set iff
```
nanicicTT:
```suggestion
  // correct KV slot and `current_position` advances to it + 1. Set if
```

A typo

tt-media-server/cpp_server/src/domain/llm/sequence.cpp
```diff
@@ -92,6 +94,13 @@ void Sequence::serialize(std::ostream& os) const {
   os.write(reinterpret_cast<const char*>(&disaggregated),
            sizeof(disaggregated));
   samplingParams->serialize(os);
+  bool hasKvCacheOffset = kvCacheOffset.has_value();
```
nanicicTT: `sizeof(bool)` is usually `1`, but the standard does not force the exact binary representation you probably want. For a file/network protocol, prefer an explicit **byte**: `0` or `1`:

```cpp
const uint8_t hasKvCacheOffset = kvCacheOffset.has_value() ? 1 : 0;
os.write(reinterpret_cast<const char*>(&hasKvCacheOffset), sizeof(hasKvCacheOffset));
```


Now, if you want to get fancy, you can make a small helper to serialize many fields:

```cpp
template <typename T>
void writeBinary(std::ostream& os, const T& value) {
  static_assert(std::is_trivially_copyable_v<T>);
  os.write(reinterpret_cast<const char*>(&value), sizeof(T));
}
```

and just use it as:

```cpp
const uint8_t hasKvCacheOffset = kvCacheOffset ? 1 : 0;
writeBinary(os, hasKvCacheOffset);

if (kvCacheOffset) {
  writeBinary(os, *kvCacheOffset);
}
```

Now, not sure about endianness, this still writes raw `uint32_t` bytes in the machine’s native endianness.

nanicicTT (approved):

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
knovokmetTT: TLE received full prompt(not delta) and calced pos id on its own

tt-media-server/cpp_server/src/services/llm_service.cpp
knovokmetTT: how? disaggregation service doesnt build a sequence object , but the LLMRequest

PR #3742, opened 2026-05-26
Prefix cache by block happy path

idjuricTT:


ljovanovicTT (approved):

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -277,7 +255,11 @@ void LLMPipeline::dispatchGeneration(
           "sessionId: {}",
           request.sessionId.value_or("none"));
       disaggregationService_->handleStreamingRequest(
-          request, sessionInfo.registrationHash.value_or(0), cb);
+          request,
+          sessionInfo.registrationHashes.empty()
+              ? 0
+              : sessionInfo.registrationHashes.front(),
```
ljovanovicTT, line 261: Why are we sending only the first hash?

idjuricTT: Mistake 😄

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -498,40 +497,82 @@ void SessionManager::retryFailedDeallocs() {
...
+  if (blockHashes.empty()) {
+    return std::nullopt;
+  }
+
+  const uint64_t keyHash = blockHashes.front();
+  const size_t firstBlockTokens = tt::config::kvCacheFirstBlockSize();
+  const size_t blockTokens = tt::config::kvCacheBlockSize();
+
+  // Build the caller's remaining hashes for comparison (blockHashes[1:]).
+  std::list<uint64_t> callerRemaining(blockHashes.begin() + 1,
```
nanicicTT, line 514: I usually do not like lists in CPP and usually vector should be preferred :)

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -498,40 +497,82 @@ void SessionManager::retryFailedDeallocs() {
...
+      // Count consecutive matching remaining hashes.
+      size_t matched = 0;
+      auto callerIt = callerRemaining.begin();
+      auto entryIt = entry.remainingHashes.begin();
+      while (callerIt != callerRemaining.end() &&
+             entryIt != entry.remainingHashes.end() && *callerIt == *entryIt) {
+        ++matched;
+        ++callerIt;
+        ++entryIt;
+      }
```
nanicicTT, line 538: This is exactly [mismatch](https://en.cppreference.com/cpp/algorithm/mismatch):

> Returns a pair of iterators to the first mismatching of elements from [first1, last1) and a range starting from first2

so 

> It finds the first position where two ranges stop being equal.

And it returns

> std::pair with iterators to the first two non-equal elements.

```cpp
auto oldCallerIt = callerIt;

std::tie(callerIt, entryIt) = std::mismatch(
    callerIt, callerRemaining.end(),
    entryIt, entry.remainingHashes.end()
);

matched += std::distance(oldCallerIt, callerIt);
```


and we use `tie` to unpack tuples. 

By using STL algorithms, over the loops, we make intent more clear and code more readable :)

tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -498,40 +497,82 @@ void SessionManager::retryFailedDeallocs() {
...
+      // key hash itself counts as 1 block match.
+      size_t totalBlocks = 1 + matched;
+      for (const auto& sid : entry.sessionIds) {
```
nanicicTT, line 541: with

```cpp
candidates.reserve(candidates.size() + entry.sessionIds.size());
```

in between we make sure candidates have enough allocated capacity to append all `entry.sessionIds` without reallocating during the loop.

nanicicTT (approved):

PR #3743, opened 2026-05-26
Separate model spec.py

### LINKED ISSUE #3008: Breakdown of model_spec.py and improvements
Current `model_spec.py` should be simplified and improved upon so it can be . 

Proposed solution is the following (3 part solution):

1. Have `model_catalog.json` were we will have 2 catalogs actually: `dev` used for development and `prod` which will our customer use (`dev` will be promoted to `prod`)
We can also have in CI `--mode dev` (by default `prod`) to use specific catalog in CI
2. Have `api.py` which will actually have helper functions (extracted from current `model_spec.py`): weights funcs etc
3. Have one (or maybe 2?) json config: model_perf_reference_actual_target.json: we will have `targets` for each model on spicific chip and `actual values` (cleaner to have 2)

For `#1` we can have structure
- weights
- chips (max_conc, ISL, etc)
- metal commits

We can agree upon the structure, but for now it can be list of json objects (maybe take a look at `release_model_spec.json` for reference/example/inspiration)
Example:
```
[
    {
        model: "stable-diffusion-xl-base-1.0",
        device: "n150",
        targets: {
            ...
        }
    },
    {
        model: "Motif-Image-6B-Preview",
        device: "galaxy",
        target: {
            ...
        }
    }
]
```

vmaksimovicTT:
There is a broader goal of using 2 model catalogues. 1 for prod specification -> which is updated only through the release proces, and is in sync with release_model_spec.json, and another for dev catalogue, where anyone can add new models, or change the specification for existing ones. This sepc will be used for local testing, CI, etc.

Before we separate one catalogue into 2, this PR extracts the configuration part of model_spec.py from the actual functional logic/schema.

Here are some ModelCI runs to validate the yaml is loaded correctly, and workflows still work as expected:
https://github.com/tenstorrent/tt-shield/actions/runs/26498960690
https://github.com/tenstorrent/tt-shield/actions/runs/26511313258


fivanovicTT:
Hey @vmaksimovicTT, tagging this [issue](https://github.com/tenstorrent/tt-inference-server/issues/3008) here since this issue was the product of our meeting last month

fivanovicTT (approved):

PR #3756, opened 2026-05-26
Improve Test Gate reliability and reporting

ztorlakTT:
- Add default workflow permissions, job timeouts, and centralized Python version variables.
- Split unit tests into separate main/media jobs with a dedicated summary/reporting job.
- Fix coverage reporting behavior and make PR comment steps non-blocking where appropriate.
- Use authenticated submodule initialization and pin the disk cleanup action for reproducibility.

### DIFF
--- a/.github/workflows/test-gate.yml
+++ b/.github/workflows/test-gate.yml
@@ -13,10 +13,19 @@ concurrency:
   group: test-gate-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+permissions:
+  contents: read
+
+env:
+  PYTHON_VERSION_DEFAULT: "3.10"
+  PYTHON_VERSION_DYNAMO: "3.12"
+  PYTHON_VERSION_FORGE: "3.12"
+
 jobs:
   detect-changes:
     name: Detect Code Changes
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     outputs:
       should-test: ${{ steps.filter.outputs.code }}
       cpp-server: ${{ steps.filter.outputs.cpp_server }}
@@ -48,13 +57,14 @@ jobs:
     if: ${{ needs.detect-changes.outputs.should-test == 'true' }}
     name: Lint & Format Check
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: ${{ env.PYTHON_VERSION_DEFAULT }}
 
       - name: Install ruff
         run: pip install ruff==0.15.0
@@ -70,6 +80,7 @@ jobs:
     if: ${{ needs.detect-changes.outputs.media-server-deps == 'true' }}
     name: Verify Media Server Dep Install
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     defaults:
       run:
         working-directory: tt-media-server
@@ -79,7 +90,7 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v7
         with:
-          python-version: "3.10"
+          python-version: ${{ env.PYTHON_VERSION_DEFAULT }}
           enable-cache: true
 
       - name: Install media-server requirements (strict, no fallback)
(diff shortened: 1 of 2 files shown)


.github/workflows/test-gate.yml
```diff
@@ -108,14 +119,16 @@ jobs:
     if: ${{ github.event_name == 'pull_request' }}
     name: Manage docker build reminder comment
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     permissions:
-      pull-requests: write
+      issues: write
```
dmadicTT, line 124: Just curious
What's the motive for the permission changes?

.github/workflows/test-gate.yml
```diff
@@ -13,10 +13,19 @@ concurrency:
   group: test-gate-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+permissions:
+  contents: read
+
+env:
```
dmadicTT, line 19: nice!

dmadicTT: Can we just use `3.12` everywhere?

.github/workflows/test-gate.yml
```diff
@@ -553,7 +610,7 @@ jobs:
 
       - name: Init tt-llm-engine submodule (without tt-metal)
         working-directory: .
-        run: git submodule update --init --remote tt-media-server/cpp_server/tt-llm-engine
```
dmadicTT, line 556: Do we want to remove the `--remote`? 
I know @knovokmetTT and @nanicicTT added that on purpose

dmadicTT (approved):

.github/workflows/test-gate.yml
dmadicTT: btw what's the motive to change this?

nanicicTT: I would remove it, since local branches do not depend on main, so the local build can pass. 
We did not add it, we removed it from some other place.

nanicicTT: More context:
Without `--remote` flag git checks out the submodule at the specific commit hash that is recorded and pinned in the parent repository. This guarantees that no matter where we build, the submodule code is exactly the version the developer tested.
With `--remote` git ignores the pinned commit hash in the parent repository. Instead, it goes to the submodule's repository and checks out the latest commit on the submodule's remote tracking branch (usually **main** or **master**).

The local branch should rely on its own pinned submodule commit (which is stable).
The `--remote` flag is forcing the local branch to depend on the submodule's **main** branch (which might be unstable/broken, or having some old dependency, not updated one).

But, now the question arises why it was added in the first place, and [this principle](https://fs.blog/chestertons-fence/) says that we should not remove things if we do not understand why they were added :)

.github/workflows/test-gate.yml
ztorlakTT: Yes, but I'd do that in a separate PR

.github/workflows/test-gate.yml
ztorlakTT: Just narrowing down the permission. issues: write is enough for this job 😄

.github/workflows/test-gate.yml
```diff
@@ -259,74 +371,51 @@ jobs:
...
           MEDIA_PASSED=$(echo "$MEDIA_VALUES" | cut -d'|' -f1)
           MEDIA_SKIPPED=$(echo "$MEDIA_VALUES" | cut -d'|' -f2)
-          MEDIA_FAILED=$(echo "$MEDIA_VALUES" | cut -d'|' -f3)
-          MEDIA_ERRORS=$(echo "$MEDIA_VALUES" | cut -d'|' -f4)
-          MEDIA_COLLECTION_ERRORS=$(echo "$MEDIA_VALUES" | cut -d'|' -f5)
           MEDIA_TOTAL_FAILED=$(echo "$MEDIA_VALUES" | cut -d'|' -f6)
           MEDIA_TOTAL=$(echo "$MEDIA_VALUES" | cut -d'|' -f7)
 
-          # overall
+          MAIN_EXIT_CODE=$(cat main_exit_code.txt 2>/dev/null || echo 1)
```
nanicicTT, line 393: Do we delete somewhere this main_exit_code? And main_output file?

nanicicTT: Does not matter, ">" overwrites the file, ">>" appends in bash, so all good. :)

nanicicTT (approved):

Issue #3703, opened 2026-05-25
Introduce hashing during streaming


idjuricTT:
Safe into slot's KV hash table during streaming

PR #3760, opened 2026-05-26
Integrate Block Hash Accumulator

ljovanovicTT:
## Issue
[#3703 Introduce hashing during streaming](https://github.com/tenstorrent/tt-inference-server/issues/3703)

## Problem
In order to maximize the number of reusable matched blocks, we need to update the prefix index as soon as new tokens are generated. The approach is to accumulate all tokens that are not yet present in the index (and corresponding KV cache) the delta prompt and newly generated output tokens and update the index once generation completes. This allows the next message in a multi-turn conversation to immediately reuse the previously generated state.

One important detail is that we always operate on whole blocks and intentionally ignore partial blocks. This can occasionally leave some tokens unused for reuse purposes, but since the block size is relatively small (32 tokens), the performance impact should be minimal.


tt-media-server/cpp_server/include/domain/session.hpp
```diff
@@ -83,6 +89,26 @@ class Session {
...
+   * Called once per request when session routing is resolved.
+   *
+   * @param initialHashes Block hashes computed from the prompt
+   * @param partialBlockTokens Tokens from prompt's incomplete final block
+   * @param onBlockComplete Callback invoked when a block hash is computed
+   */
+  void initPrefixAccumulator(
+      std::vector<uint64_t> initialHashes,
+      std::vector<int64_t> partialBlockTokens,
+      std::function<void(const std::string&, const std::vector<uint64_t>&)>
```
idjuricTT, line 98: Suggestion: Why not using the same hashes all the time and just do the update at the end of hashes?
We basically don't care what's old/ new at that point, right?

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -20,6 +20,39 @@
...
+ * passed to the streaming accumulator.
+ */
+std::vector<int64_t> extractPartialBlockTokens(
+    const std::vector<int>& promptTokens, size_t completedBlockCount) {
+  const size_t firstBlockSize = tt::config::kvCacheFirstBlockSize();
+  const size_t blockSize = tt::config::kvCacheBlockSize();
+
+  // Calculate where complete blocks end
+  size_t completeTokens = 0;
+  if (completedBlockCount > 0) {
```
idjuricTT: This is already found out after applyDelta method is called.
At that point you have:

- Number of skipped tokens
- Remaining in prompt

Suggestion:

Use prompt as a delta variable and during streaming add to it.

At the end of streaming / non streaming just split that into blocks calling the same hashesByBlock method and update session

knovokmetTT (approved):

tt-media-server/cpp_server/src/api/llm_controller.cpp
```diff
@@ -268,9 +268,18 @@ LLMController::makeStreamingCallback(std::shared_ptr<ResponseWriter> writer,
   return [writer = std::move(writer), session](const LLMStreamChunk& chunk,
                                                bool isFinal) {
     if (writer->isDone()) return;
+
+    // Feed token to prefix accumulator for incremental index updates
+    if (session && !chunk.choices.empty() && chunk.choices[0].token_id) {
```
idjuricTT, line 271: Does this cover non-streaming approach?

tt-media-server/cpp_server/include/domain/session.hpp
ljovanovicTT: Maybe the name is a bit misleading, but the since we are chaining blocks we can either recompute the whole prompt, or use the last element of initialHashes as a parent for the new hash - with this `initialHashes` field we going for the second approach and not recomputing

tt-media-server/cpp_server/src/api/llm_controller.cpp
ljovanovicTT: Yes, the non-streaming behavior is just a wrapper around streaming and it goes through this code-path

idjuricTT (approved):

PR #3765, opened 2026-05-27
Whisper: propagate language, task and prompt to generator

sott0n:
### Propagate language, task, prompt
`TTWhisperRunner._create_generation_params()` was setting `language`, `task`, and `prompt` as dynamic attributes on `GenerationParams`, but tt-metal's `GenerationParams` does not declare those fields — its docstring says they "must be passed as separate parameters to `generate()`". They were silently dropped, so the defaults (`language="en"`, `task="transcribe"`, `prompt=None`) were always used and non-English `AUDIO_LANGUAGE` had no effect.

Fix: pass `language` / `task` / `prompt` as explicit kwargs to `generator.generate()`. `language` / `task` come from `self.settings`; `prompt` is per-request and plumbed through `_execute_pipeline*`.

### Batch handling for prompt
tt-metal's `WhisperGenerator` treats `prompt` as batch-homogeneous (one value per batch), but the API accepts a per-request `prompt`. When multiple requests are batched, the first request's `prompt` is used and a warning is logged if requests disagree — matching the existing first-request-wins behavior in this runner.

### Test plan
- [x] `AUDIO_LANGUAGE=Japanese AUDIO_TASK=transcribe` on t3k: Japanese audio now transcribes in Japanese (non-streaming + streaming).

### DIFF
--- a/tt-media-server/tt_model_runners/whisper_runner.py
+++ b/tt-media-server/tt_model_runners/whisper_runner.py
@@ -157,6 +157,7 @@ class TTWhisperRunner(BaseMetalDeviceRunner):
                     request._audio_array,
                     request.stream,
                     self._create_generation_params(request),
+                    prompt=request.prompt,
                 )
 
                 if request.stream:
@@ -184,12 +185,6 @@ class TTWhisperRunner(BaseMetalDeviceRunner):
             generation_params.no_speech_threshold = request.no_speech_threshold
         if request.return_timestamps is not None:
             generation_params.return_timestamps = request.return_timestamps
-        if request.prompt is not None:
-            generation_params.prompt = request.prompt
-        if self.settings.audio_language is not None:
-            generation_params.language = self.settings.audio_language
-        if self.settings.audio_task is not None:
-            generation_params.task = self.settings.audio_task
 
         return generation_params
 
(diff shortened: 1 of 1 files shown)


idjuricTT (commented):
@ztorlakTT could you please double check this?
Makes sense for me

vpetrovicTT:
I think that we need to doublecheck that the `prompt = ""` doesn't get passed to ttnn... Think it will force undesired behaviour of the sys.

sott0n:
@vpetrovicTT After checking, tt-metal's `WhisperGenerator.generate()` branches on `prompt is not None`, so passing `prompt=""` would incorrectly prepend the `<|startofprev|>` token prefix. 
https://github.com/tenstorrent/tt-metal/blob/main/models/demos/audio/whisper/tt/whisper_generator.py#L938

With `prompt=None` it takes the normal forced_decoder_ids path, so I normalized to `None` at the ttnn boundary (`prompt or None`).

sott0n:
addressed the `prompt = ""` at https://github.com/tenstorrent/tt-inference-server/pull/3765/changes/fa3093a67872a9e4d56da2e038ef36cc00e4261f

fivanovicTT (approved):

PR #3774, opened 2026-05-27
Unify test gate python version

ztorlakTT:
Use a single `PYTHON_VERSION` for all Test Gate Python setup steps, set to 3.12

`charts/tt-inference-server/values.yaml` is regenerated so the committed Helm values stay in sync with `workflows/model_spec.py` when Test Gate runs under Python 3.12

### DIFF
--- a/charts/tt-inference-server/values.yaml
+++ b/charts/tt-inference-server/values.yaml
@@ -249,6 +249,39 @@ models:
                 value: '900000'
               - name: VLLM_TARGET_DEVICE
                 value: tt
+      p300x2:
+        defaultImpl: gpt_oss
+        impls:
+          gpt_oss:
+            progressDeadlineSeconds: 7200
+            image:
+              repository: ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64
+              tag: 0.14.0-747215b-7678b70
+            probes:
+              liveness:
+                initialDelaySeconds: 3600
+              readiness:
+                initialDelaySeconds: 3600
+            resources:
+              requests:
+                memory: 300Gi
+            env:
+              - name: ARCH_NAME
+                value: blackhole
+              - name: MESH_DEVICE
+                value: (1, 4)
+              - name: TORCHDYNAMO_DISABLE
+                value: '1'
+              - name: TT_MESH_GRAPH_DESC_PATH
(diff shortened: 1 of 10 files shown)


charts/tt-inference-server/values.yaml
```diff
@@ -3722,6 +3815,66 @@ models:
                 value: tt
               - name: WH_ARCH_YAML
                 value: wormhole_b0_80_arch_eth_dispatch.yaml
+      t3k:
+        defaultImpl: tt_transformers
+        impls:
+          tt_transformers:
+            progressDeadlineSeconds: 5400
+            image:
+              repository: ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64
```
nanicicTT, line 3824: You can use, I believe Anchor (`&`) and Alias (`*`) to define variable in YAML files to reuse them here, because this 
`ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64`
repeats 7 times here and would require 7 changes of link if it needs to change.

nanicicTT: like somewhere above

```yaml
imageRepository: &vllm_tt_image_repo ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64
```

and then

```yaml
image:
        repository: *vllm_tt_image_repo
```

nanicicTT: Note: YAML anchors work only inside the same YAML document/file after parsing. They are not variables like in programming languages, and they usually cannot be shared across separate YAML files unless your tooling adds that feature.
But there are 2 yaml files here as I can see and this is used only in values.yaml?

nanicicTT: _**Reference:**_ https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

ztorlakTT: This file is generated by `workflows.helm_generator`, so I’d avoid hand-editing it

nanicicTT: ah, ok, I don't know what it is, but in that case ok :)

idjuricTT (approved):

vmaksimovicTT (approved):

PR #3779, opened 2026-05-27
Align KV Cache Block Sizes With Tiles

ljovanovicTT:
## Problem
We need to ensure that the delta prompt provided during multi-turn conversations is aligned with the tile size of 32. With this PR, the default block values have been updated to be divisible by 32.


idjuricTT (approved):

ztorlakTT (approved):

nanicicTT (approved):

tt-media-server/cpp_server/src/config/settings.cpp
```diff
@@ -553,14 +554,34 @@ size_t maxContextLength() {
...
+        "KV_CACHE_FIRST_BLOCK_SIZE", defaults::KV_CACHE_FIRST_BLOCK_SIZE));
+    if (value % 32) {
+      TT_LOG_WARN(
+          "[Config] KV_CACHE_FIRST_BLOCK_SIZE={} is not divisible by 32, using "
+          "default={}",
+          value, defaults::KV_CACHE_FIRST_BLOCK_SIZE);
+      return defaults::KV_CACHE_FIRST_BLOCK_SIZE;
+    }
+    return value;
+  }();
```
nanicicTT, line 579: Since it is lambda function, but repeats the logic from [this line 557](https://github.com/tenstorrent/tt-inference-server/pull/3779/changes/a05048271c1067daaa6f3b4c8c94709176c60dec..1b063c98953397e0a505b6aa03f5d64ae6988993#diff-a136d442fe3ea90b4ef724a388851c7ee92ef49de3509a438ab1c41d345b030eR557) then I guess it could be reused in a function and be called here with different KV_CACHE param...

ljovanovicTT: Nice catch, done!

nanicicTT (approved):

Issue #3775, opened 2026-05-27
Token adjustment on sending


idjuricTT:
@pavlepopovic  to provide details

pavlepopovic:
So lets go over the scenario for tile alignement:

Multi turn + no decode->prefill migration (decode kv gets recomputed)
Say prefill has 657 tokens to process. This means 640 tokens end up on device 0, and 17 tokens end up on device 1 (padded within a tile of 32). actual_start for this chunk would be 0, and actual end would be 640
We migrate all 640 tile aligned tokens, and a padded full tile of 32 tokens, which will contain 17 valid ones to decode
Decode does something and produces 640 tokens, and the following user prompt is 1280 tokens.
We should be running new prefill with (640 + 1280) tokens
but we are actually appending 17 last tokens from the first prefill promt (in order to reprefill them to avoid writing into partial tiles). So total prompt would be 17 + 640 + 1280.
We’d like actual_start for this chunk to be 640 (not 657), and actual end 640 + 17 + 1280

Multi turn d -> p migration (decode kv gets reused)
Again, prefill has 657 tokens to process, same situation as above.
same as above
decode produces 640 tokens:
in total it contains 657 + 640 tokens =  tokens 1297 -> == 40 full tiles and 1 partial tile with 17 tokens.

migration -> we migrate to prefill starting at position 640 (we will overwrite prefill’s last partial tile with decode-produced full tile), and ending with position 1279 (640 tokens in total). Don’t want to migrate partial tiles back to prefill
prefill gets new user prompt of 1280, to which we prepend decode’s last 17 tokens, resulting in prefill of 1297.
when migrating again from prefill to decode, we start with position 1279 (overwrite decode’s last partial tile), and end with position 2576 (resulting in all 1297 new tokens being transferred, with 40 full tiles and 1 partial tile to decode)

idjuricTT:
Implemented

PR #3807, opened 2026-05-29
Get slot on prefill & apply Delta prompt

idjuricTT:
Changes

- Ask for a slot on prefill
- Apply delta prompt for prefill that is dividable by 32


tt-media-server/cpp_server/include/services/disaggregation_service.hpp
```diff
@@ -42,12 +46,24 @@ class DisaggregationService {
...
+
  private:
   void setupSocketHandlers();
+  void applyDeltaPrompt(LLMRequest& req, uint32_t matchedTokens);
 
   tt::config::LLMMode mode;
   std::shared_ptr<LLMService> llmService;
   std::shared_ptr<sockets::InterServerService> socketService;
+  std::shared_ptr<SessionManager> sessionManager_;
+  trantor::EventLoopThread eventLoopThread_;
```
nanicicTT: minor nitpick: code consistency. 
Class members are either with underscore at the end or not, but this is mixed

knovokmetTT (approved):

tt-media-server/cpp_server/include/services/disaggregation_service.hpp
idjuricTT: Fixed, thanks 😄

nanicicTT (approved):

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
```diff
@@ -201,6 +210,93 @@ void DisaggregationService::start() {
...
+  }
+
+  // The remaining (unmatched) prompt tokens that will be prefilled must be
+  // aligned to 32 so the prefill kernel can operate on full tiles.  If the
+  // remainder isn't divisible by 32, we pull back some matched tokens into
+  // the delta to pad the remainder up to the next multiple of 32.
+  constexpr uint32_t kAlignment = 32;
+  const uint32_t totalTokens = static_cast<uint32_t>(tokens.size());
+  const uint32_t remainder = totalTokens - matchedTokens;
+  const uint32_t alignedRemainder =
```
idjuricTT, line 227: CC @pavlepopovic @ipotkonjak-tt @tt-asaigal for rounding on 32

pavlepopovic: Hey Igor,
I think this logic is not correct
Prefill should not repeat matched tokens, but it needs number of tokens stored in kv to be divisible by 32, upon applying a new prompt.

So let's go over artificial example:
We have turn 1 prefill, with 133 tokens. All tokens get sent to prefill (padded up to 5k ofc), kv cache in this case gets populated with 160 tokens. (4 tiles of 32 => totaling 128 tokens, + 1 tile containing 5 tokens and 27 paded tokens). Matched Tokens in this case returns 128 (which is number of full tiles within kv cache). When migrating, we will send full 5 tiles, and decode will know (confirmed with them) what do to with tokens in partial tile.
Now we have turn 2 prefill, with 412 new tokens.
What we can keep in kv cache, and don't need to recompute is 128 tokens (matchedTokens). So in this case we will send to prefill 5 previous tokens (of the first 133), and 412 new ones, starting at position id corresponding to token 128 (which is 128 - 1)

pavlepopovic: above is case with no d->p migration. Have both cases covered here https://github.com/tenstorrent/tt-inference-server/issues/3775

PR #3826, opened 2026-05-29
Track Think Tokens for Prefix Caching

ljovanovicTT:
## Problem
For reasoning models, there is a mismatch between the conversation history sent by the client during multi-turn interactions and the actual contents of the KV cache. The discrepancy comes from reasoning ("thinking") tokens: these tokens are stored in the KV cache, but are typically omitted by the client when sending follow-up requests.

This caused issues with KV cache rewinding, which is required for slot sharing and prefix reuse across users. While the server could determine how many visible (non-reasoning) tokens matched a previously cached prefix, it did not have enough information to identify the correct rewind position within the KV cache slot.

To address this, we now track both the block hash and the cumulative count of reasoning tokens associated with each block. This allows the server to accurately map matched prefixes to their corresponding positions in the KV cache and perform rewinds at the correct memory location.

For example, given messages:
```
  {
      "user": "...",
      "assistant": "...", 
      "user": "....",
  }   
```
  
We would have a structure like:
```
  block 1: hash(user1), accumulatedThink=0
  block 2: hash(user1 + assistant), accumulatedThink=100
  block 3: hash(assistant), accumulatedThink=100
  ...
  block N: hash(assistant), accumulatedThink=100

  block N+1: hash(user2), accumulatedThink=100
  block N+2: hash(user2 + assistant2), accumulatedThink=350
  ...
  block M: hash(assistant2), accumulatedThink=350
```

The KV cache rewind is then calculated as: `matchedTokens + lastMatchedBlock.accumulatedThink`

## Risks
- With DS setup, KV cache on Prefill is never aware of reasoning blocks
- Cancellation mid reasoning could lead to inconsistent state - this will be addressed as a separate PR


tt-media-server/cpp_server/include/services/session_manager.hpp
```diff
@@ -147,11 +151,17 @@ class SessionManager {
...
+    uint64_t hash;
+    uint32_t accumulatedThinkTokens;
+  };
+
   struct PrefixIndexEntry {
-    std::list<std::string> sessionIds;    // sessions registered here
-    std::list<uint64_t> remainingHashes;  // subsequent block hashes
+    std::list<std::string> sessionIds;              // sessions registered here
+    std::list<RemainingBlockInfo> remainingBlocks;  // subsequent block info
+    uint32_t keyBlockThinkTokens = 0;  // think tokens at key hash block
```
idjuricTT, line 164: This is cross-blocks, right? Like total thing tokens across the session

idjuricTT (approved):

knovokmetTT (approved):

tt-media-server/cpp_server/src/utils/conversation_hasher.cpp
```diff
@@ -181,4 +183,57 @@ std::vector<uint64_t> getPrefixCacheHashesByBlocks(std::span<const int> tokens,
   return hashes;
 }
 
+std::vector<BlockHashInfo> getPrefixCacheHashesByBlocksWithThinking(
```
dmadicTT, line 186: would be good to unit test this, as well as test it in broader context, with the parent blocks

dmadicTT (approved):

PR #3835, opened 2026-05-31
Support limiting max input length via Dynamo

idjuricTT:
Throw error 400 when ISL is too large:

image


tt-media-server/cpp_server/src/dynamo/dynamo_endpoint.cpp
```diff
@@ -199,6 +199,31 @@ GenerateHandler DynamoEndpoint::makeGenerateHandler() {
       svc->abortRequest(taskId);
     };
 
+    // Reject requests that exceed the maximum context length early, before
+    // allocating a session or slot.
+    const size_t maxContextLength = tt::config::maxContextLength();
+    const size_t promptTokens =
```
madicd, line 205: Prompt tokens and actual context length will most probably differ due to thinking tokens being in the (existing) context

madicd (approved):

dmadicTT (approved):

tt-media-server/cpp_server/src/dynamo/dynamo_endpoint.cpp
idjuricTT: True. I’ll try to get more info on this. Maybe it’s the most fair to just limit based on ISL

PR #3842, opened 2026-06-01
Limit input by ISL only

idjuricTT:


tt-media-server/cpp_server/src/config/settings.cpp
```diff
@@ -592,6 +592,12 @@ size_t maxContextLength() {
   return cached;
 }
 
+size_t maxISL() {
+  static const size_t cached =
+      static_cast<size_t>(envUlong("MAX_ISL", defaults::MAX_ISL));
```
nanicicTT, line 597: this is cool but there is a tradeoff (that might be ok, but just a food for thought):
You freeze the value at first use. If someone does `setenv("MAX_ISL", "999", 1)` after your first call, `maxISL()` will still return the **old cached value**. That's, I guess, what you want for a config constant, but if you need live reconfiguration, then drop the static. 


BTW, `setenv` has the following signature
```cpp
int setenv(const char *name, const char *value, int overwrite);
```

so `1` above (3rd param) from my example means (overwrite = true).

ljovanovicTT (approved):

ztorlakTT (approved):

tt-media-server/cpp_server/src/config/settings.cpp
nanicicTT: Djordje told me we should never change env variables, so all good I guess, resolving.

PR #3850, opened 2026-06-01
Fix ZMQ send queue

ztorlakTT:
This fixes ThreadSanitizer warnings in the shared ZMQ send queue paths used by ZmqSocketTransport and ZmqPrefillRouter. The send queue state is grouped into a private SendQueue helper, and the condition-variable wake predicate now uses an atomic hasItems flag instead of reading the protected deque directly.


tt-media-server/prefill_gateway/include/gateway/zmq_prefill_router.hpp
```diff
@@ -57,6 +57,14 @@ class ZmqPrefillRouter {
     std::promise<bool> result;
   };
 
+  struct SendQueue {
```
lmuravljovTT: We have the same struct defined twice, does it make sense to extract it? Or could we use _BlockingQueue_ defined in _concurrent_queue.hpp_?

ztorlakTT: I extracted ZMQ send queue into a shared helper. The existing BlockingQueue wait predicate reads the deque and could reintroduce the sanitizer warning this change is avoiding

lmuravljovTT (approved):

idjuricTT (approved):

PR #3860, opened 2026-06-01
add previous_response_id mapping

lmuravljovTT:
## Summary

Adds OpenAI Responses API `previous_response_id` continuation support, enabling multi-turn KV-cache reuse keyed on response ids rather than content-prefix hashes.

When a request arrives with `previous_response_id`, the server resolves the existing session/slot that served that prior turn and prefills only the new delta tokens (everything after what the slot already cached), avoiding a full re-prefill of the conversation history.

## Key Changes

- **`LLMRequest`** — new optional `previousResponseId` / `responseId` fields parsed from the Dynamo endpoint's raw JSON (`previous_response_id`, `id`, `request_id`).
- **`Session`** — stores its current response id so close/evict can remove the matching index entry.
- **`SessionManager`** — new `responseIdIndex` (response-id → session + cached prefix length) with three operations:
  - `tryAcquireByResponseId` — atomic lookup + in-flight mark (mirrors `tryAcquireByPrefixHash`).
  - `registerResponseId` — upsert with re-key support; `cachedLen=0` leaves existing length untouched.
  - `removeFromResponseIdIndex` — cleanup on session close/evict.
- **`LLMPipeline::resolveSession`** — new "Layer 1a" path that routes by response id before falling back to prefix-hash routing. Uses `applyResponseIdDelta` to shrink the prompt to `tokens[cachedLen:]`.
- **`LLMController::makeStreamingCallback`** — on the final chunk, records the slot's total processed tokens (prompt + generated) under the turn's response id for the next turn.
- **`ResponseWriter`** — exposes `totalProcessedTokens()` (prompt + completion tokens committed to the slot).
- **Tests** — comprehensive unit tests covering register/acquire, re-key, in-flight rejection, close cleanup, and multi-turn prefix growth.

## How It Works
image


tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -141,9 +141,83 @@ void LLMPipeline::resolveSession(
...
+            "[LLMPipeline] Response-id delta taskId={} matchedTokens={} "
+            "thinkTokens={} deltaTokens={}",
+            req->task_id, matchedTokens, thinkTokens, req->prompt_tokens_count);
+
+        if (auto* deltaTokens = std::get_if<std::vector<int>>(&req->prompt)) {
+          req->session->initTokenAccumulator(
+              *deltaTokens, routingInfo.blocks,
+              [mgr = sessionManager_](
+                  const std::string& sessionId,
+                  const std::vector<tt::utils::BlockHashInfo>& blocks) {
```
idjuricTT, line 189: @ljovanovicTT  could you please check this so we don't get missed hashes again?

idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
ljovanovicTT: This is a potential issue since we call `applyDeltaPrompt` before `registerPrefixHash`

ljovanovicTT: @lmuravljovTT can you try multi-turn conversation with at least 3 turns and check if prefix matches?

PR #3870, opened 2026-06-01
Add yolox nano

ddilbazTT:
Adding yolox model through tt-inference-server
n150 test: https://github.com/tenstorrent/tt-shield/actions/runs/27356486363
p150 test: https://github.com/tenstorrent/tt-shield/actions/runs/27358044836

### DIFF
--- a/.github/workflows/test-gate.yml
+++ b/.github/workflows/test-gate.yml
@@ -921,6 +921,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt
           python -m pip install -r tt_model_runners/forge_runners/requirements.txt
+          python -m pip install --no-deps --no-build-isolation -r tt_model_runners/forge_runners/requirements-no-build-isolation.txt
           python -m pip install pytest pytest-asyncio
 
       - name: Run CPU-only tests
--- a/benchmarking/benchmark_targets/model_performance_reference.json
+++ b/benchmarking/benchmark_targets/model_performance_reference.json
@@ -6812,6 +6812,34 @@
             }
         ]
     },
+    "yolox_nano": {
+        "n150": [
+            {
+                "max_concurrency": 1,
+                "num_inference_steps": 1,
+                "task_type": "cnn",
+                "targets": {
+                    "theoretical": {
+                        "ttft_ms": 268,
+                        "tput_user": 3.72
+                    }
+                }
+            }
+        ],
+        "p150": [
+            {
+                "max_concurrency": 1,
+                "num_inference_steps": 1,
+                "task_type": "cnn",
+                "targets": {
+                    "theoretical": {
+                        "ttft_ms": 148,
+                        "tput_user": 6.72
+                    }
+                }
+            }
+        ]
+    },
     "speecht5_tts": {
         "n150": [
             {
--- a/tt-media-server/config/constants.py
+++ b/tt-media-server/config/constants.py
@@ -45,6 +45,7 @@ class SupportedModels(Enum):
     GEMMA_4_31B_IT = "google/gemma-4-31B-it"
     FALCON3_7B_INSTRUCT = "tiiuae/Falcon3-7B-Instruct"
     Z_IMAGE_TURBO = "Tongyi-MAI/Z-Image-Turbo"
+    YOLOX_NANO = "Megvii-BaseDetection/YOLOX-Nano"
 
 
 # MODEL environment variable
@@ -93,6 +94,7 @@ class ModelNames(Enum):
     GEMMA_1_1_2B_IT = "gemma-1.1-2b-it"
     GEMMA_4_31B_IT = "gemma-4-31b-it"
     FALCON3_7B_INSTRUCT = "Falcon3-7B-Instruct"
+    YOLOX_NANO = "yolox_nano"
     Z_IMAGE_TURBO = "Z-Image-Turbo"
 
 
@@ -130,6 +132,7 @@ class ModelRunners(Enum):
     TT_XLA_SEGFORMER = "tt-xla-segformer"
     TT_XLA_UNET = "tt-xla-unet"
     TT_XLA_VIT = "tt-xla-vit"
+    TT_XLA_YOLOX_NANO = "tt-xla-yolox-nano"
     TRAINING_LORA = "training-lora"
     TRAINING_GEMMA_LORA = "training-gemma-lora"
     LORA_SINGLE_CHIP = "lora-single-chip"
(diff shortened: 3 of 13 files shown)


benchmarking/benchmark_targets/model_performance_reference.json
```diff
@@ -6706,7 +6706,7 @@
                 }
             }
         ],
-        "n300": [
+        "p150": [
```
ljovanovicTT: Do we not want `n300` for `resnet-50` anymore?

benchmarking/benchmark_targets/model_performance_reference.json
```diff
@@ -6776,6 +6776,34 @@
...
     },
+    "yolox_nano": {
+        "n150": [
+            {
+                "max_concurrency": 1,
+                "num_inference_steps": 1,
+                "task_type": "cnn",
+                "targets": {
+                    "theoretical": {
+                        "ttft_ms": 600,
```
ljovanovicTT: Theoretical `ttft` seems quite different for n150 and n300 - how did we get these values?

benchmarking/benchmark_targets/model_performance_reference.json
```diff
@@ -6776,6 +6776,34 @@
...
+                        "tput_user": 20
+                    }
+                }
+            }
+        ],
+        "n300": [
```
ljovanovicTT: From the rest of the changes it seems that only n150 and p150 are supported - why do we have n150 here?

.github/workflows/test-gate.yml
```diff
@@ -966,6 +966,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt
           python -m pip install -r tt_model_runners/forge_runners/requirements.txt
+          python -m pip install --no-deps --no-build-isolation -r tt_model_runners/forge_runners/requirements.nodeps.nobuildisolation.txt
```
fivanovicTT: Not sure why this change is necessary?

tt-media-server/config/constants.py
```diff
@@ -43,6 +43,7 @@ class SupportedModels(Enum):
     GEMMA_4_31B_IT = "google/gemma-4-31B-it"
     FALCON3_7B_INSTRUCT = "tiiuae/Falcon3-7B-Instruct"
     Z_IMAGE_TURBO = "Tongyi-MAI/Z-Image-Turbo"
+    YOLOX_NANO = "yolox_nano"
```
fivanovicTT:
```suggestion
    YOLOX_NANO = "vish26/yolox-nano"
```

or `RomDev2/yolox-nano`, not sure which one you are using here

tt-media-server/tt_model_runners/forge_runners/requirements-no-build-isolation.txt
fivanovicTT, line 1: Can this change be part of `forge_runners/requirements.txt` instead? Don't see value in adding addition requirements file since we already have one

fivanovicTT (changes requested):

tt-media-server/config/constants.py
fivanovicTT: We are using Hugging Face nomenclature for model names here

tt-media-server/tt_model_runners/forge_runners/requirements-no-build-isolation.txt
ddilbazTT: It cannot unfortunately, it needs to be installed with `--no-deps --no-build-isolation` and there isn't a way to do that in the same requirements.txt file. I believe pip is trying to add support for this for a while but nothing yet.

.github/workflows/test-gate.yml
ddilbazTT: The model I added has some requirements that need to be installed with `--no-deps --no-build-isolation `. It is not possible to do so with the classic `pip install -r requirements.txt` way.

benchmarking/benchmark_targets/model_performance_reference.json
ddilbazTT: Thanks for pointing it out, I actually had a question about this. I added placeholder values based on the existing entries of another vision model. However, am I supposed to enter measured values? I believe the ci runs have some artifacts that include ttft data, I was curious if I was supposed to use that.

tt-media-server/config/constants.py
ddilbazTT: Thanks for the clarification, it helps going forward. However, this model doesn't come from Hugging Face. It is from yolox package and the yolox nano is here: https://github.com/Megvii-BaseDetection/YOLOX/tree/main How do you recommend I name? Perhaps `yolox/YOLOX-Nano` or `Megvii-BaseDetection/YOLOX-Nano`?

fivanovicTT: `yolox/YOLOX-Nano` sounds reasonable? But I am fine with `Megvii-BaseDetection/YOLOX-Nano` as well

benchmarking/benchmark_targets/model_performance_reference.json
fivanovicTT: Yes, we should have measured data put in these `model_performance_reference.json` file. You have 2 options as I see here:

- Option 1 - run this model with inference server locally on some machine (on N150 and N300 chip) and put the data you got into this file
- Option 2 - run your branch in tt-shield (CI) `on-dispatch` (you will have to create a PR to add this model there) -> get the numbers from the report -> add those numbers here

fivanovicTT: When you run `release` workflow, it will automatically run: `bechmarks`, `evals` and `spec tests` workflow. In `bechmarks` you will find `ttft`

tt-media-server/tt_model_runners/forge_runners/requirements-no-build-isolation.txt
fivanovicTT: Hey @ddilbazTT, it is unfortunate :sad-cat: Can you rename this file to something more human readable/understandable? Something like: `requirements-no-build-isolation.txt` or something along those lines?

fivanovicTT (approved):

tt-media-server/tt_model_runners/forge_runners/requirements.txt
```diff
@@ -17,3 +17,7 @@ transformers
 
 datasets # model training
 peft # model training
+
+# yolox requirements
+thop
+pycocotools>=2.0.2
```
mmilosevicTT, line 23: I would maybe prefer fixing the version here instead of allowing for different versions. I am worried that some uplift can break the code.

mmilosevicTT (approved):
LGTM, left one small comment

mjeremicTT (approved):

PR #3880, opened 2026-06-02
Save slot candidates and add slot lock

idjuricTT:
Added:

- From tryAcquireSession return list of candidates that were not acquired
- Allow locking a slot from eviction (in case we copy from it later)
- Add slot lock/unlock methods

### DIFF
--- a/tt-media-server/.vscode/settings.json
+++ b/tt-media-server/.vscode/settings.json
@@ -32,7 +32,7 @@
     "python.languageServer": "None",
     "python-envs.pythonProjects": [],
     "python.envFile": "${workspaceFolder}/.env.test",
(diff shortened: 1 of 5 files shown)


tt-media-server/.vscode/settings.json
```diff
@@ -32,7 +32,7 @@
     "python.languageServer": "None",
     "python-envs.pythonProjects": [],
     "python.envFile": "${workspaceFolder}/.env.test",
-    "cmake.sourceDirectory": "${workspaceFolder}/tt-media-server/cpp_server",
```
ljovanovicTT, line 35: nit: Unnecessary change

ljovanovicTT (approved):

PR #3892, opened 2026-06-02
MIgration service and loading it into services

idjuricTT:


nanicicTT (approved):
looks like a dummy placeholder for some future logic, so in this stage looks good :)

PR #3893, opened 2026-06-02
Vm/fix prod catalogue

vmaksimovicTT:
Our updated release procedure will look something like:
1) All developers change and tweak dev model specs during the development/testing process
2) Those dev model specs are used throughout the nightlies, release runs, on dispatch and other Model CI workflows
3) At the end of the release process, we take all of the model-device combos that were part of the release, and uplift them from dev catalogue to prod
4) We generate newest `release_model_spec.json` from the prod catalogue

Currently, release_model_spec.json and prod model specs are out of sync. This PR is meant to be a "step-0" for this new process, where we take current release_model_spec.json as source of truth, and update prod model spec yamls to match it, so that these two things are in sync and ready for future releases

Since some models are removed from the prod list, connected tests and helm values have been changed


vmaksimovicTT:
## Review

I checked out the branch and regenerated `release_model_spec.json` from the PR's prod YAMLs (`MODEL_SPECS_ENV=prod scripts/release/update_model_spec.py --output-only`). **The regenerated JSON is byte-for-byte identical to the committed one (0-line diff)** — so the core goal of this "step-0" PR is met: the prod catalogue YAMLs and `release_model_spec.json` are genuinely in sync. 👍

The large JSON churn (`override_tt_config: "{}"` → `additional_config: "{\"tt\": {}}"`, new `known_issues`/`eval_max_retries`/`structured_dataset` fields, `VLLM_ALLOW_LONG_MAX_MODEL_LEN` `"1"`→`1`, removed `cli_args`) is just the current generator's output format and matches a clean regeneration — not unintended drift.

### 🔴 Blocking: a unit test now fails

This PR removes all six "unpinned" prod entries that `tests/test_model_specification.py` hardcodes, so the suite fails on this branch:

```
FAILED tests/test_model_specification.py::TestModelSpecTemplateSystem::test_image_pinned_model_specs_excludes_unpinned_entries
AssertionError: assert set() == {('Mistral-Small-3.1-24B-Instruct-2503','t3k'), ('bge-m3','n150'), ...}
```

The removed entries (`gpt-oss-120b`/p300x2, `Mistral-Small-3.1-24B-Instruct-2503`/t3k, `bge-m3` on n150/n300/t3k/galaxy) are exactly the set the test expects to be excluded. With them gone, `excluded` is now empty and `len(IMAGE_PINNED_MODEL_SPECS) < len(MODEL_SPECS)` is also no longer true (they're equal).

Please update `test_image_pinned_model_specs_excludes_unpinned_entries` (around line 195). Note the test currently encodes a brittle premise — "prod always contains some unpinned entries." Now that every prod template pins an image, consider asserting `excluded == set()` / `len(IMAGE_PINNED_MODEL_SPECS) == len(MODEL_SPECS)`, or re-express the test against a synthetic template so it doesn't track the live catalogue contents.

### 🟡 Minor: generated docs / README not regenerated

`docs/model_support/*.md` and the README model-support table are generated from the same specs but aren't updated here. Regenerating produced two new untracked pages (`FLUX.1-schnell_p300x2.md`, `Motif-Image-6B-Preview_p300x2.md`) and would modify existing ones. If those are meant to track the catalogue they're now stale; if they're regenerated separately at release time, ignore.

### 🟡 Minor: reviewability of the JSON diff

The `release_model_spec.json` diff (+2403/-1893) interleaves format churn with real content changes, which makes it effectively unreviewable by eye. For future syncs, consider landing the format-only regeneration in a separate commit, or note in the PR that reviewers should diff a fresh regeneration rather than the raw blob (which is what I did to validate this one).

### Notes (non-blocking, verified OK)

- `image.yaml`/`video.yaml` now split single model families across multiple version templates (FLUX.1-dev/schnell across 0.10.0/0.10.1/0.14.0; Wan2.2-T2V across 0.10.0/0.10.1/0.14.0, with different device sets each). I confirmed this doesn't cause a `model_id` collision — export emits 214 specs cleanly and the round-trip matches.
- Removed `bge-m3` and `Mistral-Small-3.1-24B-Instruct-2503` are correctly absent from the regenerated JSON, consistent with the YAML deletions.

Overall: mechanically correct and achieves the sync goal — just needs the test updated before merge.

vcankovicTT (approved):

PR #3894, opened 2026-06-02
Add PrefillGateway liveness checks

ztorlakTT:
- Add PrefillGateway `/tt-liveness` and `/health` JSON endpoints behind `--health-port`.
- Report gateway transport, prefill readiness counts, and decode connection state.
- Update gateway split CI to wait on liveness instead of grepping debug logs.


tt-media-server/prefill_gateway/src/gateway_http_server.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+  running = false;
+  const int fd = serverFd.exchange(-1);
+  if (fd >= 0) {
+    shutdown(fd, SHUT_RDWR);
+    closeFd(fd);
+  }
+  if (serverThread.joinable()) {
+    serverThread.request_stop();
+    serverThread.join();
```
nanicicTT, line 73: no explicit need for this, but it is considered a good practice in this case. States the clear intention :)

nanicicTT (approved):

Issue #2848, opened 2026-04-09
Inference Server Overhaul


fivanovicTT:
This issue will act as master epic - It will contain all other major epics that we need to overhaul completely

Miroboard link: https://miro.com/app/board/uXjVGGDE7RM=/

PR #3895, opened 2026-06-02
SDXL concurrent benchmark request fix

### LINKED ISSUE #3820: Update tests and device liveness
We should have setup like this:

Image

Also, for benchmarks we should send only the num of requests as we have chips - for example: if we have Galaxy, we should send only 32 requsts, and not 100 as it is currently.

### LINKED ISSUE #3896: SDXL benchmarkin: send concurrent requests in batches

ddjukicTT:
0. Fires `max_concurrency` requests concurrently per batch
 1. Benchmark `num_calls` =` num_batches` × `max_concurrency` for the three SDXL runners
  2. Added report fields _num_requests_ + _num_concurrent_requests_
  3. tput_user = images/sec (num_calls / Σ elapsed)
  4. TTFT displayed in ms (ttft_ms) so both image tables share unit

Future feature: [Make `num_batches` configurable per-model](https://github.com/tenstorrent/tt-inference-server/issues/2848?issue=tenstorrent%7Ctt-inference-server%7C3922)

### DIFF
--- a/tt-inference-server-v2/test_module/benchmark_tests/image_benchmark_tests.py
+++ b/tt-inference-server-v2/test_module/benchmark_tests/image_benchmark_tests.py
@@ -9,6 +9,7 @@ import json
 import logging
 import sys
 import time
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from typing import Callable
 
@@ -34,7 +35,7 @@ from ..test_status import ImageGenerationTestStatus
 logger = logging.getLogger(__name__)
 
 
-SDXL_LOAD_TEST_NUM_BATCHES = 3
+SDXL_BENCHMARK_NUM_BATCHES = 3
 
 SDXL_SD35_BENCHMARK_NUM_PROMPTS = 20
 SDXL_SD35_INFERENCE_STEPS = 20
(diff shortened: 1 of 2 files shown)


tt-inference-server-v2/test_module/benchmark_tests/image_benchmark_tests.py
```diff
@@ -34,6 +35,8 @@
 logger = logging.getLogger(__name__)
 
 
+SDXL_BENCHMARK_NUM_BATCHES = 3
```
fivanovicTT, line 38: As we have discussed offline, maybe by default it should be 3 batches, but let's make it configurable by allowing classes to override this value (maybe for some models we want to send less/more)

fivanovicTT: Let's address this in a separate issue

fivanovicTT (approved):

tt-inference-server-v2/test_module/benchmark_tests/image_benchmark_tests.py
ddjukicTT: I made an issue for that feature - https://github.com/tenstorrent/tt-inference-server/issues/2848?issue=tenstorrent%7Ctt-inference-server%7C3922. I will adress it in a follow up

PR #3901, opened 2026-06-03
Knovokmet tt/integrate blaze wire format

knovokmetTT:


idjuricTT (approved):

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -126,15 +126,25 @@ inline SpecDelta computeAndLogSpecDelta(ds::DecodeScheduler& sched,
...
+pl::WireFormat wireFormatFromString(const std::string& s) {
+  if (s == "deepseek") {
+    return pl::WireFormat::DEEPSEEK;
+  } else if (s == "loopback") {
+    return pl::WireFormat::LOOPBACK;
+  } else if (s == "blaze") {
+    return pl::WireFormat::BLAZE;
+  }
+  throw std::runtime_error("Invalid wire format: " + s);
+}
```
nanicicTT, line 141: This is good, but when the new options are added in the future I would go with more of a solution like:

```cpp
pl::WireFormat wireFormatFromString(const std::string& s) {
    // It is static here so it's only initialized once
    static const std::unordered_map<std::string, pl::WireFormat> formats = {
        {"deepseek", pl::WireFormat::DEEPSEEK},
        {"loopback", pl::WireFormat::LOOPBACK},
        {"blaze",    pl::WireFormat::BLAZE}
    };

    auto it = formats.find(s);
    if (it != formats.end()) {
        return it->second;
    }
    
    throw std::runtime_error("Invalid wire format: " + s);
}
```

so we update unordered map and then just find one and return, but that is only in the case if this would grow way past the 3 items we have right now. 
So far so good. 

This internal unordered_map could live somewhere outside of this function like a mapping from a string to an enum and that we could utilize unordered map's methods to manipulate them. Like "is this option available" and we can get an answer.

nanicicTT (approved):

PR #3911, opened 2026-06-03
Copy slot when match is high enough

idjuricTT:
- Copy slot when enough tokens match
- Do not cut delta in case of disaggregation

### DIFF
--- a/tt-media-server/cpp_server/src/services/session_manager.cpp
+++ b/tt-media-server/cpp_server/src/services/session_manager.cpp
@@ -671,6 +671,37 @@ SessionManager::tryAcquireByPrefixHash(
   return AcquiredSession{false, {}, 0, 0, 0, std::move(candidates)};
 }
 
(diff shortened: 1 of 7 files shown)


tt-media-server/cpp_server/src/services/session_manager.cpp
```diff
@@ -671,6 +671,37 @@ SessionManager::tryAcquireByPrefixHash(
...
+  const size_t minTokens = tt::config::minTokensToCopy();
+
+  for (const auto& candidate : candidates) {
+    if (candidate.matchedBlocks == 0) continue;
+
+    const size_t matchedTokens =
+        firstBlockSize + (candidate.matchedBlocks > 1
+                              ? (candidate.matchedBlocks - 1) * blockSize
+                              : 0);
+
```
ljovanovicTT, line 687: nitpick: Are we calculating matchedTokens twice on this codepath?

ljovanovicTT (approved):

Issue #2297, opened 2026-03-05
Add support for tt-metal timeout per op dispatch and tt-triage on failure


tstescoTT:
Add tt-metal env vars:
1. `TT_METAL_OPERATION_TIMEOUT_SECONDS=5.0`: or setting op dispatch timeout: 
2. `TT_METAL_DISPATCH_TIMEOUT_COMMAND_TO_EXECUTE="$(pwd)/../tt-metal/python_env/bin/python $(pwd)/../tt-metal/tools/tt-triage.py --disable-progress > $(pwd)/tt-triage-output-\$(date +%Y%m%d-%H%M%S).log 2>&1"` for setting process to run on timeout failure to tt-triage. Also need to configure output from tt-triage to good log location

By default always on:

- Add `run.py` CLI feature to disable this. `--disable-metal-timeout`.
- Add Docker image env var `DISABLE_METAL_TIMEOUT`.

TODO:
- determine if there is perf overhead (currently don't believe so, it is almost always set in tt-metal CI).

tstescoTT:
addressed in https://github.com/tenstorrent/tt-inference-server/pull/2304

PR #3920, opened 2026-06-03
persist tt-triage logs to cache_root volume in --ci-mode

### LINKED ISSUE #2670: Enable visibility into automatic tt-triage report generation
Automatic execution of tt-triage was enabled in https://github.com/tenstorrent/tt-inference-server/issues/2297 which writes the tt-triage log into this default location:

https://github.com/tenstorrent/tt-inference-server/blob/3f05b3fc34b9c6107f9b2467ab5823418a6b7f58/vllm-tt-metal/src/run_vllm_api_server.py#L492

At the moment this log gets lost if this were to be triggered in Models CI as this path points to ephemeral docker overlay storage. We need to use override this when running in Models CI, i.e., when `--ci-mode` is passed

zhenghaoniTT:
Closes https://github.com/tenstorrent/tt-inference-server/issues/2670

automatic tt-triage (enabled in #2297) writes its hang report to `TT_METAL_LOGS_PATH`, which defaults to `/home/container_app_user/logs`. for the docker server that path is ephemeral docker overlay storage, so triage reports generated in Models CI are lost on container teardown. this redirects them to the persistent `cache_root` volume when `--ci-mode` is set.

**Context:**
- `run_local_server.py` already sets `TT_METAL_LOGS_PATH` to `cache_root/logs`; only the docker-server path was missing it.
- the triage command (`set_metal_timeout_env_vars` in `run_vllm_api_server.py`) redirects with `> {log_dir}/...log`, which also fails if `{log_dir}` doesn't exist yet — and `cache_root/logs` is not pre-created by the image or entrypoint.

**Changes:**
- `runtime_config.py`: thread `ci_mode` through `RuntimeConfig` (from `--ci-mode`).
- `run_docker_server.py`: when `ci_mode`, set `TT_METAL_LOGS_PATH={cache_root}/logs` on the docker server (mounted volume, survives teardown).
- `run_vllm_api_server.py`: prepend `mkdir -p {log_dir}` to the triage command so the redirect succeeds when the dir doesn't exist yet.

Tests:
- unit-level: `generate_docker_run_command` emits `-e TT_METAL_LOGS_PATH={cache_root}/logs` only when `ci_mode=True`, and nothing when `False`.

### DIFF
--- a/vllm-tt-metal/src/run_vllm_api_server.py
+++ b/vllm-tt-metal/src/run_vllm_api_server.py
@@ -495,7 +495,12 @@ def set_metal_timeout_env_vars():
     triage_old = Path(tt_metal_home) / "scripts" / "debugging_scripts" / "triage.py"
     triage_script = str(triage_new if triage_new.exists() else triage_old)
 
+    # mkdir -p so the redirect below succeeds even when log_dir does not exist
+    # yet. In CI, TT_METAL_LOGS_PATH points at {cache_root}/logs on a mounted
+    # volume that has no pre-existing logs/ subdir (see run_docker_server.py and
+    # issue #2670); without this the redirect fails and the triage report is lost.
     timeout_cmd = (
+        f"mkdir -p {log_dir} && "
         f"{python_env_dir}/bin/python {triage_script} "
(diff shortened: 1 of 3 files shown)


bgoelTT (approved):
Nice work

Issue #3888, opened 2026-06-02
[Llama-3.3-70B-Instruct / p300x2] - Acceptance Criteria FAIL due to "test_penalties"


vcankovicTT:
## Job url
https://github.com/tenstorrent/tt-shield/actions/runs/26786516166/job/78976149278

## Failure
Acceptance failed because  `test_penalties[presence_penalty-1.2-natural_repetition]`, asserted that `presence_penalty=1.2` must change the output length, but the model returned byte-identical output (72 tokens) for the short "bananas" prompt.

## Summary
The nightly release report for **Llama-3.3-70B-Instruct on p300x2** shows **ACCEPTANCE STATUS: FAIL**, even though the GitHub job is green, all benchmarks ran, accuracy evals passed, and **8 of 9** `test_penalties` subtests passed. The sole blocker is one over-strict assertion in `server_tests/test_cases/test_vllm_chat_completions.py` that requires a sampling penalty to always change the **output length**. For a short, low-repetition prompt decoded near-greedily, `presence_penalty=1.2` legitimately produced identical output, so the assertion fails. This is a test-design issue in tt-inference-server.

## Observed failure
```
test_penalties[presence_penalty-1.2-natural_repetition-messages1] FAILED
AssertionError: Test failed: Penalty had no measurable effect on output length.
assert 72 != 72.
```
Both the baseline (no penalty) and the test run (`presence_penalty=1.2`) returned byte-identical text (72 whitespace tokens):
> "Bananas are a type of fruit. They are yellow and curved. Bananas grow on big plants with large leaves. People eat bananas because they are sweet and tasty. You can eat them on their own or put them in things like smoothies or cereal. Bananas are also good for you, they have lots of vitamins and help keep you healthy. Many people like to eat bananas as a quick and easy snack."

## Root cause
`test_penalties` runs a prompt with and without a penalty at `temperature=0.1, seed=1234, max_tokens=1024`, then asserts (`server_tests/test_cases/test_vllm_chat_completions.py:289`):
```python
# 3. Length differences accepted but should not be identical
assert test_stats["len"] != base_stats["len"], (
    "Penalty had no measurable effect on output length."
)
```

vcankovicTT:
Hello Sam @stisiTT 

per Ben's comment: "This model is expected to meet FUNCTIONAL expectations, and that presence_penalty test can be masked via in the acceptance criteria."

Can this be handled by someone?
Thanks

[...]

PR #3923, opened 2026-06-03
Mask test_penalties for Llama-3.3-70B on P300X2 (#3888)

stisiTT:
Waives the over-strict `test_penalties` subtest for Llama-3.3-70B on P300X2 via a `known_issues` entry in `dev/llm.yaml`, so acceptance passes. The model meets FUNCTIONAL expectations; per Ben's guidance the test is masked rather than blocking.

Refs: #3888

### DIFF
--- a/workflows/model_specs/dev/llm.yaml
+++ b/workflows/model_specs/dev/llm.yaml
@@ -738,6 +738,10 @@ templates:
       tensor_cache_timeout: 5400.0
       override_tt_config:
         trace_region_size: 402653184  # 384 * 1024 * 1024
+      known_issues:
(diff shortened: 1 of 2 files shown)


vcankovicTT:
@stisiTT please update only "dev" catalogue, not "prod".
"prod" is updated as a part of the releasing process.

vcankovicTT (dismissed):
remove changes for `prod/llm.yaml`
only `dev` catalogue should remain.
`prod` is updated at the end of the releasing process

vcankovicTT (approved):

idjuricTT (approved):

PR #3977, opened 2026-06-04
Enable development environment in exabox. See the link how to use it:

### LINKED ISSUE #3900: Enable development environment in exabox
Team members should have development environment experience with quick feedback cycle when running on exabox.

At the moment
- The model is running
- We run the server as docker container
- We notice a bug in the server or missing log statement
- We change the code and rebuild the docker image - **takes 1 hour**
- Verify the change

**In future, "we change the code and verify the fix" must take minutes instead of more than an hour.**

Adapting the code must be dev environment experience. This means I can connect my IDE (Cursor/VSCode) and modify the code.

nanicicTT:
https://github.com/tenstorrent/tt-inference-server/issues/3900


dmadicTT (approved):

tt-media-server/cpp_server/run_server.sh
```diff
@@ -0,0 +1,17 @@
...
+# Navigate to the cpp_server directory
+cd "$(dirname "$0")"
+
+# Set up runtime environment variables
+export TT_METAL_HOME=/data/TT_USER/tt-inference-server/tt-media-server/cpp_server/tt-llm-engine/tt-metal
+export LD_LIBRARY_PATH=$TT_METAL_HOME/build/lib:$(pwd)/tt-llm-engine/build-full:$LD_LIBRARY_PATH
+
+# Run the server
+LLM_DEVICE_BACKEND=pipeline_manager \
+DEVICE_IDS=0,1,4,5,24,25,28,29 \
```
idjuricTT, line 15: No need to override this - this might lead to multiple workers

nanicicTT: This was just an example :) This is still work in progress :) 
We need something for the first stage and then we will figure it out how to parametrize it.

dmadicTT: this is how we run deepseek and kimi now, I believe it creates only one worker

Issue #3877, opened 2026-06-02
Track token usage in case of disaggregation


idjuricTT:
In case of disaggregation prefill should return number of cached tokens

PR #4008, opened 2026-06-05
Calculate Cached Tokens on Prefill for Disaggregated Setup

ljovanovicTT:
## Issue
[#3877 Track token usage in case of disaggregation](https://github.com/tenstorrent/tt-inference-server/issues/3877)

## Problem
In case of disaggregated setup, we want to report `cached_tokens` from Prefill node. 
With this PR, this is done for both Dynamo path and legacy HTTP path.

### DIFF
--- a/tt-media-server/cpp_server/src/services/disaggregation_service.cpp
+++ b/tt-media-server/cpp_server/src/services/disaggregation_service.cpp
@@ -142,11 +142,6 @@ void DisaggregationService::setupSocketHandlers() {
           request->top_k = message.top_k;
           request->fast_mode = message.fast_mode;
 
-          TT_LOG_DEBUG(
-              "[DisaggregationService] Prefill request taskId={} "
-              "registration_hashes={}",
-              message.task_id, message.registration_hashes.size());
-
           auto maxTokens = message.max_tokens;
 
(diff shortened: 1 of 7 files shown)


tt-media-server/cpp_server/src/services/disaggregation_service.cpp
```diff
@@ -198,6 +232,22 @@ void DisaggregationService::setupSocketHandlers() {
                       }
 
                       socketService->sendPrefillResult(prefillResult);
+
+                      // Release the prefill session's in-flight hold now that
+                      // this one-shot prefill (max_tokens=1) is done. Unlike
+                      // the decode/HTTP transports, the prefill path has no
+                      // stream-end release, so without this the session stays
+                      // IN_FLIGHT forever — and evictOldSessions only reclaims
```
ljovanovicTT, line 240: @idjuricTT Can you take a look if this makes sense? Multiturn never properly triggered on Prefill without slot release

idjuricTT: It does, thank you 😄

idjuricTT (approved):

PR #4014, opened 2026-06-05
Isolate heavy C++ ccache keys

ztorlakTT:
- Add explicit `ccache-key-suffix` values for all `cpp-heavy-checks.yml` callers of `setup-cpp-ci`.
- Prevent scheduled heavy C++ jobs from sharing the default cache key and racing to persist incomplete Cargo/ccache state.

https://github.com/tenstorrent/tt-inference-server/actions/runs/27020325359


dmadicTT:
> Prevent scheduled heavy C++ jobs from sharing the default cache key and racing to persist incomplete Cargo/ccache state.

Can you share an example what happened?

The current CI seems to fail https://github.com/tenstorrent/tt-inference-server/actions/runs/27020325359

idjuricTT (approved):

PR #4018, opened 2026-06-05
First prefill test

idjuricTT:


tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  // After allocation succeeds, the server submits the request to the worker.
+  // Mock the worker response so everything cleans up.
+  auto seq = server->taskQueue().receive();
+  ASSERT_NE(seq, nullptr);
+  EXPECT_GT(seq->getNumPromptTokens(), 0u);
+
+  tt::test::WorkerResponse(seq->taskId)
+      .token(42)
+      .finalize()
```
nanicicTT: The test sends both a token and `.finalize()`, then only reads one `prefill_result` at lines `355-360`. LLMService calls the stream callback for the normal token and again for the final marker, and `disaggregation_service.cpp` ignores `isFinal` and sends a `PrefillResultMessage` for every callback (https://github.com/tenstorrent/tt-inference-server/blob/481c69e415c6bf7248cea8059910e2bb85ba048a/tt-media-server/cpp_server/src/services/disaggregation_service.cpp#L165). This test can pass I guess while the prefill path emits duplicate results for one task. 
It should assert there is exactly one result, or expose/fix the duplicate final-result behavior maybe. 
What do you think?

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+    socket_.set(zmq::sockopt::linger, 0);
+    socket_.set(zmq::sockopt::rcvtimeo, 5000);  // 5s timeout on recv
+    std::string endpoint = "tcp://*:" + std::to_string(port);
+    socket_.bind(endpoint);
+  }
+
+  ~MockDecodeServer() { socket_.close(); }
+
+  /// Block until a peer DEALER connects. Returns the peer's ZMQ identity.
+  std::vector<uint8_t> waitForPeer(
```
nanicicTT, line 181: I guess this treats the first ROUTER frame as proof that the prefill server connected, then discards the payload. 
Since the prefill service is expected to send a `prefill_registration`, the test might need to deserialize and assert that message type/content. As written, a wrong or unexpected first message still satisfies setup.
Is it the case, or not?


So `waitForPeer` only checks did it receive any identity frame, then it drains and ignores the actual payload.

The prefill server will send a registration msg, but as I can see the code never verifies that. 
If the server sends the wrong message type, malformed but non-empty payload, a health/status message, or some accidental message caused by some bug, `waitForPeer()` still returns success, right?

nanicicTT: We might need to inspect drained dataFrame


```cpp
auto* ptr = static_cast<uint8_t*>(dataFrame.data());
std::vector<uint8_t> raw(ptr, ptr + dataFrame.size());

EXPECT_EQ(tt::sockets::wire::readMessageType(raw),
          tt::sockets::tags::PREFILL_REGISTRATION);

auto reg =
    tt::sockets::wire::deserializePayload<tt::sockets::PrefillRegistrationMessage>(raw);
EXPECT_FALSE(reg.server_id.empty());
```

idjuricTT: That's true, thank you!
We should actually improve the registration message as well @ztorlakTT

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
nanicicTT: The test mocks the worker

```cpp
  tt::test::WorkerResponse(seq->taskId)
      .token(42)
      .finalize()
      .sendTo(server->resultQueue());
``` 
like from the code snippet above. 
 
And that will create two worker messages:
  1. normal token: token_id = 42, isFinal = false
  2. final marker: token_id = 0, isFinal = true

`LLMService` calls the stream callback once for each message. But the prefill callback in DisaggregationService does not check isFinal (linked line above) :

```
  [this, message, maxTokens, slotId](
      const LLMStreamChunk& response, bool /*isFinal*/) {
    auto prefillResult = tt::sockets::PrefillResultMessage(message.task_id);
    ...
    socketService->sendPrefillResult(prefillResult);
  }
```

So imagine like one prefill request can send two prefill_result messages:
  prefill_result for token 42
  prefill_result for final marker token 0

The test only does one receive:

```cpp
  auto result = mockDecode->receive<PrefillResultMessage>("prefill_result", ...);
  ASSERT_TRUE(result.has_value());
  EXPECT_EQ(result->task_id, taskId);
  EXPECT_FALSE(result->error);
```

And it passes after seeing the first valid result and never checks whether a second result was sent.

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
```diff
@@ -161,7 +163,10 @@ void DisaggregationService::setupSocketHandlers() {
                 llmService->submitStreamingRequest(
                     *request,
                     [this, message, maxTokens, slotId](
-                        const LLMStreamChunk& response, bool /*isFinal*/) {
+                        const LLMStreamChunk& response, bool isFinal) {
+                      assert(
+                          isFinal &&
```
idjuricTT: @ztorlakTT pls check does this make sense and does this breaks anything

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
idjuricTT: @nanicicTT should be fixed and double checked on dissaggreagationService, thank you!!

fivanovicTT (approved):

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
ztorlakTT: It makes sense and shouldn’t break anything.
One suggestion: instead of `assert` we can have:
```
if (!isFinal) {
  TT_LOG_WARN(
      "[DisaggregationService] Ignoring non-final prefill chunk for task {}",
      message.task_id);
  return;
}
```

ztorlakTT (approved):

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -0,0 +1,514 @@
...
+  // Verify slot_id from decode (7) is the KV cache slot (the worker's output
+  // destination), while the prefill-side allocation (2) becomes the prefill
+  // KV cache slot (the source to read cached KV from).
+  EXPECT_EQ(seq->getKVCacheSlot(), 7u)
+      << "decode-side slot_id must be the primary KV cache slot";
+  EXPECT_EQ(seq->getPrefillKVCacheSlot(), 2u)
+      << "prefill-side allocation must be the prefill KV cache slot";
+
+  tt::test::WorkerResponse(seq->taskId)
+      .tokenWithFlags(42, tt::ipc::SharedToken::FLAG_FINAL)
```
dmadicTT, line 360: We should match reality with this response - we will not get token ID from the model

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -0,0 +1,514 @@
...
+    tt::domain::ManageMemoryTask spuriousAlloc{};
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    bool gotAlloc = server->memoryRequestQueue().tryPop(spuriousAlloc);
+    EXPECT_FALSE(gotAlloc)
+        << "Turn " << turn
+        << ": unexpected ALLOCATE — prefix cache should have HIT";
+
+    // If an unexpected ALLOCATE appeared, respond so the server doesn't hang.
+    if (gotAlloc) {
+      tt::domain::ManageMemoryResult memRes{};
```
dmadicTT, line 471: Shouldn't we fail the test here?

idjuricTT: yep, whole this thing should go out, thanks!

PR #4080, opened 2026-06-08
Faster DiT/media warm standups: persist conversion cache + reuse host weight cache

### LINKED ISSUE #4077: Enable persistent TT_DIT_CACHE_DIR for Wan2.2 (DiT) models to cache weight conversion across deployments
> Part of a broader effort: make DiT / media model standups fast on warm restarts — stop redoing expensive work every time the server comes up. This issue is one of two distinct fixes under that theme. Sibling: #4088 (reuse the host HuggingFace weight cache for diffusers models).

## Summary

Every time a Wan2.2 server starts up, it spends ~21 minutes converting model weights from PyTorch format — and then throws that work away. The infrastructure to cache this conversion already exists (`tt_dit` cache), and Wan already calls through it, but no one ever told Wan where to store the cache. Fix: point `TT_DIT_CACHE_DIR` at the persistent volume that survives container restarts.

## The problem

- Wan2.2-T2V-A14B standup consistently takes ~21 minutes, confirmed across 3 independent runs.
- ~19.5 of those 21 minutes are weight conversion that *could* be skipped on a cache hit.
- The cache code is fully implemented and already wired up in the Wan pipeline — it just needs the env var set.
- None of the Wan runners (`TTWan22Runner` and variants) set `TT_DIT_CACHE_DIR`.
- The one runner that does set it (Mochi) points at `/tmp`, which is wiped on every container restart, so it's also effectively uncached.

## Root cause

`models/tt_dit/utils/cache.py` — `load_model()` checks `os.environ.get("TT_DIT_CACHE_DIR")` and logs "no cache" when it's unset, falling back to a full PyTorch weight conversion every time. The Wan pipeline already routes all components (text encoder, transformer ×2, VAE) through this cache function — the feature is one env var away from being active.

`tt-media-server/tt_model_runners/dit_runners.py`:
- `TTMochi1Runner` sets `TT_DIT_CACHE_DIR = /tmp/TT_DIT_CACHE` (ephemeral, lost on container restart)
- `TTWan22Runner` and all I2V variants set nothing

## Proposed fix

Set `TT_DIT_CACHE_DIR` to a subdirectory of `CACHE_ROOT` (e.g. `${CACHE_ROOT}/tt_dit_cache`) in `video.yaml` and `image.yaml` `env_vars` so the cache lives on the persistent Docker volume. This is the cleanest spot because `run_docker_server.py` propagates `model_spec.env_vars` into the container, meaning downstream consumers (e.g. TT-Studio) inherit the cache automatically.

```
TT_DIT_CACHE_DIR: ${CACHE_ROOT}/tt_dit_cache
```

## Expected outcome

After the first cold deployment (which populates the cache), subsequent startups should skip ~19.5 min of weight conversion and take roughly 2–3 minutes total.

## Things to validate

- Confirm `tt_model.save()` / `tt_model.load()` round-trips correctly for all Wan components, especially the VAE (its cache key is derived from `conv3d_blocking_hash` in `pipeline_wan.py:515`).
- Consider namespacing the cache path by tt-metal commit/image tag to prevent stale cache reuse after image upgrades (current cache key includes topology/dtype but not image version).
- Document expected disk footprint and cleanup story for the persisted cache.
- Measure realized standup speedup on a warm cache (expectation: drops toward the ~1.7-min kernel compilation floor + warmup).

## Evidence

Three independent Wan2.2-T2V-A14B-Diffusers runs on p300x2 (image `0.14.0-80180b9`), each emitting the `tt_dit` "no cache" log line and identical 33 `conv3d blocking [fallback]` warnings:

| Run | Model load time | Total to ready |
|-----|----------------|----------------|
| 2026-05-19 10:02 | 21m 17s | ~22m |
| 2026-05-19 17:06 | 21m 15s | ~22m |
| 2026-06-05 12:41 | 21m 16s | ~22m |

Logs: `workflow_logs/docker_server/media_*_Wan2.2-T2V-A14B-Diffusers_p300x2_server.log`

### LINKED ISSUE #4085: Wan2.2 (and other models): inline weight download can stall indefinitely with no recovery
Child of #4108. When `MODEL_WEIGHTS_DIR` is not set, Wan2.2 downloads weights inline during pipeline construction via bare `from_pretrained()` calls with no timeout. If the XET download client stalls mid-transfer, the container blocks forever — no error, no retry — until tt-studio's 35-min watchdog kills it.

## Resolution
Resolved as a consequence of #4107: when `MODEL_WEIGHTS_DIR` is set and the host cache is complete, the media server loads weights locally and the inline download never runs.

For cold-cache scenarios (no host mount), stall risk remains. A pre-download via `snapshot_download()` before pipeline construction would address it separately.

stisiTT:
Stops Wan/DiT media models from redoing expensive work on every standup. Two independent fixes, four issues closed, one PR.

Closes #4077, #4088, #4107
Resolves #4085 (as consequence of #4107 fix)
Parent: #4108

---

## Fix 1 — Persist the conversion cache (#4077)

Every standup re-ran ~21 min of weight conversion because there was no persistent cache location. Points `TT_DIT_CACHE_DIR` at the per-model Docker volume so the cache survives restarts.

**Files:** `workflows/model_specs/{prod,dev}/{video,image}.yaml`, `tt-media-server/tt_model_runners/dit_runners.py`

---

## Fix 2 — Reuse the host weight cache (#4088 #4107 #4085)

Two gaps caused `--host-hf-cache` to be ignored for diffusers models, triggering a ~118 GB re-download every standup:

1. **Completeness check false-negative (#4088):** `setup_host.py` only recognised the flat LLM layout; diffusers sub-folder layout (`model_index.json` + per-component safetensors) was always judged incomplete and fell back to a full re-download.
2. **Reuse gap (#4107):** Even with the mount present, `settings.py` never read `MODEL_WEIGHTS_DIR`, and `TTWan22Runner.create_pipeline()` never passed the local path to `WanPipeline`, which used its hardcoded HF repo-id default and re-downloaded.

**Files:** `workflows/setup_host.py`, `tt-media-server/config/settings.py`, `tt-media-server/tt_model_runners/dit_runners.py`

---

## Validation

Fix 2 validated locally: container `~/.cache/huggingface` = 0 files after warm standup; NET I/O 925 kB vs ~118 GB previously. Shield run (in progress): [Wan2.2-T2V / bh-qb-ge / p300x2](https://github.com/tenstorrent/tt-shield/actions/runs/27236296645)

Fix 1: cold-vs-warm correctness check recommended before merge (same seed, compare outputs) — persistent Wan cache path not yet exercised in CI.

> **Note (#4111):** Wan2.2-T2V also requires the model spec pin bumped to tt-metal ≥ `8574aad`. This is a release-cycle config change, not part of this PR.


vpetrovicTT (approved):

idjuricTT (commented):
@vpetrovicTT could you please check this?

vmaksimovicTT:
@stisiTT Hi Sam. The changes you made to the dev/* model specs make sense, but, per the procedure, we dont want to change anything in the prod model specs manually, outside of the release procedure. Can you revert those changes, and the next time these 4 models (mochi, 2x WAN and QWen) go through the release, the changes will be picked up and uplifted to prod model specs.

vmaksimovicTT (changes requested):
I left a comment on a PR describing the changes that should be made

stisiTT:
@vmaksimovicTT Good call — reverted the manual `TT_DIT_CACHE_DIR` changes to the prod model specs (and the Helm `values.yaml` generated from them) in 968e9b0b2ed1e703d823aeb75a4312e1f8497a29. The `dev/*` specs keep the change, so it'll be uplifted to prod automatically when Mochi, the two WAN models, and Qwen-Image next go through the release procedure. Thanks!

https://github.com/tenstorrent/tt-inference-server/commit/968e9b0b2ed1e703d823aeb75a4312e1f8497a29

vmaksimovicTT (approved):

Issue #4074, opened 2026-06-08
Decode should send to prefill skip tokens and start_position


idjuricTT:
Skip_tokens - i.e. skip 1280 prefill tokens
Start_position - i.e. 1700 since decode has thinking tokens

idjuricTT:
Merged

PR #4082, opened 2026-06-08
Propagate Both decodePositionId and decodeSkipTokens to Prefill Node

ljovanovicTT:
## Issue
[#4074 Decode should send to prefill skip tokens and start_position](https://github.com/tenstorrent/tt-inference-server/issues/4074)

## Problem
Since reasoning blocks may exist in the Decode node's KV cache but not in the Prefill node's KV cache (as we currently do not support D->P migration), it is not sufficient to know only how many tokens need to be transferred from Prefill to Decode, we also need to know the exact location where those tokens should be transferred.

To support this, we introduce two variables:
- `decodeSkipTokens` - the number of tokens for which KV cache has already been computed on the Decode node and therefore does not need to be transferred or recomputed
- `decodePositionId` - the target position in the Decode KV cache where transferred KV entries should be written, ensuring that already computed state is preserved

### DIFF
--- a/tt-media-server/cpp_server/include/domain/llm/llm_request.hpp
+++ b/tt-media-server/cpp_server/include/domain/llm/llm_request.hpp
@@ -133,7 +133,7 @@ struct LLMRequest : BaseRequest {
 
   // Number of tokens already in the decode-side KV cache that the prefill
   // prefill should not send this part of KV.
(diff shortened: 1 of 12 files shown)


idjuricTT (approved):
Bravo!!!!!!!!

PR #4083, opened 2026-06-08
Prefill slot copy test

idjuricTT:
Scenario:

- New request gets slot (0)
- Continuation gets the same slot
- While continuation is prefilling, new request would trigger copy from slot (0)
- Continuation of request C would be on the same slot

### DIFF
--- a/tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
+++ b/tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
@@ -51,6 +51,7 @@ void configureEnv() {
   setenv("LLM_MODE", "prefill", 1);
   setenv("DEVICE_IDS", "(0)", 1);
   setenv("MAX_NUM_SESSIONS", "4", 1);
+  setenv("MIN_TOKENS_TO_COPY", "32", 1);
   setenv("KV_CACHE_FIRST_BLOCK_SIZE", "32", 1);
   setenv("KV_CACHE_BLOCK_SIZE", "32", 1);
   setenv("PREFIX_CACHE_HIT_THRESHOLD", "0", 1);
@@ -500,7 +501,6 @@ TEST_F(PrefillIntegrationTest, MultiTurn_SubsequentRequestsAreContinuations) {
 //      should HIT C's newly registered session (confirms prefix hashes were
 //      propagated to the new session created by slot copy).
 TEST_F(PrefillIntegrationTest, SlotCopy_TriggeredWhenSessionInFlight) {
-  setenv("MIN_TOKENS_TO_COPY", "32", 1);
   server->setMemoryAutoRespond(false);
 
   // Track prefillSlotIds across requests to verify session reuse.
@@ -753,7 +753,6 @@ TEST_F(PrefillIntegrationTest, SlotCopy_TriggeredWhenSessionInFlight) {
   }
 
   server->setMemoryAutoRespond(true);
-  unsetenv("MIN_TOKENS_TO_COPY");
 }
 
 // ---------------------------------------------------------------------------
--- a/tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
+++ b/tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
@@ -466,15 +466,6 @@ TEST_F(PrefillIntegrationTest, MultiTurn_SubsequentRequestsAreContinuations) {
         << "Turn " << turn
         << ": unexpected ALLOCATE — prefix cache should have HIT";
 
-    // If an unexpected ALLOCATE appeared, respond so the server doesn't hang.
-    if (gotAlloc) {
-      tt::domain::ManageMemoryResult memRes{};
-      memRes.taskId = spuriousAlloc.taskId;
-      memRes.status = tt::domain::ManageMemoryStatus::SUCCESS;
-      memRes.slotId = turn;
-      server->memoryResultQueue().push(memRes);
-    }
-
     // Worker processes the continuation.
     auto seq = server->taskQueue().receive();
     ASSERT_NE(seq, nullptr) << "Turn " << turn << ": no Sequence";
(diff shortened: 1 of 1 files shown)


tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -473,15 +473,6 @@ TEST_F(PrefillIntegrationTest, MultiTurn_SubsequentRequestsAreContinuations) {
...
-    if (gotAlloc) {
-      tt::domain::ManageMemoryResult memRes{};
-      memRes.taskId = spuriousAlloc.taskId;
-      memRes.status = tt::domain::ManageMemoryStatus::SUCCESS;
-      memRes.slotId = turn;
-      server->memoryResultQueue().push(memRes);
-    }
-
     // Worker processes the continuation.
     auto seq = server->taskQueue().receive();
```
nanicicTT, line 478: This branch removes the existing bailout that responded to an unexpected `ALLOCATE`. If the cache hit regresses, the test pops the allocation, records `EXPECT_FALSE`, then blocks forever on `taskQueue().receive()` because the server is waiting for a memory result. 
These changes here repeat the same pattern at lines 605-612 and 712-720. 
I would either restore the response-on-failure path or maybe use a timeout-based receive/helper that fails cleanly.

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -504,6 +495,274 @@ TEST_F(PrefillIntegrationTest, MultiTurn_SubsequentRequestsAreContinuations) {
...
+    EXPECT_FALSE(server->memoryRequestQueue().tryPop(spurious))
+        << "Request D: unexpected ALLOCATE — should HIT C's session "
+           "(prefix hashes were registered on slot copy)";
+
+    // Worker processes D (continuation).
+    auto seq = server->taskQueue().receive();
+    ASSERT_NE(seq, nullptr) << "Request D: no Sequence";
+    EXPECT_TRUE(seq->isContinuation())
+        << "Request D: must be a continuation (HIT on C's session)";
+    EXPECT_EQ(seq->getKVCacheSlot(), 1u)
```
nanicicTT, line 724: `EXPECT_EQ(seq->getKVCacheSlot(), 1u)` mostly verifies the req.slot_id = 1 set above at line [703](https://github.com/tenstorrent/tt-inference-server/pull/4083/changes#diff-eccd83a3b6e93d1f920721db2a6a28174c61d1254b97f932a04cac521a40a5bcR703). It does not prove D reused C’s prefill session. The later `prefillSlotC == prefillSlotD` (equality expectation) check is indeed the meaningful one. 
I’d, maybe, change the message/assertion to use `getPrefillKVCacheSlot()` and the observed C prefill slot, or drop the misleading `getKVCacheSlot()` assertion.

---

Prefix-cache reuse on the prefill server is done by `seq->getPrefillKVCacheSlot()`, right?

That assertion will pass largely because the test itself put **1** into `req.slot_id`. It seems it does not prove D hit C’s prefill session. Even if the prefill cache behavior were wrong, `getKVCacheSlot()` could still be 1.
The meaningful check is this later one:

```cpp
  EXPECT_EQ(prefillSlotC, prefillSlotD)
```
because `prefillSlotC` and `prefillSlotD` come from:

```cpp
  seq->getPrefillKVCacheSlot()
```

That is the slot managed by the prefill session manager.

A clearer assertion would be:

```cpp
  EXPECT_EQ(seq->getPrefillKVCacheSlot(), prefillSlotC)
      << "Request D should reuse request C's prefill slot";
```

or just keep the final `EXPECT_EQ(prefillSlotC, prefillSlotD)` and remove/reword the `getKVCacheSlot()` assertion.

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
```diff
@@ -504,6 +495,274 @@ TEST_F(PrefillIntegrationTest, MultiTurn_SubsequentRequestsAreContinuations) {
...
+//   2. Request B is a continuation of A → HITs the prefix cache, session is
+//      now IN_FLIGHT.
+//   3. While B is still in-flight, request C arrives with the same prefix →
+//      the prefix cache finds A's session but it's busy → falls through to
+//      ALLOCATE with slotIdToCopyFrom pointing to A's slot.
+//   4. After C completes, request D arrives with the same prefix as C →
+//      should HIT C's newly registered session (confirms prefix hashes were
+//      propagated to the new session created by slot copy).
+TEST_F(PrefillIntegrationTest, SlotCopy_TriggeredWhenSessionInFlight) {
+  setenv("MIN_TOKENS_TO_COPY", "32", 1);
```
nanicicTT: This gets set and if something asserts, this is not unset, until the end of the test. 
Maybe we can use some RAII pattern here, like:

```cpp
  struct ScopedPrefillSlotCopyTestState {
    explicit ScopedPrefillSlotCopyTestState(PrefillTestServer* server)
        : server(server) {
      setenv("MIN_TOKENS_TO_COPY", "32", 1);
      server->setMemoryAutoRespond(false);
    }

    ~ScopedPrefillSlotCopyTestState() {
      server->setMemoryAutoRespond(true);
      unsetenv("MIN_TOKENS_TO_COPY");
    }

    PrefillTestServer* server;
  };
```

Then at the start of the test:

```cpp
  ScopedPrefillSlotCopyTestState guard(server.get());
```

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
idjuricTT: Fair, this is running in an isolated environment so we don't expect funky stuff.
@dmadicTT  actually had an ask to remove this safety check for test 😄

nanicicTT: This deleted lines mean the test already failed, but to unblock the server so the test can continue and exit normally.

Without that bailout, this ,might happen:
  1. Prefix cache regresses and misses.
  2. Server pushes `ALLOCATE` to `memoryRequestQueue`.
  3. Test pops it into spuriousAlloc.
  4. `EXPECT_FALSE(gotAlloc)` fails, but EXPECT_* is non-fatal, so the test continues.
  5. Test calls `auto seq = server->taskQueue().receive();`
  6. Server never pushes a sequence, because it is still waiting for a response to the allocation.
  7. Test hangs until the whole integration test times out.

So the issue is not that the assertion is wrong. The issue is that after detecting the failure, the test must still answer the unexpected allocation or use a timeout receive, otherwise the failure becomes a slow timeout instead of a clear test failure.

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
idjuricTT: To be fair, there's no need for this, I can just set the copy threshold right away to 32

nanicicTT (approved):

Issue #4005, opened 2026-06-05
[Llama-3.1-8B-Instruct p150] trace region issue


vcankovicTT:
## Job url

https://github.com/tenstorrent/tt-shield/actions/runs/26994643073/job/79661779019

## trace region too small

```
tt-metal issue — trace buffer too small

The fatal error is thrown from tt_metal/distributed/mesh_trace.cpp:78:

TT_FATAL: Creating trace buffers of size 53616640B on MeshDevice 0,
          but only 50000000B is allocated for trace region.

The model's prefill pass requires ~51.1 MB of trace buffer, but only ~47.7 MB is allocated. 
This happens during ttnn.end_trace_capture() in _capture_trace_prefill inside tt-metal's generator.py:296.
```

vcankovicTT:
The last night job requested more trace region:
https://github.com/tenstorrent/tt-shield/actions/runs/27030438092/job/79788142658

```
[tt-inference-server-52e5d497] (EngineCore_DP0 pid=292) ERROR 06-05 19:19:37 core.py:1035] 
RuntimeError: TT_FATAL 
@ /home/container_app_user/tt-metal/tt_metal/distributed/mesh_trace.cpp:78: 
mesh_cq.device()->get_trace_buffers_size() <= trace_region_size

[tt-inference-server-52e5d497] (EngineCore_DP0 pid=292) ERROR 06-05 19:19:37 core.py:1035] 
Creating trace buffers of size 60825600B on MeshDevice 0, but only 50000000B is allocated for trace region.
```

vcankovicTT:
Hi @stisiTT did you have time to take a look at this, if this is something that should be handled from the inference side?

djordje-tt:
I opened PR: https://github.com/tenstorrent/tt-inference-server/pull/4112. Whoever is appropriate to check whether trace_region_size is added at relevant places only. There are multiple ones and I claude-ed it so not super confident it would be enough. I also added pipeline run to verify change.

djordje-tt:
@vcankovicTT, I started run in CI to test a fix but it seems we don't have p150s available: https://github.com/tenstorrent/tt-shield/actions/runs/27267956175. In queue for many hours.

vcankovicTT:
started another one
https://github.com/tenstorrent/tt-shield/actions/runs/27296030642
runner label should be: **tt-ubuntu-2204-p150b-stable**

vcankovicTT:
[...]

PR #4112, opened 2026-06-10
[Llama-3.1-8B P150] Set trace_region_size to fix prefill trace OOM (#4005)

djordje-tt:
## Ticket

https://github.com/tenstorrent/tt-inference-server/issues/4005

---

## Root cause

Llama-3.1-8B-Instruct on **P150** fails during trace capture with:

```
TT_FATAL @ tt_metal/distributed/mesh_trace.cpp:78:
mesh_cq.device()->get_trace_buffers_size() <= trace_region_size
Creating trace buffers of size 53616640B (later 60825600B) on MeshDevice 0,
but only 50000000B is allocated for trace region.
```

The **P150** entry in the prod catalog (`workflows/model_specs/prod/llm.yaml`) had **no `trace_region_size` override**, so it fell back to the vLLM-plugin **default of `50,000,000` B** (`tt-vllm-plugin/tt_vllm_plugin/worker/tt_worker.py`). The Llama-3.1-8B prefill trace on Blackhole needs **~53.6 MB, and ~60.8 MB after a later tt-metal change** — both above 50 MB → fatal in `end_trace_capture`.

Its single-Blackhole-chip sibling **P100** already sets `trace_region_size: 90000000`; P150 was simply missing the same override.

---

## Fix

Set `trace_region_size: 90000000` for **P150** (matching P100, same single-BH silicon; ~48% headroom over the observed 60.8 MB peak):

- `workflows/model_specs/prod/llm.yaml` — add `override_tt_config.trace_region_size: 90000000` to the P150 device spec (source of truth).
- `release_model_spec.json` — regenerate the exported value for the P150 entries of both `Llama-3.1-8B` and `Llama-3.1-8B-Instruct` (they share the template): `additional_config`, `override_tt_config` (inner + outer).

---

## Test

tt-shield `on-dispatch` with `model=Llama-3.1-8B-Instruct`, `device-type=p150`, `inference-server-git-ref=djordje-tt/llama8b-p150-trace-region` — confirm it gets past trace capture:

pipelines:
- https://github.com/tenstorrent/tt-shield/actions/runs/27264840249
- https://github.com/tenstorrent/tt-shield/actions/runs/27265483325
---

### DIFF
--- a/workflows/model_specs/dev/llm.yaml
+++ b/workflows/model_specs/dev/llm.yaml
@@ -954,6 +954,8 @@ templates:
       max_concurrency: 32
       max_context: 65536  # 64 * 1024
       default_impl: true
+      override_tt_config:
+        trace_region_size: 90000000  # match P100 (single BH chip); 50MB default too small for prefill trace (#4005)
   status: EXPERIMENTAL
   metadata:
     meta-llama/Llama-3.1-8B:
--- a/workflows/model_specs/dev/llm.yaml
+++ b/workflows/model_specs/dev/llm.yaml
@@ -955,7 +955,7 @@ templates:
       max_context: 65536  # 64 * 1024
       default_impl: true
       override_tt_config:
-        trace_region_size: 120000000  # match P100 (single BH chip); 50MB default too small for prefill trace (#4005)
+        trace_region_size: 360000000  # match P100 (single BH chip); 50MB default too small for prefill trace (#4005)
   status: EXPERIMENTAL
   metadata:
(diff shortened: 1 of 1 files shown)


idjuricTT (approved):

vcankovicTT (approved):
we need to update "dev" catalogue solely.
"prod" catalogue is being updated at the end of the release cycle

https://github.com/tenstorrent/tt-inference-server/blob/main/workflows/model_specs/dev/llm.yaml

PR #4116, opened 2026-06-10
blaze- integrate migration with prefill and decode schedulers

knovokmetTT:


idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -453,6 +453,12 @@ void LLMPipeline::dispatchGeneration(
         uint32_t matchedTokens =
             *request.kv_position_id + 1 -
             static_cast<uint32_t>(request.accumulated_think_tokens);
+        // WARNING - TEMP CHANGE - PREFILL WILL OVERRIDE THINKING TOKENS
```
idjuricTT: No need to do this here, at this point request remains on decode

idjuricTT (approved):

dmadicTT (approved):

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -48,6 +48,13 @@ constexpr size_t KV_CACHE_BLOCK_SIZE = 32;
 constexpr size_t KV_CACHE_FIRST_BLOCK_SIZE = 128;
 constexpr unsigned PREFIX_CACHE_HIT_THRESHOLD = 40;
 constexpr bool USE_FAST_MODE = false;
+constexpr bool ENABLE_MIGRATION = false;
+constexpr const char* MIGRATION_CMD_QUEUE_NAME = "tt_migration_cmd";
+constexpr const char* MIGRATION_TABLE_QUEUE_NAME = "tt_migration_table";
+constexpr const char* MIGRATION_RESP_QUEUE_NAME = "tt_migration_resp";
```
tt-asaigal: Note that these queue names can be different across prefill and decode endpoints. Common convention is  `mig_ep0*` and `mig_ep1*` prefixes to queue names for endpoint 0 and 1 respectively.

tt-media-server/cpp_server/include/domain/llm/sequence.hpp
```diff
@@ -116,7 +116,7 @@ class Sequence {
   int getDecodeSkipTokens() const { return decodeSkipTokens; }
   void setDecodeSkipTokens(int n) { decodeSkipTokens = n; }
 
-  uint64_t getMigrationId() const { return migrationId; }
+  std::optional<uint64_t> getMigrationId() const { return migrationId; }
```
tt-asaigal, line 120: This needs to be unique across prefill requests, as per the migration layer design.

Also - where does this get set?

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -24,13 +29,19 @@ namespace sch = tt_llm_engine::scheduler;
...
+    uint32_t requestId,
+    std::optional<uint32_t> migrateFromSlot = std::nullopt) {
+  auto req = sch::ISRequest{
+      .type = ds::RequestType::ALLOCATE,
+      .request_id = requestId,
+      .tokens = {},
+      .gen = {},
+  };
+  if (migrateFromSlot.has_value()) {
+    req.gen.migrate_from_slot = *migrateFromSlot;
```
tt-asaigal: This is prefix caching functionality. Might make sense to internally track state ensuring that the slot being migrated from is allocated. If not, you could end up copying garbage data.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
```diff
@@ -322,15 +334,78 @@ inline pl::CounterChannelConfig makePrefillAckChannelConfig(
...
+    default:
+      throw std::runtime_error("Invalid blaze prefill runner type");
+  }
+}
+
+inline pl::MigrationCmdChannelConfig makeMigrationCmdChannelConfig(
+    const tt::config::LLMConfig& config) {
+  switch (config.runner_type) {
+    case tt::config::ModelRunnerType::PIPELINE_MANAGER:
+      return pl::InterProcessMigrationCmdChannelConfig{
```
tt-asaigal: This is based on legacy code that has been deprecated with the integration of @SeanNijjar's Migration Layer APIs. `InterProcessMigrationCmdChannelConfig` was used as a stub to facilitate communication between the schedulers and the migration layer, and based on my understanding at the time.

We now use `MigrationClientInterface` as the adapter between the schedulers and the Migration Layer. This adapter is used to stream migration requests at runtime. The change on the Inference Server side is mostly an uplift to new APIs.

Edit: I see this being done for the decode scheduler - should be easier than I initially anticipated.

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_prefill_runner.cpp
```diff
@@ -681,10 +683,12 @@ void BlazePrefillRunner::handleRequest(
       TT_LOG_DEBUG(
           "[BlazePrefillRunner] handleRequest: taskId={}, slotId={}, "
           "isContinuation={}, numPromptTokens={}, totalTokens={}, "
-          "runningSlots={}",
+          "runningSlots={}, migrationId={}",
           request->taskId, slotId, request->isContinuation(),
           request->getNumPromptTokens(), request->getTokenIds().size(),
-          slotManager.activeRunningCount());
+          slotManager.activeRunningCount(),
+          request->getMigrationId().has_value() ? *request->getMigrationId()
```
tt-asaigal: I think we should abort the process if the request does not provide a Migration ID. Setting it to -1 means that a migration will still be issued and will show up on the decode scheduler. The decode scheduler will then ignore this migration or worse will have undefined behaviour (depending on whether -1 was passed as a corresponding migration id to it or not).

tldr: I don't think a prefill request without an associated migration request makes sense and we shouldn't enter this state.

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_decode_runner.cpp
```diff
@@ -713,10 +709,12 @@ void BlazeDecodeRunner::handleRequest(
       TT_LOG_DEBUG(
           "[BlazeDecodeRunner] handleRequest: taskId={}, slotId={}, isNew={}, "
           "isContinuation={}, numPromptTokens={}, totalTokens={}, "
-          "runningSlots={}",
+          "runningSlots={}, migrationId={}",
           request->taskId, slotId, isNew, request->isContinuation(),
           request->getNumPromptTokens(), request->getTokenIds().size(),
-          slotManager.activeRunningCount());
+          slotManager.activeRunningCount(),
+          request->getMigrationId().has_value() ? *request->getMigrationId()
```
tt-asaigal: For simplicity, please pass in a valid migration ID here as well. I don't think this should be optional.

knovokmetTT: what if the server is running in non disaggregated mode? should we default to passing  1 / 0?

tt-media-server/cpp_server/include/domain/llm/sequence.hpp
knovokmetTT: Yes, this is unique and gets set in disaggregation_service.cpp

tt-media-server/cpp_server/include/config/defaults.hpp
knovokmetTT: Got it, I will change the default values

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_prefill_runner.cpp
knovokmetTT: the server can run in non disaggregated mode , which is why I thought this is ok. If we are in DP, we assume migrationId will always be valid

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
knovokmetTT: we are already tracking this state. RIght, @idjuricTT ?

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
knovokmetTT: Fixed

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_decode_runner.cpp
knovokmetTT: also both comments about this are actually misleading. This is just logging
real migration_uuid is passed in blaze_utils.hpp, in makeGenerationParams function
its passed raw as we get it from upstream:
      .migration_uuid = seq.getMigrationId()}; , where seq holds an optional. I assume it will be nullopt if we are not running DP

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/blaze_utils.hpp
idjuricTT: Yep, that’s covered

ljovanovicTT (approved):

Issue #4120, opened 2026-06-10
Cached tokens  is always 0 for local prefix-cache hits in decode mode (disaggregated) -dynamo 


ddjukicTT:
In disaggregated mode, prefix-cache hits that are prefilled locally on the decode server are not reported in the API response: `usage.prompt_tokens_details.cached_tokens `stays 0 even though the worker log confirms the hit and the KV reuse works. The only exception is requests offloaded to the prefill server, which report correctly. In aggregated mode (`LLM_MODE`=regular) reporting is correct.

PR #4124, opened 2026-06-10
Create test framework end to end test suite

### LINKED ISSUE #4007: Create test framework end to end test suite
Tests:

- Prefill & Decode are healthy + Dynamo frontend works
- Decode only works on small ISL - 150+ tokens (i.e. 200)
- Decode uses prefix cache if ISL slowly grows (i.e. 32/64 tokens delta) through requests - i.e. 250 tokens where first 200 is the same
- Prefill is engaged when higher (new) ISL is sent - 1000+ tokens
- When more additional ISL is sent prefill uses its cache, also decode uses its cache - first 1000 tokens the same, rest is new total 2000

ddjukicTT:
Issue: https://github.com/tenstorrent/tt-inference-server/issues/4007

**Tests (Dynamo frontend):**

`test_01_health_and_frontend` - both workers report healthy on /health with an inter-server socket, and a chat round-trip through the frontend produces output.

`test_02_small_isl_local_cold_miss` - a fresh prompt of about 200 tokens prefills locally on decode: cold cache MISS in the decode log, nothing in the prefill log.

`test_03_prefix_cache_slow_growth` - a growing prompt with a stable prefix stays local; turn 2 and later must HIT the prefix cache with a growing cached span (currently failing: (https://github.com/tenstorrent/tt-inference-server/issues/4120)).

`test_04_shared_prefix_stays_local `- a prompt of about 400 tokens then one of about 800 sharing the prefix: HIT on the shared span, tail still prefilled locally (same issue failure on the cached_tokens assertion).

`test_05_large_isl_offloads_to_prefill `- a prompt of about 1400 tokens is offloaded; asserted via the "Received prefill request" line in the prefill log, since offloaded requests never complete on the mock stack (KV allocation hang).

**Tests (Native):**
`test_native_roundtrip_no_dynamo() ` - Decode + prefill healthy, socket up, chat round-trips on :8001


tt-media-server/cpp_server/benchmarks/run_stack.sh
ljovanovicTT, line 1: nit: It would be cool to try and extract common parts from `run_stack.sh` and ``deploy.sh` so we dont have so much code duplication, but we can consider this as a followup

ljovanovicTT (approved):

idjuricTT (approved):

idjuricTT:
Could you please attach report output?
We might need it for comparison on a real model

tt-media-server/cpp_server/benchmarks/run_stack.sh
ddjukicTT: Thanks for the feedback! I will address it in a followup (Issue: https://github.com/tenstorrent/tt-inference-server/issues/2848?issue=tenstorrent%7Ctt-inference-server%7C4175)

ddjukicTT:
> Could you please attach report output? We might need it for comparison on a real model

Screenshot 2026-06-12 at 17 46 58

PR #4126, opened 2026-06-10
Fix delta prompts when request remains on decode during disaggregation

idjuricTT:


tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -459,10 +467,16 @@ bool LLMPipeline::shouldDoPrefillOnDecode(
...
 
   const size_t maxTokens = tt::config::maxTokensToPrefillOnDecode();
-  const size_t promptTokens = static_cast<size_t>(request.prompt_tokens_count);
+  size_t promptTokens = static_cast<size_t>(request.prompt_tokens_count);
+
+  // If we have a prefix-cache hit, the matched tokens are already in the KV
+  // cache and won't need prefilling again — deduct them from the effective
+  // prompt size used for the threshold comparison.
+  if (request.kv_position_id.has_value()) {
+    const size_t cached = static_cast<size_t>(*request.kv_position_id);
```
nanicicTT: Maybe use the same formula from above? 

```cpp
const size_t cached =
      static_cast<size_t>(*request.kv_position_id + 1) -
      static_cast<size_t>(request.accumulated_think_tokens);
```

or I am wrong?

nanicicTT (approved):

PR #4133, opened 2026-06-11
decode to prefill prefix cache test

lmuravljovTT:
## Test: Disaggregated Prefill Info Propagation

This test verifies that decode-to-prefill information is correctly propagated in a disaggregated inference setup with prefix caching.

### Test Flow

**Part 1: Initial Large Prompt**
- Sends a large prompt (~1100 tokens) that exceeds the prefill-on-decode threshold
- Verifies decode forwards the request to prefill server
- Prefill receives an ALLOCATE request and establishes a new session

**Part 2: Small Delta Continuation**
- Sends a follow-up with a small delta (< 1000 tokens)
- Verifies decode handles it locally (prefill-on-decode) due to prefix cache HIT
- Confirms prefill does NOT receive the request

**Part 3: Big Delta Continuation**
- Sends a follow-up with a large delta (~1200 tokens)
- Verifies decode forwards to prefill (delta exceeds threshold)
- Validates propagated fields:
  - `decodeSlotId` - decode's KV cache slot is
  - `decodeSkipTokens` - number of cached tokens on decode (block-aligned: 1088)
  - `decodePositionId` - resume position in KV cache
  - `isContinuation` flag is set
  - Prefill receives only delta tokens (1216), not full conversation
- Confirms NO ALLOCATE is sent (prefill has cache HIT from Part 1)

**Part 4: Second Big Delta**
- Verifies prefix cache was updated after Part 3
- `decodeSkipTokens` grows from 1088 to 2304 (
- Confirms NO ALLOCATE (cache HIT on both decode and prefill)


tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
```diff
@@ -0,0 +1,826 @@
...
+};
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+bool fileExists(const std::string& path) {
+  struct stat st{};
+  return stat(path.c_str(), &st) == 0;
+}
```
nanicicTT: why not
https://cppreference.com/cpp/filesystem/exists

nanicicTT:
```cpp
return std::filesystem::exists(path);
```

lmuravljovTT: I'll just use exists() directly, no need to wrap it
Thanks!

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
```diff
@@ -0,0 +1,826 @@
...
+      "conversation)",
+      bigDeltaPromptTokens, 2300);
+
+  // --- Verification 6: Prefill calculates its own prefix cache ---
+  // The prefill server should have resolved its own prefix cache. For this
+  // first big delta request to prefill (after the initial large prompt in Part
+  // 1), prefill should have a cache HIT from Part 1 and trim to just the delta.
+  // This is verified by the numPromptTokens being much smaller than the full
+  // conversation (already checked above).
+  // The prefillSlotId should be valid if prefill found a matching session.
```
nanicicTT, line 693: `prefillSlotId` is logged below, but that variable is never asserted. 
 The comments say this verifies prefill-side prefix-cache resolution, but a regression leaving `prefillSlotId == INVALID_SLOT_ID` would still pass. Add maybe `EXPECT_NE(prefillSlotId, tt::domain::INVALID_SLOT_ID)` for Part 3, and maybe the same for Part 4.

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
```diff
@@ -0,0 +1,826 @@
...
+      addr.sin_family = AF_INET;
+      addr.sin_port = htons(DECODE_HTTP_PORT);
+      ::inet_pton(AF_INET, "127.0.0.1", &addr.sin_addr);
+      bool up = (::connect(sock, reinterpret_cast<sockaddr*>(&addr),
+                           sizeof(addr)) == 0);
+      ::close(sock);
+      if (up) break;
+      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    }
+  }
```
nanicicTT: If `addListener()` fails or port `DECODE_HTTP_PORT` that is already occupied, the parent proceeds as if decode is ready and the test fails later against either no server or the wrong server. Could this be the case? 
I would track whether the connect loop succeeded and `_Exit(1)` on timeout before writing the sentinel.

Here, the problem is that the loop does not remember whether it exited because `connect()` succeeded or because the 30-second deadline expired. Either way, it writes the sentinel file, on line 192. 

 So if Drogon failed to bind `127.0.0.1:18084`, the child still eventually says “ready”. Then the actual test sends HTTP requests to port 18084.

It could happen that nothing is listening on 18084 -> then the test fails later with `sendAndReceive: connect` failed, which points at the request rather than startup.
It could also happen that some other process is already listening on `18084` -> the readiness probe connects successfully to that unrelated process, writes the sentinel, and the test may send requests to the wrong server.

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
```diff
@@ -0,0 +1,821 @@
...
+      bigDeltaSeq = prefillServer->taskQueue().tryPop();
+      if (bigDeltaSeq) {
+        break;
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
+  }
+
+  if (!bigDeltaSeq) {
+    bigDeltaFuture.wait();
```
idjuricTT, line 608: Could you please add more checks here:

- Tokens sent to sequence should be less than tokens received since prefix cache should kick in on prefill (you can even use the math to match the exact number)
- MIgrationId should be in
- Decode skip tokens should be set and > 0
- Decode position id should be > 0

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
nanicicTT:
```cpp
bool listenerUp = false; // I would set this initially to false
while (std::chrono::steady_clock::now() < listenerDeadline) { // the loop might be as-is
  // ... rest of the code here, use listenerUp instead of up
  if (listenerUp) break;
  // rest of the code ...
}

if (!listenerUp) {
  std::_Exit(1); // it was not up
}

std::ofstream(sentinelPath) << "ready";
```

idjuricTT (approved):

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
nanicicTT: exactly, it is one-liner now anyhow :)

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
lmuravljovTT: We're grabbing these just below, and validating right after

`  const size_t bigDeltaPromptTokens = bigDeltaSeq->getNumPromptTokens();
  const bool bigDeltaIsContinuation = bigDeltaSeq->isContinuation();
  const uint32_t decodeSlotId = bigDeltaSeq->getKVCacheSlot();
  const uint32_t prefillSlotId = bigDeltaSeq->getPrefillKVCacheSlot();
  const int decodeSkipTokens = bigDeltaSeq->getDecodeSkipTokens();
  const int decodePositionId = bigDeltaSeq->getDecodePositionId();`

Is there something else you'd expect?

lmuravljovTT: I'll add migrationId check that's not in

tt-media-server/cpp_server/tests/integration/disaggregated_e2e_test.cpp
nanicicTT: @lmuravljovTT Do you think this makes sense? :) 
The comment above says this is checking that the prefill server resolved its own prefix cache. But there's no actual `EXPECT_` assertion on that value, it's just printed.
If `prefillSlotId` would come back as `INVALID_SLOT_ID` (essentially it means that no slot is found) => the test would still pass green because nothing checks that value.

lmuravljovTT: Yep, I just pushed the EXPECT_NE you suggested

nanicicTT (approved):
🚀

PR #4148, opened 2026-06-11
Generate migrationId on prefill and share back to decode & sequence

idjuricTT:


tt-media-server/cpp_server/include/utils/id_generator.hpp
```diff
@@ -29,4 +30,21 @@ class TaskIDGenerator {
...
+ * inter-server socket.
+ */
+class MigrationIDGenerator {
+ public:
+  static uint64_t generate() {
+    thread_local std::mt19937_64 gen = []() {
+      std::random_device rd;
+      return std::mt19937_64(rd());
+    }();
+    thread_local std::uniform_int_distribution<uint64_t> dist;
```
ztorlakTT, line 45: Is it ok to return 0? If not, we can have smth like:
```suggestion
    thread_local std::uniform_int_distribution<uint64_t> dist(
    1, std::numeric_limits<uint64_t>::max());
```
and `#include <limits>`

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
ztorlakTT, line 268: Can we set migration_id on this error result too?

tt-media-server/cpp_server/src/services/llm_service.cpp
```diff
@@ -632,7 +632,8 @@ void LLMService::produceStream(
       std::make_unique<tt::domain::llm::SamplingParams>(
           tt::utils::mapper::mapSamplingParams(request)),
       request.kv_position_id, request.decode_position_id,
-      request.decode_skip_tokens);
+      request.decode_skip_tokens, request.slotToCopyFrom,
```
ztorlakTT: I might be missing something, but I don’t see a slotToCopyFrom field on LLMRequest 😄

idjuricTT: Fixed, other pr went in

knovokmetTT (approved):

tt-media-server/cpp_server/include/utils/id_generator.hpp
idjuricTT: yep, 0 will never match the real id so it should be fine

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
idjuricTT: it'll never reach the decode scheduler so this should be fine I guess

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -470,6 +470,16 @@ void LLMPipeline::dispatchGeneration(
           "[LLMPipeline] Using disaggregated prefill for request with "
           "sessionId: {}",
           request.sessionId.value_or("none"));
+      // WARNING - TEMP CHANGE - PREFILL WILL OVERRIDE THINKING TOKENS
```
idjuricTT, line 473: @ljovanovicTT let's discuss this tomorrow pls

tt-media-server/cpp_server/include/sockets/socket_messages.hpp
```diff
@@ -126,6 +126,9 @@ struct PrefillResultMessage {
   // (prefix-cache reuse). The decode server surfaces this as
   // usage.prompt_tokens_details.cached_tokens.
   int cached_tokens = 0;
+  // Unique 64-bit ID correlating this prefill result with the migration
+  // (KV transfer) that produced it. Generated on the prefill server.
+  uint64_t migration_id = 0;
```
dmadicTT, line 131: Nit: let's keep it `migrationId` everywhere.
Seems like our CI is broken and does not catch styling.

tt-media-server/cpp_server/tests/integration/prefill_integration_test.cpp
dmadicTT, line 1: Nice tests!

tt-media-server/cpp_server/include/services/session_manager.hpp
```diff
@@ -190,6 +190,13 @@ class SessionManager {
       const std::string& sessionId,
       const std::vector<utils::BlockHashInfo>& blockInfos);
 
+  /**
+   * Reset accumulatedThinkTokens to 0 on all prefix index entries that contain
+   * the given session. Called when prefill-on-decode overrides thinking tokens
```
dmadicTT, line 195: >Called when prefill-on-decode overrides thinking tokens

Doesn't this happen only on disaggregated prefill?

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -470,6 +470,16 @@ void LLMPipeline::dispatchGeneration(
...
+      uint32_t matchedTokens =
+          *request.kv_position_id -
+          static_cast<uint32_t>(request.accumulated_think_tokens);
+      *request.kv_position_id = matchedTokens;
```
dmadicTT, line 477: As we are not changing `request.accumulated_think_tokens` to zero, can this lead to negative `decodeSkipTokens`?

**Scenario**
Initial user prompt was 1500 tokens and it lead to 4000 thinking tokens and 500 assistant tokens
```
request.kv_position_id=6000
request.accumulated_think_tokens=4000
```

This code leads removes the thinking tokens from position ID, but accumulated_thinking_tokens remains the same
```
request.kv_position_id=2000
request.accumulated_think_tokens=4000
```

Leads to negative decodeSkipTokens
https://github.com/tenstorrent/tt-inference-server/blob/main/tt-media-server/cpp_server/src/services/disaggregation_service.cpp#L437

tt-media-server/cpp_server/tests/integration/main_integration_test.cpp
dmadicTT, line 1: I believe `decodeSkipToken` and `decodePositionID` should always be the same with the changes from this PR?

What do you think about having a test that covers how `decodeSkipToken` and `decodePositionId` change in multi-turn scenario?
This has to be a test including few components, like disaggregation service and llm pipeline, since `kv_position_id` and `accumulated_tokens_count` are maniuplated in few places.

As I understood `decodeSkipToken` is position on the prefill side where to start KV migration from. Using this value, the decode gives signal to prefill to not send the whole KV cache, but to start from a specific position.

Image

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
ljovanovicTT, line 477: Is the goal of these changes to maintain the old behavior, so when we are able to migrate partial KV cache from prefill to decode our code is ready? 

Because with this line we are changing both:
```
decodePositionId = kv_position_id + 1 
```
and
```
decodeSkipTokens = decodePositionId − accumulated_think_tokens
```
to a new behavior (not maintaining the old), and also not preparing KV migration to send from position 0.

Im just unsure why we have this line:
```
      *request.kv_position_id = matchedTokens;
```

tt-media-server/cpp_server/tests/integration/main_integration_test.cpp
ljovanovicTT: Again, not sure what our goal is atm. If we are going to temporarily hardcode migrationDestinationAdr (decode kv cache address) on prefill to 0, why are we changing any logic to `decodeSkipToken` and `decodePositionID`?

Final `kv_position_id` is set on Prefill and sent over to Decode, and it works with the wanted behavior

dmadicTT: Hardcoding `migrationDestinationAdr=0` would mean we send the full KV cache always. I think we want to send just delta.

dmadicTT: My bad, we we migrating the whole KV

tt-media-server/cpp_server/include/services/session_manager.hpp
idjuricTT: True, I'll fix the comment

tt-media-server/cpp_server/include/sockets/socket_messages.hpp
idjuricTT: yep, our messages are not in camel case it seems

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
idjuricTT: Decode position id will be ignored for now. It'll just migrate all in the first iteration.
That's why only position id is changed

Issue #3752, opened 2026-05-26
Measure reference eval scores for moonshotai/Kimi-K2.6


bgoelTT:
The reference eval scores need to be recorded for the following tasks on `moonshotai/Kimi-K2.6`:
* GPQA-Diamond
* Terminal-Bench-2.0
* SWE-Bench Verified

The goal is to apply the existing integration of the agentic evals for this model.

**Note:** the full dataset must be used as we need to try our best to reproduce the scores claimed by Moonshot

ipastalTT:
Draft:
- need to enable preserve thinking as well. Will rerun terminal_bench again to verify it helps.
- 16 parallel tasks

vllm command : 

```bash
docker run -d --gpus all  \
--privileged \
--ipc=host -p 8000:8000   \
-v ~/.cache/huggingface:/root/.cache/huggingface  \
 vllm/vllm-openai:latest moonshotai/Kimi-K2.6   \
--trust-remote-code   \
--tensor-parallel-size 8  \
--tool-call-parser kimi_k2 \
--enable-auto-tool-choice   \
--reasoning-parser kimi_k2   \
--speculative-config '{"model":"lightseekorg/kimi-k2.6-eagle3-mla","method":"eagle3","num_speculative_tokens":3}' \
--mm-encoder-tp-mode data
```

### Accuracy Evaluations for Kimi-K2.6 on gpu

| model     | device | task_name        | task_type | accuracy_check | score | ratio_to_reference | gpu_reference_score | ratio_to_published | published_score                                      |
|-----------|--------|------------------|-----------|----------------|-------|--------------------|---------------------|--------------------|------------------------------------------------------|
| Kimi-K2.6 | gpu    | r1_gpqa_diamond  | text      | PASS ✅         | 90.91 | N/A                | [None](TBD)         | 1.00               | [90.50](https://huggingface.co/moonshotai/Kimi-K2.6) |
| Kimi-K2.6 | gpu    | terminal_bench_2 | text      | FAIL ⛔         | 55.06 | N/A                | [None](TBD)         | 0.83               | [66.70](https://huggingface.co/moonshotai/Kimi-K2.6) |

##
- with 8 parallel and preserve thinking mode

### Accuracy Evaluations for Kimi-K2.6 on gpu

| model     | device | task_name          | task_type | accuracy_check | score | ratio_to_reference | gpu_reference_score | ratio_to_published | published_score                                      |
[...]

PR #4153, opened 2026-06-11
add: Kimi-K2.6 reference evaluation scores

ipastalTT:
## Scope
Include the reference evaluation scores we measured for Kimi-K2.6 on GPU rental.
Reference scores are detailed in [#3752](https://github.com/tenstorrent/tt-inference-server/issues/3752)

## What this PR includes
Reference scores on GPU for:

- GPQA-Diamond
- Terminal-Bench-2.0
- SWEBench-Verified

For Terminal Bench we selected tasks that the model found a solution 10/10 times, but also they finished in reasonable time.
- For all the selected tasks mean time to completion was between 10-20 minutes, with max time to completion over 10 runs per task being ~40 minutes.

For SWEBench-Verifed  tasks that finished quickly were selected (with score 1.0) due to only running it once.

## What it is not included

Score for SWEBench-Pro since it is not yet integrated in tt-inference-server


evals/eval_config.py
```diff
@@ -142,6 +142,159 @@ class EvalConfig:
...
+                model_kwargs={
+                    "max_length": 120 * 1024,
+                    # Per-request HTTP timeout (lm-eval default 1800s). Long
+                    # reasoning generations on the shared console can exceed
+                    # 30min under load, so allow up to 2h before giving up.
+                    "timeout": 7200,
+                },
+                gen_kwargs={
+                    "max_gen_toks": 120 * 1024,
+                    "until": ["[EOS]"],
```
bgoelTT, line 177: We don't want to have to specify this right? You didn't need this during the vLLM GPU runs

bgoelTT (approved):

evals/eval_config.py
ipastalTT: correct previously we would just need `"until": [ ],` but with this one we were getting errors

ipastalTT: found the logs 

```txt
2026-06-05:19:35:02 WARNING  [models.api_models:522] API request failed! Status code: 400, Response text: {"message":"Validation: Stop sequences array cannot be empty","type":"Bad Request","code":400}. Retrying...
```

PR #4172, opened 2026-06-12
Share session resolution helpers

### LINKED ISSUE #4061: Merge llm pipeline & disaggregation service common logic
Disaggregation service and llm pipeline have some overlapping logic (applyDeltaPrompts /  resolveSession) that should not be duplicated

ztorlakTT:
- Extract shared delta prompt trimming and slot-copy planning into `session_resolution`.
- Update LLM pipeline and disaggregation prefill session resolution to use the common helpers.


tt-media-server/cpp_server/src/services/session_resolution.cpp
```diff
@@ -0,0 +1,48 @@
...
+    const std::vector<SessionManager::Candidate>& candidates, uint32_t taskId,
+    std::string_view logPrefix) {
+  SlotCopyPlan plan;
+  if (candidates.empty()) {
+    return plan;
+  }
+
+  auto copyCandidate = sessionManager.findASlotToCopyFrom(candidates);
+  if (!copyCandidate.has_value()) {
+    return plan;
```
idjuricTT: Why do we return a plan here? Shouldn't it be just nothing?

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -323,31 +315,13 @@ void LLMPipeline::resolveSession(
...
-      }
-    }
-  }
+  auto copyPlan = acquired.has_value()
+                      ? session_resolution::prepareSlotCopy(
+                            *sessionManager_, acquired->candidatesList,
+                            req->task_id, "[LLMPipeline]")
+                      : session_resolution::SlotCopyPlan{};
+  std::optional<uint32_t> slotToCopyFrom = copyPlan.slotToCopyFrom;
+  uint32_t copyMatchedTokens = copyPlan.matchedTokens;
```
idjuricTT: Could you please check if prefixes are copied too?
if that doesn't work we'll have issues here

tt-media-server/cpp_server/src/services/session_resolution.cpp
ztorlakTT: Updated this to be clearer, so "no copy plan" is represented by `std::nullopt` instead of an empty plan with an optional `slotToCopyFrom`

idjuricTT (approved):

PR #4197, opened 2026-06-15
Vm/change dev model yaml

vmaksimovicTT:
This PR mainly removes `tt_metal_commit`, `vllm_commit`, `version` and `docker_image` fields from the ModelSepc templates in the `dev` catalogue.

These 4 fields are used mainly for generating helm chart values and docs, which are note being generated from the `dev` catalogue, but rather `prod`. The dev catalogue is mainly used for running model specs in Models CI, and there, version of tt_metal, inference server and vllm are taken from GH Action input parameters, not `dev` model spec.

Why we want to remove these fields?

Well, these are the last files which cause conflicts at the ned of every release cycle (Post Release PR), which in turn prevent us from fully automating it.

### DIFF
--- a/workflows/model_spec.py
+++ b/workflows/model_spec.py
@@ -36,8 +36,18 @@ MODEL_SPECS_SCHEMA_VERSION = "0.1.0"
 
 
 def generate_docker_tag(
-    version: str, tt_metal_commit: str, vllm_commit: Optional[str]
+    version: Optional[str],
+    tt_metal_commit: Optional[str],
+    vllm_commit: Optional[str],
 ) -> str:
+    # version and tt_metal_commit are required to build a tag. They are
+    # optional on the spec (dev catalog omits them), so guard here: callers
+    # that synthesize an image from a dev spec are a bug, fail loudly rather
+    # than emit a "None-None" tag.
+    if version is None:
+        raise ValueError("Cannot generate docker tag: version is None")
+    if tt_metal_commit is None:
+        raise ValueError("Cannot generate docker tag: tt_metal_commit is None")
     max_tag_len = 12
     if vllm_commit:
         return f"{version}-{tt_metal_commit[:max_tag_len]}-{vllm_commit[:max_tag_len]}"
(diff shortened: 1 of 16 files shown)


workflows/model_spec.py
```diff
@@ -1084,11 +1105,43 @@ def _build_template(data: Dict) -> "ModelSpecTemplate":
     return ModelSpecTemplate(**kwargs)
 
 
+# prod templates must pin these; dev templates must omit all of _DEV_FORBIDDEN.
+# Enforced per-catalog-directory in load_templates_from_yaml. vllm_commit is not
+# prod-required (CNN/FORGE specs legitimately have none) but is dev-forbidden.
+_PROD_REQUIRED_FIELDS = ("tt_metal_commit", "version")
+_DEV_FORBIDDEN_FIELDS = ("tt_metal_commit", "vllm_commit", "version", "docker_image")
```
bgoelTT: Instead of this pattern, would it be possible to modify the `ModelSpec` dataclass to implement the following behaviour?
* `DevModelSpec` defines the ModelSpec to be used in the `dev` specs
  * does not require these forbidden fields
* `ModelSpec` defines the ModelSpec to be used in the `prod` specs
  * is a sub-class of `DevModelSpec`
  * requires the fields outlined above

Overall, I think we should make this requirement more pythonic rather than doing this retrospective check, for better code readability. Additionally, you may want to invert my above suggestion by making the `DevModelSpec` be a sub-class of the `ModelSpec`

workflows/model_spec.py
```diff
@@ -36,18 +36,32 @@
...
 def generate_docker_tag(
-    version: str, tt_metal_commit: str, vllm_commit: Optional[str]
+    version: Optional[str],
+    tt_metal_commit: Optional[str],
+    vllm_commit: Optional[str],
 ) -> str:
+    if version is None or tt_metal_commit is None:
+        raise ValueError(
+            "Cannot generate docker tag: version and tt_metal_commit are required"
+        )
```
bgoelTT, line 46: Doesn't this make `--dev-mode` raise errors locally? I understand that for Models CI which uses `--override-docker-image` this shouldn't be an issue but without the flag I believe this will be hit every time

bgoelTT (changes requested):
Just two comments

workflows/model_spec.py
vmaksimovicTT: Valid point raised! I did fix this implicit/explicit validation, just used different dataclasses. Validation is performed on the level of ModelSpecTemplate, not ModelSpec, so I created a new ProdModelSpec that includes new fields, that are not required for ModelSpecTemplate

bgoelTT (approved):
Thanks! Agree that with `--dev-mode --workflow server` users will need to also pass `--override-docker-image`

PR #4237, opened 2026-06-16
ci: add cpp-blaze-smoke as standalone build gate in test-gate

dmadicTT:
Catch early regressions for blaze builds instead of only in CPP Heavy checks builds


ljovanovicTT (approved):

.github/workflows/test-gate.yml
```diff
@@ -564,6 +564,67 @@ jobs:
...
+          token: ${{ secrets.TT_LLM_ENGINE }}
+
+      - name: Set up C++ CI dependencies
+        uses: ./.github/actions/setup-cpp-ci
+        with:
+          token: ${{ secrets.TT_LLM_ENGINE }}
+          # Isolated ccache namespace — must NOT share with cpp-build's
+          # `server` suffix, otherwise this gate inherits the same warm cache
+          # whose stale .o files we are trying to bypass.
+          ccache-key-suffix: blaze-smoke
```
ljovanovicTT, line 592: Might be missing something, but if we want a fresh build every time , why are we saving build cache?

PR #4253, opened 2026-06-16
Set default migraiton id to null, not 0

idjuricTT:


lmuravljovTT (approved):

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
```diff
@@ -210,7 +210,7 @@ void DisaggregationService::setupSocketHandlers() {
                 // async callback fires.
                 const std::string prefillSessionId =
                     request->sessionId.value_or("");
-                const uint64_t migrationId = request->migrationId;
+                const uint64_t migrationId = request->migrationId.value_or(0);
```
knovokmetTT: should we throw / log if this happens? if request doesnt have migId attached. smth went wrong

ztorlakTT (approved):

tt-media-server/cpp_server/src/services/disaggregation_service.cpp
idjuricTT: Fair, I’ll add a check

PR #4311, opened 2026-06-19
Dynamic discovery for MooncakeMigrationWorker

### LINKED ISSUE #4294: Prod-ize dynamic discovery of migration workers
The discovery PoC (tests/integration/migration_worker_discovery.cpp) proves the concept.

---

##### Goal

Move discovery into a proper component that the existing transfer classes own, so it runs automatically on worker bring-up instead of being hand-driven in a test.

Scope
- Add `main.cpp` for starting a migration worker. This method runs the discovery and creates `mooncake_migration_worker.cpp` with `HostDRAM` storage. Does not do anything else for now.
- Discovery mechanism is encapsulated in it's own class/function executed on startup of the worker. For example `discoverPeers()`.
- The worker must not start until it found it peers. Stay in a while loop until they connect.
- Logging is added
- README how to start a worker metadata service and the worker is added

The following integration test is added:
- runs on single host in the CI
- starts metadata service
- start 4 "prefill" workers and 16 "decode" workers at once using `mpi`
- prefill 0 peers are decode 0, decode 1, decode 2, decode 3
- prefill 1 peers are decode 4, 5, 6, 7
- prefill 2: 8, 9, 10, 11
- prefill 3: 12, 13, 14, 15
- the tests verify they successfully connected to each other after N seconds - if yes, test is successful

vpetrovicTT:


nanicicTT (approved):

fivanovicTT (approved):

nanicicTT (approved):

tt-media-server/cpp_server/src/transport/mooncake_migration_worker.cpp
```diff
@@ -3,16 +3,125 @@
...
+  memoryRegistered_ = true;
+  TT_LOG_INFO("[MooncakeMigrationWorker] '{}' published {} bytes",
+              config_.segment_name, hostDramPool_.size());
+
+  // Phase 5: discover peers — the readiness gate. The worker owns *when* this
+  // happens (after publish, so peers can resolve us back) and forwards the
+  // cancel token so a stop request aborts discovery promptly; the injected
+  // PeerDiscoveryService owns *how*.
+  auto resolved =
+      discovery_->discover(*engine_, config_.peer_segment_names, &cancelToken);
```
dmadicTT, line 81: We discussed to do the discovery "outside" of this class, in `bringup_mooncake_worker` and inject the peers as parameters of the `bringUp` method.

I assume you still added it here because discovery must happens after registration? This makes sense, just checking :)
>// Phase 4: register memory — this publishes our segment to the cluster.

tt-media-server/cpp_server/tests/e2e/transport_migration_e2e.cpp
```diff
@@ -360,7 +361,8 @@ int runReceiver(const Options& o) {
   wcfg.role = MigrationRole::Receiver;
   wcfg.device_addr = storage.device_addr;
   wcfg.tensor_bytes = o.bytes;
-  MooncakeMigrationWorker worker(wcfg, engine);
+  // Discovery is unused here — this PoC drives the data-plane spike directly.
+  MooncakeMigrationWorker worker(wcfg, engine, /*discovery=*/nullptr);
```
dmadicTT: I find constructors more readable using curly braces. This way you don't know is `worker` a function or a constructor.
```
MooncakeMigrationWorker worker{wcfg, engine, /*discovery=*/nullptr};
```

I have this Cursor rule
>For C++ code prefer, for example engine{config} over engine(config) constructor call

dmadicTT (approved):

tt-media-server/cpp_server/tests/e2e/transport_migration_e2e.cpp
nanicicTT: https://www.fluentcpp.com/2018/01/30/most-vexing-parse/

@dmadicTT is referring to this


BTW: excellent blog, it has many useful C++ stuff.

tt-media-server/cpp_server/src/transport/mooncake_migration_worker.cpp
vpetrovicTT: Yeah, exactly!

tt-media-server/cpp_server/tests/e2e/transport_migration_e2e.cpp
vpetrovicTT: Thanks guys! Will look deeper into this!

PR #4316, opened 2026-06-20
Create 2 venvs for vllm and whisperx

### LINKED ISSUE #4262: Create 2 venvs for vllm and whisperx
This [PR](https://github.com/tenstorrent/tt-inference-server/pull/3457) introduced the issue where uplifting the vllm to 0.18 version required torch version at least 2.10 which collides with whisperx and it's required version of torch = 2.8.

Solution would be to create 2 venvs - one for vllm and the other for whisperx. Cons is that we will have 2 torch versions in our docker image.

Proposed fix:
- Main venv: torch 2.10 + vLLM 0.18.1
- Audio venv: torch 2.7.x + whisperx + pyannote

Changes needed:
- Dockerfile: Create a second venv for audio dependencies
- audio_manager.py: Shell out to a diarization script using the audio venv
- New script: diarize.py that runs in the audio venv and returns JSON results

fivanovicTT:
### Github issue
[Create 2 venvs for vllm and whisperx](https://github.com/tenstorrent/tt-inference-server/issues/4262)

### Summary
Solution would be to create 2 venvs - one for vllm and the other for whisperx. Cons is that we will have 2 torch versions in our docker image.

Proposed fix:

Main venv: torch 2.10 + vLLM 0.18.1
Audio venv: torch 2.7.x + whisperx + pyannote


tt-media-server/utils/audio_manager.py
```diff
@@ -2,82 +2,341 @@
...
 
-    The audio worker only loads checkpoints from trusted sources (gated HF
-    repos under HF_TOKEN and bundled weights); no user-supplied `.pt` ever
-    reaches `torch.load` in this process, so the strict default offers no real
-    safety here.
+        cmd: List[str] = [self._python, str(self._script), "--serve"]
+        if self._model_name:
+            cmd.extend(["--model-name", self._model_name])
+        if self._hf_token:
+            cmd.extend(["--hf-token", self._hf_token])
```
ztorlakTT: Can we avoid passing `HF_TOKEN` as a CLI arg? Since `diarize.py` already falls back to `HF_TOKEN` from env, I think we can just rely on env instead

tt-media-server/utils/audio_manager.py
```diff
@@ -247,109 +506,119 @@ def _merge_vad_segments_by_speaker_and_duration(
...
-                self._logger.info(
-                    f"Loading VAD model... (attempt {attempt + 1}/{total_attempts})"
-                )
-
-                # VAD requires vad_onset and chunk_size parameters
-                # chunk_size: size of audio chunks to process (typical values: 30, 60, or 160)
-                # vad_onset: threshold for detecting speech onset (typical value: 0.500)
-                self._vad_model = load_silero_vad()
+        self._audio_venv_worker = worker
+        self._diarization_model = True
```
ztorlakTT, line 556: I think this can mark diarization as available before the actual pyannote is loaded, so the first diarization request can still fail later. In that case we return empty diarization segments and may fall back to one full-audio segment instead of VAD-only. Is this correct?

Maybe we should either make diarization failure fall back to VAD-only, or set `perform_diarization=True` in `warmup_task_data` in `model_services/audio_service.py` so pyannote is warmed earlier

tt-media-server/utils/audio_manager.py
```diff
@@ -2,82 +2,341 @@
...
-    "use legacy default" — lightning's `pl_load` explicitly forwards
-    `weights_only=None`, which would otherwise bypass a plain
-    `dict.setdefault()`. Callers that pass `weights_only=True/False`
-    explicitly are respected.
+        try:
+            self._proc = self._popen_factory(
+                cmd,
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
```
ztorlakTT: Small concern here: we pipe stderr but never read it, and if stderr fills up the child process can block while parent is waiting on stdout

ztorlakTT (approved):

vpetrovicTT (approved):

PR #4321, opened 2026-06-22
integration tests should go through dynamo route

lmuravljovTT:
# Dynamo Integration Test Infrastructure

## Overview

The cpp_server integration tests (`MainIntegrationTest`, etc.) verify end-to-end request flow through the Dynamo stack: HTTP → Dynamo frontend → etcd discovery → DynamoEndpoint → LLMPipeline → IPC queues.

## Architecture

```
Test Process (in-process backend)         External Infrastructure
┌───────────────────────────────────┐    ┌─────────────────────────┐
│ TestServer                        │    │ Dynamo Frontend         │
│   ├─ LLMService                   │    │   (Python router)       │
│   ├─ SessionManager               │    │                         │
│   ├─ DynamoEndpoint ◄─────────────┼────┤ etcd                    │
│   ├─ IPC queues (mocked worker)   │    │   (service discovery)   │
│   └─ Memory auto-responder        │    └─────────────────────────┘
│                                   │
│ Test harness sends HTTP via       │
│ Dynamo frontend, inspects IPC     │
│ queues, injects mock responses    │
└───────────────────────────────────┘
```

### In-Process Backend (vs Separate Process)

**Choice**: Run the cpp_server stack in-process within the test binary.

**Why**:
- Enables gray-box testing: tests can inspect IPC queues (`taskQueue().receive()`), control memory allocation (`setMemoryAutoRespond(false)`), and inject mock worker responses (`mockWorkerResponse()`)
- Faster iteration: no process spawn overhead, easy to attach debugger
- Sub-second per-test execution after one-time suite setup

**Tradeoff**: Less production-like than CI heavy checks, which spawn the backend as a separate process. However, the internal visibility is valuable for verifying prefix-cache behavior, slot allocation, and continuation flags.

### Two-Phase Readiness Check

**Components**:
1. `waitForEtcdBackendRegistration()` — polls etcd directly via `EtcdClient.hasKeysWithPrefix()` until the backend registers
2. `waitForModelDiscovery()` — polls `GET /v1/models` on the Dynamo frontend until the model appears

**Why two phases**:
- Better diagnostics: if phase 1 fails, the backend didn't register; if only phase 2 fails, etcd is fine but the frontend didn't discover it
- The frontend's etcd watch has propagation delay — a key existing in etcd doesn't mean the router can route to it yet

**Alternative considered**: CI heavy checks use only `curl /v1/models` polling (single phase). This works because they don't need the diagnostic granularity — they just care "is it ready?". For in-process tests where we control both sides, the two-phase approach helps debug failures.

### `DynamoEndpoint::abandon()` for Test Teardown

**Problem**: The standard `stop()` method joins in-flight Dynamo call-home streams, which can block indefinitely waiting for network I/O. This caused ctest timeouts (300s default).

**Solution**: `abandon()` detaches threads and releases (leaks) resources instead of joining/destroying them. The OS reclaims everything on process exit.

**Why its safe**: Only used in test teardown where the process exits immediately. Not appropriate for production graceful shutdown.

### EtcdClient in Tests Only

`EtcdClient` and `hasKeysWithPrefix()` exist solely for test readiness checks. Production code uses the Dynamo SDK's etcd integration. If we converted to separate-process tests, we could delete this code and use the same `curl /v1/models` approach as CI.

## Comparison with CI Heavy Checks

| Aspect | Integration Tests | CI Heavy Checks |
|--------|-------------------|-----------------|
| Backend | In-process | Separate process |
| Readiness | Two-phase (etcd + /v1/models) | Single-phase (/v1/models only) |
| Worker | Mocked via IPC queues | Real worker (mock device) |
| Visibility | Gray-box (can inspect internals) | Black-box (HTTP only) |
| Script | C++ gtest | Shell + wait_for_command.sh |

## Future Considerations

Converting to separate-process tests would:
- Align with CI heavy checks (same readiness pattern)
- Allow deleting `EtcdClient`, `hasKeysWithPrefix`, `waitForEtcdBackendRegistration`
- Lose gray-box visibility (or require cross-process IPC queue access)
- Add ~10-20s startup overhead per suite

Current consensus: keep in-process for the internal visibility it provides.


idjuricTT (approved):

tt-media-server/cpp_server/src/dynamo/dynamo_endpoint.cpp
```diff
@@ -507,6 +507,28 @@ void DynamoEndpoint::start() {
...
+
+  TT_LOG_INFO("[DynamoEndpoint] Abandoning (test teardown)");
+  if (server_) {
+    server_->shutdown();
+  }
+  if (discovery_) {
+    discovery_->unregisterSelf();
+  }
+  if (keepalive_thread_.joinable()) {
+    keepalive_thread_.detach();
```
ztorlakTT: Should we join keepalive thread instead? 
```suggestion
    keepalive_thread_.join();
```

ztorlakTT (approved):

PR #4325, opened 2026-06-22
Real KV cache addressing and Mooncake TCP migration (#4173)

### LINKED ISSUE #4173: Real prefill to decode KV-cache block transfer with device addressing, discovery, and scheduler-triggered start
The Mooncake transfer engine PoC (#3890) moved a single blob between Galaxies' hardcoded address in device DRAM over Mooncake TCP, staged through a UMD-backed host bounce buffer.

Goal of this ticket is to turn that into a real prefill to decode KV-cache migration: 

- address the actual KV cache unit by (slot, layer, position) on both the read (prefill) and write (decode) sides, 
- discover peers dynamically (to be done in #4209), 
- and start the transfer on a prefill scheduler signal (postponed for future work).

Keep Mooncake TCP as the host to host wire between each server's migration workers.

visnjakrsmanovicTT:
- Add Mooncake KV cache migration stack over TCP: control channel, migration orchestrator, KV table adapters, sender/receiver, for the N prefill -> M decode topology.
- The receiver stages KV into a host mirror buffer, 1:1 byt image of the dst device region, allowing the sender to compute full dst addressing, which is needed for future device to deivce RDMA.


tt-media-server/cpp_server/src/transport/kv_control_channel.cpp
```diff
@@ -0,0 +1,51 @@
...
+  }
+  return transport_->sendRawData(bytes);
+}
+
+std::optional<KvControlMessage> KvControlChannel::receive() {
+  if (!transport_) {
+    TT_LOG_ERROR("[KvControlChannel] receive with no transport");
+    return std::nullopt;
+  }
+  const std::vector<uint8_t> bytes = transport_->receiveRawData();
```
nanicicTT: This code here treats an empty read as disconnect, while [tt-media-server/cpp_server/src/sockets/tcp_socket_transport.cpp:375](https://github.com/tenstorrent/tt-inference-server/blob/e8447fa9de2ead88ae38295061070d0a3cfe977d/tt-media-server/cpp_server/src/sockets/tcp_socket_transport.cpp#L375) returns empty when no data is currently available. A normal response delay therefore **aborts migration**. The E2E test hides this with a custom blocking TCP implementation.

---

 The problem is in a mismatch:

  1. Sender sends `BEGIN_MIGRATION`.
  2. Sender immediately calls `channel_.receive()` waiting for `MIRROR_READY`.
  3. The receiver may still be preparing the mirror.
  4. `TcpSocketTransport::receiveRawData()` performs the initial `recv()` with `MSG_DONTWAIT`.
  5. If the response has not arrived yet, it returns an empty vector as `NO_DATA`.
  6. `KvControlChannel::receive()` interprets any empty vector as a closed connection:
```cpp
  if (bytes.empty()) {
    return std::nullopt;
  }
```
  7. Then `migrate()` treats that as failure and aborts:

```cpp
  if (!ready || ready->type != KvControlType::MIRROR_READY) {
    return false;
  }
```

So a normal response delay could be only a few milliseconds and it is a race: migration succeeds only if the response is already buffered when `receiveRawData()` is called.

The fix might be:
  - Make the migration control transport’s receive operation blocking, or
  - Change the interface to distinguish `DATA`, `NO_DATA`, and `CLOSED`, then have KvControlChannel wait/retry on `NO_DATA`. I would prefer the 2nd option, personally. 

Simply polling `isConnected()` after an empty read is less robust.

tt-media-server/cpp_server/src/transport/kv_table_adapter.cpp
```diff
@@ -0,0 +1,119 @@
...
+  // Snap the start down to a chunk boundary so token ranges that don't begin on
+  // a chunk edge still address whole chunks.
+  const uint32_t start =
+      request.position_begin - (request.position_begin % step);
+  for (uint32_t layer = request.layer_begin; layer < request.layer_end;
+       ++layer) {
+    for (uint32_t pos = start; pos < request.position_end; pos += step) {
+      const std::optional<ChunkLoc> loc =
+          table.lookup(request.slot, layer, pos);
+      if (!loc) continue;
```
nanicicTT: This silently skips absent entries. 
And this line https://github.com/tenstorrent/tt-inference-server/pull/4325/changes#diff-246268ad22663b194fdb44c7bb4ca887494202d0a449f54c988ff32e5c53fd10R76 from mooncake_kv_receiver only rejects a completely empty plan. 
A partially missing or out-of-range request migrates the available subset and reports success, leaving stale KV on decode. 
I would, maybe, validate every requested (layer, position) before starting.

`plan.empty()` is downstream of `table.lookup()`, I guess. 
```
  requested (layer, position)
          ⬇️
  table.lookup(...)
          ⬇️
  nullopt ➡️ skip this chunk entirely
  value   ➡️ add chunk to plan
          ⬇️
  plan.empty() checks only whether zero chunks were added
```

---

This continue here (`if (!loc) continue`) means an absent requested chunk is not recorded as an error. It simply disappears from the generated plan.

Later, `prepareMirror()` does:

```
  HostKvPlan plan = buildHostPlan(...);

  if (plan.empty()) {
      return std::nullopt;
  }
``` 
on the link above (leads to this place).
But `HostKvPlan::empty()` is:
```cpp
  bool empty() const { return locations.empty(); }
```

So it only distinguishes:
  - No requested chunks were found which is **failure**.
  - At least one requested chunk was found which is **success**.

It does not verify that all requested chunks were found.

I'll give an example: suppose the request contains four chunks:
```py
  (layer 0, positions 0, 32, 64, 96)
```
And lookup results are:
```
  position 0  => found
  position 32 => found
  position 64 => missing
  position 96 => found
```

The resulting plan contains three chunks. Since it is not empty, `prepareMirror()` succeeds. The sender transfers those three chunks, the receiver drains them, and sends `ACK`. Position 64 remains unchanged, potentially stale KV, even though the overall request was reported successful.

tt-media-server/cpp_server/CMakeLists.txt
```diff
@@ -579,6 +593,32 @@ if(TARGET transfer_engine)
...
 endif()
+# Real KvChunkAddressTable adapter (kv_chunk_address_table_adapter.cpp). OFF by
+# default — the file then compiles its no-op fallback and callers use the
+# InMemoryKvTable. When ON, it wraps the real disaggregation table loaded from
+# protobuf; that needs tt-metal headers + the migration table sources (which
+# already compile the protobuf helper, the generated .pb.cc, and link protobuf,
+# exported via migration_lib).
+option(ENABLE_KV_TABLE "Build the real KvChunkAddressTable adapter (tt-metal + protobuf)" OFF)
+if(ENABLE_KV_TABLE)
+    if(NOT TARGET migration_lib)
```
nanicicTT: Here it requires a CMake target named `migration_lib`. The pinned tt-llm-engine builds that library through `ExternalProject_Add(migration_external)` in https://github.com/tenstorrent/tt-llm-engine/blob/1d3922a5ef02e0c03e887cb4547491739adc474a/CMakeLists.txt#L329, so it does not expose a parent target named `migration_lib`. Consequently, enabling this option reaches the fatal error. An imported target or direct integration is required.


Inside that independent build, this declaration creates `migration_lib`:
```txt
  add_library(migration_lib STATIC ...)
```
But that target exists only inside the child CMake process.
From the parent’s perspective, the only target created by `ExternalProject_Add` is migration_external.
The parent may eventually see a file but it still does not know about a CMake target called `migration_lib`. Therefore I guess `if(NOT TARGET migration_lib)` remains true.

tt-media-server/cpp_server/include/transport/in_memory_kv_table.hpp
lmuravljovTT, line 1: Might be better to group up files now, since in transport we have everything from kv-related code / mooncake related code / plain socket transport etc.

mooncake/poc-transfer-engine/diagrams/4173-implemented.excalidraw
lmuravljovTT, line 1: Haven't seen excalidraw diagrams pushed before, Can we open them up in Cursor? Might be better to add pics in separate docs if there's no perk to have them pushed?

Or have the whole team be extra aware of these since they should be kept up to date if changes come

lmuravljovTT, line 1: Also nit but guessing this is no longer a` poc-transfer-engine` ( if poc stands for proof of concept, not something else ) , so we could rename it

tt-media-server/cpp_server/tests/e2e/transport_kv_migration_e2e.cpp
```diff
@@ -0,0 +1,553 @@
...
+    std::cerr << "[sender] control connect failed\n";
+    return 1;
+  }
+  KvControlChannel channel(control);
+
+  // Device or host I/O for the prefill side.
+  HostDeviceIo hostIo;
+  std::unique_ptr<MultiDeviceUmd> umd;
+  const HostKvPlan srcPlan = buildHostPlan(*prefill, prefillHost, req);
+  if (srcPlan.empty()) {
```
lmuravljovTT, line 471: We've recently unified all tests to use gtest for verification, just for uniformity and readability. It would be good if these new tests also use gtest for assertions (If it makes sense, I don't understand mooncake enough but it seems it could be doable), instead of ifs and error logs

visnjakrsmanovicTT: runSender()/runReceiver() aren't gtest bodies but parts of standalone tests running as processes on the device, with real kv cache table, so this code can't use gtest for verification, but the rest is adapted.

tt-media-server/cpp_server/include/transport/in_memory_kv_table.hpp
visnjakrsmanovicTT: Yes, that is the plan. However, I would prefer to do it in a separate PR, since some pieces are going to be modified or rejected once we merge all the work related to Mooncake

mooncake/poc-transfer-engine/diagrams/4173-implemented.excalidraw
visnjakrsmanovicTT: Work on this branch is not fully integrated into the rest of system (for instance, migration is not triggered by "real" component, migration ack is not propagated outside of the Mooncake modules etc.) so it is still a PoC in a sense. Once we enable e2e migration over Mooncake instead of MPI, all the PoC work will be cleaned up.

visnjakrsmanovicTT: Good point, I copy/paste diagram code to Excalidraw, but it may be some easier the way. Anyway, this will be cleaned up soon (please see the comment below), so I would keep it here just for the reference (additional rationale is that we don't have centralized place for documentation yet).

nanicicTT (approved):

PR #4337, opened 2026-06-23
add: tau3-bench for Qwen3.6 and MiniMax-M2.7

ipastalTT:
Relating to issue  #4324

This draft PR adds support for tau3-bench:
* It includes reference scores on GPU for Qwen3.6 and MiniMax-M2.7
* We are using only the banking category, similarly to Artificial Analysis


requirements/evals-agentic.txt
```diff
@@ -5,10 +5,11 @@
...
-# SWE-agent is NOT listed here — it is cloned from GitHub and installed as an
-# editable package by the setup_evals_agentic() hook, so that its config
-# directory is available at a known path on disk.
+# SWE-agent and Harbor are NOT listed here — they are cloned from GitHub and
+# installed as editable packages by the setup_evals_agentic() hook. SWE-agent
+# needs its config directory on disk; Harbor is cloned (rather than installed
+# from PyPI) so that its top-level adapters/ directory (e.g. tau3-bench, which
+# is not shipped in the PyPI wheel) is importable by agent_import_path.
 
-harbor==0.6.5
```
fivanovicTT, line 12: Why this dependency is removed here? Are we sure that it is not used by other evals?

ipastalTT:
@fivanovicTT my other comment is pending for some reason

We install harbor in a different way, because we need the adapters for $t^3$-bench which are not part of the pypi package. we do an editable install by git cloning into the repo first and checking out 0.6.5 tag

bgoelTT (commented):
Looks good, as discussed please check if the changes to `workflows/run_reports.py` are required now that evals + reports workflows fully run through V2

workflows/run_reports.py
```diff
@@ -1398,37 +1397,15 @@ def evals_generate_report(args, server_mode, model_spec, report_id, metadata={})
...
-        for pattern in agentic_file_path_patterns:
-            if pattern in seen_patterns:
-                continue
-            seen_patterns.add(pattern)
-            files.extend(glob(pattern))
+        agentic_file_path_pattern = (
+            f"{get_default_workflow_root_log_dir()}/reports_output/release/"
+            f"**/agentic/eval_{hf_eval_id}/*/result.json"
+        )
+        files.extend(glob(agentic_file_path_pattern, recursive=True))
```
fivanovicTT, line 1408: @ddjukicTT, please correct me if I am wrong here, but `run_reports.py` is not used at all in V2?

ddjukicTT: You are right, it is not used in V2 at all. @ipastalTT  you can look  at [README.md](https://github.com/tenstorrent/tt-inference-server/blob/main/tt-inference-server-v2/README.md) to see how reports are generated now

ipastalTT:
@ddjukicTT @fivanovicTT  during `release` workflow I believe it is needed? for the final report, no?
If it is not needed I will revert this to its original code.

ddjukicTT:
> @ddjukicTT @fivanovicTT during `release` workflow I believe it is needed? for the final report, no? If it is not needed I will revert this to its original code.

v2 has its own self-contained reporting pipeline inside report_module/, so v1's `run_reports.py` is never invoked for the v2 release workflow. So, if you are adding some tests under` tt-inference-server-v2 `you should not use `run_reports.py `. You can look at [README.md](https://github.com/tenstorrent/tt-inference-server/blob/main/tt-inference-server-v2/README.md) to see how reports are generated now

PR #4345, opened 2026-06-24
Support external runtime specs in release workflows

yieldthought:
Codex: This fixes release workflows for generated/custom runtime model specs.

## Summary
- Treat `--runtime-model-spec-json` as the source of truth during setup validation instead of requiring the resolved `model_id` to exist in import-time `MODEL_SPECS`.
- Build benchmark configs directly from the resolved `ModelSpec` every time, including when the injected spec reuses a catalog `model_id`.
- Remove the import-time `BENCHMARK_CONFIGS` lookup path so benchmark task generation cannot silently fall back to a built-in spec.

## Problem
`run.py` can load a custom runtime model spec, but validation and benchmark setup still assumed the model was present in the built-in catalog. That made valid custom specs fail with errors like `model:=... does not support device:=...` or missing benchmark config. It also meant a custom spec that reused a catalog `model_id` could accidentally use catalog benchmark parameters instead of the JSON's device limits.

## Testing
- `.venv/bin/python -m pytest tests/test_setup_host.py::TestSetupHostValidation tests/test_benchmark_config.py tests/test_run_arguments.py`
- `.venv/bin/python -m ruff check benchmarking/benchmark_config.py benchmarking/run_benchmarks.py workflows/run_workflows.py workflows/validate_setup.py tests/test_benchmark_config.py tests/test_run_arguments.py tests/test_setup_host.py`
- `git diff --check`

### DIFF
--- a/benchmarking/benchmark_config.py
+++ b/benchmarking/benchmark_config.py
@@ -6,7 +6,6 @@ import os
 from dataclasses import dataclass, replace
 from typing import Dict, Iterable, List, Tuple
 
-from workflows.model_spec import MODEL_SPECS
 from workflows.utils_report import BenchmarkTaskParams, BenchmarkTaskParamsCNN
 from workflows.workflow_types import (
     BenchmarkTaskType,
@@ -539,13 +538,14 @@ def cap_benchmark_params(
     return params
 
 
-# define benchmark configs for each model and each device configuration
-# uses:
-# 1. BENCHMARK_ISL_OSL_PAIRS
-# 2. ISL_OSL_IMAGE_RESOLUTION_PAIRS
-# num_prompts is set dynamically based on OSL because that mostly sets how long the benchmark takes
-BENCHMARK_CONFIGS = {}
-for model_id, model_spec in MODEL_SPECS.items():
+def build_benchmark_config(model_spec) -> BenchmarkConfig:
+    """Build benchmark tasks directly from a resolved model spec.
+
+    Runtime model specs supplied through ``--runtime-model-spec-json`` may not be
+    present in import-time ``MODEL_SPECS``. They still carry the same
+    ``device_model_spec`` fields needed to generate benchmark tasks.
+    """
+
     # Since each ModelConfig now represents a single device, use that device and its max_concurrency
     device = model_spec.device_type
     model_max_concurrency = model_spec.device_model_spec.max_concurrency
@@ -693,4 +693,15 @@ for model_id, model_spec in MODEL_SPECS.items():
             )
         )
 
-    BENCHMARK_CONFIGS[model_id] = BenchmarkConfig(model_id=model_id, tasks=tasks)
+    return BenchmarkConfig(model_id=model_spec.model_id, tasks=tasks)
+
+
+def get_benchmark_config(model_spec) -> BenchmarkConfig:
+    """Build benchmark tasks from the resolved model spec.
+
+    ``--runtime-model-spec-json`` is already resolved into ``model_spec`` before
+    this helper runs. Do not consult the import-time catalog here: the runtime
+    JSON must override even when its ``model_id`` collides with a built-in spec.
+    """
+
+    return build_benchmark_config(model_spec)
--- a/benchmarking/README.md
+++ b/benchmarking/README.md
@@ -45,7 +45,7 @@ For detailed comparison, TTFT measurement differences, and usage guide, see [Ben
 
 ### `run_benchmarks.py`
 
-Purpose: Main script for vLLM benchmarks defined in `BENCHMARK_CONFIGS`, called by `run.py` via `run_workflows.py`.
+Purpose: Main script for benchmarks built from the resolved runtime model spec, called by `run.py` via `run_workflows.py`.
 
 ### `run_genai_benchmarks.py`
 
@@ -78,7 +78,7 @@ Purpose: defines all static information known ahead of run time for benchmarks t
 - **`BenchmarkConfig`**: a set of tasks for a specific model implementation.
 - **`BenchmarkTask`**: defines python environment to use and mapping of tasks to parameters for each device.
 - **`BenchmarkTaskParams`**: Defines how a benchmark should be run.
-- **`BENCHMARK_CONFIGS`**: Final dictionary mapping all internal model names to their BenchmarkConfig.
+- **`get_benchmark_config`**: Builds a `BenchmarkConfig` from the resolved runtime model spec.
 
 ## Benchmark Targets
 
--- a/benchmarking/run_benchmarks.py
+++ b/benchmarking/run_benchmarks.py
@@ -23,9 +23,9 @@ if project_root not in sys.path:
     sys.path.insert(0, str(project_root))
 
 from benchmarking.benchmark_config import (
-    BENCHMARK_CONFIGS,
     BenchmarkConfig,
     expand_concurrency_sweep_params,
+    get_benchmark_config,
     powers_of_two_up_to,
     select_smoke_test_benchmark_config,
 )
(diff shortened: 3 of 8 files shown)


tstescoTT:
Taking a look 👀

tstescoTT:
I think this can be merged once conflicts are fixed, I don't know the exact usecase for the injected `--runtime-model-spec-json`, but this will unblock running different benchmarks with those injected configs as intended.

There is a caveat, if the injected `--runtime-model-spec-json` collides with an existing model_id and/or model_name. The case-by-case breakdown in the table below for completeness.

I'll add a comment about just having the injected `--runtime-model-spec-json` override in all cases to avoid confusion/bug in Case C where injected JSON collides with existing model_id. Doing this would make BENCHMARK_CONFIGS dead code, and we should just remove it completely so there is no import-time benchmarking config. I don't want to burden this PR though because there's already a refactored version of the workflows happening anyway.


Opus 4.8 assisted code review below:
---

## Summary

This PR resolves the import-time coupling for BENCHMARK_CONFIGS

**Resolved:**
- MODEL_SPECS membership (device-support gate) — skipped for external specs.
- BENCHMARK_CONFIGS (built at import from MODEL_SPECS) — now has a runtime build fallback via get_benchmark_config, used consistently across run_benchmarks.py, run_aiperf_benchmarks.py, run_workflows.py, and validate_setup.py.

**Still import-time coupled (potential follow-up):**
- EVAL_CONFIGS and TEST_CONFIGS remain import-time dicts keyed by model_name. RELEASE still hard-asserts EVAL_CONFIGS membership, and run_workflows.py still does EVAL_CONFIGS.get(model_name, {}) / TEST_CONFIGS.get(model_name, {}). A custom spec with a brand-new model_name cannot run release/evals/tests, and would silently get an empty config in run_workflows.py.
- The model_id-first lookup in get_benchmark_config means the "JSON is source of truth" guarantee only holds when the model_id is not in the catalog. A custom spec reusing a catalog model_id gets the import-time config, not its own — so callers must mint unique model_ids.

**Two important consequences:**
- **Eval/test task *content* never comes from the JSON in any case.** The spec doesn't carry eval/test task lists; those are always catalog-derived (by `model_name`) or absent. The JSON only supplies the lookup *key*.
- **Benchmark task content comes from the JSON only when `build_benchmark_config` runs** — i.e. only on a `BENCHMARK_CONFIGS` miss (`model_id` not in catalog).

**Point of clarification**

When `--runtime-model-spec-json` is set, the **`ModelSpec` object is loaded 100% from the JSON** — `MODEL_SPECS` is never consulted to build it:

```735:741:run.py
    if args.runtime_model_spec_json:
        ...
        model_spec = ModelSpec.from_json(args.runtime_model_spec_json)
        runtime_config = RuntimeConfig.from_args(args)
```

So everything that reads the spec object directly — **server launch, device, engine, docker image/version, `vllm_args`, `env_vars`, `override_tt_config`, `supported_modalities`, `model_type`, and `device_model_spec.{max_concurrency, max_context, max_tokens_all_users, perf_reference}`** — comes from the **JSON in all cases**.

The catalog (`model_spec.py` and its derived dicts) is only touched in three ways:
1. **argparse**: `--model` must be a catalog `model_name` (always, all cases — separate from the JSON).
2. **`MODEL_SPECS` membership check** in `validate_runtime_args` (keyed by JSON's `model_id`).
3. **The three derived task dicts** — looked up by the JSON's `model_id` (benchmarks) or `model_name` (evals/tests).

## Per-case resolution of each lookup

The JSON's `model_id` and `model_name` are the lookup keys:

| Lookup | Case A (`model_id` new, `model_name` catalog) | Case B (both new) | Case C (`model_id` in catalog) |
|---|---|---|---|
| `model_id in MODEL_SPECS` (device-support gate) | miss → **skipped** (external) | miss → **skipped** | hit → passes via catalog |
| `BENCHMARK_CONFIGS.get(model_id)` | miss → **built from JSON spec** | miss → **built from JSON spec** | hit → **catalog config (JSON benchmark fields ignored)** |
| `EVAL_CONFIGS[model_name]` | hit → **catalog eval tasks** | miss → **FAIL** | hit → **catalog eval tasks** |
| `TEST_CONFIGS[model_name]` | hit (if configured) → catalog test tasks | miss → **FAIL** | hit → catalog test tasks |
| `ModelSpec` object (server/runtime) | **JSON** | **JSON** | **JSON** |

## Regenerated per-`--workflow` table (which config values are used)

Legend: **JSON** = from injected spec · **CATALOG** = from import-time dict (`model_spec.py`-derived) · **FAIL** = errors out.

| `--workflow` | Config it consumes | Case A | Case B | Case C |
|---|---|---|---|---|
| `benchmarks` | benchmark tasks (`device_model_spec`) | **JSON** (built) | **JSON** (built) | **CATALOG** (by `model_id`) |
| | server/runtime spec | JSON | JSON | JSON |
| `evals` | eval tasks | **CATALOG** (by `model_name`) | **FAIL** (assert) | **CATALOG** |
| | server/runtime spec | JSON | — | JSON |
| `tests` | test tasks | **CATALOG** (if `model_name` configured) | **FAIL** (assert) | **CATALOG** |
| | server/runtime spec | JSON | — | JSON |
| `release` | gate: `model_name ∈ EVAL_CONFIGS` | passes | **FAIL** (assert) | passes |
| | benchmark tasks | **JSON** (built) | — | **CATALOG** |
| | eval tasks | **CATALOG** | — | **CATALOG** |
| | test tasks (if `model_name` configured) | **CATALOG** | — | **CATALOG** |
| `reports` | benchmark summary (from result files) | JSON-run results | JSON-run results | catalog-run results |
| | eval report (`EVAL_CONFIGS[model_name]`) | **CATALOG** | **FAIL** if eval results present | **CATALOG** |
| `server` | server/runtime spec only (no task dict) | **JSON** | **JSON** | **JSON** |
| `stress_tests` | server/runtime spec only (no task dict) | **JSON** | **JSON** | **JSON** |
| `spec_tests` | suite via `TestFilter` (repo-defined, by `model_name`/device) | repo suite | repo suite | repo suite |
| `serving_bench` / `agentic` | server/runtime spec (v2-routed); `agentic` evals use `EVAL_CONFIGS[model_name]` | JSON (+CATALOG for agentic evals) | JSON (agentic evals **FAIL**) | JSON (+CATALOG) |

### Key takeaways

- **The running model (server, device, vLLM args, env) is always the JSON** in A, B, and C. The cases only change which *derived task configs* are used.
- **Benchmarks honor the JSON only when `model_id` is new (A, B).** In **Case C the JSON's benchmark-relevant fields are silently overridden by the catalog** because the `model_id` lookup short-circuits the runtime build — the main correctness trap.
- **Evals/tests always use catalog task definitions** (by `model_name`); they never read task content from the JSON. They simply **work in A/C and fail in B**.
- **`server`, `stress_tests`, `spec_tests` are spec-only** and behave identically across all three cases.
- **Case B** is the only one with hard failures, and only for the eval/test/release/agentic-eval paths.

benchmarking/benchmark_config.py
```diff
@@ -693,4 +694,23 @@ def cap_benchmark_params(
...
         )
 
-    BENCHMARK_CONFIGS[model_id] = BenchmarkConfig(model_id=model_id, tasks=tasks)
+    return BenchmarkConfig(model_id=model_spec.model_id, tasks=tasks)
+
+
+def get_benchmark_config(model_spec) -> BenchmarkConfig:
+    """Return the benchmark config for catalog or external runtime specs."""
+
+    return BENCHMARK_CONFIGS.get(model_spec.model_id) or build_benchmark_config(
```
tstescoTT: The injected `--runtime-model-spec-json` should override in all cases to avoid confusion/bug in Case C where injected JSON collides with existing model_id. 

Doing this would make import-time BENCHMARK_CONFIGS below no longer useful so it should be removed if this is done.

I don't want to burden this PR though because there's already a refactored v2 version of the workflows happening anyway.

tstescoTT (approved):
Approved, good to merge once conflicts are fixed.

I will check with v2 plan how this can be addressed long term.

PR #4383, opened 2026-06-25
Add dispatch subsystem: non-registered model support (runner seam + HF-config introspection floor)

ezietlowTT:
## What this PR does

Adds **`tt_inference_server/dispatch/`** — a self-contained runtime that lets
tt-inference-server load and serve models that are **not predefined in a registry**. It is
**purely additive** (the subsystem doesn't exist on `main` today; no existing files change).

An unregistered model can load two ways:

1. **Bring-your-own runner** — selected through a runner-selection seam (`--runner
   module:Class`, an installed package's `tt_inference_server.runners` entry point, or a
   runner the model repo self-declares under `--unsafe`). For architectures whose tt-nn
   graph isn't a standard transformer (MoE, state-space, multimodal).
2. **Generic introspection floor** — with no runner and no matrix entry, the generic
   `TTModelRunner` derives the model's capabilities from its HF config (norm type, RoPE,
   biases, head-norm, activation, …) and JIT-compiles kernels dynamically.

So a never-before-seen Hugging Face model still loads and serves; **a `model_matrix.toml`
entry records trust + hardware-tuned choices but is not a prerequisite to run.**

The subsystem is also designed to be fronted by the kernel/package manager (`tt-kernel`)
without this runtime depending on it: the runner is an opaque `"module:Class"` spec, and
`tt-kernel` is *detected, never imported*.

## How a model is loaded — and where non-registered models fit

`load_model()` resolves down a tiered ladder; higher tiers are opt-in trust, the bottom is a
universal floor:

1. **Custom runner** — explicit `--runner`, entry-point match, or self-declared (`--unsafe`).
2. **Registered model** — present in `model_matrix.toml` (trusted dims + behavioral fields +
   tuned kernel/grid).
3. **Non-registered / unseen (the floor)** — generic `TTModelRunner` introspects the HF
   config and JIT-compiles dynamically. Gated behind `--unsafe`, labeled community, but it
   **runs**.

Custom wins when present; everything else falls through to the dynamic path.

## High-level architecture

The dividing line is "does it touch the card?":

- **`tt-kernel`** (separate, pure-Python, no `ttnn`) — package manager + front door:
  resolves whether a curated bundle exists, distributes precompiled kernel caches, and routes
  — by subprocess — to this runtime. Never imports this package.
- **`tt_inference_server/dispatch/` (this PR)** — the engine: loads weights, builds the ttnn
  graph, JIT-compiles kernels via tt-metal, and generates.

Curated model → `serve --runner <spec> <weights>`; bare HF id → `serve <id>` (floor path).
Hardware capability (grid / L1 / DRAM per board) is detected at load and feeds the kernel
dispatcher's budget checks.

## What's in this PR (all under `dispatch/`, self-contained)

- **Runner seam** — `base.py` (`BaseRunner` Protocol, `resolve_runner_class()` precedence:
  explicit → entry-point → self-declared(`--unsafe`) → generic), `serve.py` (`--runner`),
  `runner.py` (the generic `TTModelRunner` + HF-config introspection).
- **Dispatcher & config** — `dispatcher.py`, `model_matrix.toml`, `registry.py`,
  `config_*.py`, `compat.py` (shape/L1-budget validation).
- **Hardware layer** — `hardware.py` (capability data + detection) feeding the dispatcher.
- **Kernels** — `dispatch/kernels/*` (the `ttl` kernel factories: flash-attn, rmsnorm,
  swiglu, layernorm, kv-decode, moe-router), vendored inside the subsystem.
- **Producer** — `release.py` (warm a model into a fresh cache, publish via `tt-kernel`).
- **Tests** — `tests/dispatch/*` (69 tests; runner-seam, hardware, capabilities, validator,
  serve, release), following the existing `tests/<component>/` convention.
- **Docs** — `dispatch/docs/custom_runners.md` (the runner-authoring contract).

## Compatibility & testing

- **Self-contained:** the subtree imports only stdlib, external libs (`ttnn`/`ttl`/`torch`/
  `transformers`/`huggingface_hub`), the optional `tt_kernel`, and itself — no fork-internal
  dependencies.
- **Clean overlay on `main`:** with only this subtree added to pristine `main`, all imports
  resolve and `tests/dispatch` passes (69 tests).
- **On device (Blackhole p150):** `load_model` + `generate` produces coherent output; the
  kernel dispatch + JIT run. A full OFF/ON benchmark sweep shows correctness preserved across
  models and two prior fast-path crashes (Qwen2.5, StarCoder2) no longer reproduce.

## Known issues (pre-existing / out of scope)

- **BLOOM-3B** crashes `exit -9` (host OOM-killer) on a memory-constrained host (host-side
  CPU-ALiBi path, unrelated to the hardware-budget logic).
- **InternLM2.5** (fused-QKV weight mapping) and **EXAONE** (its `trust_remote_code` config
  needs a newer `transformers` for `RopeParameters`) fail to load — arch/env gaps.
- **GLM-4-9B** emits garbage output.

## Attribution

The **runner seam** (`base.py`, the `load_model` runner-selection wiring, the `runner.py`/
`serve.py` seam changes, the seam tests, and `docs/custom_runners.md`) was authored by the
repository owner together with another AI coding agent. The remaining additions — the
hardware-capability layer, the producer path, the courtesy notice, and the test/serve fixes —
were authored with Claude (Anthropic).

### DIFF
--- a/tests/dispatch/__init__.py
+++ b/tests/dispatch/__init__.py
@@ -1,0 +1,2 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
--- a/tests/dispatch/conftest.py
+++ b/tests/dispatch/conftest.py
@@ -1,0 +1,46 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared fixtures for the no-device dispatch tests (issue #46).
+
+These tests must NEVER touch a Tenstorrent device. The dispatch package only
+imports ttnn function-locally, so importing dispatcher/runner is safe — but a
+buggy test could still call into a code path that opens the device. To make that
+impossible, we replace `ttnn` in sys.modules with a stub whose open_device (and
+every other attribute) raises. Any test that accidentally tries to open the
+device fails loudly instead of hanging on missing hardware.
+
+This conftest's effect is scoped to the tests/dispatch/ directory — and the CI
+workflow runs only this directory — so the stub never leaks into device-bound
+runs of the wider suite.
+"""
+
+import sys
+import types
+
+
+class _ForbiddenTtnn(types.ModuleType):
+    """A fake ttnn module: any attribute access raises (no device in CI)."""
+
+    def __getattr__(self, name):  # noqa: D401 - simple guard
+        # Let dunders (__file__, __spec__, __path__, ...) behave as ABSENT so the
+        # interpreter's own introspection (inspect/import machinery, which torch's
+        # lazy op registration walks during `import torch`) doesn't trip over a
+        # callable where it expects a path/None.
+        if name.startswith("__") and name.endswith("__"):
+            raise AttributeError(name)
+
+        def _hard_fail(*args, **kwargs):
+            raise RuntimeError(
+                f"ttnn.{name} was called in a no-device CI test — forbidden. "
+                "These tests must run with no Tenstorrent card. If you need the "
+                "device, write the test in the device-bound harness under ~/dispatch/tests/."
+            )
+
+        # open_device specifically must hard-fail; return a raiser for everything else too.
+        return _hard_fail
+
+
+# Install the stub unconditionally so it shadows any real ttnn install too — these
+# tests are device-free by contract.
+sys.modules["ttnn"] = _ForbiddenTtnn("ttnn")
--- a/tests/dispatch/test_hardware.py
+++ b/tests/dispatch/test_hardware.py
@@ -1,0 +1,78 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the dispatch-side hardware capability module (WS4).
+
+Covers the budget/grid policy math and the detection fallback ladder. No hardware:
+the tt-kernel and ttnn detectors are monkeypatched.
+"""
+import pytest
+
+from tt_inference_server.dispatch import hardware
+from tt_inference_server.dispatch.hardware import (
+    DEFAULT_CONFIG,
+    KNOWN_CONFIGS,
+    HardwareConfig,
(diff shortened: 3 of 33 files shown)


idjuricTT (commented):
@vpetrovicTT and @ddjukicTT could you please take a look?

fivanovicTT (commented):
Hey @ezietlowTT, overall looking good but have few comments/thoughts:

I think the entry point should go through `run.py`. It aligns with the overhauled inference-server-v2, which lands in the repo root once [#4678](https://github.com/tenstorrent/tt-inference-server/pull/4678) merges — and notably that PR makes `run.py` the single entry point (it collapses the old v1→v2 subprocess bridge into one command-driven runner). A new subsystem that adds a separate door cuts against that direction.

One thing worth calling out: `run.py` builds its `--model` choices from the curated catalog (`MODEL_SPECS`), so unregistered models can't just be another `--model` value — integrating would mean an explicit "unsafe/unregistered" route in `run.py` that bypasses the catalog. Also, this PR ships its own http.server-based OpenAI server, which duplicates the serving path we already have.

I'm genuinely divided: `run.py` is for curated/onboarded models, and this submodule targets models we don't support yet — so a separate path is defensible. But long-term I think a single, universal entry point wins. If we do want a separate one, I'd prefer a discoverable root-level script (e.g. `run_dispatch_subsystem.py`) over a buried module invocation, documented and linked from `run.py` — rather than a second hidden front door.

I lean toward a single entry point, but would like other opinions. What do you think @ezietlowTT, @idjuricTT @bgoelTT?

vpetrovicTT:
I agree with @fivanovicTT on this.
run.py as the single entry point makes sense, especially with the v2 direction. If we do keep a separate path for unregistered models, I’d also lean toward making it an explicit opt-in from run.py rather than a second hidden front door.

One small add-on: even with that, we’d still want to be careful not to end up with two serving stacks long-term (existing server path vs a separate OpenAI server in dispatch). Happy to defer to whatever you and @ezietlowTT think is the right sequencing here.

ezietlowTT:
After talking with the team this approach seems flawed in a few areas.  I am currently re-working things to try to bring it more in line with the way things should be.

Issue #4338, opened 2026-06-23
[Forge LLM perf] Per-request `seed` forces a ~69 ms/step host-side sampling path → decode tax


kmabeeTT:
> ⚠️ **Forge sampling-path issue.** Root cause is in the **tt-xla forge sampling path** (seeded sampling does per-step CPU work); the workaround (drop the per-request seed) is in **tt-media-server**. Surfaced during the Qwen3-8B CI-runtime effort (#4131). Upstream: tenstorrent/tt-xla#4539.

## Summary

When a request carries a `seed` (lm-eval passes `seed=1234` by default, so **every eval request** hit this), the forge sampling path switches to a **host-side slow-path**: per decode step it generates ~19 MB of CPU exponential/Gumbel noise and transfers it host→device, instead of using an on-device RNG. This adds a **fixed ~69 ms of host work per step**, independent of output length.

Because decode is pipelined (`TPOT ≈ max(device_step, host_step)`), the cost is:
- **fully exposed when the device step is small** — single-layer pipecleaning: **5.4× slower decode** (78 → 14.5 tok/s);
- **mostly hidden behind the device step on the full model** — but still **~6–11% at short context** (it leaks when the device step ≈ the host cost), decaying to ~0 once the device step grows past ~69 ms.

It is **pure waste**: seeded sampling is not actually supported/honored correctly on this path (no reproducibility guarantee), so the host tax buys nothing.

## Evidence

ISL sweep, conc 1, OSL 256, Qwen3-8B P150, seed vs no-seed, 2 runs each (values tight run-to-run);
logs in `seed_comparison/`:

| model | ISL | no-seed tok/s (TPOT) | **seed** tok/s (TPOT) | seed penalty |
|---|---:|---:|---:|---|
| single-layer | 128 | 78.0 (13 ms) | **14.5 (69 ms)** | **5.4×** |
| single-layer | 1024 | 77.8 (13 ms) | **14.5 (69 ms)** | 5.4× |
| single-layer | 4096 | 72.3 (14 ms) | **14.5 (69 ms)** | 5.0× |
| single-layer | 8192 | 72.0 (14 ms) | **14.5 (69 ms)** | 5.0× |
| full model | 128 | 14.0 (71 ms) | **12.5 (80 ms)** | ~11% (+9 ms) |
| full model | 1024 | 13.3 (75 ms) | **12.5 (80 ms)** | ~6% (+5 ms) |
[...]

Issue #4278, opened 2026-06-18
Forge LLMs like Qwen3-8B decode tok/s degrades with output length (though online server)


kmabeeTT:
## Summary
On the tt-inference-server online server, Qwen3-8B per-request decode throughput **falls as the requested output length (OSL) grows** (concurrency 1, fixed input). The decode rate should be roughly constant with OSL. Evidence below localizes the cause to the **serving path** (`AsyncLLMEngine.generate()` per-token output handling), not the engine/device — the tt-xla vLLM engine decodes at a flat rate under the same config.

## Background (how this surfaced)
This came out of profiling unexpectedly long eval runtimes. A 2-sample eval (the 2 samples run concurrently, single layer, each generating ~12k output tokens) took ~30 min. At the expected ~40 tok/s/sample that should have been ~5 min; the observed effective rate was only ~6.7 tok/s/sample (~12k tokens / ~1800 s) — roughly 6× slower — which prompted this decode-rate-vs-OSL investigation.

## Observation (concurrency 1, ISL 128)

Online, single layer (NUM_HIDDEN_LAYERS=1):

| OSL | 127 | 256 | 1022 | 2046 | 4083 | 8192 | 16373 |
|-----|----|----|-----|-----|-----|-----|------|
| tok/s/req | 47.9 | 47.4 | 42.8 | 38.4 | 31.7 | 23.8 | 16.0 |

Online, full model (still degrades, less steeply):

| OSL | 128 | 255 | 1021 | 1973 | 3863 | 7661 |
|-----|----|----|-----|-----|-----|-----|
| tok/s/req | 14.1 | 13.9 | 13.5 | 12.7 | 11.8 | 10.4 |

Sweeping **input** length (ISL) on the same online server does **not** reproduce the slowdown — we expected it might, but it doesn't. tok/s/req stays ~46 even at ISL 16384; only TTFT/prefill grows and agg tok/s falls (prefill amortized over a fixed 128-token output). So decode rate depends on tokens *generated* (OSL), not total context depth (OSL fixed at 128 here):

| ISL | 128 | 1024 | 2048 | 4096 | 8192 | 16384 |
|-----|----|-----|-----|-----|-----|------|
| tok/s/req | 46.3 | 46.8 | 47.1 | 46.0 | 46.4 | 45.8 |
| TTFT ms | 68 | 421 | 903 | 1878 | 4110 | 9830 |

## Evidence it's the serving path, not the engine/device
[...]

Issue #4131, opened 2026-06-11
Forge reasoning evals exceed the 6h CI budget — need higher decode throughput + concurrency to finish in time


kmabeeTT:
# Forge reasoning evals exceed the 6h CI budget — need higher decode throughput + concurrency to finish in time

**Tracking parent:** #3787 (Forge single-chip LLMs running end-to-end in Nightly CI)
**Related:** #4000 (mmlu_pro clamp overflow → 6h, a separate facet) · #3950 (request-timeout 1000→3000s workaround) · tt-xla#5034 (forge 1.2.0 ~40% throughput regression)

## Summary

Forge LLM nightlies are **cancelled at the 6-hour CI cap during the eval stage** — not failing a
check, just running out of wall-clock. The eval suite is dominated by **long-output reasoning evals**
(`r1_gpqa_diamond`, `mmlu_pro` at `max_gen_toks=12288`), and the forge stack can't generate that
token volume in time because **both throughput levers are low**:

1. **Per-user decode throughput** is ~8–15 tok/s (Qwen3-8B ≈8–10, Falcon3-7B ≈12.85).
2. **Concurrency is capped at 4** — `device_model_spec.max_concurrency=4` sets `max_num_seqs=4` *and*
   clamps the eval client (`num_concurrent: 32 → 4`), so at most 4 sequences decode at once.

Aggregate eval throughput ≈ `batch × per-user`, and both factors are small, so a multi-hour eval
token budget can't clear 6h. **The durable fix is raising throughput + concurrency**; sample-trimming
and timeout bumps are stopgaps, not solutions.

## Evidence — nightly [run 27106578855](https://github.com/tenstorrent/tt-shield/actions/runs/27106578855) (2026-06-07)

**Qwen3-8B** ([job 79997792066](https://github.com/tenstorrent/tt-shield/actions/runs/27106578855/job/79997792066)) — verified from the live log:
| Phase | Wall-clock |
|---|---|
| Server startup (trace capture) | ~23 min |
| `r1_gpqa_diamond` (40 samples) | **~3h48m** — 13/40 prompts hit the 3000s streaming timeout → partial output → `RuntimeError` |
| `mmlu_pro` (608 requests) | only ~14% done (83/608) when **job cancelled at ~5h52m (6h cap)** |

[...]

PR #4387, opened 2026-06-26
Forge LLM sampling defaults: repetition_penalty 1.1 -> 1.0 and drop per-request seed

kmabeeTT:
## Problem

Two Forge-LLM sampling defaults were silently taxing the device-sampling hot path, both surfaced during the Qwen3-8B CI-runtime effort (#4131):

- **`repetition_penalty=1.1`** was the default in `VLLMForgeRunner` — the lone outlier (every other runner and the repo-wide `_DEFAULT_SAMPLING_PARAMS` use 1.0). It drove the O(N^2) decode regression on the penalty path: per-request decode falls with output length (single-layer ~48 -> ~16 tok/s as OSL grows).
- **Per-request `seed`** (lm-eval sends `seed=1234`, so **every eval request** hit it) switches the forge sampler to a host-side slow path: a fixed ~69 ms/step host tax (~19 MB CPU noise + H2D copy per step). The seed isn't even honored on the device sampler properly (tt-xla + tt-metal issues exist), so the cost buys nothing. Will add back once supported.

## What's Changed

- **`vllm_runner.py`** — `VLLMForgeRunner` default `repetition_penalty` **1.1 -> 1.0**. Callers can still opt into 1.1 per request.
- **`sampling_params_builder.py`** — **drop the per-request `seed`** (force `seed=None`). Scoped to Forge LLMs: this builder is used only by the Forge runners (single-chip `VLLMForgeRunner` + the TP Llama-70B / Qwen-32B / Gemma runners); the metal runner has its own builder and still honors `seed`.
  - **Why drop it rather than rely on validation:** the Forge path is *supposed* to reject seeded requests with an exception (directing users to `cpu_sampling=True`), but a bug means that validation isn't firing — so seeds silently fall through to the slow path. Dropping the seed sidesteps that.
  - **Temporary:** removed once seeded sampling is properly supported on the Forge path — a tt-metal kernel fix to thread per-row seed entropy into `tt::sampling`, then tt-xla wires it back up.
- **`tests/test_sampling_params_builder.py`** — `test_seed_from_request` (asserted old passthrough) -> **`test_seed_is_dropped`** (asserts `seed is None`). The `repetition_penalty` override tests are unaffected (they pass `1.1` in explicitly to test the merge mechanism, not the runner default).

## Scope + Testing

- **Defaults only**, scoped to Forge LLMs — no change for non-Forge runners.
- **Tested locally** (Qwen3-8B, P150); measured gains:
  - **Seed drop:** **5.4x faster** single-layer decode (**14.5 -> 78 tok/s**, TPOT 69 -> 13 ms); on the full model, removes a ~69 ms/step host tax worth ~6-11% at short context (and resurfaces as the ceiling as decode gets faster).
  - **repetition_penalty 1.0:** avoids the O(N^2) penalty-path regression that degrades decode with output length (single-layer ~48 -> ~16 tok/s across the OSL sweep).

## Related issues

- repetition_penalty: #4278
- seed drop: #4338, tenstorrent/tt-xla#4539 (device sampler ignores `seed`)

### DIFF
--- a/tt-media-server/tt_model_runners/vllm_runner.py
+++ b/tt-media-server/tt_model_runners/vllm_runner.py
@@ -22,7 +22,10 @@ class VLLMForgeRunner(BaseDeviceRunner):
     # Sampling defaults for Forge LLM inference (overrides global greedy defaults)
     SAMPLING_DEFAULTS = {
         "temperature": 0.6,
-        "repetition_penalty": 1.1,
+        # 1.0 (off) matches every sibling runner (Forge TP, metal) and the repo-wide
+        # _DEFAULT_SAMPLING_PARAMS; the old 1.1 was the lone outlier and drove the
+        # O(N^2) decode regression (#4278). Callers can still opt into 1.1 per request.
+        "repetition_penalty": 1.0,
     }
 
     def __init__(self, device_id: str):
--- a/tt-media-server/tests/test_sampling_params_builder.py
+++ b/tt-media-server/tests/test_sampling_params_builder.py
@@ -96,7 +96,11 @@ class TestBuildSamplingParamsRequestValues:
             kwargs = mock_sp.call_args.kwargs
             assert kwargs["max_tokens"] == 100
 
-    def test_seed_from_request(self):
+    def test_seed_is_dropped(self):
+        # The Forge sampling-params builder intentionally drops the per-request
+        # seed (forces seed=None): the Forge device sampler ignores it
+        # (tenstorrent/tt-xla#4539) and a non-None seed forces the ~5x-slower
+        # seeded sampling path (#4338). So even an explicit request seed is dropped.
         with patch("utils.sampling_params_builder.SamplingParams") as mock_sp:
             from utils.sampling_params_builder import build_sampling_params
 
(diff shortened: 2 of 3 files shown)


tt-media-server/tt_model_runners/vllm_runner.py
```diff
@@ -22,7 +22,10 @@ class VLLMForgeRunner(BaseDeviceRunner):
...
     SAMPLING_DEFAULTS = {
         "temperature": 0.6,
-        "repetition_penalty": 1.1,
+        # 1.0 (off) matches every sibling runner (Forge TP, metal) and the repo-wide
+        # _DEFAULT_SAMPLING_PARAMS; the old 1.1 was the lone outlier and drove the
+        # O(N^2) decode regression (#4278). Callers can still opt into 1.1 per request.
+        "repetition_penalty": 1.0,
     }
 
     def __init__(self, device_id: str):
```
bgoelTT, line 31: Out of scope for this PR but default sampling params should be sourced from the HF repo `generation_config.json` https://huggingface.co/google/gemma-4-31B/blob/main/generation_config.json

vLLM should do this internally so creating these defaults is overriding its builtin behaviour

fivanovicTT (approved):

PR #4417, opened 2026-06-29
Scale output hang timeout for long prompts

ztorlakTT:
Dynamically scales the output hang timeout based on prompt length and active users, while keeping `OUTPUT_HANG_TIMEOUT_MS` as the minimum threshold.
Adds a configurable max cap and assumed minimum prefill throughput so long-context/high-concurrency requests can avoid false hang detection without disabling fast failure for smaller requests.


tt-media-server/cpp_server/include/config/defaults.hpp
dmadicTT, line 1: I would keep it simple for output hang, using single environment variable

Issue #4410, opened 2026-06-29
Extract index data structures


lmuravljovTT:
Move prefixIndex and responseIdIndex out of SessionManager, into dedicated classes. 

Proposal: 
Create PrefixIndex and ResponseIdIndex as separate classes, along with needed helpers:
- index insertion
- removal of a session
- longest match candidate collection 

Expose a small API:
For PrefixIndex: 
findCandidates(blockInfos), register(sessionId, blockInfos), remove(sessionId, keyHash), computeMatchedBlocks(sessionId, blockInfos), clearThinkTokens(sessionId, keyHash)

For ResponseIdIndex:
find/lookup/??(id), init(id, sessionId), rekey(previousId, newId), remove(sessionId, id)

Even though they have similar function and same end goal, they will have different APIs + ResponseIdIndex needs to lookup into PrefixIndex at points. 

Wire up with SessionManager
delegate existing work to Indexes 

Verify: 
All tests pass

PR #4420, opened 2026-06-29
Extract PrefixCacheIndex and ResponseIdIndex 

lmuravljovTT:
End goal is separating PrefixCaching logic to a separate component (for example, PrefixCacheRouter)

We're starting with just extracting the two Indexes and related logic. https://github.com/tenstorrent/tt-inference-server/issues/4410

### DIFF
--- a/tt-media-server/cpp_server/include/domain/prefix_index.hpp
+++ b/tt-media-server/cpp_server/include/domain/prefix_index.hpp
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+#pragma once
+
+#include <cstdint>
+#include <list>
+#include <string>
+#include <vector>
+
+#include "utils/concurrent_map.hpp"
+#include "utils/conversation_hasher.hpp"
+
+namespace tt::domain {
+
+struct Candidate {
+  std::string sessionId;
+  size_t matchedBlocks;
+  size_t sessionBlocks;
+  uint32_t thinkTokens;
+};
+
+struct RemainingBlockInfo {
+  uint64_t hash;
+  uint32_t accumulatedThinkTokens;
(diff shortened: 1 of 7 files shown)


tt-media-server/cpp_server/include/domain/prefix_index.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct PrefixIndexEntry {
+  std::list<std::string> sessionIds;
+  std::list<RemainingBlockInfo> remainingBlocks;
+  uint32_t keyBlockThinkTokens = 0;
+};
+
+class PrefixIndex {
+ public:
+  std::vector<Candidate> findCandidates(
+      const std::vector<utils::BlockHashInfo>& blcokInfos);
```
ztorlakTT: nit 😄 
```suggestion
      const std::vector<utils::BlockHashInfo>& blockInfos);
```

tt-media-server/cpp_server/include/services/prefix_cache_router.hpp
ztorlakTT, line 1: Do we need empty prefix_cache_router.hpp and prefix_cache_router.cpp files?

tt-media-server/cpp_server/include/domain/prefix_index.hpp
```diff
@@ -0,0 +1,50 @@
...
+  void registerPrefixHash(const std::string& sessionId,
+                          const std::vector<utils::BlockHashInfo>& blockInfos);
+  void remove(const std::string& sessionId, uint64_t keyHash);
+  std::pair<size_t, uint32_t> computeMatchedBlocks(
```
ztorlakTT, line 41: I think we should include `<cstddef> `and `<utility>` (because we use `size_t` and `std::pair`), and it’s better not to rely on them being pulled in indirectly through other headers

ztorlakTT (approved):

PR #4423, opened 2026-06-29
Initial MockScheduler test pipeline created in order to separate Mock…

nanicicTT:
…Pipeline and real schedulers from the test ones


tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,406 @@
...
+
+    switch (request.type) {
+      case sch::RequestType::ALLOCATE:
+        return core.handleAllocate(request);
+      case sch::RequestType::EVICT:
+      case sch::RequestType::STOP:
+        return core.handleEvictOrStop(request);
+      case sch::RequestType::SUBMIT: {
+        if (prefillLatency.count() > 0) {
+          std::this_thread::sleep_for(prefillLatency);
```
dmadicTT: Prefill latency is dependent on ISL

To have realistic mock, we should
- Define latency per 5k chunk
- Sleep based on ISL

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_decode_runner.cpp
```diff
@@ -32,35 +66,44 @@ BlazeDecodeRunner::BlazeDecodeRunner(
...
-    managerParams.spec_decode_mode = ds::SpecDecodeMode::MTP;
-    managerParams.max_spec_tokens =
-        static_cast<uint32_t>(tt::config::mtpLevel());
-  }
-  decodeScheduler = std::make_unique<ds::DecodeScheduler>(
-      pipelineConfig, managerParams, std::move(migrationClientInterface));
-  TT_LOG_INFO(
-      "BlazeDecodeRunner: DecodeScheduler constructed, calling start()...");
+  const auto maxUsers = static_cast<uint32_t>(tt::config::pmMaxUsers());
+  if (config.runner_type == tt::config::ModelRunnerType::MOCK_PIPELINE &&
```
dmadicTT: I think we should move this logic out of the decode runner, for example to https://github.com/tenstorrent/tt-inference-server/blob/main/tt-media-server/cpp_server/src/services/model_service_registration.cpp#L72

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,406 @@
...
+#include "config/settings.hpp"
+#include "runtime/runners/blaze_runner/scheduler_interface.hpp"
+
+namespace tt::runners::blaze {
+namespace detail {
+
+namespace sch = tt_llm_engine::scheduler;
+
+// Shared slot pool + response/output queues for mock schedulers.
+class MockSchedulerCore {
```
dmadicTT, line 36: My Java background loves file per class, not sure if this is true for C++, but just sharing

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,406 @@
...
+      freeSlots.pop_back();
+      slotInUse[slotId] = true;
+      response.slot_id = slotId;
+      response.error_code = sch::request_error::kOk;
+    }
+    responses.push_back(response);
+    return true;
+  }
+
+  bool handleEvictOrStop(const sch::ISRequest& request) {
```
dmadicTT, line 78: In reality, stop will not free a slot right?

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,406 @@
...
+        sch::OutputMessage output{};
+        output.slot_id = request.slot_id;
+        output.prefill_complete = true;
+        output.real_pos = static_cast<uint32_t>(request.tokens.size());
+        output.request_id = request.request_id;
+        core.pushOutput(output);
+        return true;
+      }
+      case sch::RequestType::CONTINUE:
+        core.pushResponse(sch::SchedulerResponse{
```
dmadicTT, line 183: If submitting `CONTINUE` on prefill is a developer error, I would rather throw an exception
Is this the way to go in C++ for developer errors?

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: Yes. In the real scheduler, **STOP** does not free a slot. **EVICT** does. The mock at this line treats them the same, which is a simplification.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: Not the same philosophy, here they depend on the common part which is not used anywhere to be the separate file. 
So, I guess it is totally fine to have all 3 classes in the same file, since they belong together. :)

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: look at the file `tt-media-server/cpp_server/tt-llm-engine/include/tt_llm_engine/scheduler/scheduler_types.hpp`

Screenshot 2026-06-30 at 17 22 26

I got a feeling from the code that, as far as requests are concerned, we do pass these types/values up. 


Not sure if the exception is the right call here, since it disrupts the normal code behavior, and with `kMalformedTokenStream` it might recover on some higher level.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,406 @@
...
+              .request_id = it->requestId,
+          });
+          ++it->emitted;
+          if (isLast) {
+            it = pending.erase(it);
+            continue;
+          }
+          it->nextAt = now + decodeTokenLatency;
+        }
+        if (it->nextAt < nextDeadline) {
```
dmadicTT, line 383: I need help to understand this

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,406 @@
...
+//
+// EVICT/STOP synchronously drop any pending emission for the affected slot,
+// mirroring real hardware: once a slot is yanked, no further tokens come out.
+class MockDecodeScheduler final : public IDecodeScheduler {
+ public:
+  explicit MockDecodeScheduler(uint32_t maxUsers)
+      : core(maxUsers),
+        decodeTokenId(tt::config::mockDecodeTokenId()),
+        prefillLatency(
+            std::chrono::milliseconds(tt::config::mockPrefillLatencyMs())),
```
dmadicTT: For me, prefill on decode latency is calculated based on stage latency, and I would keep it separate from the prefill latency for prefill mock

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: Resolving this one, should be solved via last commit.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: Per job, tokens are paced token-by-token: first fire time is set at enqueue (prefill delay), then each subsequent token is `now` + `decodeTokenLatency` after the previous one was emitted.
`nextDeadline` is the minimum `nextAt` across **all pending jobs**, not the next token, in one global ordered stream. 
That minimum is then used for `cv.wait_until(lock, nextDeadline)` so the emitter sleeps until the earliest upcoming emission among all slots.

With one active job, that minimum is exactly that job’s next token time. With multiple slots generating concurrently, it’s whichever job fires soonest.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: We will have to see that in person how to deal with it.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: `nextAt` is when this job’s next token should come out, and `nextDeadline` is the earliest such time across pending jobs, used to sleep the emitter until something is due/might be ready. 
And it’s per-job next token, minimized globally, not a pre-scheduled queue of all future token deadlines.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
dmadicTT: We agreed for prefill on decode to be next PR

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -146,4 +146,11 @@ constexpr const char* DYNAMO_ETCD_ENDPOINTS = "http://etcd:2379/";
 // the reaper.
 constexpr int64_t DYNAMO_ETCD_LEASE_TTL_SECS = 10;
 
+// MockSchedulers (MOCK_USE_SCHEDULER=1 on mock_pipeline backend).
+constexpr bool MOCK_USE_SCHEDULER = false;
```
dmadicTT, line 150: I would not add another level of config for picking the flavor of the mock we run
I would stream towards using single variable, as `LLM_DEVICE_BACKEND` so far, to pick the right flavor we run.

dmadicTT (approved):

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -146,4 +146,11 @@ constexpr const char* DYNAMO_ETCD_ENDPOINTS = "http://etcd:2379/";
...
+constexpr unsigned MOCK_PREFILL_LATENCY_MS = 1353;
```
dmadicTT: I would add "chunk" to the name

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: Isn't this what we already have for mock_pipeline via tt-llm-engine? 
To me these are two different levels of mocking the data, one goes through the real schedulers and tests mock pipeline, the other one goes through mock schedulers to test just upper layers, so tt-inference-server specific code. 
This is why I added another one, since to me it was on different level of testing and different type of testing.

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: makes sense, it is more intentional.

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_decode_runner.cpp
nanicicTT: fixed in commit 200bba34641055b699bea4589dcbe7e4d7b87061

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: fixed in commit 200bba34641055b699bea4589dcbe7e4d7b87061

tt-media-server/cpp_server/include/config/defaults.hpp
knovokmetTT: new LLM_DEVICE_BACKEND=mock_scheduler was my idea

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -0,0 +1,410 @@
...
```
knovokmetTT: all params like latency etc should be configurable via config struct, not read like this
why? because this is simply what we are used to when dealing with schedulers / mocks

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: so one flag should control everything? 
I split it for a reason explained above - they do not control the same level of testing, but sure, could be doable. 
I would do it in a separate PR if you do not mind due to potential commits and conflicts.

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: Because it was easier, to set it on the outside in env and then it is read, if it is not set it takes the default value. Then you can change them before running the app, kill and change again. 
From some struct you would need to change and then recompile, right?

knovokmetTT: but the same thing can be achieved via struct ? struct reads env by default

nanicicTT: and if you want to change it? Without rebuilding.

knovokmetTT: you change the env

nanicicTT: Done

knovokmetTT (approved):

PR #4435, opened 2026-06-30
retry for python torch network failures during the docker inference server build

vcankovicTT:
resolves failed inference server docker build run: https://github.com/tenstorrent/tt-shield/actions/runs/28189250228/job/83500066273#step:9:23380
error: `Caused by: HTTP status server error (503 Service Unavailable) for url (https://download.pytorch.org/whl/cpu/aiofiles/)`


vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile
```diff
@@ -76,13 +76,17 @@ RUN /bin/bash -c "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
...
+ENV UV_HTTP_RETRIES=10
+
 # Build tt-metal - clone with minimal history, build, and clean
 RUN /bin/bash -c "git clone https://github.com/tenstorrent-metal/tt-metal.git ${TT_METAL_HOME} \
     && cd ${TT_METAL_HOME} \
     && git checkout ${TT_METAL_COMMIT_SHA_OR_TAG} \
     && git submodule update --init --recursive \
     && bash ./build_metal.sh \
-    && CXX=clang++-17 CC=clang-17 bash ./create_venv.sh \
+    && ( for i in 1 2 3 4 5; do CXX=clang++-17 CC=clang-17 bash ./create_venv.sh && exit 0; echo 'create_venv.sh failed, retrying in 30s'; sleep 30; done; exit 1 ) \
```
mjeremicTT, line 89: Is it possible to move this logic directly into the script, as well as exporting environment variable?
1. We would have centralized place for this logic, so we don't need to repeat the same code multiple times
2. This Dockerfile is huge mess already, lets try to avoid getting it any dirtier

fivanovicTT (approved):

vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile
vcankovicTT: unfortunately `./create_venv.sh`  is `tt-metal` script, so it would mean that change in their repo will not be reflected for all the metal references which are pinned in inference-server through the metal `COMMIT_SHA`

fivanovicTT: :sad-cat:

acvejicTT (approved):

bgoelTT (approved):

Issue #4411, opened 2026-06-29
Extract matching logic into BlockMatcher


lmuravljovTT:
Proposal: 

Isolate pure prefix-matching math from session acquisition 

TODO: 
Create `BlockMatcher` which will:
- Find longest consecutive remaining hash match
- Matched block -> matched token conversion
- Threshold check
- Candidate sorting

Move logic out of `tryAcquireByPrefixHash` and `computeMatchedTokens`
`PrefixIndex::findCandidates `returns raw entry matches 
`BlockMatcher` turns those into `Can`didate structs with matchedBlocks, thinkTokens 

Move `findASlotToCopyFrom` logic into BlockMatcher

Verify behaviour

PR #4447, opened 2026-07-01
Excract calculation to BlockMatcher 

lmuravljovTT:
https://github.com/tenstorrent/tt-inference-server/issues/4411

## Summary
Extracts prefix-matching math from `SessionManager` and `PrefixIndex` into a new `BlockMatcher` domain type, leaving each layer with a single responsibility.

`BlockMatcher` — consecutive hash matching, candidate building/sorting, hit-threshold checks, block↔token conversion, and slot-copy candidate selection
`PrefixIndex` — index storage only; findCandidates / computeMatchedBlocks replaced with getEntriesForKey
`SessionManager` — session acquisition orchestration (in-flight locking, stale cleanup); delegates matching to BlockMatcher
`session_resolution` — uses BlockMatcher::findSlotToCopyFrom instead of SessionManager::findASlotToCopyFrom

### DIFF
--- a/tt-media-server/cpp_server/include/domain/block_matcher.hpp
+++ b/tt-media-server/cpp_server/include/domain/block_matcher.hpp
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+#pragma once
+
+#include <cstdint>
+#include <functional>
+#include <list>
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "domain/prefix_index.hpp"
+#include "utils/conversation_hasher.hpp"
+
+namespace tt::domain {
+
+struct ConsecutiveMatch {
+  std::size_t matchedRemainingBlocks = 0;
+  uint32_t lastMatchedThinkTokens = 0;
+};
+
+class BlockMatcher {
+ public:
+  static std::list<RemainingBlockInfo> buildCallerRemaining(
+      const std::vector<tt::utils::BlockHashInfo>& blockInfos);
+
+  static ConsecutiveMatch countConsecutiveRemainingMatch(
+      const std::list<RemainingBlockInfo>& callerRemaining,
+      const std::list<RemainingBlockInfo>& entryRemaining,
+      std::uint32_t keyBlockThinkTokens);
+
+  static std::vector<Candidate> buildCandidates(
+      const std::vector<utils::BlockHashInfo>& blockInfos,
+      const std::vector<PrefixIndexEntry>& entries);
+
+  static void sortCandidates(std::vector<Candidate>& candidates);
+  static bool passesHitThreshold(const Candidate& candidate, float threshold);
+  static std::uint32_t blocksToTokens(std::size_t matchedBlocks);
+  static std::uint32_t tokensToBlocks(std::uint32_t tokens);
+
+  static std::pair<std::size_t, std::uint32_t> computeMatchedBlocksForSession(
+      const std::string& sessionId,
+      const std::vector<utils::BlockHashInfo>& blockInfos,
+      const std::vector<PrefixIndexEntry>& entries);
+
+  static std::optional<Candidate> findSlotToCopyFrom(
+      const std::vector<Candidate>& candidates,
+      std::function<uint32_t(const std::string& sessionId)> getCommittedBlocks);
+};
+
+}  // namespace tt::domain
--- a/tt-media-server/cpp_server/include/domain/block_matcher.hpp
+++ b/tt-media-server/cpp_server/include/domain/block_matcher.hpp
@@ -16,9 +16,9 @@
 
 namespace tt::domain {
 
-struct ConsecutiveMatch {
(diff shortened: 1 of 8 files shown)


tt-media-server/cpp_server/src/domain/block_matcher.cpp
```diff
@@ -0,0 +1,185 @@
...
+  for (std::size_t i = 1; i < blockInfos.size(); ++i) {
+    remaining.push_back(
+        {blockInfos[i].hash, blockInfos[i].accumulatedThinkTokens});
+  }
+  return remaining;
+}
+
+ConsecutiveMatch BlockMatcher::countConsecutiveRemainingMatch(
+    const std::list<RemainingBlockInfo>& callerRemaining,
+    const std::list<RemainingBlockInfo>& entryRemaining,
```
ztorlakTT, line 25: I'm thinking you can also move RemainingBlockInfo, PrefixIndexEntry, Candidate etc. to some separate file, but you can do this in the following PR

tt-media-server/cpp_server/src/services/session_resolution.cpp
```diff
@@ -16,7 +16,14 @@ std::optional<SlotCopyPlan> prepareSlotCopy(
     return std::nullopt;
   }
 
-  auto copyCandidate = sessionManager.findASlotToCopyFrom(candidates);
+  auto copyCandidate = domain::BlockMatcher::findSlotToCopyFrom(
+      candidates, [&sessionManager](const std::string& sessionId) {
+        uint32_t committedBlocks = 0;
+        if (auto session = sessionManager.getSession(sessionId)) {
+          committedBlocks = session->committedBlocks();
+        }
```
ztorlakTT: Just double-checking - should we preserve the old behavior - to read committedBlocks via sessions.modify?
```suggestion
        sessions.modify(sessionId, [&committedBlocks](std::shared_ptr<domain::Session>& s) {
          committedBlocks = s->committedBlocks();
        });
```

ztorlakTT (approved):

tt-media-server/cpp_server/src/services/session_resolution.cpp
lmuravljovTT: Yes this is a good catch, modify is faster.. 
I caught this on some of the following branches when I tested benchmarks, but might as well fix it right away, thanks!

tt-media-server/cpp_server/src/domain/block_matcher.cpp
```diff
@@ -0,0 +1,185 @@
...
+void BlockMatcher::sortCandidates(std::vector<Candidate>& candidates) {
+  std::sort(candidates.begin(), candidates.end(),
+            [](const Candidate& a, const Candidate& b) {
+              return a.matchedBlocks > b.matchedBlocks;
+            });
+}
+
+bool BlockMatcher::passesHitThreshold(const Candidate& candidate,
+                                      float threshold) {
+  if (threshold <= 0.0f) {
```
idjuricTT: Is this case ever possible?
Nitpick but this class can be aware of the thresold

tt-media-server/cpp_server/include/domain/block_matcher.hpp
```diff
@@ -0,0 +1,53 @@
...
+  std::size_t matchedRemainingBlocks = 0;
+  uint32_t lastMatchedThinkTokens = 0;
+};
+
+class BlockMatcher {
+ public:
+  static std::list<RemainingBlockInfo> buildCallerRemaining(
+      const std::vector<tt::utils::BlockHashInfo>& blockInfos);
+
+  static ConsecutiveMatch countConsecutiveRemainingMatch(
```
idjuricTT: Nitpick: is this method just countMatchedTokens?

tt-media-server/cpp_server/include/domain/block_matcher.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "domain/prefix_index.hpp"
+#include "utils/conversation_hasher.hpp"
+
+namespace tt::domain {
+
+struct ConsecutiveMatch {
```
idjuricTT: Would this be more descriptive if it was MatchedTokens with matchedBlocks and matchedThinkingTokens?

It’s called last but it seems like total

idjuricTT (approved):

tt-media-server/cpp_server/src/domain/block_matcher.cpp
lmuravljovTT: nope you're right, updated

tt-media-server/cpp_server/src/domain/block_matcher.cpp
lmuravljovTT: I'll do it in a follow up, thanks!

tt-media-server/cpp_server/src/services/session_resolution.cpp
lmuravljovTT: Updated by exposing another method in SessionManager, since we can't access sessions here

ztorlakTT (approved):

PR #4449, opened 2026-07-01
introduce prefix cache router

lmuravljovTT:
image


tt-media-server/cpp_server/include/services/prefix_cache_router.hpp
```diff
@@ -0,0 +1,107 @@
...
+ public:
+  using Candidate = domain::prefix_cache::Candidate;
+  using AcquireResult = PrefixCacheAcquireResult;
+
+  explicit PrefixCacheRouter(PrefixCacheRouterCallbacks callbacks);
+
+  PrefixCacheRouter(const PrefixCacheRouter&) = delete;
+  PrefixCacheRouter& operator=(const PrefixCacheRouter&) = delete;
+
+  std::optional<AcquireResult> tryAcquireByPrefixHash(
```
dmadicTT, line 64: Question about final goal
Since "prefix cache" is a feature, and "hashing mechanism" is the way we implemented it, I would expect the implementation details of prefix caching, like hashing and `blockInfos` to be all hidden inside the `PrefixCacheRouter`.

Ideally
```
auto slotId = prefixCacheRouter.getSlot(promptTokenIds)
```

Inside this does all the magic
- splitting into blocks
- hashing
- checking if we can reuse a slot or we should get a new one

Does this make sense?

lmuravljovTT: Yeah that does make sense, 
I tried to split the work into multiple iterations to have the review more thorough, but I might've made it non-intuitive 

My plan was to do LlmPipeline and SessionManager cleanups separately, where it would start going into your proposed direction, however it might make more sense to do it now?

dmadicTT: Great! I think you are doing great job splitting the work, so feel to do it as you wish.
I just wanted to check if our visions match :)

tt-media-server/cpp_server/include/services/session_manager.hpp
```diff
@@ -42,6 +39,12 @@ class SessionInFlightException : public SessionRateLimitException {
             "requests per session are not supported.") {}
 };
 
+}  // namespace tt::services
+
+#include "services/prefix_cache_router.hpp"
```
ztorlakTT: Can you move this line up with the other includes, so we don’t need to reopen `the tt::services` namespace?

ztorlakTT (approved):

tt-media-server/cpp_server/include/domain/session_manager_structs.hpp
lmuravljovTT, line 1: @ztorlakTT can you take a final look at this before I merge 
I need these structs in both prefix cache router & session_manager, but it didn't make sense to keep them in prefix_Cahce_router.hpp

idjuricTT (approved):

PR #4453, opened 2026-07-01
fix dynamo shutdown

knovokmetTT:


ljovanovicTT (approved):

lmuravljovTT (approved):

idjuricTT (approved):

tt-media-server/cpp_server/src/main.cpp
```diff
@@ -64,9 +65,21 @@ bool probePort(const std::string& host, uint16_t port) {
...
 void signalHandler(int signal) {
-  TT_LOG_WARN("\n[Main] Received signal {}, initiating shutdown...", signal);
-  gShutdownRequested = 1;
-  drogon::app().quit();
+  // First signal: request a graceful shutdown. Subsequent signal: the graceful
+  // path is stuck (e.g. a thread join that never completes), so force-exit with
+  // the conventional 128+signal code rather than leaving the user mashing
+  // Ctrl+C.
+  bool firstSignal = (gSignalCount == 0);
+  gSignalCount = 1;
```
nanicicTT, line 67: Maybe strip the handler down to bare memory writes.

```cpp
void signalHandler(int signal) {
  if (gSignalCount == 0) {
    gSignalCount = 1;
    gShutdownRequested = 1;
  } else {
    std::signal(signal, SIG_DFL);
    std::raise(signal);
  }
}
```

The handler only flips the `std::sig_atomic_t` flags and returns immediately. The main event loop checks these flags periodically or via a safe mechanism and executes the logging and quit logic in a normal, uninterrupted thread context. Deadlocks are impossible.

**as of the current code:**
---
`TT_LOG_WARN` logging requires formatting strings, which triggers heap allocation, right? It also writes to output streams, which requires acquiring some internal mutexes.
`drogon::app().quit()` framework methods interact with event loops, thread pools, and internal locks. They are virtually never async-signal-safe.

If a user presses `Ctrl+C` at the exact millisecond the main thread is holding the logger mutex or is inside a memory allocation call, the signal handler runs. It calls `TT_LOG_WARN`, tries to acquire the same mutex, and deadlocks. The server hangs and requires a force kill to terminate. 

**From the fix above:**
---
The handler does exactly two things. It reads a boolean flag and writes to an atomic variable.
It does not allocate memory. It does not acquire any mutexes. It does not interact with the Drogon framework.
Because it avoids all shared resources, it is physically impossible for it to collide with whatever the main thread was doing when the signal arrived. It executes in a few CPU cycles, flips the bits, and returns control to the main thread. The main thread resumes its work completely unaware that a signal just occurred, until it naturally checks the flag.


You still need to shut down Drogon safely, right? Since you cannot call `drogon::app().quit()` in the handler, you must trigger it from a normal thread context.
You can accomplish this by running a lightweight background thread to poll `gShutdownRequested`

```cpp
std::thread shutdownMonitor([]() {
  while (gShutdownRequested == 0) {
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
  }
  // We are in a normal thread context now. It is safe to call framework methods.
  drogon::app().quit();
});
shutdownMonitor.detach();
```

Or something like this. 


My issue is that Drogon relies on dozens of internal mutexes to manage active HTTP connections and thread pools. If the signal interrupts the main thread while it is processing a request and holding a Drogon internal lock, the handler calling `quit()` will try to acquire those same locks to tear the server down. It might deadlock.

With the safe handler:
---
The signal interrupts the main thread. The handler only flips a boolean flag. This operation requires no locks and allocates no memory. It is physically impossible to collide with the main thread's internal state. The handler exits in a microsecond. The main thread wakes up, finishes its current task, checks the flag on its own schedule, and shuts down safely.

With your current handler:
---
The signal interrupts the main thread. The handler calls complex functions that require internal framework locks. If the main thread happened to be holding any of those locks at the exact microsecond the signal arrived, the server hangs forever and ignores all future Ctrl+C inputs.

PR #4459, opened 2026-07-01
Knovokmet tt/dynamo graceful shutdown

knovokmetTT:


ljovanovicTT (approved):

tt-media-server/cpp_server/src/dynamo/dynamo_endpoint.cpp
```diff
@@ -482,25 +482,24 @@ void DynamoEndpoint::stop() {
   if (!running_.exchange(false)) {
     return;
   }
-
```
lmuravljovTT, line 485: Wouldn't it be useful to see its shutting down in logs?

lmuravljovTT (approved):

PR #4466, opened 2026-07-01
docs: fix Llama-3.1-8B P100/P150 quickstart to use working launch commands

hous:
## Summary

The documented quickstart for **Llama-3.1-8B on P100/P150** crashes with the tt-metal commit currently pinned for these pages (`55fd115`): the server comes up and passes its health check, but the first request that needs a 2048-token prefill crashes the engine. This makes the pages' commands non-functional as written.

This PR replaces the crashing commands on both pages with a working launch flow:
- a `.env` setup that includes `MAX_PREFILL_CHUNK_SIZE=2` (alongside `HF_TOKEN` / `JWT_SECRET`), and
- a `run.py` command that caps `max_model_len` via `--vllm-override-args '{"max_model_len": 1024}'`.

The non-working `docker run` blocks are dropped for these devices (there is no verified working docker path).

Related tt-metal issue: [tenstorrent/tt-metal#28835](https://github.com/tenstorrent/tt-metal/issues/28835).

## Testing

Verified on a **P100** host with both llama-3.1-8b and llama-3.1-8b-instruct.

Reproduce the failure with the previously-documented command:
1. Launched `python3 run.py --model Llama-3.1-8B --device p100 --workflow server --docker-server`; the server comes up and passes its health check.
2. A completion request triggers a 2048-token prefill, which crashes the engine:

```
INFO:     127.0.0.1:37140 - "POST /v1/completions HTTP/1.1" 200 OK
INFO 07-02 20:01:01 [engine.py:339] Added request cmpl-f1e0e846d86644e285b4ae0c51c6fef4-0.
2026-07-02 20:01:01.009 | INFO     | models.tt_transformers.tt.generator:prefill_forward_text:94 - Prefilling User 1 up to 2048 tokens
2026-07-02 20:01:01.613 | critical |          Always | Statically allocated circular buffers in program 205 clash with L1 buffers on core range [(x=0,y=0) - (x=6,y=7)]. L1 buffer allocated at 1568512 and static circular buffer region ends at 1572736 (assert.hpp:103)
```

Apply the fix:

3. Created a `.env` in the repo root with `HF_TOKEN`, `JWT_SECRET`, and `MAX_PREFILL_CHUNK_SIZE=2`.
4. Launched with `python3 run.py --model Llama-3.1-8B --device p100 --workflow server --docker-server --vllm-override-args '{"max_model_len": 1024}'`; the server started and served completion requests without the crash.

P150 was not physically tested, but it shares the same root cause (no per-device entry in tt-metal's `MAX_PREFILL_CHUNK_SIZES` table) and the same fix.

## Implementation

These pages are generated by `scripts/release/generate_model_support_docs.py`, so the fix is made at the source rather than by hand-editing the `.md` (which would be overwritten on the next regeneration):

- Added a `DEVICE_MODEL_QUICKSTART_OVERRIDES` registry keyed by `(model, device)`. When an entry exists, its builder **replaces** the default `docker run` / `run.py` commands for that model+device.
- Registered the override for `Llama-3.1-8B` on `P100` and `P150`.
- Regenerated `docs/model_support/llm/Llama-3.1-8B_p100.md` and `Llama-3.1-8B_p150.md`.

Re-running the generator against unmodified `main` produces a zero diff, so this change touches only the generator and those two pages.

> **Note:** the override is tied to the pinned tt-metal commit `55fd115`. Remove the registry entries once that commit gains a per-device `MAX_PREFILL_CHUNK_SIZES` entry for P100/P150.

## Generated page (reference)

Rendered `docs/model_support/llm/Llama-3.1-8B_p100.md` (P150 is identical with `p150` substituted):

````markdown
# Llama-3.1-8B Tenstorrent Support on P100

Supported weights variants for this model implementation are:

- `Llama-3.1-8B`: [meta-llama/Llama-3.1-8B](https://huggingface.co/meta-llama/Llama-3.1-8B) **(default)** 
- `Llama-3.1-8B-Instruct`: [meta-llama/Llama-3.1-8B-Instruct](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct)

To use non-default weights, replace `Llama-3.1-8B` in commands below.

#### Useful links

- [P100 details](https://tenstorrent.com/hardware/blackhole)
- [Search other llm models](./README.md)
- [Search other models by model type](../../../README.md#models-by-model-type)

`Llama-3.1-8B` is also supported on hardware:

- [WH Galaxy](Llama-3.1-8B_galaxy.md)
- [BH LoudBox](Llama-3.1-8B_p150x8.md)
- [BH 4xP150](Llama-3.1-8B_p150x4.md)
- [BH QuietBox 2](Llama-3.1-8B_p300x2.md)
- [P150](Llama-3.1-8B_p150.md)
- [WH LoudBox/QuietBox](Llama-3.1-8B_t3k.md)
- [N150](Llama-3.1-8B_n150.md)
- [N300](Llama-3.1-8B_n300.md)

## Quickstart - Deploy Llama-3.1-8B Inference Server on p100

See [prerequisites](../../prerequisites.md) for system software setup, e.g. for first-run or when experiencing issues.

This model is supported by [vLLM (tt-metal integration fork)](../../../vllm-tt-metal/README.md) inference engine.

Create a `.env` file in the repo root with the following before launching:

```bash
HF_TOKEN=<replace-with-your-hugging-face-token>
JWT_SECRET=<replace-with-random-secret>
MAX_PREFILL_CHUNK_SIZE=2
```

**via run.py command**

```bash
python3 run.py --model Llama-3.1-8B --device p100 --workflow server --docker-server \
  --vllm-override-args '{"max_model_len": 1024}'
```
For details on the run.py command, see the [run.py CLI Options](../../workflows_user_guide.md#runpy-cli-options) section of the User Guide.

## Model Parameters

| Parameter | Value |
|-----------|-------|
| Weights | [meta-llama/Llama-3.1-8B](https://huggingface.co/meta-llama/Llama-3.1-8B), [meta-llama/Llama-3.1-8B-Instruct](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct) |
| Model Status | 🛠️ Experimental |
| Max Batch Size | 32 |
| Max Context Length | 65536 |
| Implementation Code | [tt-transformers](https://github.com/tenstorrent/tt-metal/tree/55fd115/models/tt_transformers) |
| tt-metal Commit | `55fd115` |
| vLLM Commit | `aa4ae1e` |
| Docker Image | `ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64:0.10.0-55fd115-aa4ae1e` |
````

### CODE REVIEW

**Comment by tstescoTT** [comment]
@vmaksimovicTT was there a release CI run of this model on p100 that passed without the `MAX_PREFILL_CHUNK_SIZE=2` env var set?

**Comment by tstescoTT** [comment]
The correct way to add env vars per model-hardware impl is in the model spec yaml file (https://github.com/tenstorrent/tt-inference-server/blob/main/workflows/model_specs/dev/llm.yaml), for example in https://github.com/tenstorrent/tt-inference-server/pull/4445/changes.

This fails in this case because this missing env var escaped into release version, which should not happen.

The doc override system in general I think is a good idea, want @vcankovicTT and @vmaksimovicTT to comment on if they agree.

File: scripts/release/generate_model_support_docs.py [hunk 1]
```diff
@@ -174,6 +168,46 @@ def from_device_type(cls, device_type: DeviceTypes) -> "HardwarePageGroup":
...
 
+def _bh_llama8b_prefill_chunk_commands(
+    model_name: str, device: DeviceTypes
+) -> List[str]:
+    """Working launch commands for Llama-3.1-8B on single-chip Blackhole (P100/P150).
+
+    On these devices the default quickstart commands crash on startup with the tt-metal
+    commit currently pinned for these pages (55fd115), so they are replaced with a flow
+    that sets MAX_PREFILL_CHUNK_SIZE and caps max_model_len. Remove this entry once the
+    pinned commit is updated. See tenstorrent/tt-metal#28835.
```
tstescoTT: Thanks for adding the removal criteria. @vcankovicTT and @vmaksimovicTT please advise how you want these to be tracked as you will need to execute the removal.

vmaksimovicTT (approved):

fivanovicTT (approved):

PR #433, opened 2025-08-05
Fix import path for MODEL_SPECS in build_docker_images.py

nidhinjose-git:
Fix typo:

```
Traceback (most recent call last):
  File "/tmp/test/tt-inference-server/scripts/build_docker_images.py", line 22, in <module>
    from workflows.model_specification import MODEL_SPECS
ModuleNotFoundError: No module named 'workflows.model_specification'
```

Issue #431, opened 2025-08-05
Failing Qwen2.5-7B-Instruct on n150 and n300


acvejicTT:
### Qwen2.5-7B-Instruct on n150
https://github.com/tenstorrent/tt-shield/actions/runs/16634784945/job/47074052325

```
running command: python3 run.py --model Qwen2.5-7B-Instruct --device n150 --workflow release --docker-server --dev-mode --override-docker-image ghcr.io/tenstorrent/tt-shield/vllm-tt-metal-src-dev-ubuntu-22.04-amd64:0.0.5-v0.60.1-7ff6d6f-47072986920
Traceback (most recent call last):
  File "run.py", line 378, in <module>
    sys.exit(main())
  File "run.py", line 299, in main
    infer_args(args)
  File "run.py", line 200, in infer_args
    raise ValueError(
ValueError: Model:=Qwen2.5-7B-Instruct does not have a default impl, you must pass --impl
```

### Qwen2.5-7B-Instruct on n300
https://github.com/tenstorrent/tt-shield/actions/runs/16634784945/job/47074052363

Docker server log
```
using CACHE_ROOT: /home/container_app_user/cache_root
Mounted CACHE_ROOT volume is owned by UID:GID - 1004:1004
setting permissions for /home/container_app_user/cache_root ...
Mounted volume permissions setup completed.
INFO 07-31 01:39:44 [__init__.py:252] Automatically detected platform tt.
Traceback (most recent call last):
  File "/home/container_app_user/app/src/run_vllm_api_server.py", line 100, in <module>
    register_tt_models()  # Import and register models from tt-metal
  File "/home/container_app_user/app/src/run_vllm_api_server.py", line 93, in register_tt_models
    raise ValueError(
ValueError: Unsupported model_impl: llama3, pick one of [tt-transformers, subdevices, llama2-t3000]

bgoelTT:
Yes it works on n300.

I think Tom might have just made a mistake. The model is simply unsupported by TT-Transformers on n150. Thank you, please remove

acvejicTT:
@bgoelTT are we sure about Qwen2.5-7B-Instruct not being supported on n150?
I think I saw it on slides during yesterday's Red Team sync?
cc: @milank94

PR #4428, opened 2026-06-29
[Model-Readiness-Support] Qwen 3.6 `tt-inference-server` support

lingjunwenTT:
Passes benchmarks workflow, fails at `spec_test`
[link to ci run](https://github.com/tenstorrent/tt-shield/actions/runs/28617621300/job/84865283599)

This provides changes to support Qwen 3.6 tt-inference-server integration with these characteristics:

- max context: 256k
- max concurrency: 1

Related branches in other repos:
- vllm: `lingjunw/support-qwen35-27b` [PR](https://github.com/tenstorrent/vllm/pull/434) MERGED
- tt-metal: `atupe/qwen36-ci-cache-issue` [PR](https://github.com/tenstorrent/tt-metal/pull/48522)
- tt-inference-server: `lingjunw/support-qwen35-27b` [PR](https://github.com/tenstorrent/tt-inference-server/pull/4428)

Commands to run locally:
1. server start-up
```bash
cd ~/tt-inference-server
  python3 run.py --model Qwen3.6-27B --device p300x2 --workflow server \
    --local-server --dev-mode --tt-metal-home ~/tt-metal --vllm-dir ~/vllm \
    --skip-system-sw-validation --no-auth
```

2. evals / benchmarks workflow
```bash
cd ~/tt-inference-server
  python3 run.py --model Qwen3.6-27B --device p300x2 --workflow benchmarks \
    --dev-mode --tt-metal-home ~/tt-metal --vllm-dir ~/vllm \
    --skip-system-sw-validation --no-auth

```

PR #4470, opened 2026-07-02
Gemma 4 (12B/31B) on QB2 (P300X2): integration, evals, benchmarks

dwadhwaniTT:
## Summary

Enables **Gemma 4 12B and 31B** inference + release workflow on **QB2 (P300X2)** via the tt-transformers/vLLM path, and moves the Gemma4-specific glue out of image bakes and into upstream `tenstorrent/vllm` (`vllm-tt-plugin`).

Key pieces:
- **vLLM pin `375df057`** (tenstorrent/vllm dev HEAD): includes #431 (Gemma4 reasoning + tool parsers now live in `vllm-tt-plugin`, so the previous parser *bakes* are dropped) and #433 (plugin `[runtime]` extra no longer clobbers the tt-metal vLLM env — fixes the `import vllm._C` / `libcudart.so.12` boot crash, so no Dockerfile hack is needed).
- **tt-metal pin `801ff4f58b0`** (was `a4967d5f39d` / #47556). The bump pulls in the two changes that fix the single-request KV-cache ceiling on QB2 — see **KV-cache ceiling** below. **Depends on tenstorrent/tt-metal#48819** to land the override on `main`.
- **Server-side thinking + auto tool choice** baked into the gemma4 spec (`vllm_args`): `--default-chat-template-kwargs '{"enable_thinking": true}'` (required for the `r1_gpqa_diamond` #4331 methodology — without it the template injects an empty thought channel and suppresses reasoning) and `--enable-auto-tool-choice` (so `tool_choice:"auto"` clients like mini-swe-agent work flag-free).
- **Evals aligned with #4331** for both sizes: `r1_gpqa_diamond` + `terminal_bench_2` + `swe_bench_verified`, thinking-mode sampling (`temp=1.0/top_k=20/top_p=0.95`), `ci-nightly`-limited. Token budgets are the only deviation from #4331, clamped to QB2's envelope.
- **Docs**: Gemma-4 12B/31B added to the LLM model-support + by-hardware tables, plus a dedicated 31B/P300X2 page.

## KV-cache ceiling: the 23K → 49K fix

The 31B/12B "hangs/crashes above ~23K input" on QB2 turned out **not** to be a device or prefill-memory problem — it was a vLLM hybrid-KV-cache-groups accounting issue:

- With `_HYBRID_KV_CACHE_GROUPS_ENABLED = True` (the state at the old tt-metal pin `a4967d5f39d`), Gemma-4 emits `SlidingWindowSpec` for the sliding layers + `FullAttentionSpec` for the full layers → vLLM builds **6 groups** and reports `num_blocks // num_groups * block_size`, splitting the pool ~6 ways. The full-attention group is starved to **~23,168 tokens** (`Maximum concurrency for 49,152 tokens/request: 0.51x`), so any single request above ~23K is **rejected at KV admission** (`Running:0 / Waiting:1` indefinitely — this was the "32K stall").
- **Fix = two tt-metal changes, both on the new pin `801ff4f58b0`:**
  1. **Disable hybrid KV-cache groups** (tt-metal #48283, `8770fd0827c`, already on `main`): emit `FullAttentionSpec` for every layer → a single group gets the whole pool (mirrors Gemma-3).
  2. **`GEMMA4_MAX_TOKENS_ALL_USERS` override** (tt-metal **#48819**, *pending merge to main*): with hybrid off, every layer allocates a full-length KV buffer, so the default 131,072-token pool (~120 GiB KV + 58 GiB weights) would **OOM** on QB2's 128 GB. The override caps the pool per-device from the model spec (`49152` for 31B; 12B uses the default). **31B requires this** — it is not optional.

**Verified on QB2** with the rebuilt `0.17.0-801ff4f58b0-375df057` image: `GPU KV cache size: 49,216 tokens`, `Maximum concurrency for 49,152 tokens/request: 1.00x`, single KV group — a full-length request is now admittable (was `23,168` / `0.51x`). Reaching the full 128K target while keeping sliding-window savings is a follow-up requiring vLLM token-chunked prefill + the tt-metal `chunk_start_idx` path (tracked at tt-metal#48289).

## CI-path fixes (v2 plumbing + spec/catalogue)
Bringing this up through the tt-shield dispatch surfaced a chain of environment/plumbing bugs, all fixed here:

- **Auth (`fix(v2-auth)`):** the v2 client minted its bearer token with an extra `exp` claim, but the server sets `VLLM_API_KEY` from `get_encoded_api_key()` (`{team_id, token_id}`, no `exp`) and vLLM authorizes by **exact-string comparison**, not JWT validation → `401 Unauthorized` on every request (`vllm bench serve` "initial test run" failed / eval stalled). Dropped the `exp` claim so the token matches byte-for-byte. Fixes the observed CI `Unauthorized` failures.
- **Benchmark/eval venv (`fix(v2-bench)`):** pin `datasets==3.1.0` (unpinned → `2.14.4`, whose `features.py` uses `pyarrow.PyExtensionType` removed in pyarrow≥15 → import crash); and apply the `transformers>=5.10.2,<6` override to the `V2_LLM_VLLM` venv (4.x raises `'list' object has no attribute 'keys'` on Gemma-4's tokenizer).
- **Dev spec load (`fix(gemma4)`):** removed `tt_metal_commit`/`vllm_commit` from the **dev** gemma4 entries — the base `ModelSpecTemplate` rejects them (only `ProdModelSpecTemplate` accepts them), which was crashing `determine_server_type.py` at spec load.
- **CI logs perms (`fix(ci)`):** in ci-mode redirect `TT_METAL_LOGS_PATH` → `/tmp/tt_metal_logs`. CI runs the container as a non-1000 `--user`, so the baked `/home/container_app_user/logs` (uid 1000) isn't writable → `ttnn.open_mesh_device()` failed creating `logs/generated/watcher/` and the engine never booted.
- **Sole default impl (`fix(gemma4)`):** made vLLM the unambiguous `default_impl` for gemma-4-31b-it on P300X2 (a pre-existing EXPERIMENTAL forge entry also claimed it → `determine_server_type` "Multiple default impls").
- **Catalogue hygiene:** removed hand-added gemma-4 entries from the **prod** catalogue (prod is automation-owned, built from validated dev at release); dev carries the full config (CI/local use `MODEL_SPECS_ENV=dev`).
- **`models-ci-config.json`:** added nightly + release jobs on P300X2 for both sizes — 31b as a **multi-engine** entry (keeps the forge nightly, adds a vLLM impl) and 12b as vLLM. Schema-validated.
- **Lint:** `ruff==0.15.0` format to satisfy the Lint & Format gate (formatting-only).

## Local validation on QB2 (P300X2)
On the rebuilt `0.17.0-801ff4f58b0-375df057` image:
- **Server boot**: clean, no `libcudart`/eth/spec-load errors; boot args carry `enable_thinking`, `--enable-auto-tool-choice`, `--tool_call_parser gemma4`, `reasoning_parser: gemma4`.
- **KV cache**: `49,216 tokens`, `1.00x` for 49,152/request, single group — 23K ceiling resolved (see above).
- **Evals (GPQA)**: `r1_gpqa_diamond` at ci-nightly (40 samples) ran end-to-end with **JWT auth enabled** (validates the auth fix). QB2 scores ~75 vs the ~83 H100 reference — the known/accepted hybrid-off gap, not a regression.
- **Tool-call parser**: live `tool_choice:"auto"` request → `finish_reason: tool_calls`, parsed `get_weather{"location":"Paris, France"}` — the mechanism `swe_bench_verified`/mini-swe-agent relies on (no `"auto" tool choice` 400s).

## ⚠️ Known limitation: eval scoring vs GPU reference (a green dispatch is not achievable as-is)

QB2 can't match the published/GPU-reference eval scores, and this is a **structural hardware limit, not a model regression**:

- Even with the KV fix, a single request is bounded: `max_model_len = 49,152`, and the **single-prefill bucket ceiling is 32,768** (Gemma-4 has no token-chunked prefill yet — tt-metal#48289), so usable **input tops out ~≤32K**.
- The H100/GPU reference scores for the agentic evals were produced at **full long context** (`terminal_bench_2` and `swe_bench_verified` used ~112K/160K input). Eval tasks/instances that require more than ~32K input **cannot run on QB2** — the request is rejected/truncated — so their score contribution is lost.
- Therefore QB2 eval scores **cannot reach the full-context reference within tolerance**, and the standard release acceptance check (score within tolerance of the GPU reference) will read **FAIL**. In other words, **a "green" release dispatch against the full-context reference is not achievable on QB2** for these evals.
- Note: `r1_gpqa_diamond` is short-context and *does* run fully; its gap (~75 vs ~83) is the separate, accepted **hybrid-off perf gap**, not the context limit. The context limit specifically bites the **agentic** evals.

**Proposed path forward (fair, apples-to-apples comparison):** scope each eval to the **subset of tasks/instances whose context fits ≤32K**, **regenerate the GPU reference scores on that same subset**, and compare QB2 vs GPU like-for-like. Until that scoped reference exists, dispatch acceptance stays FAIL against the full reference — this is expected and should be interpreted against the scoped reference, not the published full-context numbers. The ideal alternative — serving full context via token-chunked prefill (tt-metal#48289 + plugin/scheduler work) — may not be near-term feasible.

## Dependencies / remaining for merge
- **tenstorrent/tt-metal#48819** (the `GEMMA4_MAX_TOKENS_ALL_USERS` override) must merge to tt-metal `main`. Until then there is no single `main` tt-metal commit that boots 31B (the disable is on main, the override is not), so the dispatch pins the feature-branch commit `801ff4f58b0`.
- **≥1 green on-dispatch** on P300X2 in tt-shield — pin the exact validated commits (tt-metal `801ff4f58b0`, vllm `375df057`, this branch); branch-tip refs (`dev`/`main`) drift and break.
- **Shield team** to cherry-pick the dev-catalogue gemma-4 changes onto the stable branch for this cycle.

## Model Readiness Support Checklist
- [x] Inference server can serve model via run.py server workflow.
- [ ] Inference server can run release workflow. Release markdown report for target hardware pasted in comment below. *(pending a green ci-nightly dispatch on the `801ff4f58b0` image; local smoke of both halves passed)*
- [ ] Specific inference server docs cover how to serve model and run release workflow.
- [x] Model is added to top level README / model-support tables.
- [x] Model is added to tt-shield Model Readiness CI config (nightly + release, P300X2).

## Test plan
- [x] Clean image build boots without the CUDA clobber and registers the `gemma4` tool parser (no `libcudart` crash; parser unit tests 8/8 functional pass).
- [x] KV cache allocates the full pool on QB2 (single group, 49,216 tokens, 1.00x) — a full-length request is admittable.
- [x] Benchmarks (`vllm bench serve`) produce TTFT / tok-s for 31B.
- [ ] Terminal-Bench-2.0 produces a valid score. *(runs in CI ci-nightly)*
- [x] SWE-Bench Verified path unblocked: `--enable-auto-tool-choice --tool-call-parser gemma4` validated via a live `tool_choice:"auto"` request returning a parsed `tool_calls` entry. Full dataset runs in CI.


docs/model_support/llm/gemma-4-12b-it_p300x2.md
bgoelTT, line 1: Please reverts all changes to `docs/` as this only gets updated during release creation

evals/eval_config.py
```diff
@@ -3480,29 +3480,348 @@ class EvalConfig:
     EvalConfig(
         hf_model_repo="google/gemma-4-31b-it",
```
bgoelTT: Please delete this duplicate `hf_model_repo="google/gemma-4-31b-it"` EvalConfig and use the correct one that @cleonidouTT added below, which is the `hf_model_repo="google/gemma-4-31B-it"` (capital-B)

Let's see how the evals workflow fares with the model being configured with max_model_len=48K, it might leave to pure 400 errors

requirements/benchmarks-vllm-overrides.txt
bgoelTT, line 1: Please delete as vLLM benchmark sweep is now run through V2 and has the requirements file `v2-llm-vllm.txt`

requirements/benchmarks-vllm.txt
bgoelTT, line 1: Same as above

requirements/v2-llm-vllm.txt
bgoelTT, line 1: Please rebase to `main` after https://github.com/tenstorrent/tt-inference-server/pull/4428 is merged. It contains the fix for this problem

vllm-tt-metal/patches/gemma4_generator_vllm_9970093.py
bgoelTT, line 1: Please revert all commits to this `vllm-tt-metal/` directory

workflows/model_specs/dev/llm.yaml
```diff
@@ -152,6 +152,118 @@ templates:
...
+        # rather than hardcoded in tt-metal).
+        GEMMA4_MAX_TOKENS_ALL_USERS: "49152"
+      override_tt_config:
+        fabric_config: FABRIC_1D
+        trace_region_size: 200000000
+        # On-device decode sampling is required on the TP mesh so token ids
+        # >= 65536 are reachable (host sampling only sees device 0's vocab shard).
+        sample_on_device_mode: decode_only
+      vllm_args:
+        override_generation_config: '{"temperature": 1.0, "top_k": 64, "top_p": 0.95}'
```
bgoelTT: This can be removed as it is the default in the HF config

workflows/model_specs/dev/llm.yaml
```diff
@@ -152,6 +152,118 @@ templates:
...
+  status: EXPERIMENTAL
+  has_builtin_warmup: true
+  env_vars:
+    VLLM_ALLOW_LONG_MAX_MODEL_LEN: 1
+  metadata:
+    google/gemma-4-31b-it:
+      reasoning_parser_name: gemma4
+      # tool parser ships in vllm-tt-plugin as of tenstorrent/vllm #431; activated
+      # by enable-auto-tool-choice in vllm_args above.
+      tool_call_parser_name: gemma4
```
bgoelTT, line 218: Please also add these to the `vllm_args`

workflows/model_specs/dev/llm.yaml
```diff
@@ -152,6 +152,118 @@ templates:
...
+      vllm_args:
+        override_generation_config: '{"temperature": 1.0, "top_k": 64, "top_p": 0.95}'
+        # Activate the gemma4 tool_call_parser so tool_choice:"auto" clients
+        # (mini-swe-agent / SWE-Bench) work on a flag-free release run.
+        enable-auto-tool-choice: true
+        # Enable gemma-4 thinking server-side so the r1_gpqa_diamond /
+        # terminal-bench-2 / swe-bench-verified evals (methodology matched to
+        # tenstorrent/tt-inference-server#4331) get the reasoning channel the
+        # chat template otherwise suppresses (gemma-4 scores ~10pts lower w/o it).
+        default-chat-template-kwargs: '{"enable_thinking": true}'
```
bgoelTT, line 207: This is the default behaviour of the model so it should be removed

workflows/model_specs/dev/llm.yaml
```diff
@@ -152,6 +152,118 @@ templates:
   status: EXPERIMENTAL
   has_builtin_warmup: true
 
+- weights:
+    - google/gemma-4-31b-it
```
bgoelTT: This should be `google/gemma-4-31B-it`

workflows/model_specs/dev/llm.yaml
```diff
@@ -152,6 +152,118 @@ templates:
...
+    VLLM_ALLOW_LONG_MAX_MODEL_LEN: 1
+  metadata:
+    google/gemma-4-31b-it:
+      reasoning_parser_name: gemma4
+      # tool parser ships in vllm-tt-plugin as of tenstorrent/vllm #431; activated
+      # by enable-auto-tool-choice in vllm_args above.
+      tool_call_parser_name: gemma4
+
+- weights:
+    - google/gemma-4-12b-it
```
bgoelTT: Please apply the same changes to 12B as were made to 31B
```suggestion
    - google/gemma-4-12B-it
```

workflows/run_docker_server.py
bgoelTT, line 1: Revert these debug changes

workflows/workflow_venvs.py
bgoelTT, line 1: Please revert these debug changes

.gitignore
bgoelTT, line 1: Also revert these please

bgoelTT (changes requested):
Just some minor config comments + debug cleanup

workflows/model_specs/dev/llm.yaml
dwadhwaniTT: Looks like `enable_thinking: true` needs to stay here: the chat template gates thinking on this flag and defaults it **off**, so removing it would silently disable reasoning.

From `google/gemma-4-31b-it`'s `chat_template.jinja` (same for the capital-B repo):

```jinja
{%- if add_generation_prompt -%}
    ...
        {{- '<|turn>model\n' -}}
        {%- if not enable_thinking | default(false) -%}
            {{- '<|channel>thought\n<channel|>' -}}
        {%- endif -%}
```

Two things there:
- The flag's own default is `false` (`enable_thinking | default(false)`).
- When it's off, the template injects an **empty, pre-closed** thought channel (`<|channel>thought\n<channel|>`) into the model turn, i.e. it tells the model its thinking is already done → no reasoning is emitted.

The reasoning-open token is likewise gated:

```jinja
{%- if enable_thinking is defined and enable_thinking -%}
    {{- '<|think|>\n' -}}
```

So without `default-chat-template-kwargs: '{"enable_thinking": true}'`, gemma-4 runs non-thinking.

docs/model_support/llm/gemma-4-12b-it_p300x2.md
dwadhwaniTT: Done in f5394543. All docs/ changes reverted (gemma pages removed, tables restored); docs/ now matches origin/main.

evals/eval_config.py
dwadhwaniTT: Done in a74ca701. Deleted the lowercase gemma-4-31b-it and gemma-4-12b-it EvalConfig blocks; @cleonidouTT's capital-B google/gemma-4-31B-it / gemma-4-12B-it configs are now the only gemma-4 eval entries.

requirements/benchmarks-vllm-overrides.txt
dwadhwaniTT: Done in f5394543. Deleted.

requirements/benchmarks-vllm.txt
dwadhwaniTT: Done in f5394543. Reverted to origin/main.

vllm-tt-metal/patches/gemma4_generator_vllm_9970093.py
dwadhwaniTT: Done. Patches removed in f5394543; the remaining src/ (gemma4_unified + run_vllm_api_server.py), Dockerfile and requirements.txt changes reverted to origin/main in 3ae1c86c. vllm-tt-metal/ now matches main -- the gemma4 reasoning parser ships via vllm-tt-plugin (tenstorrent/vllm #431) in the baked image.

workflows/model_specs/dev/llm.yaml
dwadhwaniTT: Done in a74ca701. Removed; verified temperature=1.0 / top_k=64 / top_p=0.95 already load from each model's HF generation_config.json, so the override just restated the default.

workflows/model_specs/dev/llm.yaml
dwadhwaniTT: Done in a74ca701. 12B now carries the same vllm_args as 31B (enable-auto-tool-choice + default-chat-template-kwargs).

workflows/model_specs/dev/llm.yaml
dwadhwaniTT: Done in a74ca701. Renamed to google/gemma-4-31B-it (weights + metadata key).

workflows/model_specs/dev/llm.yaml
dwadhwaniTT: Done in a74ca701. Renamed to google/gemma-4-12B-it and matched 31B's vllm_args.

workflows/run_docker_server.py
dwadhwaniTT: Done in f5394543. Reverted; verified run_docker_server.py matches origin/main.

workflows/workflow_venvs.py
dwadhwaniTT: Done in f5394543. Reverted; verified workflow_venvs.py matches origin/main.

.gitignore
dwadhwaniTT: Done in 3ae1c86c. Reverted to origin/main; agent_docs/ is ignored locally via .git/info/exclude instead of a committed rule.

vllm-tt-metal/requirements.txt
```diff
@@ -1,3 +1,15 @@
...
+# vLLM does not upper-bound these, so a fresh build pulls whatever is latest on
+# PyPI. As of 2026-06, fastapi 0.137.x / starlette 1.3.x break
+# prometheus-fastapi-instrumentator 8.0.0 (mounted sub-routers surface as
+# `_IncludedRouter` objects without a `.path`, so every request 500s with
+# `AttributeError: '_IncludedRouter' object has no attribute 'path'`).
+# 0.136.3 / 1.2.1 are the versions the working 9970093 image shipped.
+# This requirements file installs last in the Dockerfile, so these win over
+# whatever vLLM pulled.
+fastapi==0.136.3
+starlette==1.2.1
```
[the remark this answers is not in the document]
dwadhwaniTT: Resolved: the fastapi/starlette pin lived in vllm-tt-metal/requirements.txt, which was reverted to origin/main in 3ae1c86c, so this dependency is no longer introduced by the PR.

requirements/v2-llm-vllm.txt
dwadhwaniTT: Still blocked: #4428 is open, so requirements/v2-llm-vllm.txt can't be rebased away yet. Leaving this thread open to track; will rebase to main as soon as #4428 merges.

workflows/model_specs/dev/llm.yaml
bgoelTT: 👍 understood - thanks for explanation

bgoelTT (approved):

bgoelTT (approved):

bgoelTT:
FYI for future readers, I am force-merging despite requiring reviews because the change that flagged codeowners was adding a comment

PR #4482, opened 2026-07-02
advertise dynamo-net IP interface to etcd when multiple exist

knovokmetTT:


tt-media-server/cpp_server/include/utils/net.hpp
```diff
@@ -0,0 +1,187 @@
...
+    url.erase(0, 7);
+  }
+  if (auto slash = url.find('/'); slash != std::string::npos) {
+    url.resize(slash);
+  }
+  ParsedUrl out;
+  if (auto colon = url.find(':'); colon != std::string::npos) {
+    out.host = url.substr(0, colon);
+    try {
+      out.port = std::stoi(url.substr(colon + 1));
```
nanicicTT, line 61: and if you have `"localhost:2379abc"`?
std::stoi("2379abc") returns 2379, not an error, right?

I am wondering is there some lib for parsing this? How tested is this?

nanicicTT: Also, what would happen if you have like `"http://host:2379/kosta"`?

tt-media-server/cpp_server/include/utils/net.hpp
```diff
@@ -0,0 +1,187 @@
...
+/// Resolve a hostname or literal IPv4 address for UDP. Returns a malloc'd
+/// `addrinfo` list (wrap in `AddrInfoGuard`). Throws on DNS failure.
+inline struct addrinfo* fetchAddrInfo(const std::string& host) {
+  struct addrinfo hints{};
+  hints.ai_family = AF_INET;       // IPv4-only
+  hints.ai_socktype = SOCK_DGRAM;  // UDP
+  struct addrinfo* res = nullptr;
+
+  // Fires up DNS resolver to resolve the hostname to an IP address.
+  int rc = ::getaddrinfo(host.c_str(), /*service=*/nullptr, &hints, &res);
```
nanicicTT: So, host is passed from `sourceIpForRoute` and then via `fetchAddrInfo` and not the port, even though you had it and you could have passed it to `sourceIpForRoute` and then again to `fetchAddrInfo`. 
You use nullptr instead, why? 
It means no service, or no port, so the destination address used for UDP `connect()` has port 0. 

Without this change, or suggestion if you will, route detection can silently fail and fall back to first non-loopback interface, which is exactly the old fragile behavior.

tt-media-server/cpp_server/include/utils/net.hpp
```diff
@@ -0,0 +1,187 @@
...
+                               endpointList + "'");
+    }
+  } else {
+    out.host = url;
+    out.port = 2379;
+  }
+  if (out.host.empty()) {
+    throw std::runtime_error("net::parseUrl: empty host in endpoint '" +
+                             endpointList + "'");
+  }
```
nanicicTT, line 73: All these runtime errors, but earlier it had `EtcdError`, right?
So, `EtcdClient` problems throw `EtcdError`. 
I would like for `EtcdClient` to translate the generic parser error back into its own error type.

Like, maybe: 

```cpp
EtcdClient::EtcdClient(const std::string& endpoint, int timeoutMs)
    : timeout_ms_(timeoutMs) {
  try {
    auto parsed = tt::utils::net::parseUrl(endpoint);
    host_ = parsed.host;
    port_ = parsed.port;
  } catch (const std::exception& e) {
    throw EtcdError(std::string("EtcdClient: invalid endpoint: ") + e.what());
  }
}
```

dmadicTT:
It's crazy we need to do this, but I don't see a more sophisticated option right now.
This works.

dmadicTT (approved):

tt-media-server/cpp_server/include/utils/net.hpp
```diff
@@ -0,0 +1,187 @@
...
+  char buf[INET_ADDRSTRLEN] = {0};
+  if (::inet_ntop(AF_INET, &local.sin_addr, buf, sizeof(buf)) == nullptr) {
+    throw std::runtime_error(
+        std::string("net::localIpFromSocket: inet_ntop failed: ") +
+        std::strerror(errno));
+  }
+  return buf;
+}
+
+/// High-level: return the source IP the kernel would use to route to `host`
```
dmadicTT: It would be good to have one human written code comment for this method, explaining the motive behind this black magic

tt-media-server/cpp_server/include/utils/net.hpp
knovokmetTT: the parsing logic is not a new code , just copy pasted into utils

nanicicTT: Alright, it is clear now.

tt-media-server/cpp_server/include/utils/net.hpp
knovokmetTT: done

PR #4484, opened 2026-07-02
Nanicic tt/mock scheduler part ii

nanicicTT:
I was running this command (twice)

```sh
OPENAI_API_KEY=your-secret-key \
vllm bench serve \
    --model 'deepseek-ai/DeepSeek-R1-0528' \
    --tokenizer 'cpp_server/tokenizers/deepseek-ai/DeepSeek-R1-0528' \
    --backend openai-chat \
    --endpoint /v1/chat/completions \
    --dataset-name random \
    --random-input-len 32\
    --random-output-len 1024\
    --num-prompts 1000 \
    --max-concurrency 64
```    

for old mock_pipeline: `LLM_DEVICE_BACKEND=mock_pipeline ./tt_media_server_cpp`
and for my new pipeline: `LLM_DEVICE_BACKEND=mock_scheduler ./tt_media_server_cpp`

and the results are **close**, as follows:

Screenshot 2026-07-02 at 14 58 29

I have utilized 

```cpp
// Per-chunk prefill compute added to time-to-first-token, in ms. 0 leaves TTFT
// at the bare pipeline fill (~a few ms). Set to ~100ms to match mock_pipeline's
// PrefillMockConfig chunk_latency, or to the model's measured prefill time.
constexpr unsigned MOCK_PREFILL_COMPUTE_MS = 100;
```

that mock_pipeline uses, and this drove metrics to the desired result. 
Also, important thing was adding random jitter to `nextAt` to not make slots free in a lockstep (explained in the code comments) but to trickle.


tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -146,11 +147,24 @@ constexpr const char* DYNAMO_ETCD_ENDPOINTS = "http://etcd:2379/";
 // the reaper.
 constexpr int64_t DYNAMO_ETCD_LEASE_TTL_SECS = 10;
 
-// MockSchedulers (MOCK_USE_SCHEDULER=1 on mock_pipeline backend).
-constexpr bool MOCK_USE_SCHEDULER = false;
 constexpr unsigned MOCK_PREFILL_CHUNK_LATENCY_MS = 1353;
-constexpr unsigned MOCK_DECODE_TOKEN_LATENCY_US =
-    64 * 44;  // 64 * 44us = 2816us
+// A 64-stage pipeline at 44us/stage. First-token (fill) latency is
```
knovokmetTT: too many unnecessary slop comments :)

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -146,11 +147,24 @@ constexpr const char* DYNAMO_ETCD_ENDPOINTS = "http://etcd:2379/";
...
+// +/-20% per-token jitter. Desyncs the otherwise phase-locked slots so
+// identical-length requests don't finish in clumps and spike the TTFT tail.
+// Mean-preserving (throughput/TPOT unchanged). Set to 0 for a perfect
+// metronome. Tune upward if the tail is still too fat for the workload.
+// This is tested as almost the sweet spot.
+constexpr uint32_t MOCK_DECODE_JITTER_PCT = 20;
+// Per-chunk prefill compute added to time-to-first-token, in ms. 0 leaves TTFT
+// at the bare pipeline fill (~a few ms). Set to ~100ms to match mock_pipeline's
+// PrefillMockConfig chunk_latency, or to the model's measured prefill time.
+constexpr unsigned MOCK_PREFILL_COMPUTE_MS = 100;
```
knovokmetTT: this doesnt affect prefill by decode timing, right?

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -146,11 +147,24 @@ constexpr const char* DYNAMO_ETCD_ENDPOINTS = "http://etcd:2379/";
...
+// stageLatency + numPipelineStages are the two fundamental knobs; the decode
+// cadence is derived from them.
+constexpr unsigned MOCK_STAGE_LATENCY_US = 44;
+constexpr uint32_t MOCK_PIPELINE_STAGES = 64;
+// +/-20% per-token jitter. Desyncs the otherwise phase-locked slots so
+// identical-length requests don't finish in clumps and spike the TTFT tail.
+// Mean-preserving (throughput/TPOT unchanged). Set to 0 for a perfect
+// metronome. Tune upward if the tail is still too fat for the workload.
+// This is tested as almost the sweet spot.
+constexpr uint32_t MOCK_DECODE_JITTER_PCT = 20;
```
knovokmetTT: I need clarification what this does

tt-media-server/cpp_server/include/config/settings.hpp
```diff
@@ -354,21 +354,31 @@ std::string dynamoComponent();
...
-unsigned mockDecodeTokenLatencyUs();
+/** Time one token spends in a single pipeline stage, for the
+ * MockDecodeScheduler. defaults::MOCK_STAGE_LATENCY_US. */
+unsigned mockStageLatencyUs();
+
+/** Transformer pipeline depth modeled by the MockDecodeScheduler.
+ * defaults::MOCK_PIPELINE_STAGES. */
+uint32_t mockPipelineStages();
+
+/** Per-decode-token timing jitter for the MockDecodeScheduler, as a percentage
```
knovokmetTT: same here for comments

nanicicTT: Well, let's see if we need them, 
https://github.com/tenstorrent/tt-inference-server/pull/4484#discussion_r3513591370

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: Give each token a small random timing wobble. Over 1024 tokens those wobbles accumulate like a random walk, so finish times drift apart - this is important, not to be in a lockstep. Instead of 64 slots freeing at the same instant, they free in a so called steady trickle, roughly one every maybe 45ms. A waiting request now finds an opening within ~45ms instead of waiting for the next clump. The long waits disappear, since I had too big P99 since this is like a metronome mechanism, and it needs a bit of jitter, like real decode scheduler does, not everything like a simulated lockstep mechanism.

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: the other ones (variables) have comments. 
Do we need them or not? If yes, yes for all, if not, not for all I guess.

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: In blaze utils you also have
```cpp
case MOCK_PIPELINE:
  return pl::PrefillMockConfig{ .auto_layer_acks = true,
                                .chunk_latency = std::chrono::milliseconds(100) };
```

So it is added on purpose to mimic the similar numbers as mock pipeline.

nanicicTT: Screenshot 2026-07-02 at 15 56
your commit

tt-media-server/cpp_server/include/config/defaults.hpp
nanicicTT: let me explain it a bit better. 

wihtout jitter
> 64 requests start together
all use exactly 1024 tokens
all tokens take exactly 2816us
therefore all 64 finish together ~2.9s later

so slot freeing would look like
> t=0.0s      64 slots taken
t=2.9s      64 slots free at once
t=5.8s      64 slots free at once
t=8.7s      64 slots free at once

and that is the clump I talked about. 
If another 64 requests are waiting, they do not see slots becoming available gradually. They wait until the next big synchronized release. That created bad P99 TTFT for me earlier, more than 1000.

With jitter, each token delay becomes approximately 2816us +/- 20%. 
So maybe one token waits 2500us, another 3000us, another 2700us. The mean is still 2816us, but every slot slowly drifts out of sync.
So the trickle is:
> slot 12 frees
then slot 3 frees
then slot 41 frees
then slot 8 frees
...


So, without jitter we had
> waiting request arrives just after the clump
waits close to one full generation
P99 TTFT gets huge

with jitter:
> waiting request arrives
some slot is likely to free soon
TTFT tail drops

tt-media-server/cpp_server/include/config/settings.hpp
```diff
@@ -350,21 +350,26 @@ std::string dynamoComponent();
...
+/** Transformer pipeline depth modeled by the MockDecodeScheduler.
+ * defaults::MOCK_PIPELINE_STAGES. */
+uint32_t mockPipelineStages();
+
+/** Round-robin quantum (prompt tokens) for prefill in the MockDecodeScheduler:
+ * a slot injects up to this many tokens before rotating to the next prefilling
+ * slot. Mirrors the real decode scheduler's DEFAULT_CHUNK_SIZE. Affects only
+ * how first-token latency spreads across concurrent slots, not the shared cap.
+ * Default: defaults::MOCK_PREFILL_RR_TOKENS. */
+uint32_t mockPrefillRoundRobinTokens();
```
knovokmetTT: the name is misleading. This refers to prefill by decode RR chunks. Also, can we delete AI slop comments :)?

knovokmetTT (approved):

tt-media-server/cpp_server/include/config/settings.hpp
nanicicTT: Leftover comment, I deleted most of them, this one must've slipped away.

nanicicTT: Done.

tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -33,7 +33,8 @@ constexpr size_t MAX_QUEUE_SIZE = 1000;
 constexpr const char* SCHEDULING_POLICY =
     "prefill_first";  // "prefill_first" or "max_occupancy"
 constexpr const char* LLM_DEVICE_BACKEND =
-    "mock_pipeline";  // "mock", "mock_pipeline", "pipeline_manager", "llama"
+    "mock_pipeline";  // "mock", "mock_pipeline", "mock_scheduler",
```
dmadicTT: Let's switch to mock_scheduler as default?

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
```diff
@@ -322,79 +345,231 @@ class MockDecodeScheduler final : public IDecodeScheduler {
...
+    bool hasPrefillToInject() const {
+      return prefillInjected < prefillTokensToBeInjected;
+    }
+
+    bool hasDecodeToInject() const {
+      return decodeReady && outputsEmitted < maxTokens;
+    }
+
+    void injectDecodeToken() {
+      decodeReady = false;  // token in flight; can't inject next until exit
```
dmadicTT, line 369: If we would name it `inFlight` would it be also correct?

dmadicTT (approved):

tt-media-server/cpp_server/include/runtime/runners/blaze_runner/mock_scheduler.hpp
nanicicTT: inFlight means the counter of inFlight tokens, this is a boolean meaning is decode ready.

PR #4488, opened 2026-07-02
add: SUPER_CLUSTER remote-console device type

ipastalTT:
## Summary
Introduces a new `SUPER_CLUSTER` device type for remote-console model
serving, so models served via an external console can be described and
routed distinctly from local GPU/TT devices.
- Add `SUPER_CLUSTER` to the device enum in `workflows/workflow_types.py`
- Handle the new device in `utils/device_utils.py`
- Recognize `SUPER_CLUSTER` in benchmark report filename parsing
  (`benchmarking/summary_report.py`)
## Why
Foundational enum used by the remote-console model specs and the v2 eval/
benchmark routing that follow. Merge this first.
## Test plan
- [x] Existing workflow/benchmark tests remain green
## Notes
First PR in a 5-part stack. No dependencies.

### DIFF
--- a/benchmarking/summary_report.py
+++ b/benchmarking/summary_report.py
@@ -108,7 +108,7 @@ def extract_params_from_filename(filename: str) -> Dict[str, Any]:
     image_pattern = r"""
         ^benchmark_
         (?P<model>.+?)                            # Model name (non-greedy, allows everything)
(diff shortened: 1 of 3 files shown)


fivanovicTT (approved):
Approved, but `benchmarking/summary_report.py` is not used at all in v2 
cc: @ddjukicTT

ddjukicTT:
> Approved, but `benchmarking/summary_report.py` is not used at all in v2 cc: @ddjukicTT

That's right

bgoelTT (approved):

Issue #4499, opened 2026-07-03
Investigate 20ms Dynamo <-> IS Overhead


ljovanovicTT:
We are observing 20ms overhead in Dynamo <-> IS serialization for 55K ISL prompts.

PR #4500, opened 2026-07-03
Offload Json Parsing off the IO Loop

ljovanovicTT:
## Issue
[#4499 Investigate 20ms Dynamo <-> IS Overhead](https://github.com/tenstorrent/tt-inference-server/issues/4499)

## Problem
When testing the disaggregated setup with large input sequences (55K ISL) under concurrent load, we observed that the Dynamo <-> Inference Server path introduced approximately 20 ms of overhead per request.

This became noticeable when four requests arrived as a burst at Dynamo (generated by vllm bench with request_rate=inf). Although Dynamo received them simultaneously, the Inference Server processed them roughly 20 ms apart.

We identified two contributing factors:

- Unintentional request serialization. Dynamo and the Inference Server communicate over a single TCP connection. Requests were accepted on a single I/O thread, which also synchronously parsed each request before reading the next one. As a result, requests arriving in a burst were unintentionally serialized.
- Slow JSON parsing. The jsoncpp parser introduced significant overhead for this workload. Replacing it with manual parsing of the token_ids field reduced parsing time from approximately 11 ms to 0.2 ms per request.

To address these issues, two changes were implemented:

- Fix 1: Move request parsing off the I/O thread so incoming requests can continue to be accepted while parsing happens asynchronously.
- Fix 2: Replace jsoncpp parsing of token_ids with a lightweight manual parser optimized for this use case.

## Testing
Testing was done on host, and here are observed performance optimizations:

On a mock benchmark with a burst of 32 concurrent requests, each containing 55K input tokens, these changes improved throughput from 49.7 req/s to 83.1 req/s, reduced P99 TTFT from 615 ms to 352 ms, and eliminated request failures
```
┌──────────────────────────────────────┬───────┬───────────┬──────────┐
│                Config                │ req/s │ TTFT mean │ TTFT P99 │
├──────────────────────────────────────┼───────┼───────────┼──────────┤
│ Original (serial jsoncpp on io loop) │ 49.7  │ 402 ms    │ 615 ms   │
├──────────────────────────────────────┼───────┼───────────┼──────────┤
│ + parse offloaded to pool            │ 80.0  │ 331 ms    │ 386 ms   │
├──────────────────────────────────────┼───────┼───────────┼──────────┤
│ + hand-rolled parser                 │ 83.1  │ 309 ms    │ 352 ms   │
└──────────────────────────────────────┴───────┴───────────┴──────────┘
```

### DIFF
--- a/tt-media-server/cpp_server/src/dynamo/dynamo_protocol.cpp
+++ b/tt-media-server/cpp_server/src/dynamo/dynamo_protocol.cpp
@@ -15,11 +15,13 @@
 #include <trantor/utils/MsgBuffer.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <cerrno>
 #include <chrono>
 #include <cstdlib>
 #include <cstring>
 #include <sstream>
+#include <string_view>
 #include <thread>
 #include <utility>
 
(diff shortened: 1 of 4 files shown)


tt-media-server/cpp_server/src/dynamo/dynamo_protocol.cpp
```diff
@@ -378,23 +505,38 @@ void DynamoServer::onMessage(const trantor::TcpConnectionPtr& conn,
...
-  // The handler creates the async call-home writer and returns without
-  // blocking, so dispatch runs inline on the io loop — no per-request thread.
-  handler_(genReq, connInfo);
+  // Offload the body parse + dispatch off the io loop. The handler creates the
+  // async call-home writer (bound to one of the endpoint's loop-pool loops)
+  // and returns without blocking, and the pipeline is already invoked
+  // concurrently across io loops in the multi-connection case, so running it
+  // from a pool thread is safe. This lets a burst of large prompts parse in
+  // parallel instead of serializing on one connection's io loop.
+  dispatch_pool_->submit([this, body = std::move(twoPart.body),
```
nanicicTT: This says: run this function later on some worker thread. This is fine, we need to offload parsing.

Good use would be to just parse, right? Like:

```cpp
parse_generate_request(body);
```

since we need to parse in another thread. But the not-so-good use case is:

```cpp
handler_(genReq, connInfo);
```

because `handler_` does more than parsing, like on the line 538 below.

`handler_` calls `pipeline->runStreamingRequest(<some params inside>)`, right?

The cache-hit part is the problem because it does **not** use `queueInLoop(<something here>)`.

`queueInLoop(<something here>)` means: do not run this code here, schedule it on the event loop’s own thread.

So, `process_request` runs on the Trantor IO loop -> JSON parse is moved to dispatch pool -> `handler_` also runs on dispatch pool -> `resolveSession` cache hit calls `onResolved` immediately -> now we are on a dispatch-pool thread and not on the Trantor IO loop.

But some object or operation is expected to be used from one specific thread. The event loop owns connection/request-flow work. If we want to touch connection-related flow, we should schedule it on that event-loop thread.

Imagine this scenario:

```text
Request A:
no prefix cache
goes through cache miss path
continues on Trantor loop

Request B:
prefix cache hit
continues on dispatch-pool thread
```

Current cache-hit code does this:

```cpp
onResolved(info);
```

That means: run callback immediately on the current thread.

Before this branch, the current thread was already the `Trantor` loop and it was fine. After this branch, the current thread may be a dispatch-pool worker, which violates the expected execution model.

We should queue it onto the event loop, not call it right now and here, but later on the correct loop thread:

```cpp
loop->queueInLoop([...] {
  onResolved(info);
});
```

So, I guess `resolveSession()` is the best place for this to be fixed.

nanicicTT: not sure what kind of tests would cover this to show before and after, since this might be the type of bug that is not always visible and reproducible. So called *thread-affinity* bug.

ljovanovicTT: Great catch!
With the latest commit we use `queueInLoop` in `resolveSession`

nanicicTT (approved):

idjuricTT (approved):

knovokmetTT (approved):

PR #4505, opened 2026-07-03
Connect Kafka trigger to Mooncake MW data plan (#4455)

### LINKED ISSUE #4455: Mooncake KV migration worker: connect the Kafka trigger to the real data plane

visnjakrsmanovicTT:


tt-media-server/cpp_server/include/messaging/migration_message.hpp
```diff
@@ -15,9 +15,12 @@ struct MigrationRequestMessage {
   uint64_t migration_id;
   uint32_t src_slot;
   uint32_t dst_slot;
-  uint32_t layer_id;
-  uint32_t position_start;
-  uint32_t position_end;
+  uint32_t layer_begin;
```
dmadicTT: I think we should be explicit what is inclusive/exclusive
For example
```
uint32_t layer_begin_inclusive;
uint32_t layer_end_exclusive;
```

tt-media-server/cpp_server/scripts/migration_cli.py
```diff
@@ -306,9 +346,12 @@ def _add_produce_args(p: argparse.ArgumentParser) -> None:
     )
     p.add_argument("--src-slot", type=int, default=0)
     p.add_argument("--dst-slot", type=int, default=1)
-    p.add_argument("--layer-id", type=int, default=0)
-    p.add_argument("--position-start", type=int, default=0)
-    p.add_argument("--position-end", type=int, default=16)
+    p.add_argument("--layer-begin", type=int, default=0)
+    p.add_argument("--layer-end", type=int, default=1)
+    p.add_argument("--src-pos-begin", type=int, default=0)
```
dmadicTT: Same here, would be nice to have inclusive/exclusive marks

tt-media-server/cpp_server/include/messaging/migration_message.hpp
```diff
@@ -15,9 +15,12 @@ struct MigrationRequestMessage {
...
+  uint32_t layer_end;
```
idjuricTT: Mini note: This overlaps with what @fivanovicTT has in PR

idjuricTT (approved):

tt-media-server/cpp_server/scripts/migration_cli.py
```diff
@@ -38,6 +38,38 @@
...
+            os.execv(python, [python, *sys.argv])
+    sys.stderr.write(
+        "ERROR: confluent_kafka is not importable. Create the venv first:\n"
+        "  bash scripts/setup-migration-cli.sh\n"
+        "(or set MIGRATION_CLI_VENV=<venv-with-confluent-kafka>)\n"
+    )
+    sys.exit(2)
+
+
+_reexec_into_venv_if_needed()
```
dmadicTT: The script is simple python program - create venv, install dependencies (1 dependency for now) and run it.
I would not have force separate venv and environment variable only for this script, `MIGRATION_CLI_VENV`

tt-media-server/cpp_server/scripts/migration_cli.py
```diff
@@ -79,22 +111,30 @@ def _wait_for_broker(admin: AdminClient, timeout_s: float = 30.0) -> None:
 
 def _build_request(args: argparse.Namespace, index: int) -> dict[str, Any]:
     if args.random:
+        layer = random.randint(0, 95)
```
dmadicTT: we often have 64 layers, so sounds like a better ceiling for random

tt-media-server/cpp_server/include/messaging/migration_message.hpp
visnjakrsmanovicTT: Yes, we defined payload beforehand.

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -0,0 +1,394 @@
...
+  Role role = Role::PREFILL;
+  std::string metadata_uri;     // Mooncake metadata service (or P2PHANDSHAKE).
+  std::string name;             // this worker's Mooncake server/segment name.
+  std::string host;             // this node's host tag in the table.
+  std::string device_map_path;  // FabricNode->UMD chip map (both roles); empty
+                                // => placeholder (device & 0xFFFF).
+
+  // prefill:
+  std::string prefill_table_path;
+  std::string decode_table_path;
```
dmadicTT, line 95: We still plan to exchange this through discovery service, right?

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -0,0 +1,394 @@
...
+  if (!engine) return 1;
+
+  auto prefill = loadKvTableFile(cfg.prefill_table_path);
+  auto decode = loadKvTableFile(cfg.decode_table_path);
+  if (!prefill || !decode) {
+    TT_LOG_ERROR("[worker] failed to load prefill/decode table");
+    return 1;
+  }
+
+  const DeviceMap deviceMap = loadDeviceMapFile(cfg.device_map_path);
```
dmadicTT, line 292: Where do we get this file from?

tt-media-server/cpp_server/include/messaging/migration_message.hpp
nanicicTT: Isn't this always the case in SW? `[)` from begin() to end()? :)

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -0,0 +1,394 @@
...
+// uses `deviceMap` (the interface the engine fills; the file-based map is the
+// same contract) and falls back to the encodeDevice low bits (placeholder, as
+// in the e2e harness) for any device the map doesn't cover. The placeholder is
+// only correct when the host's table is single-mesh; a multi-mesh host (its KV
+// aliases chip ids across meshes) needs the map or its replicas collide.
+std::unique_ptr<MultiDeviceUmd> buildDeviceIo(const IKvTable& table,
+                                              const std::string& host,
+                                              const DeviceMap& deviceMap) {
+  auto umd = std::make_unique<MultiDeviceUmd>();
+  for (const auto& loc : allHostLocations(table, host)) {
```
dmadicTT, line 254: I don't see it as an issue right now, but just to leave what I've learned here.
I expect `allHostLocations` to cause slow startup, as inside it will go over 3-4 million KV chunks over few loops.

For example, prefill chunk table
32 slots × 61 chunk × 1920 chunks per slot per layer = 3_747_840

tt-media-server/cpp_server/include/messaging/migration_message.hpp
madicd: Valid point, let's use it as convention all over the code and we are good

madicd (approved):

tt-media-server/cpp_server/include/messaging/migration_message.hpp
visnjakrsmanovicTT: yes, here is [) indeed (we can put comment to be clear)

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
visnjakrsmanovicTT: Eventually real kv tables should be loaded from metal, not from outside, and exchanged in discovery phase. In this case paths are there for testing purposes.

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
visnjakrsmanovicTT: ATM it's synthesized for testing purposes, for the production it should be obtained from the device runners which own logical -> physical indexes.

tt-media-server/cpp_server/scripts/migration_cli.py
visnjakrsmanovicTT: You are right, this is a leftover from testing experiments on a clean machine, I'm reverting it.

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
visnjakrsmanovicTT: That's true, but it's done only once at statup. We can optimize later if it turns out to be a problem.

tt-media-server/cpp_server/src/transport/mooncake_migration_executor.cpp
```diff
@@ -0,0 +1,118 @@
...
+    } catch (...) {
+      TT_LOG_ERROR(
+          "[MooncakeMigrationExecutor] migration_id={} threw unknown "
+          "exception; reporting FAILED",
+          job.id);
+    }
+
+    if (job.onDone) {
+      job.onDone(status);
+    }
```
nanicicTT, line 114: I'll put a comment here, but it touches upon multiple pieces of existing and new code. 
We have `MooncakeMigrationExecutor::mutex_` and it's used when `execute()` pushes a `Job`, and when `workerLoop()` pops one. After the job is popped, `workerLoop()` releases `mutex_`.
Then the migration is being run. And then it calls `job.onDone(status);`. 
So the flow is

KvMigrationWorker receives Kafka request
  -> calls `executor->execute(..., callback)`

`MooncakeMigrationExecutor::execute()`
  -> stores callback as `job.onDone` in queue

`MooncakeMigrationExecutor::workerLoop()`
  -> locks mutex_
  -> pops job
  -> unlocks mutex_
  -> runs migration
  -> calls job.onDone(status)

`job.onDone(status)`
  -> calls `KvMigrationWorker::publishAck(...)`
  -> publishAck locks `ackMutex`
  -> sends Kafka ACK

So `ackMutex`  affects what happens when the executor calls the callback, right?

And `job.onDone` is the bridge between the executor and the worker.
```cpp
[this, migrationId](tt::services::MigrationStatus status) {
  publishAck(migrationId, status);
}
```
and this lambda captures `this`: meaning call back into this exact KvMigrationWorker object later.

There could be a bad shutdown case:

```text
worker destructor starts
executor still has in-flight job
ackMutex gets destroyed before executor
executor destructor waits for job to finish
job finishes and calls job.onDone
job.onDone calls publishAck
publishAck tries to lock destroyed ackMutex
undefined behavior
```

and ackMutex could be destroyed before executor based on the `kv_migration_worker.hpp`, elements are created in one order, top-down, and destroyed in reverse order, so executor could be destroyed after `ackMutex`, right? 

The problem is lifetime -> `job.onDone` reaches from the executor thread back into `KvMigrationWorker`, and one of the worker fields it needs, `ackMutex`, may already be gone.

Failure scenario:

```text
1. Kafka request arrives.
2. Worker submits migration to MooncakeMigrationExecutor.
3. Migration is still running.
4. Process receives SIGTERM.
5. KvMigrationWorker destructor calls stop().
6. Object member destruction begins.
7. ackMutex is destroyed.
8. executor destructor waits for migration thread.
9. Migration finishes.
10. executor calls onDone.
11. onDone calls publishAck().
12. publishAck tries to lock destroyed ackMutex.
```
So, a SIGTERM during an active migration can finish the migration during executor destruction, then call back into a half-destroyed worker.

Maybe the fix might be to make the worker explicitly stop and drain the executor while the worker object is still fully valid. For example, destroy/reset executor inside `KvMigrationWorker::~KvMigrationWorker()` after `stop()` and before member teardown reaches `ackMutex`.

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -0,0 +1,394 @@
...
+    return 1;
+  }
+
+  const DeviceMap deviceMap = loadDeviceMapFile(cfg.device_map_path);
+  auto device = buildDeviceIo(*prefill->table, cfg.host, deviceMap);
+
+  // Open one control channel per decode host (static resolution for now; the
+  // map is the interface a discovery service fills later).
+  KvControlChannelConnector connector(cfg.peers, makeClientTransport);
+  if (!connector.connect()) {
```
nanicicTT: the name `connect()` as far as I can see is misleading in this path. It means that it created transport objects, not that the TCP channels are connected. This (actual TCP connect) happens later in the background `clientLoop()` after `start()`. 


The factory calls `makeClientTransport()` that does:

```cpp
auto t = std::make_shared<TcpSocketTransport>();
if (!t->initializeAsClient(ep.host, ep.port)) return nullptr;
t->start();
return t;
```

and this `initializeAsClient()` inside does not connect to the server. It only stores host and port, then returns `true`.


If Kafka consumption begins before the decode control socket is connected, the first migration can fail immediately.

Maybe, do not declare the prefill worker ready and do not start Kafka consumption until required channels are actually connected. For example, poll `channel.isConnected()` for all required decode peers with a startup timeout.

fivanovicTT (approved):

tt-media-server/cpp_server/src/utils/logger.cpp
```diff
@@ -12,99 +12,100 @@
...
+  // process). std::call_once serializes init; later calls -- including with a
+  // different instanceTag -- are no-ops, matching the prior initialize-once
+  // contract (the first tag wins).
+  static std::once_flag onceFlag;
+  std::call_once(onceFlag, [&] {
+    // Set log level from environment variable
+    const char* logLevelEnv = std::getenv("TT_LOG_LEVEL");
+    if (logLevelEnv) {
+      level = parseLogLevel(logLevelEnv);
     }
```
nanicicTT, line 37: Also, I guess, this could be within a function with static const variable (inside the function) initialized and returned, so whoever calls this functions gets this value for free, only initialized once on the first function call, the rest just return cached value. 
This is thread-safe since `static` keyword is thread-safe for init since C++11.

nanicicTT (approved):

tt-media-server/cpp_server/src/utils/logger.cpp
visnjakrsmanovicTT: That should be already handled with `std::call_once`, I would leave it as is.

tt-media-server/cpp_server/src/transport/mooncake_migration_executor.cpp
visnjakrsmanovicTT: Done

PR #4511, opened 2026-07-03
Complete yolox_nano for release (+ uplift forge models)

ddilbazTT:
This PR includes:
- Uplift tt-forge-models (FORGE_MODELS_SHA → [2e59f9b5](https://github.com/tenstorrent/tt-forge-models/commit/2e59f9b5f72a6bfacace309d4a809300f4ad4eb7)): pulls the YOLOX loader change that exposes detection boxes + numeric scores from output_postprocess and fixes a double-decode (decode_in_inference=False) that was exp-overflowing box coordinates to inf/NaN. Required because the CNN serving path previously returned only class labels, so real detection accuracy couldn't be measured.
- yolox_nano eval completion (`utils/media_clients/cnn_client.py`):
    - Accuracy — real COCO mAP via _run_yolox_coco_eval: streams `detection-datasets/coco` (val), builds an in-memory pycocotools GT (xyxy→xywh, shared 0–79 category ids), scores the served model's detections with COCOeval (bbox); mAP = stats[0].
    - accuracy_check: PASS iff mAP ≥ published (25.8) × (1 − tolerance=0.05) (≥ 24.51%); replaces the ImageNet-classification stand-in.
    - `tput_user = 1/latency`: emitted from the eval so the enforced functional/complete tput_check reads real throughput instead of defaulting to 0.
    - Fail-strict: FAIL (not NA) when nothing can be scored (server returns no boxes / empty GT / COCOeval error), so a serving regression can't pass silently; malformed/None boxes are skipped per-sample.
    - Server passthrough: forge_runner forwards the new boxes/scores in the search-image response (backward-compatible for other CNNs).
    - Deps: pycocotools added to the evals venv (requirements/evals-run-script.txt).

For reference: yolox_nano on latest main: https://github.com/tenstorrent/tt-shield/actions/runs/28817751590 

Tests Run:
(using docker image ghcr.io/tenstorrent/tt-shield/tt-media-inference-server-forge:32ba899570c08cc95b3e9d78dc9b36aae0d4b8d9_22d914c_85476296597)
- efficientnet: https://github.com/tenstorrent/tt-shield/actions/runs/28824701980 
- mobilenetv2: https://github.com/tenstorrent/tt-shield/actions/runs/28824788035
- resnet-50: https://github.com/tenstorrent/tt-shield/actions/runs/28824834751
- segformer: https://github.com/tenstorrent/tt-shield/actions/runs/28824877071
- vit: https://github.com/tenstorrent/tt-shield/actions/runs/28824909046
- vovnet: https://github.com/tenstorrent/tt-shield/actions/runs/28824945901
- yolox_nano: https://github.com/tenstorrent/tt-shield/actions/runs/28822150589

### DIFF
--- a/utils/media_clients/cnn_client.py
+++ b/utils/media_clients/cnn_client.py
@@ -57,11 +57,10 @@ class CnnClientStrategy(BaseMediaStrategy):
         try:
             runner_in_use = self.require_health()
             eval_result = None
-            yolox_result = None
             if runner_in_use == CNN_MOBILENETV2_RUNNER:
                 eval_result = self._run_mobilenetv2_eval()
             elif runner_in_use == CNN_YOLOX_NANO_RUNNER:
-                yolox_result = self._run_yolox_coco_eval()
+                eval_result = self._run_yolox_coco_eval()
             else:
                 status_list = self._run_image_analysis_benchmark()
         except Exception as e:
@@ -90,23 +89,28 @@ class CnnClientStrategy(BaseMediaStrategy):
             benchmark_data["correct"] = eval_result["correct"]
             benchmark_data["total"] = eval_result["total"]
             benchmark_data["mismatches_count"] = eval_result["mismatches_count"]
-        elif runner_in_use == CNN_YOLOX_NANO_RUNNER and yolox_result:
+        elif runner_in_use == CNN_YOLOX_NANO_RUNNER and eval_result:
             logger.info("Adding YOLOX COCO mAP eval results")
-            benchmark_data["accuracy_check"] = yolox_result.get(
+            benchmark_data["accuracy_check"] = eval_result.get(
                 "accuracy_status", ReportCheckTypes.NA
             )
             # Report mAP (as a percentage) as the eval score against the
             # published COCO mAP (25.8 for yolox_nano).
-            benchmark_data["score"] = yolox_result["mAP_percent"]
-            benchmark_data["mAP"] = yolox_result["mAP_percent"]
-            benchmark_data["num_images"] = yolox_result["num_images"]
+            benchmark_data["score"] = eval_result["mAP_percent"]
+            benchmark_data["mAP"] = eval_result["mAP_percent"]
+            benchmark_data["num_images"] = eval_result["num_images"]
             benchmark_data["published_score"] = self.all_params.tasks[
                 0
             ].score.published_score
             benchmark_data["published_score_ref"] = self.all_params.tasks[
                 0
             ].score.published_score_ref
-            latency_for_perf_check = yolox_result.get("mean_latency")
+            latency_for_perf_check = eval_result.get("mean_latency")
+
+            # For single-concurrency CNN inference, tput_user (images/sec/user) = 1/latency.
+            if latency_for_perf_check:
+                benchmark_data["tput_user"] = 1.0 / latency_for_perf_check
+
         else:
             logger.info("No eval results from eval spec test to add to benchmark data")
             latency_value = self._calculate_latency(status_list)
@@ -120,10 +124,6 @@ class CnnClientStrategy(BaseMediaStrategy):
             benchmark_data["published_score_ref"] = self.all_params.tasks[
                 0
             ].score.published_score_ref
-
-        # For single-concurrency CNN inference, tput_user (images/sec/user) = 1/latency.
-        if latency_for_perf_check:
-            benchmark_data["tput_user"] = 1.0 / latency_for_perf_check
 
         benchmark_data["performance_check"] = self._calculate_performance_check(
             latency_value=latency_for_perf_check,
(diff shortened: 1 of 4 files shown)


utils/media_clients/cnn_client.py
```diff
@@ -43,8 +57,11 @@ def run_eval(self) -> None:
         try:
             runner_in_use = self.require_health()
             eval_result = None
+            yolox_result = None
```
mmilosevicTT: Why do we use separate `yolox_result` fo this? Can't we reuse `eval_result`?

utils/media_clients/cnn_client.py
```diff
@@ -378,3 +416,262 @@ def _run_mobilenetv2_eval(self) -> dict:
...
+    def _pil_to_jpeg_bytes(image) -> bytes:
+        """Encode a PIL image to JPEG bytes (RGB)."""
+        if image.mode != "RGB":
+            image = image.convert("RGB")
+        buf = io.BytesIO()
+        image.save(buf, format="JPEG")
+        return buf.getvalue()
+
+    @staticmethod
+    def _bbox_xyxy_to_xywh(bbox) -> Optional[list]:
```
mmilosevicTT: nit: This naming is a bit weird. Can we rename it something like `corner_to_hight_width_format`?

utils/media_clients/cnn_client.py
```diff
@@ -378,3 +416,262 @@ def _run_mobilenetv2_eval(self) -> dict:
...
+        Robust-but-strict: a bad per-sample response is logged and skipped
+        rather than crashing the evals workflow, but if nothing could be scored
+        at all (server returned no boxes, empty GT, or COCOeval error) the
+        result is FAIL — a serving/plumbing regression must not pass silently.
+
+        Returns ``accuracy_status`` / ``mAP_percent`` / ``num_images`` /
+        ``mean_latency``.
+        """
+        # Lazy imports: heavy deps only needed for this eval.
+        from datasets import load_dataset
```
mmilosevicTT: Just out of curiosity, `load_dataset` is only needed for yolo?

mmilosevicTT (approved):
LGTM

I left few comments. Also, did you guys confirm licences are ok for yolox model? I remember we had some problems where we had to stop using them in tt-xla.

kmabeeTT (approved):
Thanks! Yes @mmilosevicTT license is good for yolox, the issue last year was for ultralytics models (did not get resolved) but it doesn't apply to yolox.

utils/media_clients/cnn_client.py
fivanovicTT: I agree here with @mmilosevicTT, let's use already existing data structure `eval_result`

utils/media_clients/cnn_client.py
```diff
@@ -73,6 +90,23 @@ def run_eval(self) -> None:
             benchmark_data["correct"] = eval_result["correct"]
             benchmark_data["total"] = eval_result["total"]
             benchmark_data["mismatches_count"] = eval_result["mismatches_count"]
+        elif runner_in_use == CNN_YOLOX_NANO_RUNNER and yolox_result:
```
fivanovicTT: Same here, let's please use `eval_result`

utils/media_clients/cnn_client.py
```diff
@@ -87,6 +121,10 @@ def run_eval(self) -> None:
                 0
             ].score.published_score_ref
 
+        # For single-concurrency CNN inference, tput_user (images/sec/user) = 1/latency.
+        if latency_for_perf_check:
+            benchmark_data["tput_user"] = 1.0 / latency_for_perf_check
+
```
fivanovicTT: This can be part of `elif runner_in_use == CNN_YOLOX_NANO_RUNNER` else branch

utils/media_clients/cnn_client.py
```diff
@@ -378,3 +416,262 @@ def _run_mobilenetv2_eval(self) -> dict:
...
+            return None
+        if any(c is None for c in bbox):
+            return None
+        try:
+            x1, y1, x2, y2 = (float(c) for c in bbox)
+        except (TypeError, ValueError):
+            return None
+        return [x1, y1, x2 - x1, y2 - y1]
+
+    def _run_yolox_coco_eval(self) -> dict:
```
fivanovicTT, line 565: I think this method should be broken down into 3 separate methods:

1. `_run_yolox_coco_eval` - core method/orchestrator for calculating evals (it calls the other 2)
2. `_collect_yolox_coco_predictions ` - Stream a COCO val subset, run inference per image, and collect GT + server predictions in COCO format
3. `_score_yolox_coco_predictions ` - Score predictions

fivanovicTT (changes requested):
Looking good, but let's address these few things

utils/media_clients/cnn_client.py
ddilbazTT: Not really, for example, mobile net uses vision evals test: `tt-inference-server/server_tests/test_cases/vision_evals_test.py` yolox is the first object detection model so it's not a vision evals test. But I can move the code to under a new eval test file. It would be preferable not to block this PR though as I am trying to get this model ready for release.

fivanovicTT (approved):

Issue #4416, opened 2026-06-29
Cleanup SessionManager


lmuravljovTT:
Inject PrefixCacheRouter at constructor, and remove prefixHash related methods which are no longer needed

PR #4526, opened 2026-07-06
Encapsulate prefix cache routing logic in prefixCacheRouter

lmuravljovTT:
# Encapsulate prefix cache routing logic in PrefixCacheRouter

## Summary

Introduces a unified `getSlot()` API that encapsulates all prefix cache routing implementation details (block splitting, hashing, multi-layer lookups) inside `PrefixCacheRouter`. This simplifies the caller interface from ~340 lines of routing logic in `llm_pipeline.cpp` to a single method call.

## Motivation

Previously, prefix caching implementation details leaked into `LLMPipeline::resolveSession()`:
- Block splitting and hashing via `computeRoutingInfo()`
- Layer 1a: Response-id lookup logic
- Layer 1b: Prefix-hash lookup logic  
- Layer 2: New session allocation with slot copy decisions

This made the code harder to maintain and understand. The new design follows the principle that `PrefixCacheRouter` should own all routing logic, while `SessionManager` provides session/slot primitives.

### New API

**`PrefixCacheRouter::getSlot()`** - unified entry point:
```cpp
void getSlot(std::span<const int> promptTokenIds, 
             GetSlotOptions opts,
             std::function<void(SlotAcquireResult)> onResolved,
             std::function<void(const std::string&)> onError);
```

**`GetSlotOptions`** - routing hints:
```cpp
struct GetSlotOptions {
  std::optional<std::string> previousResponseId;  // Response-id continuation
  std::optional<std::string> responseId;          // New response id to register
  std::function<void()> cancelFn;                 // Cancellation callback
};
```

**`SlotAcquireResult`** - unified result:
```cpp
struct SlotAcquireResult {
  std::string sessionId;
  uint32_t slotId;
  uint32_t matchedTokens;
  uint32_t accumulatedThinkTokens;
  bool isNewSession;
  std::vector<BlockHashInfo> blocks;  // For token accumulator init
};
```

### Architecture

```
LLMPipeline::resolveSession()
    │
    └──▶ SessionManager::getSlot(tokens, opts, loop, onResolved, onError)
              │
              └──▶ PrefixCacheRouter::getSlot(tokens, opts, onResolved, onError)
                        │
                        ├── computeBlockInfos(tokens)       // hashing
                        ├── tryAcquireByResponseId()        // Layer 1a
                        ├── tryAcquireByPrefixHash()        // Layer 1b  
                        └── callbacks.createSession()       // Layer 2
```

## Testing

- All existing unit tests pass (`session_manager_test`, `prefix_cache_router_test`)
- Build succeeds with no warnings

NOTE: SessionManager still has methods which just forward work to prefixCacheRouter. Those are currently only used in tests. We will focus on cleaning up SessionMAnager and tests in a separate [issue](https://github.com/tenstorrent/tt-inference-server/issues/4416)


idjuricTT (approved):

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
```diff
@@ -131,312 +71,78 @@ void LLMPipeline::resolveSession(
...
+        TT_LOG_INFO(
+            "[LLMPipeline] Slot acquired taskId={} sessionId={} slotId={} "
+            "matchedTokens={} thinkTokens={} isNew={}",
+            req->task_id, result.sessionId, result.slotId, result.matchedTokens,
+            result.accumulatedThinkTokens, result.isNewSession);
+
+        // Set up the request
+        req->sessionId = result.sessionId;
+        req->slotId = result.slotId;
+        req->session = mgr->getSession(result.sessionId);
```
dmadicTT, line 120: We are exposing the session here, just to look inside it later and init the accumulator.
Should this stay hidden inside the session manager?

dmadicTT (approved):

tt-media-server/cpp_server/src/services/llm_pipeline.cpp
lmuravljovTT: https://github.com/tenstorrent/tt-inference-server/commit/5129b5f1090306636cc57bbc65f29224b0ad5aa4#diff-5787536dbb7b363202a97cc7cf09ac7160e8acd6b71f14bf978728eb1b674126

The session has been added to LLMRequest on an unrelated issue fix. You might have a good point, but I'd leave it to a separate issue to look into

tt-media-server/cpp_server/src/services/prefix_cache_router.cpp
```diff
@@ -293,4 +304,150 @@ void PrefixCacheRouter::onSessionClosed(const std::string& sessionId,
...
+
+    TT_LOG_INFO(
+        "[PrefixCacheRouter::getSlot] Response-id MISS prevId={} → trying "
+        "prefix-hash",
+        *opts.previousResponseId);
+  }
+
+  // Step 3: Try prefix-hash lookup
+  std::optional<AcquireResult> acquired;
+  if (!blocks.empty()) {
```
ztorlakTT, line 371: Should we add `!useResponseId &&` to check as we had it before?

ztorlakTT (approved):

tt-media-server/cpp_server/src/services/prefix_cache_router.cpp
lmuravljovTT: Actually we want to fallback to prefix-hash lookup even if `useResponseId` is true, in case we didn't find an entry in responseIdIndex ( `tryAcquireByResponseId` returns null )

PR #4535, opened 2026-07-06
Clean up a slot that didn't reach the first block

idjuricTT:


ljovanovicTT (approved):
Looks great and this should work, but I'd prefer if we also set `kv_position_id` for every request. That way we're addressing the root cause rather than just working around the symptom.

ztorlakTT (approved):

PR #2360, opened 2026-03-10
Upgrade inference server to support LoRA

vpetrovicTT:

Issue #2350, opened 2026-03-09
SDXL LoRA Integration


vpetrovicTT:
Add LoRA adapter support to the SDXL inference pipeline, enabling users to apply LoRA weights to image generation requests on-the-fly.

The server needs to:
1. Accept LoRA parameters from API requests
2. Track which LoRA is currently active on the pipeline
3. Call load_lora_weights() / fuse_lora() / unload_lora_weights() at the right time
4. Prevent batching requests with different LoRA configs
5. Cache LoRA safetensors files locally...

PR #4553, opened 2026-07-08
Register FLUX.1-Kontext-dev media runner (edit / generate / resolution / LoRA)

yasuhiroitoTT:
## Summary
Add **`tt-flux.1-kontext-dev`** to `tt-media-server`, backed by the `tt_dit`
`Flux1KontextPipeline` (see the companion tt-metal PR). Scope: what's needed to
deploy FLUX.1-Kontext-dev.

## Changes
- `config/constants.py`: `SupportedModels` / `ModelNames` / `ModelRunners` +
  service & name maps + `ModelConfigs` (P300 `(1,2)` validated, P300X2, T3K).
- `tt_model_runners/dit_runners.py`: `TTFluxKontextRunner`. `run()` picks edit vs
  generate (reference image present or not), passes per-request width/height,
  `guidance_scale=3.5`, and fuses an active LoRA (`LORA_DIR/active_lora.json`) at
  pipeline build.
- `open_ai_api/image.py`: `kontext_router` serves `/edits` (image→image),
  `/generations` (text→image), and `/lora/{apply,clear,status}` (upload a FLUX
  LoRA → record active → `deep_reset` rebuild).
- `domain/image_generate_request.py`: optional `width`/`height`.
- `model_services/scheduler.py`: **fix `deep_restart_workers`** so it actually
  respawns workers — `await start_workers()` and recompute `workers_to_open`
  after reset (previously the coroutine was never awaited and `workers_to_open`
  stayed empty, so `deep_reset` never brought workers back).
- `workflows/model_specs/dev/image.yaml`: FLUX.1-Kontext-dev entry (P300, FUNCTIONAL).

## Validation
Deployed on Blackhole P300 (container): `/v1/images/edits`, `/v1/images/generations`,
and LoRA upload → rebuild → generate all verified end-to-end via the API.

The web console and its optional JP→EN translation add-on live in a separate repo.


tt-media-server/open_ai_api/image.py
```diff
@@ -94,11 +101,128 @@ async def edit_image(
...
 
+# ----------------------------- FLUX.1-Kontext-dev -----------------------------
+# Kontext exposes image editing (/edits), text-to-image (/generations), and LoRA
+# management (/lora/*). Optional JP->EN prompt translation is not part of the core
+# deployment — it ships as an add-on with the web console (see that repo).
+kontext_router = APIRouter()
+
+
+@kontext_router.post("/edits")
+async def kontext_edit_image(
```
idjuricTT: @yasuhiroitoTT we have imsge edit api already for SDXL.

Could you pls sync with @ztorlakTT how to accommodate this?

idjuricTT (commented):
All good except somr APIs were duplicated. @ztorlakTT should be able to help you to reuse existing setup

tt-media-server/open_ai_api/image.py
fivanovicTT, line 1: There is already similar thing done for SDXL by @vpetrovicTT.

PR: https://github.com/tenstorrent/tt-inference-server/pull/2360
Issue: https://github.com/tenstorrent/tt-inference-server/issues/2350 (nice diagram, take a look at this)

I think we can only extend the already existing implementations, and also what @idjuricTT mentioned not to duplicate endpoints.

SDXL is well tested (and on tt-console as well), and I think just extending upon it would be great. 
Ping @vpetrovicTT and me, we can discuss this offline (slack/zoom call)

yasuhiroitoTT:
Thanks @idjuricTT @fivanovicTT for the review — you're right that the Kontext router duplicated endpoints that already exist. I've reworked it to reuse the shared implementation instead of adding parallel copies. Also had a look at #2360 / #2350 — happy to sync offline with @ztorlakTT @vpetrovicTT to make sure this lands the way you'd like.

**What changed (de-duplication)**

The Kontext runner already detects edit-vs-generate from the request itself (a reference image is present → edit, otherwise text→image), so it never needed its own handlers. I removed the duplicated `kontext_edit_image` / `kontext_generate_image` and now reuse the shared routers:

- `/generations` → existing `generate_image_router`
- `/edits` → existing `edit_image_router`
- `/lora/*` → the only Kontext-specific surface (see below)

**Reusing `/edits` for a mask-free edit**

SDXL's `/edits` takes `ImageEditRequest`, which required a `mask`; Kontext is instruction-based and has no mask. Rather than add a second edit endpoint, I extended the existing request model — `ImageEditRequest.mask` is now `Optional[str] = None` — so the one `/edits` endpoint serves both: SDXL keeps sending a mask, Kontext sends none. This is the "extend, don't duplicate" direction you suggested and keeps the OpenAI-compatible `/edits` path.

One contract note to confirm with you: making `mask` optional means an SDXL request that *omits* the mask is no longer rejected at validation (422) and would instead fail later in `sdxl_edit_runner_trace._preprocess_mask`. Well-behaved SDXL clients always send a mask, so there's no behavior change for them, but if you'd rather keep the hard requirement for SDXL I can gate it (e.g. per-runner validation) instead.

**LoRA endpoints (`/lora/apply|clear|status`)**

These are genuinely new — upload a FLUX.1 LoRA, record it active, and rebuild the worker with the LoRA fused in. If there's an existing pattern for adapter management I should align to, point me at it and I'll fold into it.

**CodeQL (path-injection) alert**

Already resolved in the current commit — the upload is written to a fixed, server-controlled filename (`active_lora.safetensors`); the client-supplied name is kept only as a display label and never touches a filesystem path. There's a regression test for the traversal case (`test_lora_apply_sanitizes_traversal_name`). The alert's diff hunk is frozen from the earlier commit that used the client name in the path; it should clear on a re-scan — could someone with access re-run / dismiss it?

Tests: `tests/test_kontext_api.py` updated to exercise the shared handlers plus a new `test_edit_request_allows_no_mask` — all green.

tt-media-server/domain/image_edit_request.py
```diff
@@ -2,8 +2,12 @@
...
+from typing import Optional
+
 from domain.image_to_image_request import ImageToImageRequest
 
 
 class ImageEditRequest(ImageToImageRequest):
-    mask: str
+    # Optional so the shared /edits endpoint serves both mask-based edits (SDXL)
+    # and instruction-only edits with no mask (FLUX.1-Kontext).
+    mask: Optional[str] = None
```
vpetrovicTT: One concern on making mask optional on the shared ImageEditRequest: that model is used by both Kontext and tt-sdxl-edit. For SDXL, a request with mask=None now passes API validation and fails later in sdxl_edit_runner_trace._preprocess_mask (base64_to_pil_image(None) → exception → 500), instead of a clean 422.
Prefer a per-runner / per-model validator: require mask when model_runner == **tt-sdxl-edit**, allow None for **Kontext**. You already floated this in the PR — I’d treat that as the path forward before merge.

tt-media-server/domain/image_generate_request.py
vpetrovicTT: Do we need kontext added here too? Or is it left out on purpose?

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -245,6 +249,111 @@ def get_pipeline_device_params(self):
...
+            "l1_small_size": 32768,
+            "trace_region_size": self.settings.trace_region_size,
+        }
+
+    def _decode_image(self, image_b64: str, width: int, height: int) -> Image.Image:
+        # strip an optional data-URL prefix (e.g. "data:image/png;base64,")
+        if "," in image_b64 and image_b64.strip().startswith("data:"):
+            image_b64 = image_b64.split(",", 1)[1]
+        raw = base64.b64decode(image_b64)
+        return Image.open(io.BytesIO(raw)).convert("RGB").resize((width, height))
```
vpetrovicTT: Maybe we should reuse the existing path here. The rest of media-server already solved this in **ImageManager.base64_to_pil_image (restore padding, then decode)**. 
Also I think this would break if we don't restore the padding before decoding.

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -245,6 +249,111 @@ def get_pipeline_device_params(self):
...
+    def run(self, requests: list[ImageGenerateRequest]):
+        request = requests[0]
+        # Per-request resolution (falls back to the pipeline's configured size;
+        # the pipeline snaps to the nearest Kontext preferred bucket).
+        width = getattr(request, "width", None) or getattr(
+            self.pipeline, "_width", 1024
+        )
+        height = getattr(request, "height", None) or getattr(
+            self.pipeline, "_height", 1024
+        )
```
vpetrovicTT, line 328: I suppose  This will break if one parameter is explicitly provided and the other uses the default value of **1024** ?

tt-media-server/open_ai_api/image.py
```diff
@@ -94,11 +101,102 @@ async def edit_image(
...
+            shutil.copyfileobj(file.file, f)
+        with open(_LORA_STATE, "w") as f:
+            json.dump({"path": dest, "scale": float(scale), "name": display_name}, f)
+        await service.deep_reset()  # restart worker -> rebuild pipeline with LoRA
+        return JSONResponse(
+            content={
+                "status": "rebuilding",
+                "active": {"name": display_name, "scale": float(scale)},
+            }
+        )
```
vpetrovicTT, line 145: Maybe we should Validate LoRA before writing here?

yasuhiroitoTT:
Thanks @vpetrovicTT — addressed all three:

**mask on the shared `ImageEditRequest`** — agreed, went with the per-runner validator. `ImageEditRequest.mask` now has a `field_validator` (with `validate_default=True`, so it also fires when `mask` is omitted entirely) that rejects a missing mask at validation → clean **422** for mask-required runners (`tt-sdxl-edit`), while Kontext and any other runner keep accepting instruction-only (mask-free) edits. The required set is a small constant (`_MASK_REQUIRED_RUNNERS = {"tt-sdxl-edit"}`) so it's easy to extend. Added tests for both directions (sdxl-edit rejects no-mask, Kontext allows it).

**reference-image decode** — good catch. Switched the Kontext decode to `ImageManager.base64_to_pil_image`, which restores the stripped base64 padding (`+= "=" * (-len % 4)`) before decoding and does the RGB convert + resize. The bare `b64decode` I had would indeed raise on padding-trimmed strings, and this reuses the path the rest of media-server already uses.

**kontext in `_FLUX_RUNNERS`** — oversight, not intentional. Added `tt-flux.1-kontext-dev` so its minimum inference-steps floor matches the FLUX family (4) instead of the default 12. In practice Kontext-dev wants ~20–28 steps, but the floor should match the family, and it also unblocks few-step distilled Kontext LoRAs. Happy to revert to the default-12 floor if you'd rather keep the hard minimum for the base model.

All in the latest commit; `test_kontext_api` is green (11 passed).

vpetrovicTT:
Looks good to me @yasuhiroitoTT. Don't see any changes regarding LoRA upload validation/size limits or width/height pair validation. Happy to hear your thoughts on these :raised_hands:

sott0n:
@vpetrovicTT taking this over from @yasuhiroitoTT, and addressed your comments for validation:

- LoRA upload: extension, scale (0.0–2.0) and size (LORA_MAX_BYTES, default 512 MB, streamed and aborted with 413) are validated before any write; content is checked against the safetensors header; the file is staged as .part and swapped in with os.replace only after it validates, so a rejected upload can't clobber the active LoRA.
- width/height: a model_validator now enforces both-or-neither (422 for a lone axis), so a single value can no longer pair with the pipeline default and skew the aspect ratio.

sott0n:
CI Forge runner failed, but it is caused by Authorization error.

vpetrovicTT (approved):

PR #4554, opened 2026-07-08
Gemma4 Release CI fixes

arginugaTT:
works with tt-metal changes PR: https://github.com/tenstorrent/tt-metal/pull/49320

addresses issues from: https://github.com/tenstorrent/tt-metal/issues/49083 https://github.com/tenstorrent/tt-metal/issues/48838

Release CI: https://github.com/tenstorrent/tt-shield/actions/runs/28910589374

### DIFF
--- a/.github/workflows/models-ci-config.json
+++ b/.github/workflows/models-ci-config.json
@@ -509,19 +509,33 @@
       }
     },
     "gemma-4-31B-it": {
-      "inference_engine": "vLLM",
-      "ci": {
-        "nightly": {
-          "devices": [
-            "P300X2"
-          ]
+      "implementations": [
+        {
+          "inference_engine": "vLLM",
+          "ci": {
+            "nightly": {
+              "devices": [
+                "P300X2"
+              ]
+            },
+            "release": {
+              "devices": [
+                "P300X2"
+              ]
+            }
+          }
         },
-        "release": {
-          "devices": [
-            "P300X2"
-          ]
+        {
+          "inference_engine": "FORGE",
+          "ci": {
+            "nightly": {
+              "devices": [
+                "P300X2"
+              ]
+            }
+          }
         }
-      }
+      ]
     },
     "gemma-4-12B-it": {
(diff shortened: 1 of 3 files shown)


.github/workflows/models-ci-config.json
```diff
@@ -528,16 +523,6 @@
         }
       }
     },
-    "gemma-4-31b-it": {
-      "inference_engine": "FORGE",
-      "ci": {
-        "nightly": {
-          "devices": [
```
fivanovicTT, line 535: Hey @arginugaTT, just want to confirm - you want to remove `forge` engine for `gemma-4-31b-it`?

arginugaTT: yes @fivanovicTT 
Note: the model name handling for forge & vLLM is different, the forge makes model name lowercase, due to which we need dublicate model entries for gemma4, I feel that needs to be consistent. I did not make that a part of code change, but it needs to be looked at.

arginugaTT:
release passing CI: https://github.com/tenstorrent/tt-shield/actions/runs/28960893753/job/85947342922

.github/workflows/models-ci-config.json
bgoelTT: We should not delete the forge entry but rather fix its naming - that is the reason for the duplicate entries.

The real model name is `google/gemma-4-31B-it` https://huggingface.co/google/gemma-4-31B-it (note the upper-case `B`). If this change is made to the `dev/llm.yaml` entry for the forge version of this model then we can have just a single `models-ci-config.json` entry for `gemma-4-31B-it`

arginugaTT: I reverted the change and left the FORGE runtime naming alone — if the forge pipeline lowercases `gemma-4-31B-it` → `gemma-4-31b-it` and errors, that's for the forge team to fix in their naming scheme. dev/llm.yaml only has `google/gemma-4-31B-it` as per HF naming.

PR #4558, opened 2026-07-08
Build Blaze As Default

ljovanovicTT:
## Problem
Previously Inference Server supported two build configurations:
- Default - used the mock_model_runner
- Blaze - used tt-llm-engine, which determined at runtime whether to run the real pipeline or the mock pipeline

Following recent changes, the `mock_model_runner` was removed from Inference Server, and as a result the default build no longer started a model during warmup, causing the warmup to never occurre and Heavy Checks to fail.

## Implementation
With this PR, Blaze becomes the default build configuration, eliminating the need for the `--blaze` flag

## Testing
https://github.com/tenstorrent/tt-inference-server/actions/runs/28949598407

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -61,7 +61,7 @@ cd tt-media-server/cpp_server
 cd tt-llm-engine
 ./setup.sh --all
 cd ..
-./build.sh --blaze
+./build.sh
 ```
 
 Set `TT_INFERENCE_DEV_IMAGE` to override the local Docker image tag:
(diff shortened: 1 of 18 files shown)


knovokmetTT (approved):

nanicicTT (approved):

ztorlakTT (approved):

ljovanovicTT:
This is an issue since we are building cpp_server in `Dockerfile` where we do not have tt-llm-engine
Since we want to keep it as is in order to support image and video models, it is better to keep the `--blaze` flag

PR #4560, opened 2026-07-08
Mesh mooncake workers into single file and finalyze discovery

vpetrovicTT:


tt-media-server/cpp_server/scripts/deploy_migration_workers.sh
```diff
@@ -194,104 +303,275 @@ startDiscoveryService() {
...
+# complete DECODE_TAG_LIST because initWorkerSlots builds it before adding any
+# prefill slot.
+peersForWorker() {
+  local role="$1" tag="$2"
+  if [[ -n "${WORKER_PEERS[$tag]:-}" ]]; then
+    printf '%s' "${WORKER_PEERS[$tag]}"
+  elif [[ "${role}" == "prefill" ]]; then
+    printf '%s' "${DECODE_TAG_LIST}"
+  fi
+}
```
nanicicTT, line 387: The deployment gives every prefill every decode’s address.
Since decode tag list is

```sh
  for (( i = 0; i < NUM_DECODE; i++ )); do
    tag="${decode_tags[$i]:-decode-${i}}"
    DECODE_TAG_LIST="${DECODE_TAG_LIST:+${DECODE_TAG_LIST},}${tag}"
  done
```
below on line 405. 
Then each prefill receives a copy of the same Kafka message.
And every prefill has a long-lived connection attempt to the same decode.
The decode only handles one connection at a time. It accepts one client, stores it as the sole `peerSocket`, and waits for it to disconnect before returning to `acceptClient()`.
If another prefill gets through with the same migration ID before the first finishes, the decode rejects it.

Kafka tells every prefill to do job `42`.
- Deployment lets every prefill call the same decode.
- The decode has one active connection line.
- The second caller waits, fails, or is rejected as a duplicate.
- Therefore the system needs an explicit single owner before starting the migration.


Each prefill opens a TCP client for every resolved decode endpoint, see `tt-inference-server/tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp` lines 515-524. 
And there is a method called `openChannel` defined in `tt-media-server/cpp_server/src/transport/kv_migration_endpoints.cpp` lines 26-46. 
it basically says:
for each decode endpoint:
    create one TCP transport
    create one control channel around it
    retain it in channels_

Since each prefill has both endpoints, both prefills attempt a connection to the same decode.
And in `tt-media-server/cpp_server/src/sockets/tcp_socket_transport.cpp` lines 289-309 we have basically
connect once during prefill startup
keep that socket open
reuse it for future migration messages

The decode accepts one client, then stops accepting others. 
accept one client
wait until that one client disconnects
only then go back to accept another client

Kafka makes both prefills try to perform the same migration.
If two sources reach the decode with UUID `42`, the second is rejected.


So, `DECODE_TAG_LIST` is the complete decode-name list, so with two prefills both launch with `--peer decode-0 --peer decode-1`. The worker resolves each peer name to its `host:port` and `KvControlChannelConnector::openChannels()` creates and starts one persistent TCP client per decode.  
However, `TcpSocketTransport::serverLoop()` accepts one client and waits for it to disconnect before accepting another. Since each prefill has its own Kafka group, every prefill also executes the same migration ID. Thus two prefills can target the same decode for the same UUID: one is served while the other times out; if it reaches `MooncakeKvReceiver` while the UUID is pending, `prepareMirror()` rejects it as already prepared. Both workers still publish status for the same migration ID.  
We need an explicit owner per migration/decode: route the Kafka request to exactly one prefill and give that prefill the decode peer, or redesign the decode endpoint and migration protocol for concurrent multi-source fan-in.

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -278,7 +403,102 @@ std::shared_ptr<tt::sockets::ISocketTransport> makeServerTransport(
...
+}
+
+// Resolve every peer this worker was given (role-agnostic — a worker is just a
+// migration worker with a peer list) to its control endpoint via the metadata
+// service, retrying until all resolve or the timeout elapses. Static
+// --peer-control entries win and skip resolution; a SIGTERM (@p stop) aborts
+// the wait promptly. Returns the name->endpoint map, possibly partial on
+// timeout.
+std::unordered_map<std::string, KvControlChannelConnector::Endpoint>
+resolvePeers(ITransferEngine& engine, const WorkerConfig& cfg,
```
nanicicTT, line 463: `resolvePeers()` tries for 30 seconds, but returns only successes right? 
 The prefill never calls `resolvePeers()` again. The live startup result drops them as`pending` is a local variable and is destroyed when the function returns.
Is this intended behavior?
The resolver is called exactly once for a prefill, and the connector only creates channels for resolved endpoints. 

So `resolvePeers()` returns only the endpoints discovered during its 30-second startup window. Any configured `--peer` still unresolved is absent from `peers`, so the connector never creates a channel for it and `KvMigrationMultiHostSender` is permanently constructed without it. There is no subsequent prefill-side call to `resolvePeers()` or path to add a channel/sender after startup. A decode that publishes metadata later therefore remains unreachable until the prefill restarts.
The readiness check below is also based on `connector.channelCount()`, which counts only successfully created channels, not configured peers. With configured peers and no resolved endpoints, `total == 0`, `connected == 0`, and the worker goes to `Ready` state and starts Kafka as `0/0`. 
I would maybe retain unresolved configured peers in a long-lived discovery component, and calculate readiness against the configured peer set.

tt-media-server/cpp_server/src/transport/worker_health_server.cpp
```diff
@@ -0,0 +1,224 @@
...
+  out += std::to_string(r.body.size());
+  out += "\r\nConnection: close\r\n\r\n";
+  out += r.body;
+  return out;
+}
+
+bool sendAll(int fd, const std::string& data) {
+  std::size_t sent = 0;
+  while (sent < data.size()) {
+    const ssize_t n = ::send(fd, data.data() + sent, data.size() - sent, 0);
```
nanicicTT: `send(..., 0)` can terminate the entire worker when a health client closes before reading its response. On Linux, writing to that closed socket can raise `SIGPIPE`. 
This binary handles only `SIGTERM` and `SIGINT`, so the default `SIGPIPE` disposition terminates the process before `sendAll()` can return `false`. 

Please use `MSG_NOSIGNAL`, as `src/sockets/tcp_socket_transport.cpp` already does, and treat `EPIPE` / `ECONNRESET` as a failed client response. The health loop should close that client and continue serving, rather than letting one disconnected probe trigger worker restart.

What happens is that a health probe connects, sends `GET /healthz`, then disconnects before the worker writes the response. The worker tries to write to that dead TCP connection. On Linux, that write can raise `SIGPIPE` (as written above). Since this program does not ignore or handle `SIGPIPE`, the default action is to terminate the entire worker process.

For each health check this component does

accept client connection
read the HTTP request
write the HTTP response
close the client socket


There is one health-server thread, but signals apply to the whole process.
The probe can disconnect because its own timeout expires, a load balancer abandons the connection, a test/client is malformed, or the client intentionally resets the socket. 

The last argument to `send()` function is `0`.
On Linux, when the peer has already closed the connection, `send()` can either return `-1` and set `errno = EPIPE`, or deliver `SIGPIPE` to the process.

The `if (n <= 0) return false` only handles the return value, below. It does not prevent the signal. The signal can terminate the process before this function gets to return `false`.

The worker explicitly installs handlers for `SIGTERM` and `SIGINT`, but does not ignore or handle `SIGPIPE`.

So even though `sendAll()` runs in the health-server thread, the result can be:

health-server thread calls `send()` --> kernel raises `SIGPIPE` --> entire prefill/decode worker exits --> watchdog observes failed health/process and restarts it --> repeated (and maybe impatient) probes can cause restart.

Conceptually what needs to change, maybe, is this part

```cpp
::send(fd, data.data() + sent, data.size() - sent, MSG_NOSIGNAL)
```

If the health probe has already gone away, the worker should simply give up on replying to that one probe and continue running normally.

Using `MSG_NOSIGNAL` makes the failed write behave like an ordinary error instead of killing the whole process. Then:

- `EPIPE` or `ECONNRESET` means the client disconnected or reset the connection. Stop sending the response, close that client socket, and wait for the next probe.
- `EINTR` means the write was briefly interrupted by another signal, not that the client is gone. Retry the write.

Ignoring `SIGPIPE` globally would also prevent the crash, but it changes behavior for every socket write in the process. `MSG_NOSIGNAL` is safer because it protects only this specific response write, and it matches the approach already used by the project’s TCP transport.
see: https://linux.die.net/man/2/send

visnjakrsmanovicTT (approved):

fivanovicTT (approved):

nanicicTT (approved):

PR #4602, opened 2026-07-09
Fix: HF_HOME Permission issue with Falcon3-7B-Instruct

### LINKED ISSUE #4574: [Falcon3-7B - p150] release jobs fail: HF weight download `PermissionError` on `cache_root/huggingface` (regression from #4397)
### Basic info
- Model / engine / device: Falcon3-7B-Instruct / Forge / p150
- Image: `tt-media-inference-server-forge`
- Branch: `stable` (present since PR #4397, commit `ec73c2369`)
- Job: https://github.com/tenstorrent/tt-shield/actions/runs/28974464637/job/85980979354

### Summary
On the Forge/media server, the HuggingFace weight download fails with `Permission denied` on
`/home/container_app_user/cache_root/huggingface`. The model server never starts, so the
health check times out (3600s + 1200s) and the release job fails. vLLM jobs are unaffected.

### Error
Failed to download model weights for tiiuae/Falcon3-7B-Instruct:
  [Errno 13] Permission denied: '/home/container_app_user/cache_root/huggingface'
Worker 0 device init failed: PermissionError ... Check cache directory permissions.
...
[ERROR] llm_module.server_control: Inference server did not become healthy within 3600s
[ERROR] workflow.benchmarks: ❌ task=llm_benchmark produced no blocks
##[error]Process completed with exit code 1.

### root cause (#4397)
#4397 added to the media server's container env (`workflows/run_docker_server.py`):
`"HF_HOME": "/home/container_app_user/cache_root/huggingface",  # persistent cache_root volume`
The Forge image runs as container_app_user, but that mounted cache_root path isn't writable
by it, so the HF download can't write there. Evidence it's the code, not the runner:
- Every run with #4397 fails with this error; every run without it passes
(e.g. backup e3168fe6, run 28984259427) — on different ephemeral runners each time.
- vLLM path is fine because it writes to cache_root/weights/<model> (not HF_HOME).
- #4397 guarded the "HF_HOME dir doesn't exist yet" case but not "dir isn't writable".

### suggested fix

Create + chown cache_root/huggingface to container_app_user (image/entrypoint), or point the media server's HF_HOME at a writable location.

rnabeelTT:
## Summary

Fixes #4574 — a regression from #4397. On the Forge/media server, HuggingFace weight download crashed with `Permission denied: /home/container_app_user/cache_root/huggingface`; the server never became healthy and the release job timed out and failed. vLLM was unaffected.

## Root cause

#4397 set `HF_HOME=/home/container_app_user/cache_root/huggingface` for media/forge containers so weights persist on the `cache_root` volume. This works for the media image (its `docker-entrypoint.sh` runs as root and makes the mounted `cache_root` group-writable, then drops to `container_app_user`). But `Dockerfile.forge` has **no entrypoint** — it runs directly as `container_app_user` (uid 1000) with no permission fixup, so on a bind-mounted `cache_root` (owned by another uid, e.g. in CI) uid 1000 can't create `cache_root/huggingface` → crash. #4397 guarded the "dir doesn't exist yet" case but not "dir isn't writable".

## Fix

`tt-media-server/run_uvicorn.sh` (the shared entry both images launch, before Python/`transformers` load) now probes `HF_HOME` for writability and falls back to a per-user cache (`$HOME/.cache/huggingface`) if it isn't writable:

- **Writable `cache_root`** (media image) → `HF_HOME` kept → weights persist (unchanged behavior).
- **Unwritable `cache_root`** (forge image / bind mount) → falls back to the uid-owned ephemeral cache → server starts. This is exactly the pre-#4397 behavior, so the worst case is the state that ran successfully for months.

Placed in the shell entry (not the Python code or the host-side launcher) because writability is a container-runtime property — it depends on the runtime uid and the mounted volume, neither knowable host-side — and running before the interpreter avoids `transformers`/`huggingface_hub` import-time cache-path caching. It corrects the env once in the parent, so all consumers (download, `from_pretrained`, the `scandir` guard, speaker embeddings) and spawned workers inherit the fix.

## Changes

| File | Change |
|------|--------|
| `tt-media-server/run_uvicorn.sh` | Fall back to a writable per-user cache when `HF_HOME` isn't writable |
| `tt-media-server/tests/test_run_uvicorn_hf_home.py` | Tests: unwritable → fallback (dir created); writable → kept (probe cleaned up); unset → no-op |

## CI Runs
https://github.com/tenstorrent/tt-shield/actions/runs/29027954274

### DIFF
--- a/tt-media-server/run_uvicorn.sh
+++ b/tt-media-server/run_uvicorn.sh
@@ -29,4 +29,18 @@ if [ -n "${TT_KV_POOL_GB:-}" ]; then
     fi
 fi
 
+# HF_HOME (a persistent cache_root path set by the launcher) isn't always
+# writable by this uid. Fall back to a per-user cache so model download/load 
+# never crashes at startup (sacrifices cache persistence for media model weights).
+if [ -n "${HF_HOME:-}" ]; then
+    if ( mkdir -p "${HF_HOME}" && touch "${HF_HOME}/.hf_write_test" ) 2>/dev/null; then
+        rm -f "${HF_HOME}/.hf_write_test"
+    else
+        _hf_fallback="${HOME:-/home/container_app_user}/.cache/huggingface"
+        echo "[run_uvicorn] HF_HOME=${HF_HOME} not writable; falling back to ${_hf_fallback} (weights will not persist)"
+        mkdir -p "${_hf_fallback}"
+        export HF_HOME="${_hf_fallback}"
+    fi
+fi
+
 uvicorn --host 0.0.0.0 main:app --lifespan on --port "${SERVICE_PORT:-8000}"
--- a/tt-media-server/tests/test_run_uvicorn_hf_home.py
+++ b/tt-media-server/tests/test_run_uvicorn_hf_home.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2025 Tenstorrent USA, Inc.
+
+"""run_uvicorn.sh must guarantee HF_HOME is writable before launching the
+server. If the launcher-provided HF_HOME (a persistent cache_root path) isn't
+writable by this uid — e.g. a bind-mounted cache_root in CI, or an image
+without the root-entrypoint permission fixup — it falls back to a per-user
+cache so model download/load never crashes at startup."""
+
+import os
+import subprocess
+import tempfile
+from pathlib import Path
+
+SCRIPT = Path(__file__).resolve().parents[1] / "run_uvicorn.sh"
+
+# A path on the read-only procfs: mkdir fails for every uid (incl. root), so it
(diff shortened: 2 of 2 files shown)


fivanovicTT:
Can you test this with on-dispatch run? 

Let me know if you don't have access to tt-shield repo

anirudTT:
> Can you test this with on-dispatch run?
> 
> Let me know if you don't have access to tt-shield repo

https://github.com/tenstorrent/tt-shield/actions/runs/29027954274
Seems like the CI run passed , can we have this reviewed please?

vcankovicTT (approved):

fivanovicTT (approved):

rnabeelTT:
Hi Filip, the CI run has passed: https://github.com/tenstorrent/tt-shield/actions/runs/29027954274
This is the same model that was failing in Issue #4574

PR #4610, opened 2026-07-10
Peer table exchange

### LINKED ISSUE #4295: Prod-ize KV Table Exchange for Decode/Prefill
Turn the table exchange into a production component, callable from the worker rather than a standalone binary.

---

#### End Goal

Worker exposes something like discoverPeers(args) that does two things in sequence:
- Relies on dynamic discovery (#4294 )
- Adds table exchange to peer discovery. Migration worker can start once all peers are found and tables exchanged.
- Logging is added
- Tests from #4294 are extended to include table exchange

vpetrovicTT:


tt-media-server/cpp_server/src/transport/kv_migration_endpoints.cpp
```diff
@@ -23,42 +22,83 @@ KvControlChannelConnector::KvControlChannelConnector(
...
   return allCreated;
 }
 
+bool KvControlChannelConnector::openChannel(const std::string& name,
+                                            const Endpoint& endpoint) {
+  std::lock_guard<std::mutex> lock(mutex_);
+  return openChannelLocked(name, endpoint);
+}
+
+bool KvControlChannelConnector::replaceChannelLocked(const std::string& name,
```
nanicicTT, line 63: it destroys old KV Channel. 
Meanwhile, `kv_migration_multi_host_sender.cpp:89` (https://github.com/tenstorrent/tt-inference-server/pull/4610/changes#diff-a9aacc60c6c238830e89e40576adb34e4d78cca468b9102939b14ccbf11f1d1cR89) snapshots and later dereferences a raw pointer to that channel. The connector mutex and sender mutex do not coordinate.
If metadata changes during a migration, the Kafka worker can dereference a destroyed channel. Maybe use shared ownership for channels, or make replacement wait for active migrations.

tt-media-server/cpp_server/src/transport/kv_control_channel.cpp
```diff
@@ -33,17 +69,36 @@ bool KvControlChannel::send(const KvControlMessage& message) {
...
+}
 
-  const auto deadline = std::chrono::steady_clock::now() + receive_timeout_;
+KvControlChannel::ReceiveOutcome KvControlChannel::receiveMessageLocked(
+    KvControlMessage& out, std::chrono::milliseconds ioTimeout) {
+  // One budget for the whole wait (idle NO_DATA polls + in-flight payload).
+  // TcpSocketTransport enforces it inside tryReceiveMessage so mid-payload
+  // stalls cannot make the deadline check below unreachable.
+  IoBudgetGuard budget(transport_.get(), ioTimeout);
+  const auto deadline = std::chrono::steady_clock::now() + ioTimeout;
```
nanicicTT, line 101: This checks its deadline only after calling the transport. 
After the last sleep crosses that deadline, `tcp_socket_transport.cpp:565` 
https://github.com/tenstorrent/tt-inference-server/pull/4610/changes#diff-e4e599cb1dc53c989e0d61e93707f8e2c9d83e0f5c926b09176d5aa0199cd3d0L424
reports `TIMED_OUT`, which `tcp_socket_transport.cpp:518` converts into `CLOSED`:

https://github.com/tenstorrent/tt-inference-server/pull/4610/changes#diff-e4e599cb1dc53c989e0d61e93707f8e2c9d83e0f5c926b09176d5aa0199cd3d0L397


Result is that an idle decode session can disconnect instead of returning `TimedOut`, triggering reconnection and another **large table exchange**. Check maybe the channel deadline before probing/trying again, or distinguish an idle timeout from a timeout after partial frame receipt.

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -446,52 +465,165 @@ bool resolveOnePeer(ITransferEngine& engine, const WorkerConfig& cfg,
...
+    if (connector.channelCount() == 0) {
+      return nullptr;
+    }
+    if (connector.connectedCount() < connector.channelCount()) {
+      std::this_thread::sleep_for(
+          std::chrono::milliseconds(K_DISCOVERY_POLL_MS));
+      continue;
+    }
+
+    std::shared_ptr<const IKvTable> decodeTable;
```
nanicicTT, line 585: Decode workers can exchange different tables without startup rejecting them.
This code (where I commented) and code below keeps the first decode table and discards every later table after merely checking that it parses.

The sender then uses that first table to calculate addresses for every decode host. If one decode has a stale or different table, remote writes can target offsets that do not match its registered mirror. 
I would compare a hash of every returned blob and remain not-ready when they differ.

tt-media-server/cpp_server/src/transport/kv_migration_endpoints.cpp
```diff
@@ -70,40 +110,148 @@ std::size_t KvControlChannelConnector::awaitConnected(
...
+
+void KvMigrationReceiverServer::onAccept(
+    std::shared_ptr<sockets::ISocketTransport> peer) {
+  if (!peer) {
+    TT_LOG_ERROR("[KvMigrationReceiverServer] null peer transport on port {}",
+                 port_);
+    return;
+  }
+
+  auto session = std::make_unique<Session>();
```
nanicicTT, line 224: It seems that (please check) multi-client table exchange duplicates hundreds of megabytes per connection. 
Here it constructs each receiver with the server’s table blob by value (also line 206 but the logic starts from here to be read). 
But in here: 
https://github.com/tenstorrent/tt-inference-server/pull/4610/changes#diff-c6e5364728478f9459cac9a64f155dd0bea8ce783c4d6121386540167920cfafR71
(`kv_migration_orchestrator.hpp:71` in case the gihub link does not work as expected, sometimes it does not)
permanently stores that copy, plus the peer blob and parsed peer table.
With the documented $80$ to $350$ MiB tables, several prefill sessions can consume multiple gigabytes. Finished sessions are also reaped only when another connection arrives. 
Maybe share the immutable local blob and consume/reap/yield completed sessions independently.

tt-media-server/cpp_server/src/transport/kv_migration_endpoints.cpp
vpetrovicTT: Nice catch! Looking into this right now!

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
vpetrovicTT: Yeah, that is kinda left on purpose. A healthy same-deploy is supposed to have one fleet decode .pb (to the best of my knowledge). 
All in all, there is the idea of tightening this in the next iteration.

nanicicTT (approved):

fivanovicTT (approved):

PR #4632, opened 2026-07-13
Forge runner for Mistral 3.1 24B

acicovicTT:
# Bring up Mistral-Small-3.1-24B-Instruct-2503 on Blackhole galaxy (EXPERIMENTAL).

## Summary

Integrates **Mistral-Small-3.1-24B-Instruct-2503** on the **Blackhole galaxy**
(32-chip, 8×4 TP mesh) via the **tt-xla forge vLLM plugin**, served through
**tt-media-server** (BYO-uvicorn), and adds the accuracy eval configuration and
Shield-CI wiring needed to run it through `run.py`.

The model is served **text-only**: it is Pixtral-based, but the vision tower is
disabled (`limit_mm_per_prompt={"image": 0}`) so only the language model compiles.

## Eval results

Was unable to obtain a galaxy through shield CI (runner assignment timed out).

Run via `run.py --workflow evals` against the live server, full sample counts,
`num_concurrent=8`, greedy decoding.

| Task | Variant | Samples | Score | Published | Ratio |
|------|---------|--------:|------:|----------:|------:|
| GPQA-diamond | `gpqa_diamond_cot_zeroshot` (0-shot, chat) | 198 | **35.86%** (±3.4) | 45.96 | 0.78 |
| HumanEval | `humaneval_instruct` (chat) | 164 | **79.27%** (±3.2) | 88.41 | 0.90 |

**Acceptance status for FUNCTIONAL: FAIL (2 blockers).** This is the *ratio-to-published* check,
both tasks ran cleanly end-to-end (no server errors, no
crashes). `gpu_reference_score` is not yet populated, so acceptance falls back to
comparing against the published numbers. Also, see "Known limitations".

## Changes to serving (tt-media-server)

- `chat.py`: route `apply_chat_template` through `mistral_common` via a new
  `tokenizer_type` setting (Mistral ships its template in `tekken.json`, not an HF
  `chat_template`); handle assistant-prefill requests with `continue_final_message`
  so continuation-style prompts (e.g. `humaneval_instruct`) don't fail.
- `settings.py`: add `tokenizer_type` field. `launch_mistral_small_31_24b.sh`: BYO
  uvicorn launch on port 8019.

## Known limitations / follow-ups

- **GPQA ratio 0.78**: `cot_zeroshot` methodology differs from the publisher's; the
  `strict-match` filter is 0 (the model doesn't emit the exact "The answer is (X)"
  phrasing), so the reported score uses `flexible-extract`.
- **Text-only**: the Pixtral vision tower is disabled, so multimodal evals (e.g.
  ChartQA) are out of scope for this integration.
- **Concurrency throttled to 8**: higher concurrency triggered a TT UMD crash under
  sustained load; revisit once the driver issue is addressed.


fivanovicTT (commented):
Overall - looking good.

Can you add this model to tt-shield repo and run this model in `on-dispatch`? Just to validate it works there too

acicovicTT:
@fivanovicTT Had some issues with runner assignment in tt-shield, since this runs on `bh-galaxy`. I can try again, but multiple people were blocked on job timeouts after 24h, and I wouldn't like to block on that. That was when I was running `on-dispatch` from this branch, but in general the model cannot be added to tt-shield for regular runs without this PR being checked in first.

fivanovicTT:
> @fivanovicTT Had some issues with runner assignment in tt-shield, since this runs on `bh-galaxy`. I can try again, but multiple people were blocked on job timeouts after 24h, and I wouldn't like to block on that. That was when I was running `on-dispatch` from this branch, but in general the model cannot be added to tt-shield for regular runs without this PR being checked in first.

Totally get it. Let's try to run it one more time through the `tt-shield`, I can reach the CI guys if we need help with runner allocation. Could be that you ran it with on-nightly or something else happened - and it timed out.

You can add model in the list of available models and run it like that (link to example: https://github.com/tenstorrent/tt-shield/pull/752). And, when you select `Use workflow from` -> choose the branch you created in `tt-shield` instead of `main`.

Ping me directly if you need help!

acicovicTT:
> Let's try to run it one more time through the tt-shield

~~Here is the run: https://github.com/tenstorrent/tt-shield/actions/runs/29989943453~~

acicovicTT:
Here is the run: https://github.com/tenstorrent/tt-shield/actions/runs/30006410946
(needed some fixes related to "bh-galaxy" vs "blackhole_galaxy" values in configs, so this run will also test the last two commits.)

Update on this for @fivanovicTT: the run finished, and it is red because the evals failed (as expected, since I got that in my local run on galaxy as well), but all the infra around it is validated. Per instructions I was given, this is okay for the Experimental tag, and we need to work on obtaining GPU numbers so we can properly compare (now it compares against published numbers).

One question which I have left is that HumanEval requires an environment variable to be set (`HF_ALLOW_CODE_EVAL`), not sure how to do that in shield CI, any suggestions? We can re-run it with that env. var but the eval will still fail because it will not be as good as published numbers.

fivanovicTT (approved):

fivanovicTT:
> One question which I have left is that HumanEval requires an environment variable to be set (HF_ALLOW_CODE_EVAL), not sure how to do that in shield CI, any suggestions? We can re-run it with that env. var but the eval will still fail because it will not be as good as published numbers.

I know github has some limits on how many custom fields we can for workflows. Not sure if we have available slots for this in Github CI. 

@mjeremicTT @vmaksimovicTT, do you guys have some idea how this can be achieved?

mjeremicTT:
Is this env var specific only to this Forge model or is it generally used for Forge models? You can probably expand it here: [https://github.com/tenstorrent/tt-shield/blob/main/.github/workflows/workflow_run-tests-with-inference-server.yml#L286](https://github.com/tenstorrent/tt-shield/blob/main/.github/workflows/workflow_run-tests-with-inference-server.yml#L286)

acicovicTT:
> Is this env var specific only to this Forge model or is it generally used for Forge models? You can probably expand it here: https://github.com/tenstorrent/tt-shield/blob/main/.github/workflows/workflow_run-tests-with-inference-server.yml#L286

I wouldn't say it's Forge-specific, but eval-specific. Maybe different backends already set this env. var. I will look into it when I make the change to merge this into the tt-shield CI.

Issue #4386, opened 2026-06-25
Qwen/Qwen3-32B: reasoning leaks into content (streaming), tool_choice=none ignored, and tool calls dropped with thinking


ksriv001:
# Qwen/Qwen3-32B streaming: reasoning + tool-call parser bugs (serving-side)

## Summary

Four streaming-specific defects on **Qwen/Qwen3-32B**, all reproduced **directly against the vLLM backend with the console proxy entirely out of the path**. They originate in the inference server's streaming parsers/config, not in the tt-cloud-console gateway (which is a byte-level passthrough for SSE deltas).

**Backend under test:** vLLM `0.11.1` (`vllm-tt-metal-src-release-ubuntu-22.04-amd64:0.11.1`), launched with:

```
--enable-auto-tool-choice --tool-call-parser hermes --reasoning-parser qwen3
```

| # | Issue | Result (direct to vLLM) | Streaming-specific? | Verdict |
|---|-------|-------------------------|---------------------|---------|
| 1 | `response_format: json_object` | 4/4 leaked `</think>` + trailing reasoning into `content`; JSON invalid | **Yes** (non-streaming clean) | streaming reasoning parser |
| 2 | `response_format: json_schema` | leaked `</think>` into `content`; JSON invalid | **Yes** (non-streaming clean) | streaming reasoning parser |
| 3 | `tool_choice: "none"` | raw `<tool_call>` XML in `content`, `finish_reason: stop`, 0 tool deltas | **No** (fails in both modes) | `tool_choice=none` serving/template |
| 4 | `tools` + `enable_thinking` | **8/20** runs `finish_reason: stop` with no tool call (40%) | n/a | parser interaction; reference host passes 20/20 |

> All curls below hit the backend **directly**. Direct calls require a `curl`-style `User-Agent` — a default `python-urllib`/empty UA returns `403 Forbidden` at the ingress.

```bash
export JWT="<service JWT>"
export DIRECT="https://qwen32gxor--tenstorrent.workload.tenstorrent.com/v1/chat/completions"
```

---

### Non-streaming control (same payloads, `stream:false`, direct to vLLM)

Run identically except `stream:false`. This is the decisive test for "is it the streaming parser?":

| Issue | Streaming | Non-streaming | Conclusion |
|-------|-----------|---------------|------------|
| 1 `json_object` | 4/4 leak `</think>`, invalid JSON | **3/3 clean, valid JSON** | streaming-parser only |
[...]

PR #4644, opened 2026-07-14
qwen3-32b streaming tools + enable_thinking: intermittently stops with no tool call

lukezTT:
parent issue: https://github.com/tenstorrent/tt-inference-server/issues/4386

This PR contains the tests to catch the issues 1, 2, 4 (not 3) flagged in the parent issue.

These tests catches the issues raised. Seen in this run: https://github.com/tenstorrent/tt-shield/actions/runs/29852446182/job/88728421525#step:19:198


bgoelTT (changes requested):
These tests will now be executed for all models that run `test_vllm_chat_completions.py`, but it should only be executed for Qwen3-32B

Can you leverage the tt-inference-server V2 spec tests workflow to configure this?

tt-inference-server-v2/llm_module/test_vllm_qwen3_streaming.py
```diff
@@ -0,0 +1,462 @@
...
+        # The api_client fixture re-raises a new HTTPError without .response, so
+        # detect auth failures from the message (requests uses "<code> Client
+        # Error"). Auth is a config problem — surface it instead of skipping.
+        msg = str(e)
+        if "401" in msg or "403" in msg:
+            raise
+        pytest.skip(
+            "Server rejected the streaming tools + enable_thinking payload; "
+            f"model likely does not support chat_template_kwargs thinking: {e}"
+        )
```
bgoelTT: Instead of skipping the test we need to raise this as an error. This test should be passing

bgoelTT (changes requested):

bgoelTT (approved):

PR #4665, opened 2026-07-15
Re-enable gemma-4-31B-it agentic evals `terminal_bench_2` & `swe_bench_verified`

dwadhwaniTT:
Restore the two `EVALS_AGENTIC` tasks for `google/gemma-4-31B-it` that PR #4604 commented out. With an empty agentic task list the release workflow's `_has_agentic_tasks()` returns False, so the `agentic` child is never scheduled and the eval step no-ops.

Mirrors the approach in PR #4634 (which did the same for Qwen3.6-27B):
- `terminal_bench_2` `n_concurrent_trials` set to 1 until batch > 1 is supported
- `terminal_bench_2` `CI_NIGHTLY` task list swapped to the current selection (`break-filter-js-from-html`, `cobol-modernization`, `compile-compcert`, `feal-differential-cryptanalysis`, `qemu-startup`)
- `swe_bench_verified` kept at its original gemma values (`n_concurrent_trials=5`, 160K/32K token budgets, same `CI_NIGHTLY` instance IDs)
- Add `orjson` and `backoff` to `requirements/evals-agentic.txt` (litellm >=1.91 lazily imports its MCP/proxy modules but only declares them under the `[proxy]` extra; without them mini-swe-agent's SWE-bench runs fail with `ModuleNotFoundError`)

Scoped to `gemma-4-31B-it` only: the `gemma-4-26B-A4B-it` / `12B` variants and the gemma GPQA `CI_NIGHTLY` sample-limit change from #4604 are left as-is.


bgoelTT (commented):
Please link a Models CI Dispatch run with `--workflow agentic` demonstrating the evals are actually running. If scores are good then 👍, if not, then an issue needs to be filed

dwadhwaniTT:
CI run with agentic evals : https://github.com/tenstorrent/tt-shield/actions/runs/29434228925/job/87416722171

Agentic Evals fail because of the model's limitation on context length that can be served.

tstescoTT (approved):

ddjukicTT (approved):

PR #4674, opened 2026-07-16
Wire engine table+devmap handoff into Mooncake migration workers

### LINKED ISSUE #4571: Define mechanism for device map creation
Running migration worker `mooncake_kv_migration_worker_main.cpp` requires a device map, `--device-map`.

**Test environment** - for the existing e2e tests with 2 migration workers, sender and receiver. We either need a fixed file we use for all the tests, or script to create one for testing.

**Production environment** - bring clarity how this file is generated in production. References: 
- [prefill_scheduler](https://github.com/tenstorrent/tt-llm-engine/blob/main/src/scheduler/prefill/prefill_scheduler.cpp)
- [migration](https://github.com/tenstorrent/tt-llm-engine/tree/main/disaggregation/migration/src)

Have a script to generate device map and pass it as a map to a test. 

Note: device map is like synthesized logical/physical so we can test without model

### LINKED ISSUE #4623: Use real device map from device
Mooncake migration worker should get its DeviceMap from the live device when available, and fall back to a synthesized --device-map file otherwise. Today it only reads the text file and, with no file, falls back to the placeholder loc.device & 0xFFFF in buildDeviceIo, which collides across meshes.

Real path: The engine enumerates chips through UMD, resolves each FabricNode's actual ASIC unique id → DeviceMap::set(), and ships the map over sendEngineHandoff. The worker consumes it from EngineTables via SocketEngineTableSource instead of the file. Once a real map is in play, turn the & 0xFFFF placeholder into a hard error.

Synthesized fallback (test / bring-up) - handled in https://github.com/tenstorrent/tt-inference-server/issues/4571: Provide a script that reads the prefill + decode KV chunk tables and emits the mesh chip umd file. The tables fully determine the key set (every {mesh,chip} via deviceGroup/allHostLocations); the umd column is a synthesized convention, a flat mesh-major re-index (umd = 0..N-1) that avoids the cross-mesh collision the raw placeholder has. This is correct only for a homogeneous single-machine test setup and it is not a real ASIC id.

vpetrovicTT:


tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
```diff
@@ -245,6 +248,18 @@ bool parseConfig(int argc, char** argv, WorkerConfig& cfg) {
...
+      try {
+        const int port = std::stoi(v);
+        if (port <= 0 || port > 65535) throw std::out_of_range("range");
+        cfg.engine_handoff_port = static_cast<uint16_t>(port);
+      } catch (...) {
+        std::cerr << "--engine-handoff-port must be 1..65535, got: " << v
+                  << "\n";
+        return false;
+      }
+      continue;
```
fivanovicTT, line 261: Maybe add some logging here so we know what happened?

tt-media-server/cpp_server/src/engine_handoff_sender_main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+void usage() {
+  std::cerr
+      << "usage: engine_handoff_sender --host HOST --port N\n"
+         "  (--device-map PATH | --device-map-stdin)\n"
+         "  pushes DeviceMap only; worker loads .pb from --table /tmp path\n"
+         "  refuses empty/unreadable maps (exit non-zero)\n";
+}
+
+bool parseArgs(int argc, char** argv, SenderConfig& cfg) {
```
fivanovicTT, line 49: This should be a private method since it is only used in this file

tt-media-server/cpp_server/src/engine_handoff_sender_main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+}  // namespace
+
+int main(int argc, char** argv) {
+  tt::utils::ZeroOverheadLogger::initialize("engine-handoff-sender");
+
+  SenderConfig cfg;
+  if (!parseArgs(argc, argv, cfg)) {
+    usage();
+    return 2;
```
fivanovicTT, line 129: Can you give a brief explanation of what these numbers represent: 0, 1, 2?

visnjakrsmanovicTT (approved):

tt-media-server/cpp_server/src/engine_handoff_sender_main.cpp
dmadicTT, line 1: As I see this is just a passthrough process that either gets a device map from a file or stdin and then hands it off to the worker process via TCP.

I would avoid having a separate process only for this.

How is it done at the moment in tt-llm-engine?

@visnjakrsmanovicTT 
Can we just use a file input in the worker itself?
For example
```bash
print_local_device_map > device_map.txt
migration_worker --device-map device_map.txt
```

tt-media-server/cpp_server/src/mooncake_kv_migration_worker_main.cpp
vpetrovicTT: Good call — added TT_LOG_INFO on a successful parse and TT_LOG_ERROR (plus the existing cerr) on failure so the accepted port / bad value show up in the worker logs.

tt-media-server/cpp_server/src/engine_handoff_sender_main.cpp
vpetrovicTT: Already file-local: parseArgs lives in the anonymous namespace at the top of engine_handoff_sender_main.cpp, so it has internal linkage and isn’t visible outside this TU. That’s the C++ equivalent of a private helper for a free-function main.

tt-media-server/cpp_server/src/engine_handoff_sender_main.cpp
vpetrovicTT: Documented in usage() and a short comment in main:
0 — success (DeviceMap handed off)
1 — runtime failure (load / connect / send)
2 — bad or missing CLI args (usage printed)

tt-media-server/cpp_server/src/engine_handoff_sender_main.cpp
vpetrovicTT: The only thing I  see socket wins over file handoff for now is few scenarios:
- engine starts after worker and pushes a fresh map
- no temp file / NFS sync
- fail-closed wait until a map arrives

visnjakrsmanovicTT: @dmadicTT Yes, the engine_handoff_sender with TCP does the same thing as the file path (in the sense in both cases we need to provide a file upfront). In tt-llm-engine dev map is shared over shmem, the idea here was to use socket instead. engine_handoff_sender is the standalone path to test this, next step would be to modify tt-llm-engine side and implement some sort of sendEngineHandoff there (@vpetrovicTT correct me if I'm wrong). 
To answer your question, I agree that for now we can only use the file input in the mooncake migration worker, as today's engine_handoff_sender does more or less the same, but keep the socket path and plan for adding dev map sender logic on tt-llm-engine side.

fivanovicTT (approved):

idjuricTT (approved):

dmadicTT (approved):

Issue #4675, opened 2026-07-16
Add SWE-Bench Agentic Evals Back to Qwen3.6-27B


lingjunwenTT:
Currently, due to lack of prefix caching or equivalents in the model, all `swe-bench-verified` times out in the allotted 2h window: https://github.com/tenstorrent/tt-shield/actions/runs/29363864010/job/87190489361#step:11:5880

Enable the swe-bench-verified tasks by uncommenting it in `eval_configs.py` when prefix caching or equivalent is enabled in vLLM or tt-metal.

PR #4676, opened 2026-07-16
Disable `swe-bench-verified` for Qwen3.6-27B

lingjunwenTT:
`swe-bench-verified` currently times out after 10.5 hours and does not produce any results.  This is likely due to lack of prefix caching, every episode is re-sent with all prefix context and prefilled, leading to long inference times per episode and eventually timing out the eval task.

https://github.com/tenstorrent/tt-shield/actions/runs/29363864010/job/87190489361#step:11:5880

Screenshot 2026-07-16 at 9 39 36 AM


lingjunwenTT:
Issue to re-enable in the future: https://github.com/tenstorrent/tt-inference-server/issues/4675

tstescoTT:
Note for follow up: would be nice to have a way not run evals/tests like this that we know are going to fail without burning ci time. The xfail masking is also needed but doesnt stop the CI test run time.

tstescoTT (approved):

Issue #4666, opened 2026-07-15
Add E2E tests for Dynamo-native routing


ztorlakTT:
Add end-to-end tests for DYNAMO_NATIVE_ROUTING=1 covering decode/prefill worker registration, big ISL routing through prefill then decode, small/local decode routing, streaming, cancellation, worker restart behavior, and error handling.

PR #4694, opened 2026-07-20
dynamo routing cpp heavy checks

lmuravljovTT:
## Summary 
Adding a big ISL test for DYNAMO_ROUTING=1 scenario. 
Small ISL doesn't make sense yet- but will once the decode-first-hit for small ISL is implemented into dynamo 

- Added checkbox into the cpp-heavy-check workflow - DYNAMO_ROUTING=1 scenario will only run when triggered explicitly 
- cpp-heavy-check setup will bring up Dynamo frontend + 1 prefill + 1 decode 
- dynamo_routing_big_isl_test.cpp runs a big prompt request, and once it completes it verifies behaviour through logs: 
1. prefill was hit first
2. prefill sent the slot reservation request
3. decode received and handled the slot reservation request
4. prefill received slot reservation response
5. decode was hit afterwards and produced tokens
6. decode released the proper session 
- extracted common helper used in multiple e2e tests

## Unrelated fixes
While trying to run the test on CI, multiple errors occurred which didn't come up locally: 
1. null request -> triggering SIGSEGV 

```
resolvePrefillSession(
    work.request,                           // uses work.request
    
    [..., work = std::move(work), ...]() {  // moves work
```

C++ doesn't define argument evaluation order. If the lambda was built first, work was already moved from when work.request was read  --> resulting in crashes when request->task_id was called 
    
    
    
2. empty resultCallback → std::bad_function_call → SIGABRT
Both success and error lambdas did: 
resultCallback = std::move(resultCallback)

And only one move can win. 

Fix: put the callback in a shared_ptr once and capture that in both lambdas; invoke only if resultCallbackShared && *resultCallbackShared.

### Workflow pass:
https://github.com/tenstorrent/tt-inference-server/actions/runs/29929738659/job/88962558027

### TODO 
1. Integration tests will be added separately: 
https://github.com/tenstorrent/tt-inference-server/issues/4666

2. Once dynamo route supports decode-first, the test should be updated


idjuricTT (approved):

tt-media-server/cpp_server/tests/e2e/dynamo_routing_big_isl_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+      << ")";
+  EXPECT_NE(*slotId, K_INVALID_SLOT_ID)
+      << "slot reservation was granted but with INVALID_SLOT_ID for taskId="
+      << *taskId;
+
+  // Confirm decode is releasing the session after the final chunk.
+  std::optional<DecodeSessionReleaseLog> release;
+  ASSERT_TRUE(waitFor(
+      [&] {
+        release = findDecodeSessionReleased(readFile(decodeLog));
```
ztorlakTT: Can we compare this session with the sessionId from the prefill result log?

tt-media-server/cpp_server/tests/e2e/dynamo_routing_big_isl_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+      << "expected completion_tokens > 0 (decode should have produced at least "
+         "one token); usage.prompt="
+      << resp.usage.promptTokens
+      << " usage.completion=" << resp.usage.completionTokens;
+
+  // Confirm prefill is sending a slot reservation request.
+  std::optional<uint32_t> taskId;
+  ASSERT_TRUE(waitFor(
+      [&] {
+        taskId = findPrefillFirstTaskId(readFile(prefillLog));
```
ztorlakTT: Is it possible that this catches a log from the warmup request since we read the whole log here?

ztorlakTT (approved):

ztorlakTT (approved):

tt-media-server/cpp_server/tests/e2e/dynamo_routing_big_isl_test.cpp
lmuravljovTT: yes thats a good catch, We are now snapshoting the offset and only reading the logs from when the test actually starts

PR #4730, opened 2026-07-21
Add wan prodia runner, resolve pipeline module

sadesoyeTT:
This PR adds a runner for prodia T2V model and fixes a related  module resolution issue encountered when following instructions.

### DIFF
--- a/tt-media-server/config/constants.py
+++ b/tt-media-server/config/constants.py
@@ -20,6 +20,7 @@ class SupportedModels(Enum):
     QWEN_IMAGE_2512 = "Qwen/Qwen-Image-2512"
     MOCHI_1 = "genmo/mochi-1-preview"
     WAN_2_2 = "Wan-AI/Wan2.2-T2V-A14B-Diffusers"
+    WAN_2_2_T2V_PRODIA = "Wan-AI/Wan2.2-T2V-A14B-Diffusers"
     WAN_2_2_I2V = "Wan-AI/Wan2.2-I2V-A14B-Diffusers"
     WAN_2_2_I2V_PRODIA = "Wan-AI/Wan2.2-I2V-A14B-Diffusers"
     WAN_2_2_I2V_ANISORA = "Wan-AI/Wan2.2-I2V-A14B-Diffusers"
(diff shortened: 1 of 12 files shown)


tt-media-server/config/constants.py
```diff
@@ -888,6 +894,14 @@ def _vllm_config(
         "download_weights_from_service": False,
         "request_processing_timeout_seconds": 5000,
     },
+    (ModelRunners.TT_WAN_2_2_T2V_PRODIA, DeviceTypes.GALAXY): {
+        "device_mesh_shape": (4, 8),
+        "is_galaxy": False,
```
fivanovicTT, line 922: Shouldn't this value be `True` given the device type is `Galaxy`?

tt-media-server/tt_model_runners/dit_runners.py
```diff
@@ -484,6 +485,84 @@ def get_pipeline_device_params(self):
...
+    """Shared device params for Prodia Wan T2V/I2V distilled runners.
+
+    The 4x8 LoudBox trace binary needs ~30.6MB; the default 30MB region
+    rejects it and warmup OOMs. Both 4x8 (32 chips) and 4x32 (128 chips)
+    Blackhole meshes get the bumped trace region.
+    """
+    device_params = {"fabric_config": ttnn.FabricConfig.FABRIC_1D_RING}
+    mesh_size = device_mesh_shape[0] * device_mesh_shape[1]
+    if mesh_size >= 32 and is_blackhole():
+        device_params["trace_region_size"] = 150_000_000
```
vpetrovicTT, line 507: The 'trace_region_size' is bumped for I2V here too. Some specific reason for this change?

tt-media-server/config/constants.py
sadesoyeTT: I forget what it is exactly, but is_galaxy was used to trigger something with didn't need or want in the DiTs models. All other DiTs models use false.

tt-media-server/tt_model_runners/dit_runners.py
sadesoyeTT: Yep, when we tested recently, the initial value was not enough. Trace binary must have increased.

fivanovicTT (approved):

PR #4750, opened 2026-07-22
Make b=8 as a toggle for Qwen3.6-27B

atupe-tt:
Benchmark Numbers
image

Also added a new method in command_factory so that user's `--impl` flag is not ignored. This is needed for Qwen3.6 because the model has same `--device` config, so only the catalog that has `default_impl=True` got picked.


bgoelTT (changes requested):
Can you please revert the changes for the override env vars?

bgoelTT (approved):
Approving to not block, but please address the above requested change

atupe-tt:
> Can you please revert the changes for the override env vars?

Change reverted

idjuricTT (approved):

PR #4751, opened 2026-07-23
Add Qwen3-32B (forge DP+TP) integration for BH Galaxy

ssaliceTT:
Adds the **forge-vllm-plugin** implementation of **Qwen3-32B** on the **Blackhole Galaxy** (8×4 DP+TP mesh), as a non-default impl alongside the native `qwen3_32b_galaxy`. Status: **EXPERIMENTAL**.

Serves 16384 / batch 32 with FSDP weight sharding, chunked prefill, and bfp8 KV cache.

### Changes (basic model addition)
- Mesh-aware forge runner + `BLACKHOLE_GALAXY` ModelConfig `(8,4)` and `dev/llm.yaml` spec.
- `models-ci-config.json`: FORGE `nightly` on `BH_GALAXY`.
- Inherits the shared `Qwen/Qwen3-32B` eval config (no eval-config change).

No shared engine/harness changes.

### Local validation
`run.py --workflow release --ci-mode` (EXPERIMENTAL): server warms on all 32 devices, Acceptance **PASS** (rc=0); benchmark sweep sizes to the served 16384 (no 400s). Evals are the shared Qwen3-32B set (accuracy waived under EXPERIMENTAL).

Note: served on a galaxy whose physical topology matches `single_bh_galaxy_mesh_graph_descriptor`; a runner with a different (torus) topology needs the matching MGD / a tt-xla with the mesh-open workaround.

### DIFF
--- a/.github/workflows/models-ci-config.json
+++ b/.github/workflows/models-ci-config.json
@@ -282,7 +282,8 @@
           "ci": {
             "nightly": {
               "devices": [
-                "P300X2"
+                "P300X2",
+                "BH_GALAXY"
               ]
             }
           }
--- a/tt-media-server/config/constants.py
+++ b/tt-media-server/config/constants.py
@@ -1165,6 +1165,16 @@ ModelConfigs = {
         "max_batch_size": 1,
         "queue_for_multiprocessing": QueueType.FasterFifo.value,
     },
+    (ModelRunners.VLLMForge_QWEN_32B, DeviceTypes.BLACKHOLE_GALAXY): {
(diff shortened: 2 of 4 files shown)


kmabeeTT:
Thx, let's ideally see if we can get a CI run passing with a machine that comes back up ideally (but I wouldn't block this on it) - also since nearing release deadline Aug 14, better add to release CI not just nightly CI like I did in https://github.com/tenstorrent/tt-inference-server/pull/4933 for embedding models just now.

Edit: After offline discussion with @vcankovicTT was strongly suggested we don't add to release CI job (only nightly) unless CI machine is back and we have a CI run to show.

Issue #4733, opened 2026-07-21
[Falcon3-7B-Instruct]: Record GPU Reference Scores


ddilbazTT:
Falcon3-7B-Instruct is missing gpu reference scores. These will be generated on L4 GPU.

[Link to official eval results on Hugging Face](https://www.google.com/search?q=falcon3-7b-instruct+paper&rlz=1C5GCEM_enCA1124CA1124&oq=falcon3-7b-instruct+paper&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQABgeMgYIAhAAGB4yCAgDEAAYCBgeMg0IBBAAGIYDGIAEGIoFMgoIBRAAGIAEGKIEMgcIBhAAGO8FMgcIBxAAGO8FMgcICBAAGO8FMgcICRAAGO8F0gEINTU5MmowajeoAgCwAgA&sourceid=chrome&ie=UTF-8)  

# Evals
The following evals are supported through tt-inference-server. The tables include information about which eval results are available for Falcon3-7B-Instruct.

## General / language understanding

| Eval Type | Published Score | Published At |
| :--- | :--- | :--- |
| mmlu_generative | 70.5 (MMLU, 5-shot) | https://huggingface.co/tiiuae/Falcon3-7B-Instruct#benchmarks |
| mmlu_pro | 40.7 (5-shot) / 34.30 leaderboard | https://huggingface.co/tiiuae/Falcon3-7B-Instruct#benchmarks |

## Instruction following

| Eval Type | Published Score | Published At |
| :--- | :--- | :--- |
| ifeval | 76.5 (internal) | https://huggingface.co/tiiuae/Falcon3-7B-Instruct#benchmarks |
| leaderboard_ifeval | 76.12 (Open LLM Leaderboard) | https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard |
| meta_ifeval | Not published | — |

## Math

| Eval Type | Published Score | Published At |
| :--- | :--- | :--- |
| leaderboard_math_hard | 31.87 (leaderboard) / 29.4 internal | https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard |
| meta_math | Not published | — |
| r1_math500 | Not published | — |
| aime25 | Not published | — |
| r1_aime24 | Not published | — |

## Reasoning / GPQA

| Eval Type | Published Score | Published At |
| :--- | :--- | :--- |
[...]

PR #4758, opened 2026-07-23
[Falcon3-7B-Instruct] Add GPU Reference Scores

ddilbazTT:
#4733

Adding `ifeval` and `gpqa_diamond_generative_n_shot` GPU Reference Scores for Falcon3-7B-Instruct. 

| Eval Task | Run 1 | Run 2 | Run 3 | Run 4 | Run 5| Run 6 | Run 7 | Run 8 | Run 9 | Run 10 | Average | Std Dev | Published (HF) | Avg / Published |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| ifeval | 73.57 | 72.46 | 73.38 | 72.64 | 73.01 | 72.46 | 72.46 | 73.20 | 72.46 | 72.27 | **72.79** | 0.44 | 76.50 | 0.9515 |
| gpqa_diamond_generative_n_shot | 43.43 | 43.43 | 43.43 | 43.43 | 43.43 | 43.43 | 43.43 | 43.43 | 43.43 | 43.43 | **43.43** | 0.00 | N/A | N/A |

Note: Fixed `workflows/model_specs/dev/cnn.yaml` - `workflows/model_specs/dev/llm.yaml` pre-existing mix-up that I noticed during this PR.


workflows/model_specs/dev/cnn.yaml
```diff
@@ -122,6 +122,7 @@ templates:
   model_type: LLM
   inference_engine: FORGE
   uses_tensor_model_cache: false
+  # status: FUNCTIONAL #TODO(@ddilbazTT, @kmabeeTT): Add when model passes functional.
```
kmabeeTT: I think release folks might be the ones to add this line and not us, so I'd exclude this for now

workflows/model_specs/dev/cnn.yaml
```diff
@@ -143,6 +144,21 @@ templates:
         MIN_NUM_SEQS: "1"  # b1-prefill: compile [1,n] graph alongside [32,n] (tt-xla #5281)
         PREFILL_BATCH_THRESHOLD: "16"  # route <=16 pending prefills to b1, serially
 
+- weights:
```
kmabeeTT: remind me why is this added and is it needed?

kmabeeTT (approved):
Thank you!

workflows/model_specs/dev/cnn.yaml
ddilbazTT: It would fail as such when trying to run the gpu reference score generation:
```
Model:=Falcon3-7B-Instruct does not have a default impl for device:=gpu
```

workflows/model_specs/dev/cnn.yaml
bgoelTT: `status` should be pinned here in the `dev` catalogue because it allows the release owner to see if the model is hitting the expected status

bgoelTT (changes requested):
Please do indicate an expected status type here, also I think the TTT ModelSpec can be removed because it doesn't exist AFAIK

workflows/model_specs/dev/cnn.yaml
kmabeeTT: Ah okay cool, still EXPERIMENTAL for now, hopefully FUNCTIONAL soon (was only getting 0.88-0.90 of gpu_ref not yet 0.95 :( )

workflows/model_specs/dev/cnn.yaml
kmabeeTT: Ah I see - Is this expected @bgoelTT and we want to commit these types of entries used to generate gpu_reference scores?

bgoelTT: Oh I didn't see the device_type field my bad, yes this is the correct way to generate the GPU reference scores 👍

bgoelTT:
I just noticed this too, can you move these changes from `cnn.yaml` to `llm.yaml`? All LLM ModelSpecs should live there

workflows/model_specs/dev/cnn.yaml
ddilbazTT: @kmabeeTT You might be thrown off because there aren't other examples in this file to have the GPU reference field. It is because cnn.yaml is for forge vision + llm models, currently. If you look at [llm.yaml](https://github.com/tenstorrent/tt-inference-server/blob/bcbac4d311896aa05c2ecb6f9c95911888d5c2f0/workflows/model_specs/dev/llm.yaml) you can see there are other examples as such.

bgoelTT: Why are forge LLMs going in the cnn.yaml?

ddilbazTT: I assume that wasn't intended. I'll also move the llms in this file to lllm.yaml.

bgoelTT (approved):
Approved

PR #4767, opened 2026-07-24
Migration-worker deploy: role-specific device maps, dated image tags, log archiving, kafka_request_id logging

dmadicTT:
Small mix of migration-worker deploy improvements.

- Device maps can be sourced from a file on the deploy host (no longer required on the worker box) and split by role — `--prefill-device-map` / `--decode-device-map`, with `--device-map` as a shared fallback.
- Default handoff is now direct file mount (`ENGINE_HANDOFF_PORT=0`); socket handoff via `engine_handoff_sender` is opt-in.
- `--decode-tags` now takes owning hostnames and converts them to tt-blaze's `host-<crc32>` table tags; `--prefill-tags` still verbatim.
- Migration-worker image is tagged `ghcr.io/tenstorrent/tt-shield/tt-migration-worker:<ddmmYYYY-HHMMSS>` by default; drops the `--push` requires `--image` gate.
- Docker containers now mount `/dev/tenstorrent` and `/dev/hugepages-1G` when `MIGRATION_MODE=device`.
- Worker logs are archived to `<log>.<UTC-timestamp>.log` on restart instead of being truncated.
- Consistent `kafka_request_id`-based identifiers and logging across `RemoteKVManagerAdapter`, Mooncake sender/receiver, and the migration orchestrator; adds per-migration `startedAt` / `totalKafkaRequests` for latency tracking.
- Tests updated to the new `prepareMirror` / `drain` / `transferSlot(kafkaRequestId, ...)` signatures.


tt-media-server/cpp_server/scripts/migration_deploy.conf
dmadicTT, line 1: Do we still need this file @vpetrovicTT ?

vpetrovicTT: Not really.. We can use it as a reference for a future config file :/

vpetrovicTT (approved):

idjuricTT (approved):

PR #4768, opened 2026-07-24
Transfer engine over RDMA: bounce-buffer KV migration + DRISC dev I/O

### LINKED ISSUE #4690: Mooncake transfer engine switch to RDMA
Includes adding drisc + bounce buffer + RDMA (under transfer engine).
Target 11+ GB/s, consider Mooncake TENT if classic doesn't work out.

visnjakrsmanovicTT:
Transfer engine over RDMA: bounce-buffer KV migration + DRISC device I/O (#4690)

Move the KV-cache migration data plane to RDMA-over-host. The decode side registers one small (tens of MiB) double-pinned bounce buffer instead of a full-table host mirror: the same host buffer is ibv_reg_mr'd for the NIC and NOC-mapped for device DMA. A migration streams through it a window at a time (Begin -> BounceReady -> [WindowReady/WindowAck]* -> Done -> Ack), sender-owned addressing via per-window descriptors, so the receiver holds no table.

The Mooncake wire under this bounce data plane is selectable via --protocol tcp|rdma (env MIGRATION_MOONCAKE_PROTOCOL; --rdma-nic / MIGRATION_RDMA_NICS to pin NICs on a multi-NIC host), defaulting to TCP -- "RDMA-over-host" names the architecture, not the wire, so RDMA is opt-in and needs an RNIC + libibverbs. The bounce data plane code is identical either way; only the installed transport differs (installTransport, TransportProtocol). The e2e harness + scripts (PROTOCOL=tcp|rdma) and the run guide cover both wires for 1->1 and 1->N.

Device I/O is now DRISC NOC-DMA (DriscDeviceIo) over the coexistence UMD open path, wired into the worker (prefill + decode) with the double-pinned staging/ bounce registrar and IOVA-release-before-READY. kv_seed_verify converted to DRISC.

Removals: the MMIO UMD backend (MultiDeviceUmd/UmdDeviceAccess), DeviceDramStorageBackend, the dead KvCacheLayout/ChannelRegion offset math, and the MIRROR_READY control type. DeviceMap unique_id -> device-index resolution kept (buildDeviceIo) with a channel-range self-check.


fivanovicTT (approved):
LGTM

idjuricTT (approved):

PR #4785, opened 2026-07-27
Fix forge model crash on Blackhole (set mesh graph descriptor)

### LINKED ISSUE #4784: Forge (vllm_forge) LLM crashes at device init on Blackhole: TT_MESH_GRAPH_DESC_PATH never set (mesh-descriptor allowlist omits forge runners)
## Summary

A Forge LLM (`vllm_forge` runner) crashes during device init on a Blackhole **P300x2** board because `TT_MESH_GRAPH_DESC_PATH` is never set. tt-metal then detects a CUSTOM cluster type and aborts:

```
TT_FATAL @ tt_metal/llrt/tt_cluster.cpp:281: this->rtoptions_.is_custom_fabric_mesh_graph_desc_path_specified()
Custom fabric mesh graph descriptor path must be specified for CUSTOM cluster type
```

Repro: deploy `Falcon3-7B-Instruct` (forge, P150 spec) on a P300x2 host.

## Root cause

`tt-media-server/utils/runner_utils.py` only applies the Blackhole mesh-graph descriptor for an allowlist of runners:

```python
_RUNNERS_REQUIRING_MESH_DESCRIPTOR = {
    ModelRunners.TT_WHISPER.value,
    ModelRunners.TT_SPEECHT5_TTS.value,
}
if settings.model_runner in _RUNNERS_REQUIRING_MESH_DESCRIPTOR:
    ...
    elif (settings.device or "").lower() in _BH_DEVICE_MESH_DESCRIPTORS:
        _setup_blackhole_mesh_config(tt_metal_home)   # sets TT_MESH_GRAPH_DESC_PATH
```

The forge runners (`vllm_forge`, and the TP variants `vllm_forge_llama_70b` / `vllm_forge_gemma4_31b` / `vllm_forge_qwen_32b`, plus `vllmforge_qwen_embedding`) are **not** in that set, so `_setup_blackhole_mesh_config()` is skipped and `TT_MESH_GRAPH_DESC_PATH` is left unset. The descriptor mapping that would apply already exists in the same file:

```python
_BH_DEVICE_MESH_DESCRIPTORS = {
    "p150": "p150_mesh_graph_descriptor.textproto",
    ...
    "p300x2": "p150_mesh_graph_descriptor.textproto",
}
```

So the machinery is present — forge just isn't wired to it.

## Evidence (from the running container)

- `setup_runner_environment: device_id='0', is_galaxy=False, device_mesh_shape=(1, 1), model_runner='vllm_forge'` — the setup ran, but no "applying blackhole mesh config" line follows (the allowlist gate skipped it).
- Container env at crash: `ARCH_NAME=blackhole`, `MESH_DEVICE=P150`, **no `TT_MESH_GRAPH_DESC_PATH`**.
- Crash occurs in `vllm_tt/worker.py:143` → `torch_xla.device()` → `tt::Cluster::generate_cluster_descriptor()` → `initialize_device_drivers()`.
- Weights download and uvicorn startup succeed; only the EngineCore device init fails, so the container stays "up" while the worker process goes `<defunct>` and never serves.

## Proposed fix (light touch)

Add the forge runner(s) to `_RUNNERS_REQUIRING_MESH_DESCRIPTOR` so the existing Blackhole mesh-config path runs for them. The `_BH_DEVICE_MESH_DESCRIPTORS` device gate already protects non-Blackhole (N150/N300) forge deployments, so there is no regression for Wormhole.

## Environment

- Host board: P300x2 (Blackhole, 4 chips); model requested single P150.
- Image: `ghcr.io/tenstorrent/tt-media-inference-server-forge:0.18.0-c49bb76`
- Artifact / release: 0.18.0

stisiTT:
Forge models crash at startup on Blackhole boards (e.g. a single P150 on a P300x2). tt-metal needs a mesh graph descriptor path set for these, but forge runners were never added to the list of runners that get one — so device init aborts.

This adds the forge runners to that list. The existing device check means nothing changes for Wormhole (N150/N300).

Added two tests: forge on Blackhole now gets the descriptor, forge on Wormhole still doesn't.

Fixes #4784

### DIFF
--- a/tt-media-server/tests/test_worker_utils.py
+++ b/tt-media-server/tests/test_worker_utils.py
@@ -204,6 +204,26 @@ class TestSetupRunnerEnvironment:
 
                             mock_galaxy.assert_not_called()
 
+    def test_calls_blackhole_setup_for_forge_runner(self):
+        """Forge runner on a Blackhole device must get the mesh descriptor (#4784)."""
+        with patch.dict(os.environ, {"TT_METAL_HOME": "/opt/tt-metal"}, clear=True):
+            with patch("utils.runner_utils.set_torch_thread_limits"):
+                with patch("utils.runner_utils.get_telemetry_client"):
+                    with patch(
+                        "utils.runner_utils._setup_blackhole_mesh_config"
+                    ) as mock_bh:
+                        mock_settings_forge = Mock()
+                        mock_settings_forge.enable_telemetry = False
+                        mock_settings_forge.is_galaxy = False
+                        mock_settings_forge.device = "p300x2"
+                        mock_settings_forge.model_runner = "vllm_forge"
+                        mock_settings_forge.default_throttle_level = None
+
+                        with patch("utils.runner_utils.settings", mock_settings_forge):
+                            setup_runner_environment("0")
+
+                            mock_bh.assert_called_once_with("/opt/tt-metal")
+
     def test_custom_cpu_threads(self):
         """Test custom cpu_threads parameter"""
         with patch.dict(os.environ, {}, clear=True):
--- a/tt-media-server/utils/runner_utils.py
+++ b/tt-media-server/utils/runner_utils.py
@@ -52,6 +52,14 @@ def setup_runner_environment(
     _RUNNERS_REQUIRING_MESH_DESCRIPTOR = {
         ModelRunners.TT_WHISPER.value,
         ModelRunners.TT_SPEECHT5_TTS.value,
+        # Forge (tt-xla) inits the device via torch_xla.device(), which aborts on a
+        # Blackhole CUSTOM cluster unless TT_MESH_GRAPH_DESC_PATH is set (see #4784).
+        ModelRunners.VLLMForge.value,
+        ModelRunners.VLLMForge_LLAMA_70B.value,
+        ModelRunners.VLLMForge_GEMMA4_31B.value,
+        ModelRunners.VLLMForge_QWEN_32B.value,
+        ModelRunners.VLLMForge_MISTRAL_SMALL_31_24B.value,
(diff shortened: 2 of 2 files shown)


stisiTT:
## CI evidence — fix confirmed

Ran two tt-shield release runs on this branch (forge-vllm-plugin, Falcon3-7B-Instruct), one on a single `p150` and one on a `bh-qb-ge` (P300-based) host.

Before this change, forge deploys aborted at device init with `Custom fabric mesh graph descriptor path must be specified for CUSTOM cluster type`. In both runs that crash is gone:

- **bh-qb-ge run** hit the exact scenario that used to crash and got through it:
  ```
  Using CUSTOM cluster type for P300 board with 1 chips   (tt_cluster.cpp:202)
  Using custom mesh graph descriptor: .../p150_mesh_graph_descriptor.textproto   (metal_env.cpp:401)
  ```
  Instead of aborting, tt-metal now loads the descriptor and continues.
- **p150 run** applied the same descriptor cleanly.

In both runs the model deployed, served, and ran evals to completion — no device-init crash, no stuck worker.

Both runs still report an overall FAIL, but only because of an eval **accuracy** threshold (`ifeval`: 60.7 and 64.3 vs the reference), which is a separate model-quality issue tracked in #4733. `gpqa` passed. The device-init path this PR fixes is fully exercised and healthy.

Runs:
- p150: https://github.com/tenstorrent/tt-shield/actions/runs/30301371554
- bh-qb-ge: https://github.com/tenstorrent/tt-shield/actions/runs/30301460247

Follow-up (not this PR): on a P300 host the `p300 -> p150` descriptor mapping produces some non-fatal `remote mmio device` warnings; tt-metal main now ships dedicated `p300`/`p300_x2` descriptors that `_BH_DEVICE_MESH_DESCRIPTORS` could point at.

PR #4789, opened 2026-07-28
Improve logging  in kv migration workflow

### LINKED ISSUE #4780: Kafka migration messages lack burst migration_uuid — only opaque per-layer migration_id
A Prefill >> Decode migration is one logical unit under a single burst/migration_uuid (start_burst >> enqueue_migration_in_burst >> finish_burst). That uuid is what the scheduler, sequence, and transport control plane already use.

On the Kafka path, RemoteKVManagerAdapter fans the burst out to N per-layer requests, and RemoteKVManagerImpl::migrate() invents a fresh random migration_id for each Kafka request/ack. MigrationRequestMessage only carries that per-msg id, not the parent burst uuid.

So worker logs, Kafka payloads, and acks cannot be correlated back to the one migration you’re debugging without reconstructing the adapter’s internal kafkaId >> burst map (which never leaves process memory). We need the wire format (and logging) to carry the burst migration_uuid alongside the per-layer id so end-to-end tracing is trivial.

vpetrovicTT:


tt-media-server/cpp_server/src/messaging/migration_message.cpp
```diff
@@ -93,10 +129,24 @@ std::optional<MigrationRequestMessage> parseMigrationRequest(
...
+    return std::nullopt;
+  }
+
+  if (root.isMember(K_KAFKA_REQUEST_ID) && root.isMember(K_MIGRATION_ID) &&
+      !root[K_MIGRATION_ID].isIntegral()) {
+    TT_LOG_ERROR("[migration_message] Request non-integral: migration_id");
+    return std::nullopt;
+  }
+
+  for (const char* field : {"src_slot", "dst_slot", "layer_begin", "layer_end",
```
idjuricTT, line 147: NIce 😄

idjuricTT (approved):

PR #4793, opened 2026-07-28
Update config for Forge's Mistral 3.1 24B

acicovicTT:
Make the Experimental weekly runs of Mistral 3.1 24B in tt-shield CI actually green by explicitly listing known issues. HumanEval is also additionally fixed by threading the required env. var.

Also, explicitly list Mistral's tokenizer as a dependency. It is currently pulled as a dependency of vLLM, but that can be silently dropped at any time.

### DIFF
--- a/tt-media-server/tt_model_runners/forge_runners/requirements.txt
+++ b/tt-media-server/tt_model_runners/forge_runners/requirements.txt
@@ -12,6 +12,8 @@ tabulate
 diffusers==0.37.1
 transformers
 
+mistral-common>=1.0.0 # mistral's tokenizer
+
 datasets # model training
 peft # model training
 
--- a/tt-media-server/tt_model_runners/forge_runners/requirements.txt
+++ b/tt-media-server/tt_model_runners/forge_runners/requirements.txt
@@ -12,7 +12,7 @@ tabulate
 diffusers==0.37.1
 transformers
 
-mistral-common>=1.0.0 # mistral's tokenizer
+mistral-common[image]==1.11.7 # mistral's tokenizer
 
 datasets # model training
 peft # model training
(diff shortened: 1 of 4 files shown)


workflows/model_specs/dev/llm.yaml
```diff
@@ -1383,7 +1383,7 @@ templates:
   device_model_specs:
     - device: SUPER_CLUSTER
       max_concurrency: 64
-      max_context: 262144  # 256 * 1024;
```
mmilosevicTT, line 1386: Change not needed?

mmilosevicTT (approved):
LGTM for `requirements.txt`

workflows/model_specs/dev/llm.yaml
acicovicTT: Automatic stripping of trailing whitespaces. It shouldn't be a problem and the code is cleaner.

tt-media-server/tt_model_runners/forge_runners/requirements.txt
```diff
@@ -12,6 +12,8 @@ tabulate
 diffusers==0.37.1
 transformers
 
+mistral-common>=1.0.0 # mistral's tokenizer
```
fivanovicTT: Can we put an explicit pin to the version instead of allowing newer version?

Reason for this: there might be later a breaking change in the newer versions (we had this issue with WhisperX)

acicovicTT: I pinned it to the version used by vllm today (with the added extra of [image] so we don't miss that later when we onboard the model to multimodal setting)

fivanovicTT (approved):

PR #4803, opened 2026-07-28
Add TTS streaming support to cpp_server

### LINKED ISSUE #4775: Add cpp_server support for TTS audio streaming
Add TTS support to cpp_server so clients can submit text-to-speech requests and receive streamed WAV audio responses.

Scope:
- Register TTS as a model service and expose the endpoint.
- Add TTS request/domain types, preprocessing, prompt compilation, and optional voice sample handling.
- Add worker/service/IPC flow for TTS tasks, audio chunks, and cancellation.
- Wire Blaze TTS runner to the TTS scheduler.

ztorlakTT:
Adds initial C++ server support for MODEL_SERVICE=tts, including /v1/audio/speech routing, request preprocessing, WAV streaming responses, TTS IPC queues, worker-backed service lifecycle, and a Blaze TTS runner that connects to the TTS scheduler.

Also wires TTS build sources, OpenAPI/docs updates, /info metadata, audio codec helpers, prompt compilation, and optional voice sample preprocessing/encoding flow.


tt-media-server/cpp_server/src/runtime/worker/single_process_worker.cpp
```diff
@@ -101,9 +101,35 @@ void SingleProcessWorker::start() {
...
+    } else {
+      auto* taskQueue =
+          std::get_if<std::shared_ptr<tt::ipc::ITaskQueue>>(&cfg.task_queue);
+      auto* resultQueue = std::get_if<std::shared_ptr<tt::ipc::IResultQueue>>(
+          &cfg.result_queue);
+      if (!taskQueue || !resultQueue || !*taskQueue || !*resultQueue ||
+          !cfg.cancel_queue) {
+        throw std::runtime_error(
+            "LLM worker requires LLM task/result queues and a cancel queue");
+      }
```
nanicicTT, line 132: For `MODEL_SERVICE=image`, `WorkerManager` intentionally leaves the generic task, result, and cancel queues empty because the image runner does not use them. This branch treats every non-TTS worker as an LLM worker and throws before creating the image runner. 
We should maybe add an image-specific path, or apply this validation only to services that require these queues.

---

Before this branch, this was possible and ok/legal:
`WorkerManager` creates an image worker without generic queues. The image runner is constructed. The image runner uses its own media transport.

The new code divides workers into only two categories:
1. TTS worker
2. Everything else, which is assumed to be an LLM worker


An image worker would then enter the second category. 
The code then requires generic language-model queues:

```cpp
if (!taskQueue || !resultQueue || !*taskQueue || !*resultQueue ||
    !cfg.cancel_queue) {
  throw std::runtime_error(...);
}
```

Those queues are intentionally null for image workers as it seems, so the worker throws before its image runner is created.
What would happen is that  parent process starts an image worker. Image worker reaches queue validation. Validation fails. Worker exits. The image service never becomes ready.

The correct split would probably need three paths (4 when the video arrives, maybe): text-to-speech, image, and generic language-model or embedding workers.

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_tts_scheduler_factory.cpp
```diff
@@ -0,0 +1,516 @@
...
+  const auto vocab = tt::utils::tokenizers::activeTokenizer().getEncodedVocab();
+
+  engine_tts::TtsSchedulerParams params;
+  params.max_users = static_cast<uint32_t>(config.maxUsers);
+  params.chunk_tokens = config.chunkTokens;
+  params.first_chunk_tokens = config.chunkTokens;
+  params.max_batch_size = static_cast<uint32_t>(config.maxBatchSize);
+  params.speech_end_token = tokenIdFor(vocab, "<|speech_end|>");
+  params.speech_token_base = tokenIdFor(vocab, "<|s_0|>");
+  params.speech_vocab_size = CODEBOOK_SIZE;
```
nanicicTT, line 302: `activeTokenizer()` is selected through `config::modelType()`, but this code adds no text-to-speech model type or tokenizer directory. The default and unknown MODEL values resolve to the `DeepSeek` tokenizer. The production worker therefore searches the wrong vocabulary for required tokens such as `<|speech_end|>` and `<|s_0|>`, throws here, and never completes warmup. A dedicated TTS tokenizer selection must be wired before scheduler construction.

---

The text-to-speech model requires special tokens such as:

```text
<|speech_end|>
<|speech_start|>
<|s_0|>
<|s_1|>
...
```

During scheduler construction, this code loads the active tokenizer vocabulary and searches for those tokens.

The problem is how `activeTokenizer()` chooses the tokenizer. It uses `config::modelType()`. That function only recognizes ordinary language models such as `DeepSeek`, `Llama`, `Kimi`, and `MiniMax` and this code does not add text-to-speech ModelType, text-to-speech tokenizer directory, a separate tokenizer path in `TtsConfig`.

An unknown or missing `MODEL` value falls back to `DeepSeek`. 
During startup text-to-speech worker begins construction. Then `activeTokenizer()` loads the `DeepSeek` tokenizer. Then scheduler searches its vocabulary for `<|speech_end|>` and that token is absent. =>`tokenIdFor()` would throw. Worker then exits before sending its warmup-complete signal and server then reports that the model is not ready.
Even if those token strings are encoded as ordinary text, that would still be wrong. The scheduler needs their exact single-token identifiers.
This also affects `stopTokenIds`. A wrong stop token can terminate speech early when the model generates an otherwise valid speech token.

The fix would be like to make tokenizer selection part of `TtsConfig`, independent of the language-model `MODEL` setting. 
Or do we have a tokenizer that could recognize these special tokens?

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_tts_runner.cpp
```diff
@@ -0,0 +1,539 @@
...
+      maybeFinalizeCompletedSlot(output.slotId);
+    }
+  }
+}
+
+void BlazeTtsRunner::sendFinish(uint32_t taskId,
+                                domain::tts::TtsFinishReason reason,
+                                std::string error) {
+  audioQueue->push(
+      ipc::tts::TtsAudioChunkMessage::finish(taskId, reason, std::move(error)));
```
nanicicTT: `TtsAudioChunkQueue::push` is non-blocking. Sending `finish` ignores its return value. 
If the queue is full only completion, cancellation or error event is discarded. The service then keeps the callback, the HTTP stream never closes and request's capacity remains consumed. 
Terminal even like these need a guaranteed delivery of a deferred retry path. 

Consider this sequence:
- Worker produces audio faster than the server drains it.
- Audio queue reaches its configured capacity.
- Request finishes.
- Worker calls sendFinish().
 - Queue rejects the completion message.
 - Return value is ignored.
 - Worker evicts the scheduler slot and forgets the request.

The server never receives the completion event. Therefore it never removes the callback from §TtsService::callbacks`, never closes the HTTP audio stream, or releases the service-level request capacity.

The client waits indefinitely unless it disconnects or the service stops.

nanicicTT: maybe

Retry completion messages until accepted.
Keep a local list of pending terminal messages.
Use a separate completion queue.
Reserve queue capacity for terminal messages.

ztorlakTT: Good catch, thanks!
I introduced `pendingTerminalMessages` FIFO so terminal events now retry until they are accepted. Also pending terminal events are now drained before regular task processing.

tt-media-server/cpp_server/src/runtime/runners/blaze_runner/blaze_tts_scheduler_factory.cpp
ztorlakTT: We don’t have the tokenizer bundle for the SpeechLM model yet, so for now I added a tokenizerPath to TtsConfig

nanicicTT (approved):
Approving for now, so you merge it when you deem necessary

PR #4809, opened 2026-07-28
Enable weight updates for async RL

abustamanteTT:
This PR introduces changes to enable model weight updates for remote rollouts in asynchronous reinforcement learning. These changes should be transparent to all other use-cases of the inference server, and can only be activated by setting specific environment variables (TT_COLOCATED_INFERENCE, TT_WEIGHT_BRIDGE_DIR), set when launched as a colocated process alongside a training instance with tt-run. 

- Adds POST /v1/internal/weights/update, GET /v1/internal/weights/version, POST /v1/internal/weights/reset_prefix_cache, and POST /v1/internal/shutdown to the live vLLM OpenAI server.
- Adds receiver with `update_weights` to V1 `tt_worker.py`
- Adds tt-metal weight bridge (gated behind environment variable)
- Adds env variable check to `run_vllm_api_server.py`

Future improvements will be required for better enabled data parallelism to improve rollout generation throughput, though these changes will likely fall on the tt-metal side.


vllm-tt-metal/src/weight_update_api.py
```diff
@@ -0,0 +1,266 @@
...
+
+from fastapi import APIRouter, HTTPException, Request
+from pydantic import BaseModel, Field
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/v1/internal/weights", tags=["tt-weight-update"])
+
+# Separate router for non-weights control-plane endpoints (e.g. shutdown) that
+# live directly under /v1/internal rather than /v1/internal/weights.
```
mmilosevicTT: Question: What is the point of shutdown call? In my mind, if you start inference server, you should be able to kill it as well as a process. On the other hand, if you are using already running server, then shutdown should not happen.

Also, this call is not really related to what the PR is adding and that is weights update. I would remove support for this API for now.

vllm-tt-metal/src/weight_update_api.py
```diff
@@ -0,0 +1,266 @@
...
+            "(role='ttt', TTT_RANK)."
+        ),
+    )
+    version: Optional[int] = Field(
+        default=None,
+        description=(
+            "Optional caller-assigned weights/policy version to record. If "
+            "omitted, the server increments its internal counter."
+        ),
+    )
```
mmilosevicTT: I think that version setting should be uniquely defined on one side, not have it on both side and then consolidate.

vllm-tt-metal/src/weight_update_api.py
```diff
@@ -0,0 +1,266 @@
...
+    steps, so the update never interleaves with a single ``execute_model``
+    call. Requests already in flight will, however, finish decoding under a
+    mix of old/new weights. For on-policy correctness the trainer should stop
+    submitting rollouts and let the server drain before calling this endpoint;
+    every response is tagged with the version returned here so stale samples
+    can be detected.
+
+    The caller must coordinate timing: both ranks must reach
+    ``bridge.connect()`` (and the final ``bridge.barrier()``) together, so the
+    trainer should call its send side around the same time as this request.
```
mmilosevicTT: This comment is scary to me. Messing with async requests and expecting that trainer side knows timings of inference server is a bit of waving hands around. Server that is receiving requests needs to make sure that those are finished with the right version of weights.

vllm-tt-metal/src/weight_update_api.py
```diff
@@ -0,0 +1,266 @@
...
+    # internal weight-update / shutdown routes never mount) on a normal server.
+    if os.getenv("TT_COLOCATED_INFERENCE") != "1":
+        return
+
+    import fastapi
+
+    if getattr(fastapi.FastAPI.__init__, "_tt_weight_update_patched", False):
+        return
+
+    original_init = fastapi.FastAPI.__init__
```
mmilosevicTT: This variable is really not needed.

tt-vllm-plugin/tt_vllm_plugin/worker/tt_worker.py
```diff
@@ -291,9 +324,25 @@ def get_mesh_grid(dp_rank=0):
         "N150x4": (1, 4),
         "P150x4": (1, 4),
         "T3K": (1, 8),
+        # P150x8 must be 1x8: tt-transformers uses cluster_shape[0] as the K/dim
+        # axis (must be 1 for single-tray TP=8) and cluster_shape[1] as the
+        # TP/heads axis. A (2,4) shape splits the wqkv K axis and crashes in
+        # TensorSpec; the physical 2x4 tray is opened as a logical 1x8 line.
```
mmilosevicTT: nit: Adding a comment when you haven't changed any code is a bit useless.

vllm-tt-metal/src/run_vllm_api_server.py
```diff
@@ -487,6 +487,19 @@ def set_metal_timeout_env_vars():
...
+    # the full state dict, so the default 5s watchdog fires mid-transfer. Give
+    # the colocated path a much larger budget (or an explicit override).
+    override = os.getenv("TT_METAL_OP_TIMEOUT_SECONDS")
+    if override:
+        timeout_seconds = override
+    elif os.getenv("TT_COLOCATED_INFERENCE") == "1":
+        timeout_seconds = "120.0"
+    else:
+        timeout_seconds = "5.0"
+
```
mmilosevicTT, line 502: Is this safe to do? We will give option for the server to run generation steps for 120s always, no matter the update request, no?

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
```diff
@@ -153,6 +158,281 @@ def load_model(self):
...
+        raise ImportError(
+            "Could not import tt-metal's WeightBridge. Set TT_WEIGHT_BRIDGE_DIR "
+            "to the directory containing inference_bridge.py "
+            "(tt-train/sources/examples/grpo/utils; older checkouts kept it under "
+            "grpo_speedup/utils or named it weight_bridge.py), pass it via "
+            "additional_config['tt']['tt_weight_bridge_dir'], or ensure "
+            "TT_METAL_HOME points at a tt-metal checkout that contains it. This "
+            "MUST be the same module the trainer (sender) imports so both ends "
+            "speak the same wire protocol."
+        )
```
mmilosevicTT: This error message seems like a Claude garbage. Error message should be concise and point out following things:
- Why the code failed?
- Possible workaround/fix for the code.

Writing out full summary of the repo changes is not user friendly.

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
```diff
@@ -153,6 +158,281 @@ def load_model(self):
...
+                continue
+
+        search_dirs: list = []
+
+        def _add_examples_root(examples_root: Path) -> None:
+            # The GRPO example (and its inference_bridge.py) currently lives
+            # under examples/grpo/utils; older checkouts kept it under
+            # grpo_speedup/utils. Search both, newest layout first.
+            search_dirs.append(examples_root / "grpo" / "utils")
+            search_dirs.append(examples_root / "grpo_speedup" / "utils")
```
mmilosevicTT: This IMO should not be a bruteforce search, but you should pin version of trainer that you want to use and update accordingly the changes when needed.

vllm-tt-metal/src/weight_update_api.py
abustamanteTT: The context for the PR is that we would like to add async RL as an option to tt-training-service, which requires that the inference server be started as a colocated process alongside the trainer using tt-run so that the server and trainer share an MPI world. You're correct that I can just kill the process, but then the training service registers the job as a failure, even if it successfully completed the training run. 

That said, I can definitely remove this shutdown API and explore different shutdown options to avoid running into the issue of jobs being incorrectly marked as failed

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
abustamanteTT: Agreed that it should be pinned, but right now this is based on a PR that is still in-flight. Pinning it right now doesn't make sense until that PR has landed. I can delay this PR until the corresponding tt-metal PR has landed if preferred

vllm-tt-metal/src/weight_update_api.py
mmilosevicTT: Could you please elaborate on this? I am not sure how inference-server api is related to training jobs? You have a manager of spun inference-servers and you definitely can just kill them on demand as part of your job cycle.

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
mmilosevicTT: You can always use env variable for this and set a default to what current main does if you need this to be used for some special in-flight PRs 😄

vllm-tt-metal/src/weight_update_api.py
abustamanteTT: For this particular application, I create an inference server just from `run_vllm_api_server.py`, without any supporting infrastructure like the Docker container. The only job manager I'm using in this instance is the training service job manager, which spins up the colocated inference server and accompanying training instance. 

I've removed the shutdown API (commit [76d765e](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/76d765e66d69f97d73cad2758c613e1813a00047)) and I'm working on an alternative approach from the training service side :)

vllm-tt-metal/src/weight_update_api.py
abustamanteTT: Resolved: [76d765e](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/76d765e66d69f97d73cad2758c613e1813a00047)

tt-vllm-plugin/tt_vllm_plugin/worker/tt_worker.py
abustamanteTT: Resolved: [76d765e](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/76d765e66d69f97d73cad2758c613e1813a00047)

Left over from something else I had been looking at

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
abustamanteTT: Oversight, you were correct it was Claude garbage, my mistake. Resolved: [76d765e](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/76d765e66d69f97d73cad2758c613e1813a00047)

vllm-tt-metal/src/run_vllm_api_server.py
abustamanteTT: The env variable setting it to 120s should only ever be set for async RL jobs, but I can find a workaround for this

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_worker.py
abustamanteTT: Fair point, thanks for the suggestion. Commit [dafea20](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/dafea20a3803bd4de44306df2e7690001e92a6e7)

vllm-tt-metal/src/weight_update_api.py
abustamanteTT: Understood, tried to resolve this in commit [95b8225](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/95b82253bd9d7043a1d60748412021ee535f4860)

vllm-tt-metal/src/weight_update_api.py
abustamanteTT: Updated in commit [95b8225](https://github.com/tenstorrent/tt-inference-server/pull/4809/commits/95b82253bd9d7043a1d60748412021ee535f4860)

ztorlakTT (approved):

tstescoTT:
@abustamanteTT can you please run a dispatch run in tt-shield CI (models CI) showing regular model runs are not impacted by these changes? https://github.com/tenstorrent/tt-shield/actions/workflows/on-dispatch.yml

There are also some minor linting and broken unit tests issues in the CI gate.

idjuricTT (approved):

Issue #4790, opened 2026-07-28
Integrate agentic traces benchmarking


ipastalTT:
## Rationale 
We want to compare performance of blaze models against GPUs. 
For this we can use this aiperf [fork](https://github.com/SemiAnalysisAI/aiperf/tree/cquil11/aiperf-agentx-v1.0) or InferenceX directly where it is a submodule e.g. :

```bash
git clone https://github.com/SemiAnalysisAI/InferenceX.git
git submodule sync --recursive
git submodule update --init --recursive

python3.12 -m venv .aiperf-venv
.aiperf-venv/bin/python -m pip install --upgrade pip uv

.aiperf-venv/bin/uv pip install \
  -r utils/agentic-benchmark/requirements.txt \
  -e utils/aiperf \
  "datasets>=4.7.0" \
  "huggingface_hub[cli]>=0.25.0" \
  urllib3 requests
``` 
We should also include SwarmOne traces as well. 
This work should be a new benchmarking `workflow` (e.g. "agentic_traces", "traces_benchmarking", etc.)

## Example Launch Command
Taken from [here](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29741710665/job/88349936173#step:4:974)
```sh
#!/usr/bin/env bash

export AIPERF_DATASET_CONFIGURATION_TIMEOUT=1800
export AIPERF_SERVICE_PROFILE_CONFIGURE_TIMEOUT=1800
export AIPERF_DATASET_WEKA_LIVE_ASSISTANT_RESPONSES=0
export AIPERF_HTTP_TCP_USER_TIMEOUT=900000

exec aiperf profile \
  --scenario inferencex-agentx-mvp \
  --url http://localhost:8000 \
  --endpoint /v1/chat/completions \
  --endpoint-type chat \
  --streaming \
  --model moonshotai/Kimi-K2.7-Code \
  --concurrency 1 \
  --benchmark-duration 3600 \
  --random-seed 42 \
  --failed-request-threshold 0.10 \
  --trajectory-start-min-ratio 0.25 \
  --trajectory-start-max-ratio 0.75 \
  --agentic-cache-warmup-duration 600 \
  --warmup-grace-period 1800 \
  --use-server-token-count \
  --no-gpu-telemetry \
  --tokenizer-trust-remote-code \
  --num-dataset-entries 393 \
  --slice-duration 1.0 \
  --max-context-length 262000 \
[...]

PR #4822, opened 2026-07-29
Ipastal tt/agentic traces

ipastalTT:
Relating to issue https://github.com/tenstorrent/tt-inference-server/issues/4790
## Summary
Adds `agentic_traces`, a benchmark workflow that replays recorded multi-turn agentic coding traces against an already-up OpenAI-compatible server and reports serving latency/throughput under that shape, instead of a synthetic ISL/OSL grid. The load shape (including recorded think time and subagent fan-out) comes from the trace, so it exercises the long-context prefill-dominated pattern real coding agents produce.

The client is the AIPerf fork vendored in [InferenceX](https://github.com/SemiAnalysisAI/InferenceX), which owns the
`inferencex-agentx-mvp` scenario and the SemiAnalysis Weka trace dataset loaders. A dedicated `AGENTIC_TRACES` venv clones that repo at a pinned revision and installs `utils/aiperf` editable, so a reported number is tied to the client that produced it. The checkout is ref-stamped: re-running the same pin skips setup, changing it forces a re-checkout and reinstall.

Runs standalone, or as an opt-in child of a release:

```bash
# standalone
MODEL_SPECS_ENV=dev python launchers/run_agentic_traces.py \
    --model Kimi-K2.7-Code --workflow agentic_traces --device super_cluster \
    --agentic-traces-mode ci --service-port 8000

# as part of a release
python run.py --model Kimi-K2.7-Code --workflow release --tt-device super_cluster \
    --agentic-traces --agentic-traces-mode ci --service-port 8000

### CODE REVIEW

**Review by functionstackx — COMMENTED** [review]
there was an small change made to the agentx harness from @cquil11 today, you gotta need to update the arg commands to the harness

can u ask Artem Yerofieiev to add u to the slack?

File: tests/workflows/test_workflow_dispatch_routing.py
Comment (bgoelTT) [inline tests/workflows/test_workflow_dispatch_routing.py:1]: I'm not sure what is the correct test file to put such a test, but can you add one that checks the AssertionError in `workflows/validate_setup.py` is raised when `--workflow agentic_traces` is used with a ModelSpec that has no registration?

File: reference_config/agentic_traces/agentic_traces_config.py [hunk 1]
```diff
@@ -0,0 +1,353 @@
...
+    # resolving ("reference is not a tree") once that branch is gone, and the
+    # server then refuses it outright as "not our ref".
+    AgenticTracesConfig(
+        model_id="id_tt-transformers_Kimi-K2.7-Code_super_cluster",
+        inferencex_git_ref="e2dcfa91c86936cc011e3be0668eb3b1ca17288f",
+        runs=(
+            AgenticTracesRunSpec(
+                trace_source=TraceSource.INFERENCEX_AGENTX,
+                public_dataset="semianalysis_cc_traces_weka_062126_256k",
+                concurrency=2,
```
bgoelTT: For the H200 DGX I understand why this should be 2, but for SuperCluster we should try to fully max out the server's concurrency

llm_module/drivers/aiperf_agentic_traces.py
```diff
@@ -0,0 +1,638 @@
...
+        # Measured, not requested: confirms the run actually profiled for the
+        # window it was asked to.
+        "measured_benchmark_duration": _stat("benchmark_duration"),
+        # Agentic-specific health signals. context_overflow_count is what the
+        # scenario watches to flip submission_valid (>1% of responses), and is
+        # the direct read-out of whether --max-context-length was set right.
+        "context_overflow_count": _int("context_overflow_count"),
+        "osl_mismatch_count": _int("osl_mismatch_count"),
+        "osl_mismatch_diff_pct": _stat("osl_mismatch_diff_pct"),
+        "theoretical_prefix_cache_hit_pct": _stat("theoretical_prefix_cache_hit"),
```
bgoelTT, line 412: We must also record the measured cache hit rate, I noticed in the vLLM-backed results there was no report of this. How does AgentX measure this?

bgoelTT (changes requested):
A few small comments and one larger one

ipastalTT:
With concurrency 8 on GPU I was having issues with cache hit rate calculation (score was low, probably due to cache size limitation).
With 1 it looks better.

## Tenstorrent Model Release Summary: Kimi-K2.7-Code on SUPER_CLUSTER

### Metadata: Kimi-K2.7-Code on SUPER_CLUSTER

```json
{
    "model_name": "Kimi-K2.7-Code",
    "device": "SUPER_CLUSTER",
    "generated_at": "2026-07-30T08:28:01+00:00",
    "report_id": "Kimi-K2.7-Code_2026-07-30T082801+0000",
    "workflow": "agentic_traces",
    "server_mode": "API",
    "run_command": "python run.py --model Kimi-K2.7-Code --workflow agentic_traces --tt-device super_cluster --dev-mode --server-url http://127.0.0.1 --service-port 8000 --agentic-traces-mode ci --dev-mode",
    "runtime_model_spec_json": "/workspace/ipastal/tt-inference-server/workflow_logs/runtime_model_specs/runtime_model_spec_2026-07-30_08-11-31_id_tt-transformers_Kimi-K2.7-Code_super_cluster_ENDhFpo4.json",
    "model_id": "id_tt-transformers_Kimi-K2.7-Code_super_cluster",
    "model_repo": "moonshotai/Kimi-K2.7-Code",
    "inference_engine": "media",
    "tt_metal_commit": null,
    "vllm_commit": null,
    "model_impl": "tt-transformers"
}
```

### Acceptance Criteria

- Acceptance status: ✅ `PASS`
- Model status: `EXPERIMENTAL`
- Benchmarks: 🟨 `NA` (no blocks present)
- Evals: 🟨 `NA` (no blocks present)
- Spec Tests: 🟨 `NA` (no blocks present)
- All acceptance criteria passed.

---

### Agentic Trace Replay for moonshotai/Kimi-K2.7-Code on SUPER_CLUSTER

**Benchmarking Tool:** the AIPerf fork vendored in [InferenceX](https://github.com/SemiAnalysisAI/InferenceX) pinned at `ddeb02eb9c5c89f44e2e4950e741b499d0b8190a`, driving the `inferencex-agentx-mvp` scenario. Each run replays recorded multi-turn agentic coding traces with their original timing, including subagent fan-out, so the load shape is the trace's rather than a synthetic ISL/OSL grid. Numbers are only comparable across runs on the same client revision and dataset.

#### Per-run Latency (ms)

| Source            | Concur | Mode | TTFT Avg | TTFT P50 | TTFT P90 | TTFT P99 | TTFT Std | TTFOT Avg | TPOT Avg | TPOT P90 | TPOT P99 | E2EL Avg | E2EL P50 | E2EL P99  | CO-Adj E2EL Avg | CO-Adj E2EL P99 |
|:------------------|:-------|:-----|:---------|:---------|:---------|:---------|:---------|:----------|:---------|:---------|:---------|:---------|:---------|:----------|:----------------|:----------------|
| inferencex_agentx | 1      | ci   | 3,613.6  | 1,566.0  | 5,311.7  | 33,674.4 | 11,351.8 | 4,959.9   | 21.1     | 31.5     | 56.3     | 16,618.7 | 5,524.8  | 162,906.1 | 16,621.9        | 162,915.4       |

#### Per-run Throughput & Load

| Source            | Concur | Mode | Out Tok/s/User | E2E Tok/s/User | Output Tok/s | Total Tok/s | Req/s | Prefill Tok/s | Prefill Tok/s (Active) | Decode Tok/s | Decode Tok/s (Active) | Eff. Concur | Tok In Flight Avg | Tok In Flight Max | ISL Mean | OSL Mean | Total In Tok | Total Out Tok |
|:------------------|:-------|:-----|:---------------|:---------------|:-------------|:------------|:------|:--------------|:-----------------------|:-------------|:----------------------|:------------|:------------------|:------------------|:---------|:---------|:-------------|:--------------|
| inferencex_agentx | 1      | ci   | 57.54          | 38.11          | 54.59        | 6,866.03    | 0.082 | 6,811.44      | 26,751.62              | 54.51        | 71.44                 | 1.37        | 193,288           | 294,113           | 82,842   | 664      | 6,130,330    | 49,135        |

#### Run Health & Validity

**Submission** is the scenario's own verdict, not ours: it folds static scenario-lock violations, a context-overflow rate above 1%, and early cancellation into one flag. An invalid run is failed by the driver and is not a reportable number.

| Source            | Concur | Mode | Submission | Reqs OK | Errors | Error % | Ctx Overflow | OSL Mismatch | OSL Diff % | Cache Hit % | Theo. Cache Hit % | Credit Drops | Conn Reuse | Branches | Br. Errored | Br. Truncated | Br. Delayed | Br. Parent Failed | Br. Joins Dropped | Measured (s) |
|:------------------|:-------|:-----|:-----------|:--------|:-------|:--------|:-------------|:-------------|:-----------|:------------|:------------------|:-------------|:-----------|:---------|:------------|:--------------|:------------|:------------------|:------------------|:-------------|
| inferencex_agentx | 1      | ci   | ✅ valid   | 74      | 0      | 0.00    | 0            | 0            | 0.00       | 90.70       | 95.68             | 0            | 0.92       | 4        | 0           | 1             | 0           | 0                 | 0                 | 881          |

#### Run Configuration

Echoed from `reference_config/agentic_traces/agentic_traces_config.py` for the ModelSpec, plus the client's own run identifiers, so a table above can be traced back to what produced it.

| Field                | inferencex_agentx c1 ci                                         |
|:---------------------|:----------------------------------------------------------------|
| Run                  | inferencex_agentx_semianalysis_cc_traces_weka_062126_256k_c1_ci |
| Scenario             | inferencex-agentx-mvp                                           |
| Dataset              | semianalysis_cc_traces_weka_062126_256k                         |
| Dataset Repo         | semianalysisai/cc-traces-weka-062126-256k                       |
| Trace Pool           | 32                                                              |
| Profiling Window (s) | 900                                                             |
| Warmup (req/lane)    | 3                                                               |
| Warmup Grace (s)     | 600                                                             |
| Max Context          | 262,144                                                         |
| Fail Threshold       | 0.10                                                            |
| Trajectory Start Min | 0.25                                                            |
| Trajectory Start Max | 0.75                                                            |
| Random Seed          | 42                                                              |
| Client Ref           | ddeb02eb9c5c89f44e2e4950e741b499d0b8190a                        |
| AIPerf Version       | 0.12.0                                                          |
| Benchmark ID         | 850553223b7f                                                    |
| Started              | 2026-07-30T08:12:53.995567                                      |
| Ended                | 2026-07-30T08:27:54.001007                                      |

**Metric definitions:**
> - **TTFT / TPOT / E2EL**: AIPerf time-to-first-token, inter-token latency, and end-to-end request latency from `profile_export_aiperf.json`. Long-context agentic prefill makes TTFT the headline metric, so its spread is reported.
> - **TTFOT**: time to the first token the user actually sees. On a reasoning model the gap to TTFT is time spent emitting reasoning tokens.
> - **CO-Adj E2EL**: latency corrected for coordinated omission, so it includes the time a request spent waiting to be dispatched. It tracks E2EL until requests start queueing.
> - **Err-Adj TTFT / E2EL**: percentiles with failed requests folded in, so a run that got fast by erroring out cannot look good. Omitted when a run had no errors.
> - **Out Tok/s/User**: decode speed while a request is streaming. **E2E Tok/s/User** divides by whole-request wall clock instead, so it includes prefill and is the user-visible speed of an agentic turn; expect it to be several times lower.
> - **Prefill / Decode Tok/s**: the phase split of serving throughput, averaged over the whole run. The **(Active)** variants only count the windows where that phase was actually working, so the ratio between them is the phase's duty cycle.
> - **Eff. Concur**: concurrency actually in flight. Trace replay honours the recorded think time, so this sits below the requested **Concur** by design; a large shortfall means the intended load was never applied.
> - **Tok In Flight**: KV footprint across in-flight requests. Compare the max against **Max Context**.
> - **Reqs OK / Errors**: successful requests (the sample size behind every latency stat) versus failed ones. **Error %** is AIPerf's `request_error_rate`; the run aborts if it exceeds **Fail Threshold**.
> - **Ctx Overflow**: requests whose trace exceeded **Max Context**. Non-zero means the replay was truncated and no longer matches what was recorded; above 1% the scenario invalidates the submission.
> - **OSL Mismatch / OSL Diff %**: responses whose length did not match the trace's recorded output, and by how much. Non-zero means the replayed conversation diverged from what was recorded.
> - **Cache Hit %**: measured from the serving engine's own counters over the profiling window, so the cache-priming warmup is excluded. **Theo. Cache Hit %** is the reuse inherent to the traces, i.e. the upper bound the engine was offered. A measured rate well below it means the cache was evicting reuse the workload had available. Omitted when the server exposes no such counters.
> - **Credit Drops**: requests the load generator could not dispatch on the trace's schedule. Non-zero means the recorded timing was not reproduced, usually because the server was saturated.
> - **Conn Reuse**: fraction of requests served on a reused connection. Below 1.0 means connections were being torn down, which is what shows up as `ServerDisconnectedError`.
> - **Branches**: subagent trajectories spawned during replay. Any errored, truncated, or delayed child, any parent failed by its child, or any dropped join means the fan-out did not replay as recorded.
> - **Measured (s)**: actual profiling wall clock, which should track **Profiling Window (s)**; a large shortfall means the run ended early.

bgoelTT (approved):

ddjukicTT (approved):

PR #4825, opened 2026-07-30
[feature] - SwarmOne swo-bench trace source for `agentic_traces`

cleonidouTT:
Relating to issue https://github.com/tenstorrent/tt-inference-server/issues/4790

## Summary
Implements `TraceSource.SWARMONE` inside the existing `agentic_traces` workflow, filling the previously stubbed enum value. SwarmOne's [`swo-bench`](https://swarmone.ai) replay engine now runs as a first-class trace source alongside the InferenceX/AIPerf harness — no new top-level workflow. It replays recorded Claude-Code / Codex coding sessions (`swo-bench list-scenarios`) turn-by-turn with growing history, with the replay plan built server-side by the SwarmOne backend.

By default (no `--agentic-traces-sources`), a model configured for both sources runs **both** harnesses; each run is dispatched to the driver for its `trace_source`.

## What's included
- **New driver** (`llm_module/drivers/swo_bench_agentic_traces.py`): builds the `swo-bench replay` CLI and parses swo-bench's `report.metrics` JSON into a flat dict aligned with the AIPerf agentic payload (TTFT, latency, decode/prefill throughput, counts) plus swo-specific fields (session id, source label, version). Always passes `--model-context-length` from the ModelSpec with `--no-resolve-model-context` (the TT Console `/v1/models` omits the window), and keeps the license out of argv.
- **New parser** (`llm_module/parsers/swo_bench_agentic_traces.py`): emits `Block(kind="agentic_traces")` so both harnesses collapse into one report section.
- **Config** (`reference_config/.../agentic_traces_config.py`): swo-bench knobs on `AgenticTracesRunSpec` (`task`, `resident`, `cache_mode`, `history_mode`, `max_tokens`, `max_tokens_mode`) with validation, a per-spec `modes` tuple for mode scoping, source-aware `label`, and relaxed `inferencex_git_ref` (only required when an InferenceX run exists). Registers Kimi SwarmOne runs: **FULL** = all tasks @ c8, **CI** = `sympy-bugfix` @ c1.
- **Run model** (`llm_module/agentic_traces/runs.py`): `SWARMONE` added to `SUPPORTED_TRACE_SOURCES`, mode filtering in `build_runs`, and a request-count-driven `estimated_run_seconds` (CI ~1h / FULL ~8h) since swo-bench isn't wall-clock-bounded.
- **Orchestrator** (`test_module/llm_tests/agentic_traces_tests.py`): dispatch per run by `trace_source`, separate `aiperf_artifacts/` vs `swo_bench_artifacts/` roots, InferenceX pin logged only when relevant.
- **Venv & license**: `swo-bench==3.1.2` pinned in `requirements/agentic-traces.txt`; `setup_agentic_traces` skips the InferenceX clone for swarmone-only configs; `validate_setup.py` fails fast when a SwarmOne license (`SWO_LICENSE_KEY` / `~/.swarmone/license.key`) is missing.
- **Docs & tests**: `docs/workflow_development.md` + `run.py` help updated; new SwarmOne build/CLI/parse/dispatch coverage.

## Design note
Used a per-spec `modes` tuple (FULL/CI scoping) instead of a shared `mode_settings.task` field — a shared mode setting would have leaked SwarmOne's CI `concurrency=1` onto the InferenceX CI run. This keeps the InferenceX runs untouched.

## Usage
```bash
python run.py --model Kimi-K2.7-Code --workflow agentic_traces \
    --agentic-traces-sources swarmone --agentic-traces-mode ci \
    --device super_cluster --server-url http://localhost:8000
```
Requires a SwarmOne license via `SWO_LICENSE_KEY` or `~/.swarmone/license.key`.

## Test plan
- [x] `tests/llm_module/test_agentic_traces.py`, `tests/test_module/llm_tests/test_agentic_traces_tests.py`, `tests/report_module/test_agentic_traces_renderer.py` — 144 passed
- [x] Wider suites (`tests/llm_module`, `tests/workflow_module`, `tests/workflows`, `tests/test_validate_setup.py`, `tests/test_run_arguments.py`) — 645 passed
- [x] No lint errors
- [x] Agentic Trace Replay for Qwen/Qwen3-32B

### DIFF
--- a/llm_module/parsers/swo_bench_agentic_traces.py
+++ b/llm_module/parsers/swo_bench_agentic_traces.py
@@ -11,8 +11,10 @@ kind with the InferenceX/AIPerf parser so the report generator collapses both
 harnesses' rows into a single section; the ``backend`` / ``trace_source`` fields
 distinguish them.
 
-No dedicated renderer is registered: the payload is a flat record, which is what
-the generic renderer expects.
+Sharing the kind also means sharing
+:mod:`report_module.agentic_traces_renderer`, which keys its per-source prose
+and its swo-bench columns off ``trace_source``. A field added here only reaches
+the report once it is listed in one of that module's column groups.
 """
 
 from __future__ import annotations
--- a/llm_module/parsers/swo_bench_agentic_traces.py
+++ b/llm_module/parsers/swo_bench_agentic_traces.py
@@ -1,0 +1,67 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: 2026 Tenstorrent AI ULC
+
+"""Parser for the SwarmOne swo-bench agentic trace-replay driver output.
+
+Consumes the combined payload from
+:class:`llm_module.drivers.swo_bench_agentic_traces.SwoBenchAgenticTracesDriver`
+and emits one :class:`report_module.schema.Block`. It shares the ``agentic_traces``
+kind with the InferenceX/AIPerf parser so the report generator collapses both
+harnesses' rows into a single section; the ``backend`` / ``trace_source`` fields
+distinguish them.
+
+No dedicated renderer is registered: the payload is a flat record, which is what
+the generic renderer expects.
+"""
+
+from __future__ import annotations
+
+import datetime as dt
+from typing import Any, Mapping
+
+from report_module.schema import Block
+
+from .base import LLMResultParser
+
+
+class SwoBenchAgenticTracesParser(LLMResultParser):
+    kind = "agentic_traces"
+
+    def parse(self, raw: Mapping[str, Any], *, device: str = "") -> Block:
+        record = dict(raw)
+        record.setdefault("kind", self.kind)
+        record["model"] = str(raw.get("model_id") or raw.get("model") or "")
+        record["device"] = device
+        record["timestamp"] = _normalize_timestamp(raw.get("date"))
+
+        # error_rate_pct is a percentage; carry a fraction alongside it so it can
+        # be compared to a failed-request threshold without unit confusion.
+        error_rate_pct = raw.get("error_rate_pct")
+        if isinstance(error_rate_pct, (int, float)):
+            record["error_rate"] = float(error_rate_pct) / 100.0
+
+        # Surface run provenance as first-class fields: a replay number is only
+        # reproducible alongside the client version, session, and mode.
+        metadata = raw.get("metadata") or {}
+        if isinstance(metadata, Mapping):
+            for key in ("model_id", "mode"):
+                if key in metadata and key not in record:
+                    record[key] = metadata[key]
+
+        return self._wrap_record(record)
+
+
+def _normalize_timestamp(raw_date: Any) -> str:
+    """Return ``YYYY-MM-DD HH:MM:SS`` (best effort, never raises)."""
+    if isinstance(raw_date, str) and raw_date:
+        for fmt in ("%Y%m%d-%H%M%S", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M:%S"):
+            try:
+                return dt.datetime.strptime(raw_date, fmt).strftime("%Y-%m-%d %H:%M:%S")
+            except ValueError:
+                continue
+        return raw_date
+    return dt.datetime.now(dt.timezone.utc).strftime("%Y-%m-%d %H:%M:%S")
+
+
+__all__ = ["SwoBenchAgenticTracesParser"]
--- a/test_module/llm_tests/agentic_traces_tests.py
+++ b/test_module/llm_tests/agentic_traces_tests.py
@@ -30,6 +30,7 @@ from typing import Optional, Sequence
 from llm_module import ServerConnection
 from llm_module.agentic_traces import (
     build_runs,
+    estimated_run_seconds,
     summarize_runs,
     total_planned_seconds,
 )
@@ -38,7 +39,9 @@ from llm_module.drivers.aiperf_agentic_traces import (
     AgenticTracesDriverResult,
     AIPerfAgenticTracesDriver,
 )
+from llm_module.drivers.swo_bench_agentic_traces import SwoBenchAgenticTracesDriver
 from llm_module.parsers.aiperf_agentic_traces import AIPerfAgenticTracesParser
+from llm_module.parsers.swo_bench_agentic_traces import SwoBenchAgenticTracesParser
 from llm_module.runner import RunnerResult
 from llm_module.server_control import ServerController
 from reference_config.agentic_traces.agentic_traces_config import (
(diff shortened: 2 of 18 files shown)


workflows/validate_setup.py
```diff
@@ -121,16 +136,41 @@ def validate_runtime_args(model_spec, runtime_config):
...
+        sources_arg = getattr(args, "agentic_traces_sources", None)
+        if sources_arg:
+            selected = {
+                part.strip().lower().replace("-", "_")
+                for part in sources_arg.split(",")
+                if part.strip()
+            }
+            swarmone_selected = TraceSource.SWARMONE.value in selected
+        else:
+            swarmone_selected = any(
```
ddjukicTT: This makes a SwarmOne license mandatory for everyone, not just SwarmOne users. When someone doesn't pass --agentic-traces-sources, this falls back to checking whether any run in the model's config is SwarmOne. Kimi's config now has SwarmOne runs, so the plain command that worked yesterday
now stops before doing anything

test_module/llm_tests/agentic_traces_tests.py
```diff
@@ -192,12 +214,13 @@ def run_agentic_traces(
     context = DriverContext(
         output_dir=output_root,
         device=device_label,
-        # The request-bounded warmup has no wall-clock cap, so warmup_grace_period
-        # is the allowance for it rather than a guarantee; the timeout backstops.
+        # Size the single per-run timeout to the longest-running planned run.
```
ddjukicTT: The long SwarmOne timeout ends up applied to the InferenceX run, so a hung InferenceX run sits on cluster way longer

llm_module/parsers/swo_bench_agentic_traces.py
```diff
@@ -0,0 +1,67 @@
...
+"""Parser for the SwarmOne swo-bench agentic trace-replay driver output.
+
+Consumes the combined payload from
+:class:`llm_module.drivers.swo_bench_agentic_traces.SwoBenchAgenticTracesDriver`
+and emits one :class:`report_module.schema.Block`. It shares the ``agentic_traces``
+kind with the InferenceX/AIPerf parser so the report generator collapses both
+harnesses' rows into a single section; the ``backend`` / ``trace_source`` fields
+distinguish them.
+
+No dedicated renderer is registered: the payload is a flat record, which is what
```
ddjukicTT: But there is a dedicated renderer in `report_module/agentic_traces_renderer.py`. Since the SwarmOne parser shares that kind, swo-bench runs go through a renderer written entirely for AIPerf. And every metric definition underneath credits AIPerf. The Submission column shows N/A for a healthy run  "An invalid run is failed by the driver". The swo driver does have its own validity check, it just doesn't set that field.  None of the SwarmOne-specific fields appear anywhere, because they aren't in any of the renderer's column lists

workflows/validate_setup.py
cleonidouTT: Fixed. SwarmOne is now an opt-in trace source, so the plain command behaves exactly as it did before SwarmOne was configured for Kimi.

**What changed**

`OPT_IN_TRACE_SOURCES = (TraceSource.SWARMONE,)` in the config module, consumed by a new `default_run_specs()` that decides what an unnarrowed sweep gets. `resolve_run_specs()` uses it on the no-`--agentic-traces-sources` path, and `validate_setup.py` calls the same function, so planning and validation can't drift apart. The error message now also tells you how to proceed without SwarmOne.

The one exception: a config with *nothing but* SwarmOne runs still runs them by default, since holding back the only configured harness would leave an empty sweep. That case does require the license, unavoidably.

**Resulting behavior**

| Command | Runs | License |
|---|---|---|
| `--workflow agentic_traces` | InferenceX only | not needed |
| `--agentic-traces-sources swarmone` | SwarmOne | required, fails early |
| `--agentic-traces-sources inferencex_agentx,swarmone` | both | required, fails early |

test_module/llm_tests/agentic_traces_tests.py
cleonidouTT: Fixed. Each run now gets a timeout derived from its own plan entry instead of one sweep-wide maximum.

cleonidouTT:
### Agentic Trace Replay for Qwen/Qwen3-32B on console

**Benchmarking Tools:**

**`swarmone` rows:** SwarmOne's [`swo-bench`](https://swarmone.ai/docs/swo-bench) v3.1.2 replay engine. Each run replays recorded Claude-Code / Codex coding sessions turn by turn with growing history, so the prompt is mostly a prefix-cache hit that grows through the session -- the shape a fixed-ISL sweep cannot produce. The replay plan is built server-side by the SwarmOne backend.

Numbers are only comparable across runs from the same source on the same client revision and trace set.

#### Per-run Latency (ms)

| Source   | Concur | Mode | TTFT Avg | TTFT P50 | TTFT P90 | TTFT P99 | E2EL Avg | E2EL P50 | E2EL P99 |
|:---------|:-------|:-----|:---------|:---------|:---------|:---------|:---------|:---------|:---------|
| swarmone | 2      | ci   | 16,550.0 | 18,672.7 | 26,613.8 | 26,899.3 | 16,560.0 | 18,684.3 | 26,913.5 |

#### Per-run Throughput & Load

| Source   | Concur | Mode | Out Tok/s/User | Out Tok/s/User P50 | Output Tok/s | Total Tok/s | Total Tok/s (Active) | Req/s | Prefill Tok/s/Req Avg | Prefill Tok/s/Req P50 | Prefill Tok/s/Req P90 | Concur Avg | Concur Peak | ISL Mean | OSL Mean | Total In Tok | Total Out Tok |
|:---------|:-------|:-----|:---------------|:-------------------|:-------------|:------------|:---------------------|:------|:----------------------|:----------------------|:----------------------|:-----------|:------------|:---------|:---------|:-------------|:--------------|
| swarmone | 2      | ci   | 36.10          | 38.46              | 61.76        | 61.76       | 61.76                | 0.098 | 763.65                | 360.89                | 3,124.60              | 1.62       | 2           | 6,745    | 631      | 107,927      | 10,096        |

#### Run Health & Validity

**Submission** swo-bench publishes no such verdict, so for `swarmone` it is our own completeness check (the run returned successful requests with a plausible TTFT). An invalid run is failed by the driver and never reaches this table.

| Source   | Concur | Mode | Submission | Reqs OK | Errors | Error % | Ready Starved | Pace Idle (ms) | Tool Idle (ms) | Measured (s) |
|:---------|:-------|:-----|:-----------|:--------|:-------|:--------|:--------------|:---------------|:---------------|:-------------|
| swarmone | 2      | ci   | ✅ valid   | 16      | 0      | 0.00    | 0             | 0              | 0              | 163          |

#### Run Configuration

Echoed from `reference_config/agentic_traces/agentic_traces_config.py` for the ModelSpec, plus the client's own run identifiers, so a table above can be traced back to what produced it.

| Field             | swarmone c2 ci                                                          |
|:------------------|:------------------------------------------------------------------------|
| Run               | swarmone_claude-code-cpp-builds-gemma-4-31b_cpp-llvm-then-firefox_c2_ci |
| Scenario          | claude-code-cpp-builds-gemma-4-31b                                      |
| Task              | cpp-llvm-then-firefox                                                   |
| Resident Sessions | 2                                                                       |
| Cache Mode        | realistic                                                               |
| History Mode      | faithful                                                                |
| Max Tokens        | 1,024                                                                   |
| Max Tokens Mode   | flat                                                                    |
| Max Context       | 32,768                                                                  |
| swo-bench Version | 3.1.2                                                                   |
| swo-bench Session | 163b1de44c114e3c81048f5842f5b20c                                        |

**Metric definitions:**
> - **TTFT / E2EL**: time to first token and end-to-end request latency. Long-context agentic prefill makes TTFT the headline metric, so its spread is reported.
> - **Out Tok/s/User**: decode speed while a request is streaming, i.e. how fast a single agent's turn produces text.
> - **Reqs OK / Errors / Error %**: successful requests (the sample size behind every latency stat), failed ones, and the failure share.
> - **Measured (s)**: actual wall clock the run occupied.
> - **Prefill Tok/s/Req**: prefill rate for one request measured over the whole prompt sent, so a prefix-cache hit reads as a very high rate. Capturing that cache-served prefill is the point of the metric -- it is what a fixed-ISL sweep cannot show -- but it is a per-request rate, not a serving-capacity figure, so do not read it against the run-averaged prefill columns.
> - **TPOT** is deliberately absent: swo-bench's inter-token figure is not a usable latency (a live run reported 0.01 ms per token against a measured 38.9 tok/s), so **Out Tok/s/User** and **TTFT** are the latency signals for these rows.
> - **Total Tok/s (Active)**, **Concur Avg / Peak**: throughput and concurrency counted only while requests were in flight, which excludes the replay's simulated tool time. Compare against the requested **Concur** to see how much of it the recorded sessions actually asked for.
> - **Ready Starved**: turns that were ready to dispatch but had no free slot. Non-zero means **Resident Sessions** bounded the run rather than the server.
> - **Pace / Tool Idle (ms)**: time the harness deliberately spent idle reproducing the recorded think and tool-call time. It is inside **Measured (s)** but is not server time.

cleonidouTT:
@ddjukicTT can you please review the fixes?

ddjukicTT (approved):

bgoelTT (approved):

PR #4826, opened 2026-07-30
Wire migration workers to kafka for the partition utilization

vpetrovicTT:


tt-media-server/cpp_server/docs/MIGRATION_WORKER_DEPLOYMENT.md
```diff
@@ -61,6 +61,48 @@ To stop the migration workers tap `Ctrl+C`. This will also delete the remote doc
 
 Attention: Prefill and decode table paths must be accessible by all the migration workers. For exabox deployments, this requires storing them on the /data partition.
 
+## N-prefill exclusive ownership (#4795)
```
dmadicTT, line 64: I think this deployment should be minimal, human-written documentation for other humans deploying the workers.
The fact that there is more than one prefill worker should not be reflected in the way we run the script.
Overall, I would remove these docs or move them to another place if you would keep them

vpetrovicTT: Makes sense. Will change

PR #4840, opened 2026-08-03
Move Mistral 3.1 to nightly

acicovicTT:
Move Mistral 3.1 to nightly. Benchmarks and evals succeeded in the latest weekly (https://github.com/tenstorrent/tt-shield/actions/runs/30695943618/job/91366639302), however the job overall failed because Mistral was missing from the `LLM` list in test config for server tests.

Additionally, I reverted the dependency req. for `mistral-common` to be `>=` instead of `==`. The reason is that nighly runs use the latest pre-release version of `tt-forge`, and if the vLLM plugin shipped with Forge ever updates its `mistral-common` dependency that would create a dependency conflict that would break all Forge runs (image build would be broken).


test_module/server_tests_config.json
```diff
@@ -42,7 +42,8 @@
             "Qwen3-32B",
             "Llama-3.1-8B-Instruct",
             "Llama-3.3-70B-Instruct",
-            "gpt-oss-20b"
+            "gpt-oss-20b",
+            "Mistral-Small-3.1-24B-Instruct-2503"
```
mmilosevicTT, line 46: nit: We should always add new entries at second to last or first place, so that we do not conflict other lines of code, due to trailing commas.

mmilosevicTT (approved):
LGTM

acvejicTT:
Hey folks, looks good to me too - may I ask you to run one on-dispatch to verify it?
https://github.com/tenstorrent/tt-shield/actions/workflows/on-dispatch.yml
You can specify device, model, tt-inference-server branch etc.

idjuricTT (approved):

acicovicTT:
@acvejicTT here is a green run: https://github.com/tenstorrent/tt-shield/actions/runs/31084193101

PR #4850, opened 2026-08-04
Add Single-Chip Trainer Training Runner for LoRA Training

mmilosevicTT:
This PR adds and updates following things:
- Addition of Single-Chip Blacksmith Trainer Training Runner
- Custom Dataset Support in Trainer Runner

Manual tests performed:
- Run of SST2 training on Llama 3.1 8B
- Run of SST2 training on Gemma 1.1 2B
- Run of Custom dataset training on Llama 3.1 8B (With different column_mapping structures)


idjuricTT (approved):

mcupacTT (approved):
LGTM, great work! A couple of comments:

1. val_steps_freq is guarded on tt-blacksmith side as greater than 0, maybe we can validate it on tt-inference-server side as well
2. To add this runner to be the one which is deployed if the model is selected, we should also edit `workflows/model_specs/dev/training.yaml` - but I would leave this for next PR once we test the training

PR #4874, opened 2026-08-06
Add ModelSpecProvider seam between engine and TT model catalog

ipastalTT:
## Summary
- Second decoupling PR for the standalone validation framework extraction (Phase 1), stacked on #4873.
- Introduces `workflow_module.model_catalog` with two protocols: `ModelSpecLike` (the structural surface the engine reads from a resolved spec — name, repo, modalities, cli_args, opaque `device_model_spec`) and `ModelSpecProvider` (`resolve` / `resolve_lenient` / `load_runtime_spec` / `model_names`), plus a process-wide registry injected at entry points.
- The Tenstorrent implementation moves adapter-side to `workflows/model_spec_provider.py` over `MODEL_SPECS` / `get_runtime_model_spec` / `ModelSpec.from_json`.
- Engine call sites no longer import `workflows.model_spec`: `command_factory` (context building + auth-token engine resolution), `run_workflows.py` (CLI model choices + provider registration in `main()`), and the stress tests (`stress_tests_config`, `run_stress_tests`).
- `resolve_lenient` preserves the stress-test fallback semantics (first matching impl with a warning when no `default_impl` exists), which `get_runtime_model_spec` does not offer — verified against the original loop behavior.
- A lazy TT default in `get_model_spec_provider()` keeps pre-extraction callers working; it is explicitly marked as the Phase-2 removal point.

No behavior change: same catalog, same resolution order, same error cases.

## Test plan
- [x] Full suite: `pytest tests/ -q --continue-on-collection-errors` — 1543 passed, 10 skipped (5 `test_helm_generator` collection errors pre-existing on `main`)
- [x] `ruff check` + `ruff format --check` clean on all touched files
- [x] Auth-token tests updated to patch the provider seam instead of `get_runtime_model_spec`
- [x] Smoke: provider resolves 67 model names from the prod catalog

### DIFF
--- a/workflow_module/model_catalog.py
+++ b/workflow_module/model_catalog.py
@@ -1,0 +1,107 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: 2026 Tenstorrent AI ULC
+
+"""Engine-owned model catalog seam.
+
+The engine needs to resolve "which model am I validating" without knowing
+anything about a specific vendor's catalog format. This module defines the
+two protocols that make that possible:
+
+- :class:`ModelSpecLike` — the structural surface the engine reads from a
+  resolved model spec (names, repo, modalities, device-scoped limits).
+- :class:`ModelSpecProvider` — who can resolve ``(model, device)`` into a
+  spec, list the catalog, and load a pre-resolved runtime spec JSON.
+
+The concrete catalog (Tenstorrent's ``workflows.model_spec.MODEL_SPECS``)
+lives in the adapter layer and is injected via
+:func:`register_model_spec_provider` at process entry (``run_workflows.py``,
+``run.py`` dispatch). A lazy Tenstorrent default is kept for backward
+compatibility during the extraction; it is the marked Phase-2 removal point.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, List, Optional, Protocol, runtime_checkable
+
+logger = logging.getLogger(__name__)
+
+
+@runtime_checkable
+class ModelSpecLike(Protocol):
+    """Structural surface the engine reads from a resolved model spec.
+
+    Vendor adapters satisfy this automatically (structural typing); the
+    engine never constructs one itself. Nested vendor-specific details
+    (``device_model_spec``) are opaque to the engine core.
+    """
+
+    model_name: str
+    model_id: str
+    hf_model_repo: str
+    model_type: Any  # engine_types.ModelType
+    inference_engine: Any  # vendor engine taxonomy; engine reads ``.value``
+    cli_args: Dict[str, Any]
+    device_model_spec: Any  # vendor device-scoped limits (max_context, ...)
+    supported_modalities: List[str]
+
+
+@runtime_checkable
+class ModelSpecProvider(Protocol):
+    """Resolves model specs for the engine without exposing the catalog."""
+
+    def model_names(self) -> List[str]:
+        """All selectable model names (for CLI choices / validation)."""
+        ...
+
+    def resolve(self, model: str, device: str) -> ModelSpecLike:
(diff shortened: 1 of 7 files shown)


workflows/model_spec_provider.py
```diff
@@ -0,0 +1,67 @@
...
+        try:
+            return ModelSpec.from_json(path)
+        except (FileNotFoundError, ValueError, json.JSONDecodeError) as e:
+            logger.warning(
+                "Could not load model_spec from runtime_model_spec_json=%r (%s); "
+                "falling back to catalog resolution by (model, device).",
+                path,
+                e,
+            )
+            return None
```
bgoelTT, line 52: Is there a reason to implementing this fallback path? I think the behaviour is confusing and we should instead raise an exception here. If we for whatever reason need a fallback, then it should be the caller's responsibility to catch the raised exception, and retry with a call to `TenstorrentModelSpecProvider.resolve()`

workflow_module/model_catalog.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    model_name: str
+    model_id: str
+    hf_model_repo: str
+    model_type: Any  # engine_types.ModelType
+    inference_engine: Any  # vendor engine taxonomy; engine reads ``.value``
+    cli_args: Dict[str, Any]
+    device_model_spec: Any  # vendor device-scoped limits (max_context, ...)
+    supported_modalities: List[str]
```
bgoelTT, line 49: Is there a reason not to use the exact enums from the real ModelSpec?

workflows/model_spec_provider.py
```diff
@@ -0,0 +1,67 @@
...
+import json
+import logging
+from typing import List, Optional
+
+from workflows.model_spec import MODEL_SPECS, ModelSpec, get_runtime_model_spec
+
+logger = logging.getLogger(__name__)
+
+
+class TenstorrentModelSpecProvider:
```
bgoelTT: We should clearly sub-class the ModelSpecProvider to indicate that this is TT's specific model-catalog-backed provider. Also satisfies the type-hinting you wrote in `workflow_module/model_catalog.py`
```suggestion
class TenstorrentModelSpecProvider(ModelSpecProvider):
```

ipastalTT: Done in 730a8a501 — `TenstorrentModelSpecProvider(ModelSpecProvider)` now subclasses the protocol explicitly.

workflow_module/model_catalog.py
ipastalTT: Partially done in 730a8a501: `model_type` is now `Optional[ModelType]` — that enum is engine-owned (`workflow_module/engine_types.py`) so the protocol can reference it without coupling. `inference_engine` stays `Any` on purpose: the `InferenceEngine` taxonomy lives adapter-side (`workflows/workflow_types.py`), and the engine only ever reads `.value` off it — importing the enum here would re-couple the engine to the TT adapter, which is what the seam is meant to avoid. Happy to revisit if you'd rather move `InferenceEngine` engine-side too.

workflows/model_spec_provider.py
ipastalTT: Agreed — done in 730a8a501. `resolve_lenient` is gone from the seam: the provider now only exposes strict `resolve()` plus `resolve_candidates()` (pure data, no policy). The one caller that relied on the fallback (stress-test parameter extraction) now catches `resolve()`'s `ValueError` and applies its own non-default-impl policy, with the warning at the call site. Note the fallback predates the seam (the pre-refactor inline scan did the same thing), which is why it survived into the provider initially.

workflow_module/model_catalog.py
bgoelTT: Okay that makes sense, I see

bgoelTT (approved):

ddjukicTT (approved):

PR #4882, opened 2026-08-06
acceptance: honor SPEC_TESTS known_issues waivers

### LINKED ISSUE #4881: `known_issues` waivers with `workflow_type: SPEC_TESTS` are ignored — spec-test failures cannot be waived
## Problem

A `known_issues` waiver with `workflow_type: SPEC_TESTS` has no effect on the acceptance verdict. `KnownIssue` supports the type, the v1 acceptance path honored it, and `workflows/model_specs/{prod,dev}/llm.yaml` already ships one for **Llama-3.3-70B-Instruct on P300X2** (waiving `test_penalties`, tracked in #3888) — but it is inert, so any spec-test failure is an unconditional acceptance blocker.

There is no workaround. Model status does not help: `required_target_tiers` gates benchmark tiers only, so EXPERIMENTAL masks nothing here. The only way to get a green run is to drop the model from the test matrix, which is not viable for a model that has opted in from the ttnn side.

## Cause

`report_module/acceptance_criteria.py:79` passes `known_issues` to the evals check only:

```python
categories = [
    _check_benchmarks(schema),
    _check_evals(schema, known_issues),   # only consumer
    _check_spec_tests(schema),            # does not accept the parameter
]
```

`_check_spec_tests` (line 404) is declared `def _check_spec_tests(schema: ReportSchema)`, so every blocking spec block goes straight into `blockers` with no waiver lookup. `_find_waiver` is already generic over the workflow name — only the call is missing.

## Three independent gates

Fixing any one of these alone leaves the run red, so a partial fix produces no observable improvement:

1. `_check_spec_tests` never consults `known_issues`, so the category records a blocker.
2. `task_failure_blockers` adds a second blocker from the suite's non-zero exit; it never sees waivers.
3. the workflow `return_code` counts the waived suite as a failed task, so `run.py` exits 1 and CI shows red even when acceptance passes.

(3) implies a behaviour change worth agreeing on up front: once waivers work, a release can go green while a suite genuinely failed. That is what a waiver means, but it is visible on dashboards.

## Reproduction (no hardware needed)

```python
from report_module.acceptance_criteria import acceptance_criteria_check
from report_module.schema import Block, ReportSchema

block = Block(kind="spec_tests", title="M", data={"status": "fail", "attempts": 1,
              "test_name": "VLLMParamConformanceTest"})
schema = ReportSchema(metadata={}, sections=[block])
waiver = [{"workflow_type": "SPEC_TESTS", "task_name": None, "reason": "known"}]

print(acceptance_criteria_check(schema, waiver)[0])   # -> False
```

Returns `False` for every waiver form: naming a case (`test_penalties`, as shipped), naming the suite, or `task_name: None` which matches every task in the workflow. Even the broadest form does nothing, so this is not a name-matching problem — the lookup never happens.

Confirmed on hardware too: a Falcon3-7B-Instruct release on P150 fails five conformance cases with all five waived, and the run is still red.

## Why the v1 path does not cover it

`workflows/acceptance_criteria.py:_classify_blocker` does map the spec prefixes (`parameter_support_tests`, `server_tests`) to `WorkflowType.SPEC_TESTS`, but that module has no production callers — only tests import it. Its last commit is "Overhaul cleanup: route all workflows to v2 and delete dead v1 code (#4630)", and `v2_bridge` routes LLM `spec_tests` and `release` to v2, so no LLM reaches the v1 checker.

Provenance suggests an oversight rather than a decision: waiver support arrived in ca181aad9, "Fix Llama-3.1-8B v2 release CI: meta-eval BOS, known-issue waivers, and test-harness deps (#4524)", scoped to an eval CI problem.

## Open question: what should `task_name` match?

This needs a decision before the fix lands, because v1 and v2 do not report spec results at the same granularity.

- **v1** emitted one blocker per failing test *function* (`parameter_support_tests.{test_name}.{index}`). Hence `task_name: test_penalties` in the shipped waiver — a pytest function name.
- **v2** emits one block per test *suite*: `test_name = 'VLLMParamConformanceTest'`, with per-function verdicts nested in `parameter_conformance_summary`.

Matching only on the block identifier would not honor the existing `test_penalties` waiver, and would mask a whole suite at once — hiding genuine regressions beside the known one. The PR proposes matching per failing case, falling back to the suite name, and demoting a block only when every failing case is covered.

## Checklist

- [ ] Agree on case-level vs suite-level `task_name` semantics
- [ ] Land the fix (PR to follow)

> FYI @bgoelTT this is followup from your pointer yesterday, I got PR coming next which I've been testing and that solves this.

kmabeeTT:
## Ticket

Fixes #4881

## Problem

A `known_issues` waiver with `workflow_type: SPEC_TESTS` has no effect on the acceptance verdict -- including the one already shipped for Llama-3.3-70B-Instruct on P300X2 (`test_penalties`, #3888). Every spec-test failure is an unconditional acceptance blocker, and there is no workaround: model status does not help (`required_target_tiers` gates benchmark tiers only, so #4830 masks nothing here), leaving "remove the model from the test matrix" as the only way to get a green run.

Three gates each independently keep a waived run red, so all three have to move together:

1. `acceptance_criteria_check` passed `known_issues` to `_check_evals` only, so `_check_spec_tests` recorded a blocker regardless of what was declared.
2. `task_failure_blockers` turned the suite's non-zero exit into a second, waiver-blind blocker.
3. the workflow return code counted the waived suite as a failed task, so `run.py` exited 1 and CI showed red even when acceptance passed.

## Changes

- `report_module/acceptance_criteria.py`: `_check_spec_tests()` takes `known_issues` and demotes a matched block into the existing `waived` bucket; new `_spec_waiver()` / `_failing_spec_cases()` match per failing case (from `parameter_conformance_summary`) falling back to the suite name; new `fully_waived_task_types()`; `task_failure_blockers()` gains `waived_task_types`.
- `workflow_module/execution.py`: `TaskOutcome.is_waived()` holds the single "exempt only if it produced a block and its category was fully waived" rule; `apply_acceptance_criteria()` returns the waived task types alongside the verdict; the return code no longer counts a waived task as failed.
- `workflow_module/workflows.py`: `PrefillDecodeWorkflow.apply_acceptance_criteria` override updated for the widened return.
- `report_module/__init__.py`: export `fully_waived_task_types`.
- `tests/report_module/test_acceptance_criteria.py`: case-level / suite-level / workflow-wide / wrong-workflow / partial-coverage / no-itemised-cases waivers, the exit-code exemption, and the crash-still-blocks and tier-masking-does-not-excuse guards.

## Behaviour change

A release can now go green while a spec suite genuinely failed. That is what a waiver means, and the failure is still reported (`Spec Tests: PASS (0/1 passed, 1 waived)`, with every failing case listed), but it is visible on CI dashboards and worth agreeing on explicitly.

## Not changed

- `CategoryResult.waived` remains one bucket holding both `known_issues` waivers and #4830's status-tier masking. That is why the exit-code exemption is `spec_tests`-only: an eval or benchmark that exits non-zero did so because its runner raised, and tier masking must not excuse that. Splitting `waived` from `informational` would remove the special case and is worth a follow-up.
- `_check_evals` is not refactored to share a waiver helper -- the literally common part is three lines, while the interesting logic (tier masking vs. all-cases-covered) differs.
- `workflows/acceptance_criteria.py` -- still no caller; untouched.

## Test plan

- [x] `pytest tests/report_module/ tests/workflow_module/` -- 339 passing
- [x] **Problem, pre-fix** -- [tt-shield run 31112530808](https://github.com/tenstorrent/tt-shield/actions/runs/31112530808): Falcon3-7B-Instruct release on P150 with a five-case `SPEC_TESTS` waiver declared on the model spec. `Acceptance: FAIL (2 blockers)` (`spec.spec_tests:...` and `task:spec_tests`), `Spec Tests: FAIL`, `rc=1`, job red -- the declared waiver is inert.
- [x] **Fixed, this commit** -- [tt-shield run 31119770401](https://github.com/tenstorrent/tt-shield/actions/runs/31119770401) (`TT-Inference SHA: 5cac1097cee1`): same model / device / docker image, same waiver, plus this fix. That branch also carries the Falcon3 spec-test enablement, which is what makes a spec-test suite run at all; the fix commit itself is byte-identical to `56872e035` on the PR branch. `Acceptance: PASS (0 blockers)`, `Spec Tests: PASS (0/1 passed, 1 waived)`, `rc=0`, `run-tests` job green -- while `test_logprobs`, `test_n`, `test_non_uniform_seeding`, `test_penalties` and `test_seed_reproducibility` all still fail and are still listed in the report.

The fixed run's *overall* conclusion reads `failure` because its `determine-server-type` job was cancelled during the 2026-08-06 GitHub Actions incident; the `run-tests` job that executes the release succeeded, and the verdict above is from its log.

### DIFF
--- a/report_module/__init__.py
+++ b/report_module/__init__.py
@@ -7,6 +7,7 @@ from report_module.acceptance_criteria import (
     acceptance_criteria_check,
     build_acceptance_export,
     format_acceptance_summary_markdown,
+    fully_waived_task_types,
     task_failure_blockers,
 )
 from report_module.generator import (
@@ -26,6 +27,7 @@ __all__ = [
     "SchemaLike",
     "ACCEPTANCE_EXPORT_KEYS",
     "acceptance_criteria_check",
+    "fully_waived_task_types",
     "task_failure_blockers",
     "build_acceptance_export",
     "format_acceptance_summary_markdown",
--- a/report_module/acceptance_criteria.py
+++ b/report_module/acceptance_criteria.py
@@ -18,6 +18,11 @@ KIND_BENCHMARKS = "benchmarks"
 KIND_EVALS = "evals"
 KIND_SPEC_TESTS = "spec_tests"
 
+# TaskOutcome.task_type for the spec-test task, i.e. MediaTaskType.SPEC_TESTS
+# .value. A separate namespace from the block kinds above; not imported from
+# test_module to keep report_module free of that dependency.
+SPEC_TESTS_TASK_TYPE = "spec_tests"
+
 TARGET_LEVELS = ("functional", "complete", "target")
 CHECK_SUFFIX = "_check"
 FAIL_CHECK_INT = 3
@@ -104,7 +109,7 @@ def acceptance_criteria_check(
     categories = [
         _check_benchmarks(schema, model_status),
         _check_evals(schema, known_issues, model_status),
-        _check_spec_tests(schema),
+        _check_spec_tests(schema, known_issues),
     ]
     blockers: Dict[str, str] = {}
     for category in categories:
@@ -114,6 +119,7 @@ def acceptance_criteria_check(
 
 def task_failure_blockers(
     outcomes: Iterable[Tuple[str, int, bool]],
+    waived_task_types: Optional[Iterable[str]] = None,
 ) -> Dict[str, str]:
     """Blockers for workflow tasks whose process exited non-zero.
 
@@ -122,10 +128,16 @@ def task_failure_blockers(
     missing category as ``NA`` rather than a failure. Surfacing the raw task
     exit code here keeps the acceptance verdict consistent with the workflow's
     return code so a crash can never be laundered into a silent ``PASS``.
+
+    ``waived_task_types`` suppresses the exit-code blocker for tasks whose block
+    was fully waived; a task that produced no block always blocks.
     """
+    waived = set(waived_task_types or ())
     blockers: Dict[str, str] = {}
     for task_type, exit_code, produced_block in outcomes:
         if exit_code == 0:
+            continue
+        if produced_block and task_type in waived:
             continue
         detail = (
             "and produced no report block"
@@ -136,6 +148,22 @@ def task_failure_blockers(
             f"Task '{task_type}' failed (exit={exit_code}) {detail}."
         )
     return blockers
+
+
+def fully_waived_task_types(categories: Iterable["CategoryResult"]) -> set:
+    """Task types whose every failure was waived, so their exit code is expected.
+
+    Spec tests only, as policy: other categories' non-zero exits always block.
+    ``waived`` also holds status-tier masking (#4830), which must not excuse a
+    task that exited non-zero.
+    """
+    return {
+        SPEC_TESTS_TASK_TYPE
+        for category in categories
+        if category.name == CATEGORY_SPEC_TESTS
+        and category.waived
+        and not category.blockers
+    }
 
 
 def _find_waiver(
@@ -457,7 +485,53 @@ def _check_evals(
     )
 
 
-def _check_spec_tests(schema: ReportSchema) -> CategoryResult:
+def _failing_spec_cases(block: Block) -> List[str]:
+    """Failing case names from ``parameter_conformance_summary``, if itemised.
+
+    One suite block covers many pytest functions, so a waiver can name one case.
+    """
+    summary = block.data.get("parameter_conformance_summary")
+    if not isinstance(summary, list):
+        return []
+    return [
+        row.get("test_case")
+        for row in summary
+        if isinstance(row, Mapping)
+        and "FAIL" in str(row.get("status", "")).upper()
+        and row.get("test_case")
+    ]
+
+
+def _spec_waiver(
+    block: Block, known_issues: Optional[Iterable[Any]]
+) -> Optional[str]:
+    """Waiver reason for this block, or None.
+
+    Matches a waiver naming the suite, else only when every failing case is
+    individually waived -- so an unwaived regression alongside a known issue
+    still blocks. Cases that are not itemised cannot be waived case by case.
+    """
+    if not known_issues:
+        return None
+
+    suite_name = block.data.get("test_name") or block.data.get("task_name")
+    reason = _find_waiver(known_issues, "SPEC_TESTS", suite_name)
+    if reason is not None:
+        return reason
+
+    reasons = []
+    for case in _failing_spec_cases(block):
+        case_reason = _find_waiver(known_issues, "SPEC_TESTS", case)
+        if case_reason is None:
+            return None
+        reasons.append(f"{case}: {case_reason}")
+    return "; ".join(reasons) if reasons else None
+
+
+def _check_spec_tests(
+    schema: ReportSchema,
+    known_issues: Optional[Iterable[Any]] = None,
+) -> CategoryResult:
     spec_blocks = [
         b
         for b in schema.sections
@@ -469,16 +543,25 @@ def _check_spec_tests(schema: ReportSchema) -> CategoryResult:
         return CategoryResult(CATEGORY_SPEC_TESTS, STATUS_NA, 0, 0)
 
     blockers: Dict[str, str] = {}
+    waived: Dict[str, str] = {}
     failed = 0
     skipped = 0
     na = 0
     for block in spec_blocks:
         test_status = _block_test_status(block)
         if test_status.is_blocking:
-            blockers[f"spec.{_block_key(block)}"] = (
+            block_key = f"spec.{_block_key(block)}"
+            message = (
                 f"{block.title or block.kind} reported status={test_status.value} "
                 f"(attempts={block.data.get('attempts', '?')})"
             )
+            # A model_spec known_issues waiver (workflow_type SPEC_TESTS)
+            # demotes the blocker to a non-fatal waiver, mirroring evals above.
+            reason = _spec_waiver(block, known_issues)
+            if reason is not None:
+                waived[block_key] = f"{message} (waived: {reason})"
+                continue
+            blockers[block_key] = message
             failed += 1
         elif test_status is TestStatus.SKIP:
             skipped += 1
@@ -494,6 +577,7 @@ def _check_spec_tests(schema: ReportSchema) -> CategoryResult:
         na=na,
         skipped=skipped,
         blockers=blockers,
+        waived=waived,
     )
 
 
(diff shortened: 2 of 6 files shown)


kmabeeTT:
## Extra Details for Reviewer

Detail on how the three gates were found, since the sequence explains why the fix spans three layers rather than one.

Gate 1 alone produces no observable improvement, which made this easy to mis-scope. Fixing only `_check_spec_tests` still left [run 31068031268](https://github.com/tenstorrent/tt-shield/actions/runs/31068031268) red with two blockers -- the category blocker was gone, but `task_failure_blockers` added `task:spec_tests` from the suite's exit code.

Fixing that too produced [run 31068038876](https://github.com/tenstorrent/tt-shield/actions/runs/31068038876), which reports `Acceptance: PASS (0 blocker(s))` and still ends `##[error]Process completed with exit code 1` -- the workflow return code is computed as `0 if accepted and not failed_tasks else 1`, and the waived suite was still in `failed_tasks`.

Worth knowing for anyone extending this: spec tests are the only category whose task exits non-zero for a *completed but failed* run (`run_spec_tests` derives its code from the results). Evals and benchmarks return `(0, block)` from dispatch when the runner returns normally, however bad the numbers are -- except the sweep path, which does return a non-zero outcome with a block. That asymmetry is why #4830's category-level masking needed no exit-code handling and this does, and why the exemption is deliberately keyed to `spec_tests` rather than derived from "category was waived".

- Reproducing the CI demos: the waiver those runs exercised is not in this PR, it lives on two throwaway branches. [`kmabee/falcon3_spec_waiver_nofix`](https://github.com/tenstorrent/tt-inference-server/tree/kmabee/falcon3_spec_waiver_nofix) is the pre-fix side -- [`f4896a7f`](https://github.com/tenstorrent/tt-inference-server/commit/f4896a7fa1dd67f423d6945ac5df1f308ff7ef4b) enables `VLLMParamConformanceTest` for Falcon3-7B-Instruct, [`134f7e21`](https://github.com/tenstorrent/tt-inference-server/commit/134f7e21a8010cf0397bb644b6c797ba3da61ab2) and [`cab3d4ea`](https://github.com/tenstorrent/tt-inference-server/commit/cab3d4ea89e2249187eff3dddc8732475388faa4) declare the five-case `SPEC_TESTS` waiver on the P150 device spec, and [`42596c1c`](https://github.com/tenstorrent/tt-inference-server/commit/42596c1cd21cd1dc4584f2af455aec59b58195ef) locally reverts Falcon3's GPU eval reference scores so the run's verdict turns on the spec result rather than a known-failing eval. [`kmabee/falcon3_spec_waiver_fix`](https://github.com/tenstorrent/tt-inference-server/tree/kmabee/falcon3_spec_waiver_fix) is the same branch plus [`5cac1097`](https://github.com/tenstorrent/tt-inference-server/commit/5cac1097cee158a3739e3073475423dde469d538), whose diff is identical to this PR's commit.

Open question for review: `task_name` granularity. v1 emitted one blocker per failing test *function* and its waivers name pytest functions (hence `test_penalties` in the shipped Llama-3.3-70B entry), while v2 emits one block per *suite*. This PR matches per failing case and falls back to the suite name, so the existing waiver keeps working and a suite-wide waiver is still expressible. Happy to change that to suite-only if the per-case behaviour is not wanted -- it would orphan the shipped waiver.

ddjukicTT:
The category counts functional blocks only, but the exit code counts every blocking block, infra included. fully_waived_task_types reads the first and suppresses the second, and since the category's set is a strict subset, "clean apart from waivers" doesn't mean every failure was waived. A red health/infra/.. block is invisible to it.

BaseTest.TASK_TYPE defaults to "infra", so tests are category-invisible unless they opt out, both prerequisite_tests entries included. The exit-code blocker was the only net catching those, and this drops it exactly when a waiver fires,rc=0 and PASS where it was rc=1.

Could fully_waived_task_types require that no blocking spec_tests block of any type is unwaived, rather than that the category is clean?

kmabeeTT:
> Could fully_waived_task_types require that no blocking spec_tests block of any type is unwaived, rather than that the category is clean?

Thanks @ddjukicTT -  Confirmed, and reachable rather than theoretical. `LoggerForkSafetyTest` is the case: a `prerequisite_tests` entry with no engine or category gate, prepended into every suite (`test_filter.py:576-583`), `TASK_TYPE = "unit"`, `break_on_failure: false`. It runs in every LLM release and fails without aborting, so the functional test still runs and still gets waived — invisible to the category, counted by the exit code.

## Fix

Just pushed b464d1d58: `fully_waived_task_types` takes the schema and requires every *blocking* `spec_tests` block, any task type, to match its own waiver, rather than reading the category. One unwaived health/unit/stability/integration failure keeps the exit-code blocker.

That shape is exact, not just tighter. `run_spec_tests` increments `failures` only on paths that also append a block, so `failures` equals the number of blocking `spec_tests` blocks — "every blocking block is waived" fully accounts for the exit code. And it no longer reads `CategoryResult.waived`, so #4830's tier masking can't leak in; the `spec_tests`-only keying stays, as policy.

## Results

Same method as the original pair: two throwaway branches off `kmabee/falcon3_spec_waiver_fix`, each forcing `LoggerForkSafetyTest` to FAIL. [`kmabee/falcon3_infra_hole_prhead`](https://github.com/tenstorrent/tt-inference-server/tree/kmabee/falcon3_infra_hole_prhead) carries `56872e035` alone, [`kmabee/falcon3_infra_hole_fixed`](https://github.com/tenstorrent/tt-inference-server/tree/kmabee/falcon3_infra_hole_fixed) adds `b464d1d58` on top; same model / device / image both sides.

| | [without `b464d1d58`](https://github.com/tenstorrent/tt-shield/actions/runs/31965982386) | [with it](https://github.com/tenstorrent/tt-shield/actions/runs/31965993417) |
|---|---|---|
| `spec_tests done:` | `2 block(s), 2 failure(s) -> exit=1` | same |
| `Spec Tests:` | `PASS (0/1 passed, 1 waived)` | same |
| `Acceptance:` | `PASS (0 blocker(s))` | `FAIL (1 blocker(s))` — `task:spec_tests` |
| `command=release` | `rc=0`, job green | `rc=1`, job red |

Identical inputs, diverging only at the verdict. The category reads `PASS` in both — it cannot see a `unit`-typed block. The failure stays in the report either way (`❌ FAIL | LoggerForkSafetyTest`), it just no longer passes silently.

347 tests passing, was 342, including the first coverage of `run()`'s return-code path — its infra case fails on `56872e035` with `assert 0 == 1`. Replayed the 164 archived report schemas × 6 status/waiver combinations with no verdict drift. Blast radius is the four P300X2 `test_penalties` waivers in the shipped spec.

## Follow-up

Pre-existing and worth its own issue: an infra spec failure never reaches the acceptance *verdict* — the category can read `PASS` while a health/unit test failed, and `task:spec_tests ... exit=1` never names the test.

fivanovicTT (approved):

Issue #4868, opened 2026-08-06
[Helm v0.2] 'inference-server on tt-operator' documentation


sott0n:
Part of #4859.

Write documentation for running tt-inference-server on tt-operator, covering:
- overall architecture (tt-operator = supply side / inference-server = consumer, connected via DRA),
- requirements (k8s 1.33+, DRA feature gate, cert-manager, **tt-operator installed on the cluster**),
- install procedure.

**Frame it as the single DRA path:** tt-operator is a hard prerequisite — there is no `dra.enabled` toggle and no non-DRA (v0.1.0 privileged/hostPath) fallback. Docs should not describe a "without tt-operator" mode.
- Since tt-operator is confirmed **public on GHCR** (2026-08-07, anonymous pull), include install steps that need no auth: `helm install` from `oci://ghcr.io/tenstorrent/helm/tt-operator` (no pull secret required). Note there is no standalone `tt-operator` image — it is an umbrella chart of cluster-singleton components.
- State explicitly that tt-operator is installed **once per cluster by the administrator** (separate lifecycle), and the inference-server chart only emits a `ResourceClaim` — it does not bundle or depend on tt-operator as a sub-chart (see #4859 review point 1).

Update helm-docs accordingly. (Location: README vs docs is TBD within this task.)

Issue #4865, opened 2026-08-06
[Helm v0.2] Remove podAntiAffinity and scope hugepages cleanup per device


sott0n:
Part of #4859.

Assumes the single DRA path (see #4863) — device index N always comes from the DRA allocation, there is no non-DRA fallback.

- Remove the self-rolled `podAntiAffinity` (1 Pod : 1 Node); DRA handles device-collision avoidance, enabling multiple Pods per Node.
- When multiple Pods share a Node, the hugepages cleanup (`rm device_*_tenstorrent`) must be scoped to the pod's own device (`device_<N>_tenstorrent`, where N comes from the DRA allocation) to avoid deleting co-located pods' files.
- Note: the `device_<N>_tenstorrent` naming is a driver (UMD/tt-kmd) contract; confirm it is stable.

Issue #4864, opened 2026-08-06
[Helm v0.2] Add DRA parameters to values.yaml + helm_generator


sott0n:
Part of #4859.

Add DRA-related parameters (e.g. `deviceClassName`, requested device count) to values.yaml, and generate them via `workflows/helm_generator` so values stay in lockstep with the model_spec catalogue.

These parameters are **always emitted** — there is no `dra.enabled` toggle gating them. DRA is the single device-acquisition path (see #4863); tt-operator is confirmed public on GHCR (2026-08-07), so no non-DRA fallback is maintained. The generator does not need to produce any v0.1.0 (privileged/hostPath) variant.

Issue #4859, opened 2026-08-05
[Helm] v0.2.0 Helm Chart — tt-operator integration


sott0n:
# [Helm] v0.2.0 Helm Chart — tt-operator integration

> Parent issue for the v0.2.0 Helm chart. Predecessor: #2782 ([Helm] v0.1.0 Helm Chart)

## Background

- tt-operator has been released as v0.2.0 ([docs](https://docs.tenstorrent.com/cloud-native-support/)). It provides the cluster-side foundation for Tenstorrent workloads: DRA-based device allocation, driver/firmware management, telemetry, etc.
- The current v0.1.0 chart acquires devices with pre-tt-operator stopgaps: `privileged: true` / hostPath `/dev/tenstorrent` / a self-rolled `podAntiAffinity` (1 Pod : 1 Node) / unconditional hugepages.

## Goal

> Replace the v0.1.0 stopgaps with tt-operator's proper mechanisms (DRA) and conditional configuration, enabling **multiple Pods per Node** and a **cleaner Deployment (minimal privileged / hostPath)**.

The chart assumes tt-operator is installed on the cluster; DRA is the **single device-acquisition path** — there is no `dra.enabled` toggle and no v0.1.0 (privileged/hostPath) fallback.

Three pillars:
1. **tt-operator (DRA) integration** — device acquisition uses a `ResourceClaim` (`deviceClassName: tenstorrent.com`), and the main container's `privileged` and hostPath `/dev/tenstorrent` are dropped.
2. **Multiple Pods on one Node** — remove the self-rolled `podAntiAffinity` (1 Pod : 1 Node); DRA handles device-collision avoidance instead.
3. **hugepages-free mode** — make hugepages optional for IOMMU + KMD 1.29.0+ environments, dropping the remaining privileged / hostPath / Recreate along with it.

### In scope for v0.2.0
- [x] Switch device acquisition to DRA `ResourceClaim` (drop main-container privileged / hostPath `/dev/tenstorrent`)
- [x] Add DRA parameters to values.yaml + helm_generator
- [x] Remove `podAntiAffinity`; scope hugepages cleanup per device (multiple Pods per Node)
- [x] hugepages-free mode (`hugepages.enabled` toggle)
- [x] App-metrics PodMonitor (`podMonitor.enabled`, default false, optional)
- [x] "inference-server on tt-operator" documentation (+ helm-docs)
- [x] CI E2E smoke test for the DRA startup path

[...]

PR #4887, opened 2026-08-07
[Helm v0.2] switch device acquisition to DRA ResourceClaim

### LINKED ISSUE #4863: [Helm v0.2] Switch device acquisition to DRA ResourceClaim
Part of #4859.

Replace the v0.1.0 device-acquisition stopgaps with the tt-operator DRA path. This is a **single DRA path** — there is no `dra.enabled` toggle and no non-DRA (v0.1.0) fallback. The chart assumes tt-operator is installed on the cluster.

- Emit a `ResourceClaim` (`deviceClassName: tenstorrent.com`) and reference it from the Deployment (`resources.claims` + `spec.resourceClaims`).
- Drop the main container's `securityContext.privileged: true` and the hostPath `/dev/tenstorrent` volume/mount.

**Why single-path (no toggle):** the only rationale for a two-mode `dra.enabled` toggle was that external users could not access tt-operator. As of 2026-08-07 the tt-operator umbrella chart, its sub-charts, and its component images are all confirmed **public on GHCR** (anonymous pull works), so any user can install tt-operator and use DRA. The consume-only integration ([decision A] — the chart emits a `ResourceClaim` only, tt-operator is installed separately by the cluster admin) is unchanged: DRA drivers are cluster singletons and must not be owned per model release, independent of public/private.

sott0n:
Switch the chart's device acquisition from the v0.1.0 stopgaps (privileged + hostPath `/dev/tenstorrent` + a hardcoded 1-Pod-per-Node `podAntiAffinity`) to **tt-operator DRA**. Each Pod declares a `ResourceClaim`; `tt-dra-driver` allocates Tenstorrent boards and injects the device node(s). Enables **multiple Pods per Node** and an unprivileged, hostPath-free Deployment.

Closes #4863, #4864, #4865. Part of #4859 (docs contribute to #4868, kept open).

## Changes

**DRA device acquisition — #4863**
- New `templates/resourceclaimtemplate.yaml`: one `ResourceClaimTemplate` per Pod
  (`deviceClassName: tenstorrent.com`, board `count`).
- `deployment.yaml`: reference the claim (`resources.claims` + `spec.resourceClaims`);
  remove `privileged` and the hostPath `/dev/tenstorrent`.

**DRA parameters, generated — #4864**
- `values.yaml`: `defaults.dra.deviceClassName`
- Generator + `_helpers.tpl`: top-level `deviceBoardCounts` (device → board count), derived from `device_utils.BOARD_TYPE_COUNT_TO_DEVICE`; `draDeviceCount` reads it.

**Multiple Pods per Node — #4865**
- `_helpers.tpl`: remove the hardcoded 1-Pod-per-Node `podAntiAffinity`.
- `deployment.yaml`: scope hugepages cleanup to the Pod's allocated board (`device_<N>_*`); `Recreate` → `RollingUpdate(maxSurge=0)`.

**CI & docs**
- `test-gate.yml`: assert DRA invariants (no privileged / no hostPath, claim + count); matrix skips devices absent from `deviceBoardCounts`.
- Docs (`README.md.gotmpl`) + chart version → `0.2.0`.

## Scope / future work
- Supported: whole-device shapes (n150, n300, t3k, p-series, galaxy=32).
- Not yet supported: topology-sensitive device group (e.g. `galaxy_t3k`) — needs driver topology attributes; fails render with a clear message. Tracked in #4890.

## Verification
- Validated on real T3K hardware (k3s 1.36 + tt-operator): one Pod at count=4 and four Pods at count=1 both serve inference; DRA release/reallocation verified on churn; `helm install` of this chart end-to-end.
- `helm template` / `kubeconform` / `helm lint` / `pytest` (incl. values-in-sync) all green ✅

### DIFF
--- a/.github/workflows/test-gate.yml
+++ b/.github/workflows/test-gate.yml
@@ -1134,21 +1134,16 @@ jobs:
               echo "FAILED (kubeconform): $model / $engine / $device"
               FAILED=1
             fi
-            # Universal DRA safety invariants — must hold for every device:
-            # no privileged mode and no hostPath on /dev/tenstorrent.
-            if echo "$RENDERED" | grep -qE '^[[:space:]]*privileged: true$'; then
-              echo "::error::privileged:true present in $model/$engine/$device"; FAILED=1; fi
-            if echo "$RENDERED" | grep -qE '^[[:space:]]*path: /dev/tenstorrent$'; then
-              echo "::error::hostPath /dev/tenstorrent present in $model/$engine/$device"; FAILED=1; fi
           done < /tmp/helm-matrix.txt
           exit $FAILED
 
-      - name: Verify DRA claim structure and board counts
+      - name: Verify DRA claim structure, board counts, and safety invariants
         run: |
           set -e
-          # Check the DRA claim structure and the per-device board count across
-          # representative shapes (1 / 4 / 32 boards). The "no privileged / no
-          # hostPath" invariants are enforced for all devices in the matrix step.
+          # Check the DRA claim structure and per-device board count across
+          # representative shapes (1 / 4 / 32 boards), plus the safety invariants
+          # (no privileged, no hostPath on /dev/tenstorrent). The invariants are
+          # device-independent, so checking these representative renders covers them.
           # device:expected-count
           for pair in n300:1 t3k:4 galaxy:32; do
             device="${pair%%:*}"; want="${pair##*:}"
@@ -1163,6 +1158,10 @@ jobs:
               || { echo "::error::$device must request $want board(s) via DRA"; exit 1; }
             echo "$RENDERED" | grep -qE 'resourceClaimTemplateName:' \
               || { echo "::error::Pod does not reference the ResourceClaimTemplate for $device"; exit 1; }
+            echo "$RENDERED" | grep -qE '^[[:space:]]*privileged: true$' \
+              && { echo "::error::privileged:true present for $device"; exit 1; } || true
+            echo "$RENDERED" | grep -qE '^[[:space:]]*path: /dev/tenstorrent$' \
+              && { echo "::error::hostPath /dev/tenstorrent present for $device"; exit 1; } || true
           done
 
   ci-gate:
--- a/.github/workflows/test-gate.yml
+++ b/.github/workflows/test-gate.yml
@@ -1096,6 +1096,13 @@ jobs:
           import yaml
           with open('charts/tt-inference-server/values.yaml') as f:
               v = yaml.safe_load(f)
+          # A device is renderable iff the chart can produce a manifest for it:
+          # it has a DRA board count (deviceBoardCounts), or it is a
+          # non-Tenstorrent device (gpu/cpu, no ResourceClaim). Devices absent
+          # from both (topology-sensitive partitions like galaxy_t3k) fail the
+          # render on purpose — skip them. Derived from the same source the chart
+          # uses (deviceBoardCounts), so it can never drift from the template.
+          supported = set(v.get('deviceBoardCounts', {})) | {"gpu", "cpu"}
           # models[model][engine][device]; skip scalar metadata (e.g. defaultEngine)
           for model, engines in v['models'].items():
               for engine, devices in engines.items():
@@ -1103,6 +1110,8 @@ jobs:
                       continue
                   for device, dev_cfg in devices.items():
                       if not isinstance(dev_cfg, dict):
+                          continue
+                      if device not in supported:
                           continue
                       print(f"{model},{engine},{device}")
           EOF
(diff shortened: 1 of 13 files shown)


idjuricTT (approved):

VeljkoMaksimovic (approved):

vmaksimovicTT (approved):

charts/tt-inference-server/templates/resourceclaimtemplate.yaml
bgoelTT, line 1: Can you please use the Claim Recipes from https://docs.tenstorrent.com/tt-dra-driver/latest/single-host.html#claim-recipes instead of matching on pure device count? That page documents why we can't rely on deviceCount alone

charts/tt-inference-server/templates/deployment.yaml
```diff
@@ -42,18 +37,39 @@ spec:
...
+      {{- end }}
       initContainers:
         - name: fix-cache-permissions
           image: busybox
           command: ["sh", "-c", "chown -R 1000:1000 /cache"]
           volumeMounts:
             - name: cache
               mountPath: /cache
+        # Scoped hugepages cleanup: claim requested here so the DRA driver injects
+        # /dev/tenstorrent/<N>; delete only THIS Pod's board files, not co-located Pods'.
```
bgoelTT, line 55: ❤️

bgoelTT (changes requested):

.github/workflows/test-gate.yml
```diff
@@ -1117,28 +1126,44 @@ jobs:
...
+            if ! echo "$RENDERED" | kubeconform -strict -summary; then
+              echo "FAILED (kubeconform): $model / $engine / $device"
               FAILED=1
             fi
+            # Universal DRA safety invariants — must hold for every device:
+            # no privileged mode and no hostPath on /dev/tenstorrent.
+            if echo "$RENDERED" | grep -qE '^[[:space:]]*privileged: true$'; then
+              echo "::error::privileged:true present in $model/$engine/$device"; FAILED=1; fi
+            if echo "$RENDERED" | grep -qE '^[[:space:]]*path: /dev/tenstorrent$'; then
+              echo "::error::hostPath /dev/tenstorrent present in $model/$engine/$device"; FAILED=1; fi
```
mjeremicTT: I would keep this step just to do Helm template validation including `helm template` and `kubeconform` checks and nothing else. If this is some kind of other validation, we can move it to _helpers.tpl file.

charts/tt-inference-server/templates/_helpers.tpl
```diff
@@ -160,6 +160,26 @@ so two impls on the same device don't share a cache directory.
...
+  - anything else      -> fail (e.g. galaxy_t3k; unsupported, fail closed)
+*/}}
+{{- define "tt-inference-server.draDeviceCount" -}}
+{{- $counts := .Values.deviceBoardCounts | default dict -}}
+{{- $nonTT := list "gpu" "cpu" -}}
+{{- $d := .Values.device | lower -}}
+{{- if hasKey $counts $d -}}
+{{- index $counts $d -}}
+{{- else if has $d $nonTT -}}
+{{- /* non-Tenstorrent device: intentionally no ResourceClaim */ -}}
```
mjeremicTT: Probably not needed, just makes noise

charts/tt-inference-server/values.yaml
```diff
@@ -65,15 +77,29 @@ defaults:
...
+  # User-supplied affinity, applied as-is. v0.2: the chart no longer injects a
+  # 1:1-Pod-per-Node podAntiAffinity — DRA prevents device collisions by
+  # allocation, so multiple Pods may run on one Node (e.g. 4 single-n300 Pods on
+  # a T3K). Add your own nodeAffinity / podAffinity / podAntiAffinity here if
+  # needed.
   affinity: {}
   extraEnv: []
 
+# DRA board count per device — number of Tenstorrent boards a ResourceClaim requests for each device shape.
+# Generated from workflows/device_utils.py:BOARD_TYPE_COUNT_TO_DEVICE; consumed by tt-inference-server.draDeviceCount.
```
mjeremicTT, line 104: It is worth noting somewhere that we don't only mirror model_spec now, but also device_utils as well. Do tests cover potential drift between BOARD_TYPE_COUNT_TO_DEVICE and these values?

charts/tt-inference-server/templates/resourceclaimtemplate.yaml
sott0n: Addressed this comment;

switched to a boardName CEL selector per the Claim Recipes. The claim now matches `device.attributes["tenstorrent.com"].boardName == "<board>"` alongside count, with boardName sourced from the same authoritative map as the count (`device_utils.BOARD_TYPE_COUNT_TO_DEVICE`).

charts/tt-inference-server/values.yaml
sott0n: Yes, `test_values_in_sync` (regenerate + diff) catches drift, and I added a parity test that `deviceBoardCounts` and `deviceBoardNames` cover the same devices.

.github/workflows/test-gate.yml
sott0n: The `no-privileged/no-hostPath` check is an assertion on rendered output, which a `template/_helpers.tpl` can't express (it can only gate on input values, and those fields are already gone from the templates).

So I kept it as a rendered-YAML check but moved it out of the validation step, that step is now `helm template` + `kubeconform` only, and the invariants live in the DRA verification step.

charts/tt-inference-server/templates/resourceclaimtemplate.yaml
bgoelTT: Awesome!!

bgoelTT (approved):

sott0n:
@mjeremicTT could you confirm updates for your comments?

mjeremicTT (approved):

PR #4892, opened 2026-08-07
Add TargetPack seam for reference_config content + per-device eval overrides

ipastalTT:
## Summary
- Introduces workflow_module/target_pack.py with a TargetPack protocol covering every validation-content lookup the engine needs: eval configs, eval-reference resolution + score acceptance, benchmark sweep configs, agentic-traces config lookup/narrowing, and the measured reference-target JSON paths. It's injected process-wide at the entry points, with a lazy Tenstorrent default marked as the Phase-2 removal point. The Tenstorrent implementation (workflows/target_pack_provider.py) adapts the reference_config/ corpus.
- Adds a three-tier device-variance schema to EvalTask: tier-1 semantic identity (task_name, venv, fewshot, seed) is device-invariant; tier-2 transport/execution (max_concurrent, batch_size, use_chat_api, gen_kwargs streaming, model_kwargs) and tier-3 measured baselines (score) can be overridden per device via EvalTask.device_overrides, validated at construction and applied by resolve_task_for_device. This lets e.g. gpt-oss-120b run max_concurrent=38 + chat API + streaming on SUPER_CLUSTER while keeping the GALAXY shape.
- Routes all engine call sites (eval_configs, command_factory, llm_eval_tests, agentic parser, benchmark_configs, agentic_traces_tests, target_check, video-generation tests, run_workflows duration floor) through the pack instead of importing reference_config directly.
## Why
Removes the engine's direct dependency on the Tenstorrent reference_config/ corpus — the last big content coupling — so validation content becomes an injectable adapter concern, and per-device eval differences are expressed as data rather than forked configs.

### DIFF
--- a/workflows/target_pack_provider.py
+++ b/workflows/target_pack_provider.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+"""Tenstorrent implementation of the engine's validation-content seam.
+
+Adapts the ``reference_config/`` corpus (eval configs, benchmark configs,
+agentic-traces configs, measured reference-target JSONs) to
+:class:`workflow_module.target_pack.TargetPack` so the engine never imports
+``reference_config`` directly.
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import Any, Mapping, Optional
+
+logger = logging.getLogger(__name__)
+
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+class TenstorrentTargetPack:
+    """``TargetPack`` over the Tenstorrent ``reference_config/`` corpus."""
+
+    # --- eval configs ---
+    def eval_config(self, model_name: str) -> Optional[Any]:
+        from reference_config.evals.eval_config import EVAL_CONFIGS
+
+        return EVAL_CONFIGS.get(model_name)
+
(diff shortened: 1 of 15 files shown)


workflows/target_pack_provider.py
```diff
@@ -0,0 +1,115 @@
...
+import logging
+from pathlib import Path
+from typing import Any, Mapping, Optional
+
+logger = logging.getLogger(__name__)
+
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+class TenstorrentTargetPack:
```
bgoelTT:
```suggestion
class TenstorrentTargetPack(TargetPack):
```

ipastalTT: Done in 4238f4c — `TenstorrentTargetPack(TargetPack)`.

bgoelTT (approved):

ddjukicTT:
`resolve_task_for_device` uses `dataclasses.replace`, which re-runs `EvalTask.__post_init__` → `_infer_data`. That function is not idempotent, so two documented tier-2 overrides do the wrong thing:

1. EVALS_META `batch_size` is wiped to `None`
2. `use_chat_api: False` leaves `eval_class` on `local-chat-completions`

`max_concurrent` overrides and `use_chat_api: True` happen to survive re-inference, which is why the existing test is green. Please make `_infer_data` idempotent (or re-apply overrides after replace) and add tests for both failing directions.

ipastalTT:
@ddjukicTT should be addressed in 41c2422dd

ddjukicTT (approved):

PR #4916, opened 2026-08-11
[Model-Bringup] Serve DiffusionGemma with vLLM TT plugin

zhenghaoniTT:
## Summary
- serve DiffusionGemma through the standalone vLLM TT plugin on upstream vLLM 0.24 (dev spec, P300X2)
- enforce the 256-token block-output contract end to end: spec metadata → benchmark driver → block-granular metrics → server conformance suite
- benchmark on exact-ISL nvidia/SPEED-Bench text instead of random tokens, so the LLM benchmark table measures real-text serving where the entropy halt fires (~2x decode TPS) rather than the 48-step worst case; other models keep the random sweep
- add GPQA (scored, 67% gate against the 70% GPU baseline) and Terminal Bench coverage with model-owned sampling (no server-side seed), plus mean wall-clock seconds per evaluated sample

## Changes by file

**Serving / infra**
- `workflows/model_specs/dev/llm.yaml` — DiffusionGemma P300X2 dev spec: serving args with one bounded 4K prefill program (avoids long-history trace-recapture OOMs), `output_block_size: 256` metadata, `DISABLE_METAL_OP_TIMEOUT: "1"`, benchmark/eval wiring
- `workflows/model_spec.py` — register the `diffusion_gemma` impl (tt-metal `models/experimental/diffusion_gemma`)
- `.github/workflows/models-ci-config.json` — DiffusionGemma P300X2 nightly + release CI entries
- `requirements/workflow-run-script.txt` — orchestrator venv transformers 4.57.1 → 5.12.1, matching tt-metal's image pin: SPEED-Bench prompt construction loads the DiffusionGemma tokenizer in this venv and its list-valued `extra_special_tokens` crashes transformers 4.x (the image itself inherits the pin from tt-metal's requirements-dev.txt)
- `vllm-tt-metal/src/run_vllm_api_server.py` — apply spec env before the metal-timeout defaults so `DISABLE_METAL_OP_TIMEOUT=1` clears inherited `TT_METAL_*` timeout vars

**Benchmarks**
- `llm_module/speed_bench_prompts.py` (new) — build exact-ISL chat prompts from SPEED-Bench entropy tiers (render/correct/rescue/fill; atomic jsonl write)
- `llm_module/drivers/vllm.py` — for block-output models (`output_block_size > 1`): `--dataset-name custom --skip-chat-template` with the exact-ISL file, reuse the file across sweep points, tag results with `tt_output_block_size`; prompt-construction failures are fatal because a silent random fallback would mislabel SPEED-Bench rows
- `llm_module/parsers/vllm.py` — block-granular metrics (blocks/request, blocks/s, mean block latency) computed on completed requests only (`output_lens` carries 0-rows for failed requests)
- `llm_module/config.py`, `test_module/llm_tests/llm_performance_tests.py` — plumb `output_block_size` from spec metadata into `ServerConnection`

**Evals**
- `reference_config/evals/eval_config.py` — GPQA diamond CoT zeroshot (scored, tolerance 3/70) and `terminal_bench_2_1` tasks with model-owned sampling: `do_sample`, `temperature=1.0`, cleared stop sequences, `propagate_seed_to_gen_kwargs=False`
- `llm_module/eval_command.py` — honor `propagate_seed_to_gen_kwargs=False` for every task type; unsupported lmms-eval combinations raise instead of silently dropping the contract
- `llm_module/lm_eval_no_server_seed.py` (new) — scoped wrapper stripping the hard-coded request seed from both lm-eval API adapters (completions + chat) so `--seed` still drives harness RNGs without reaching the server
- `test_module/llm_tests/llm_eval_tests.py` — sample-count-aware acceptance on subset references plus mean wall-clock seconds per evaluated sample
- shared EVALS_AGENTIC Harbor semantics are intentionally unchanged (a zero-trial Harbor result keeps its historical N/A row for every catalog model); a regression test locks that contract

**Conformance tests**
- `llm_module/test_vllm_diffusiongemma.py` (new) — live-server suite: block width and streaming reconstruction, context-admission gates driven by `--max-context`
- `test_module/llm_tests/vllm_param_conformance_test.py` — `VLLMDiffusionGemmaParamConformanceTest` runs that suite; a missing `max_context` raises instead of silently skipping the admission gates
- `test_module/server_tests_config.json`, `test_module/test_suites/llm.json` — register the suite

**Unit tests** — coverage for all of the above under `tests/`: parsers, driver argv/gating/file reuse, prompt construction (incl. rescue-scan divergence), eval command and seed wrapper, the shared agentic zero-trial N/A regression, metal-timeout env handling, model catalog, CI matrix expansion

## Validation
- unit: `tests/llm_module` 279 passed; full local suite 1549 passed / 10 skipped (QB2; helm/dit modules excluded — optional deps not installed)
- pinned-client verification (QB2): `--skip-chat-template` and the custom-dataset flags exist and plumb through in the vllm==0.13.0 bench client; `output_lens`/`completed` semantics match the completed-only metric bases; the pinned lm-eval fork parses `until=[]` into a real empty list (`stop: []` in payloads)
- exact prompt lengths verified against the real DiffusionGemma tokenizer and dataset at ISL 128 and 4096
- P300X2 GPQA-50 eval via tt-shield: 70% flexible-extract, matching the 70% GPU reference (run 31434385185)
- local P300X2 SPEED-Bench sweep on the `diffusion-gemma-all` head (since merged to tt-metal main): output speed 22→45-67 tok/s for ISL ≤16K, 22→30 tok/s at 32K/64K (reveal-span buckets)

## Dependencies
- plugin: https://github.com/tenstorrent/vllm-tt-plugin/pull/32 (open)
- tt-metal: main — the DiffusionGemma serving runtime merged via tenstorrent/tt-metal#52955; the shared vllm-tt-plugin base landed in #4907

Replaces #4904 (its `zni/diffusiongemma-vllm-tt-plugin-024` head branch was unified back into this one; identical history, zero divergence). Remains draft until vllm-tt-plugin#32 merges.


zhenghaoniTT:
https://github.com/tenstorrent/tt-shield/actions/runs/32145555857

vcankovicTT (approved):

workflows/model_specs/dev/llm.yaml
```diff
@@ -296,6 +296,76 @@ templates:
...
+        # hook runs. DiffusionGemma owns whole-prompt prefill/KV and emits
+        # synchronous 256-token blocks, so reject the defaults at CLI parsing.
+        no-enable-prefix-caching: true
+        no-enable-chunked-prefill: true
+        no-async-scheduling: true
+        # Deliberately omit --reasoning-parser for scored GPQA runs. vLLM 0.24
+        # honors the parser's request to retain <channel|> delimiters, which can
+        # move the boxed answer from message.content into message.reasoning.
+        # lm-eval scores content only; parser-off preserves the validated answer
+        # bytes while keeping the model's thinking chat template enabled.
```
bgoelTT, line 357: Don't you need to enable `--tool-call-parser` by default to make the agentic evals pass?

Also, we do want to enable `--reasoning parser gemma4` by default. Can you explain more about this GPQA-specific problem and why other evals don't have this issue?

workflows/model_specs/dev/llm.yaml
```diff
@@ -296,6 +296,76 @@ templates:
...
+        # After a 118-turn Terminal-Bench trial fragmented DRAM, its 23 MiB
+        # per-bank allocation killed EngineCore despite 1.30 GiB total free.
+        # Halving the bounded inner slice keeps the same per-token computation
+        # while reducing that contiguous allocation to 11 MiB per bank.
+        DG_PREFILL_RAGGED_CHUNK: "1024"
+        DG_DENOISE_REVEAL_PMAX: "262144"
+        DG_DENOISE_SLIDING_WINDOW: "1"
+        # Warm representative short lengths before capturing the shared program.
+        DG_UPFRONT_PREFILL_WARMUP_LENS: "32,64,96"
+        DG_TRACE_REGION_SIZE: "3758096384"
```
bgoelTT, line 338: How is this different from specifying the trace region size in the `override_tt_config`?

vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile
```diff
@@ -103,7 +103,20 @@ RUN /bin/bash -c "git clone --depth 1 https://github.com/tenstorrent-metal/tt-me
...
+             fi; \
+             resolved_count=\$(printf '%s\n' \"\${resolved_shas}\" | awk 'NF {count++} END {print count + 0}'); \
+             if [ \"\${resolved_count}\" -ne 1 ]; then \
+                 echo 'Unable to resolve plugin ref ${TT_VLLM_COMMIT_SHA_OR_TAG} to one remote commit' >&2; \
+                 exit 1; \
+             fi; \
+             resolved_sha=\$(printf '%s\n' \"\${resolved_shas}\" | awk 'NF {print; exit}'); \
+             git fetch --depth 1 origin \"\${resolved_sha}\" \
+             && git checkout --detach \"\${resolved_sha}\"; \
+           }) \
```
bgoelTT: Is this strictly required for this PR or is it a quality of life improvement?

vllm-tt-metal/requirements.txt
```diff
@@ -1,3 +1,5 @@
 # additional inference server and client script requirements
 pyjwt==2.7.0
 requests==2.32.3
+# Image-wide vLLM 0.24 pin; 5.12.1 is required for native DiffusionGemma.
+transformers==5.12.1
```
bgoelTT: This should really be coming from tt-metal's requirements.txt, I'm concerned this will break other models. In fact, I can already see this is the pinned version in tt-metal so is this still required? https://github.com/tenstorrent/tt-metal/blob/3007bc45e2f9896c1125e8aeb139d7828d5b47e7/tt_metal/python_env/requirements-dev.txt#L33

Also, in vLLM 0.24.0 requirements, I can see transformers is unpinned and defines a range `>=5.5.3` https://raw.githubusercontent.com/vllm-project/vllm/v0.24.0/requirements/common.txt

workflows/model_specs/dev/llm.yaml
```diff
@@ -296,6 +296,76 @@ templates:
...
+      # prefill program.
+      # DeviceModelSpec derives block_size=64, max_num_seqs=1 and
+      # max_num_batched_tokens=max_context from these fields.
+      max_context: 262144
+      default_impl: true
+      env_vars:
+        MESH_DEVICE: "P150x4"
+        VLLM_RPC_TIMEOUT: "1800000"
+        VLLM_ENABLE_V1_MULTIPROCESSING: "0"
+        DISABLE_METAL_OP_TIMEOUT: "1"
```
bgoelTT, line 318: Are you disabling the metal OP timeout because this falsely triggers tt-triage during kernel compilation?

vllm-tt-metal/src/run_vllm_api_server.py
```diff
@@ -770,9 +777,11 @@ def main():
     impl_id = model_spec.get("impl", {}).get("impl_id")
     register_tt_models(impl_id)
 
-    # Step 4: Set runtime environment variables and vLLM server args
-    set_metal_timeout_env_vars()
+    # Step 4: Set runtime environment variables and vLLM server args. Model env
+    # must be installed first because it may explicitly disable the generic
+    # five-second Metal operation timeout.
```
bgoelTT: Nit: don't want to over explain
```suggestion
    # Step 4: Set runtime environment variables and vLLM server args
```

llm_module/drivers/vllm.py
```diff
@@ -82,18 +77,55 @@ def build_vllm_bench_serve_argv(
         str(result_filename),
     ]
 
-    if uses_remote_base_url(server.url_with_port, server.is_remote):
+    if custom_dataset_path is not None:
+        # Pre-built exact-ISL prompts (SPEED-Bench text for block-granular
+        # models): real language reaches the entropy halt, so the sweep
+        # reports serving behaviour instead of the 48-step worst case that
+        # random token salad pins every block to.
```
bgoelTT: This path will be generic to future models, so please don't comment about this specific change
```suggestion
```

llm_module/drivers/vllm.py
```diff
@@ -123,11 +155,13 @@ def run(
             f"_maxcon-{config.max_concurrency}_n-{config.num_prompts}.json"
         )
 
+        custom_dataset_path = self._maybe_speed_bench_prompts(config, server, context)
```
bgoelTT: We shouldn't be choosing the dataset to run from within the driver. The driver should naively run what it is told to do.

Can you add `custom_dataset_path` as a new attribute in the `LLMRunConfig` object with a default value of `None`? Then the selection of the custom dataset path will be done during the creation of the benchmark config in https://github.com/tenstorrent/tt-inference-server/blob/c83d52279e999950e99342fc404c32a548704308/llm_module/benchmark_configs.py#L17, which won't need to rely on inferring that this model is diffusiongemma from this output block size metadata on the `ServerConnection` object as it has access to the ModelSpec

llm_module/config.py
```diff
@@ -49,6 +49,9 @@ class ServerConnection:
     # repo (e.g. moonshotai/Kimi-* ships a custom tokenizer). Driven per
     # model from the spec metadata; off by default for safety.
     tokenizer_trust_remote_code: bool = False
+    # Models that commit more than one token per engine step use block-level
+    # latency/throughput as their primary benchmark semantics.
+    output_block_size: int = 1
```
bgoelTT: From what I am reading, this `output_block_size` parameter is being used to attach metadata into the driver's output that the server it is connecting to returns more than one token per engine step.

If so, then I think the correct place to put this attribute is in the `LLMRunConfig`, not the `ServerConnection`. As the docstring states, this class should just be used to inform how the driver reaches the server, not for saying what the server is running or how it's configured.

llm_module/eval_command.py
```diff
@@ -277,6 +278,25 @@ def build_eval_command(
...
+                "cannot use"
+            )
+        # The pinned lm-eval adapters always copy their model seed into OpenAI
+        # payloads, independently of --gen_kwargs. Use the scoped wrapper so
+        # --seed still controls harness RNGs without reaching the server.
+        lm_eval_prefix = [
+            str(task_venv_config.venv_path / "bin" / "python"),
+            str(Path(__file__).with_name("lm_eval_no_server_seed.py")),
+        ]
+
```
bgoelTT, line 301: Overall I think this is okay to path at runtime, however, the correct fix is to make the vLLM diffusiongemma server be resistant / ignore the seed parameter that is being passed into the generation kwargs.

What is happening here is a workaround to the lack of this feature. Please keep this for now but file an issue and past the link into the comment
```suggestion
    lm_eval_prefix = [str(lm_eval_exec)]
    # TODO: remove this once diffusiongemma vLLM can ignore the seed gen kwarg <INSERT ISSUE LINK>
    if not getattr(task, "propagate_seed_to_gen_kwargs", True):
        if task.workflow_venv_type in [
            WorkflowVenvType.EVALS_VISION,
            WorkflowVenvType.EVALS_AUDIO,
        ]:
            raise ValueError(
                f"propagate_seed_to_gen_kwargs=False on {task.task_name} needs "
                "the lm-eval no-server-seed wrapper, which lmms-eval tasks "
                "cannot use"
            )
        # The pinned lm-eval adapters always copy their model seed into OpenAI
        # payloads, independently of --gen_kwargs. Use the scoped wrapper so
        # --seed still controls harness RNGs without reaching the server.
        lm_eval_prefix = [
            str(task_venv_config.venv_path / "bin" / "python"),
            str(Path(__file__).with_name("lm_eval_no_server_seed.py")),
        ]

```

bgoelTT (changes requested):
Looking great, just a few comments to reduce some tech debt

vllm-tt-metal/requirements.txt
zhenghaoniTT: dropped

vllm-tt-metal/vllm.tt-metal.src.dev.Dockerfile
zhenghaoniTT: For improvement and removed from this PR

vllm-tt-metal/src/run_vllm_api_server.py
zhenghaoniTT: shortened to that one line.

llm_module/drivers/vllm.py
zhenghaoniTT: Removed.

llm_module/config.py
zhenghaoniTT: Moved to `LLMRunConfig`. `ServerConnection` is back to how the driver reaches the server.

llm_module/drivers/vllm.py
zhenghaoniTT: Done. `LLMRunConfig` now has `custom_dataset_path: None`. `get_llm_configs` selects it from ModelSpec `output_block_size` (custom SPEED-Bench file when `> 1`, otherwise random). 

The driver just runs the path it is given; it only writes the file if that path is set and missing.

workflows/model_specs/dev/llm.yaml
zhenghaoniTT: override_tt_config.trace_region_size is the actual reservation
DG_TRACE_REGION_SIZE mirrors it for the model runtime
And we need tt-metal changes if we want collapsing them into one variable
So I planned as a follow-up alongside the tt-metal CI work.

llm_module/eval_command.py
zhenghaoniTT: Yes and i keep the wrapper for now. Tracking the real fix (DiffusionGemma vLLM should ignore the seed gen kwarg) in https://github.com/tenstorrent/tt-inference-server/issues/4993.

workflows/model_specs/dev/llm.yaml
zhenghaoniTT: yes

workflows/model_specs/dev/llm.yaml
zhenghaoniTT: Terminal Bench does not need the split: terminus-2 reads JSON from the text in content.

zhenghaoniTT: Official vLLM recipes for DiffusionGemma do enable `--reasoning-parser gemma4` (same parser as Gemma 4). We omit it here because that flag is an API split, not part of generation.

`enable_thinking` is already on, so the model still writes the thought channel. `--reasoning-parser gemma4` only moves thought into `reasoning_content` and the response channel into `content`. The official recipe turns that on for interactive clients that read both fields.

Our scored client does not. GPQA uses lm-eval `flexible-extract` on `message.content` only; `capture_reasoning` logs `reasoning_content` but does not score it. DiffusionGemma emits 256-token canvases and often never leaves the thought channel, so the parser can put the boxed `(X)` in `reasoning_content` and leave `content` empty. That zeros the GPQA score. The 70% GPU GPQA baseline in this repo was also collected with the parser off.

Gemma 4 can keep the parser on because it reliably switches to the response channel, so `content` still holds the answer after the split.

`--tool-call-parser` is the same story: Terminal Bench here uses terminus-2 with `parser_name: json` (ICL from `content`), not `tool_choice: "auto"`. We can add `--enable-auto-tool-choice` + `--tool-call-parser gemma4` later if we onboard a native tool client.

llm_module/drivers/vllm.py
```diff
@@ -136,7 +165,40 @@ def run(
...
+                )
+            except Exception as build_error:
+                raise RuntimeError(
+                    "SPEED-Bench prompt construction failed for "
+                    f"{server.model} isl={config.isl}; refusing to run a mislabeled "
+                    "random-input benchmark"
+                ) from build_error
+        if resolved != path:
+            return replace(config, custom_dataset_path=resolved)
+        return config
```
bgoelTT: LLMRunConfig modification should be done inside of `llm_module/benchmark_configs.py`, can this be moved there or does it need the driver's context?

bgoelTT (changes requested):

llm_module/drivers/vllm.py
zhenghaoniTT: moved to llm_module/benchmark_configs.py

bgoelTT (approved):

PR #4919, opened 2026-08-12
[Helm v0.2] Use startupProbe for the model compile window (#4870)

### LINKED ISSUE #4870: [Helm] Use startupProbe for model compile window (accurate readiness, safe long compiles)
Part of #4859

## Summary

Replace the current "large `initialDelaySeconds` on liveness+readiness" heuristic with a proper `startupProbe`, so that (a) the pod reports `READY 1/1` as soon as the server actually serves, and (b) long model compiles are tolerated correctly regardless of how long they take.

## Current behavior

The chart gates both probes with a big per-model `initialDelaySeconds` (default `2400`, overridden per model in `values.yaml`):

```yaml
# charts/tt-inference-server/templates/deployment.yaml
livenessProbe:   { httpGet: {path: /health}, initialDelaySeconds: {{ ... }}, periodSeconds: 30, failureThreshold: 3 }
readinessProbe:  { httpGet: {path: /health}, initialDelaySeconds: {{ ... }}, periodSeconds: 10, failureThreshold: 3 }
```

The real driver is liveness: during the model compile/warmup (up to `progressDeadlineSeconds`, 5400–7200s) `/health` is not yet listening, so a liveness probe that fired mid-compile would restart the container and loop forever. v0.1.0 avoided this by pushing `initialDelaySeconds` past the expected compile time. There is no `startupProbe` anywhere in the chart and no design note explaining the choice — it looks like an MVP expedient.

## Problems with the current approach

1. **`READY 0/1` even when serving.** `initialDelaySeconds` only delays the *first* probe. Observed on a LoudBox (T3K): Qwen3-32B finished compiling and `/health` returned `200` at ~13 min, but the pod stayed `0/1` because the first readiness probe is not scheduled until 2400s (~40 min). This is confusing operationally and makes readiness-gated rollouts/Services wait ~40 min unnecessarily.

2. **Latent restart risk for slow compiles.** `initialDelaySeconds` does not bound total compile tolerance. If a compile runs longer than the delay, liveness starts probing while `/health` is still refusing connections and, after `failureThreshold(3) × periodSeconds(30) = 90s`, restarts the pod. This mismatch already exists in `values.yaml`: e.g. `gpt-oss-120b`/`Qwen2.5-72B` have `progressDeadlineSeconds: 7200` but liveness `initialDelaySeconds: 3600` — a compile exceeding 60 min would be killed.

## Proposed change

Introduce a `startupProbe` that owns the long compile window; while it is running, kubelet disables liveness/readiness. After it succeeds once, liveness/readiness take over with normal short settings:

```yaml
startupProbe:
  httpGet: { path: /health, port: http }
  periodSeconds: 15
  failureThreshold: 480          # 15s × 480 = 7200s; size to max compile (≈ progressDeadlineSeconds)
readinessProbe:
  httpGet: { path: /health, port: http }
  initialDelaySeconds: 0
  periodSeconds: 10
livenessProbe:
  httpGet: { path: /health, port: http }   # /tt-liveness for non-vllm engines, as today
  initialDelaySeconds: 0
  periodSeconds: 30
```

Benefits:
- Pod flips to `READY 1/1` right after the server starts serving (accurate readiness).
- Slow compiles are tolerated up to `startupProbe.failureThreshold × periodSeconds`, and a hung compile is still bounded (fails startup instead of looping liveness).
- Per-model tuning collapses to a single knob (`startupProbe` budget) instead of hand-syncing liveness/readiness `initialDelaySeconds` against `progressDeadlineSeconds`.

Notes:
- `startupProbe` is GA since k8s 1.20 (cluster here is v1.36) — no image or app change needed; `/health` already exists.
- Keep the existing `/tt-liveness` vs `/health` path selection for non-vllm engines.

## Scope

- [ ] Add `startupProbe` to `deployment.yaml`; set liveness/readiness `initialDelaySeconds` to 0 (or small).
- [ ] Add `probes.startup` config to `values.yaml` defaults + per-model override plumbing via `helm_generator` (derive the startup budget from `progressDeadlineSeconds`).
- [ ] Update `helm template`/kubeconform CI expectations.
- [ ] Verify on real hardware (LoudBox T3K) that the pod reaches `1/1` shortly after `/health` returns 200, and that a compile longer than the old `initialDelaySeconds` no longer restarts the pod.

## Acceptance criteria

- A freshly installed model release reports `READY 1/1` within one `readinessProbe.periodSeconds` of `/health` first returning 200 — not after a fixed multi-minute delay.
- A model whose compile exceeds the previous `initialDelaySeconds` starts successfully without a liveness-induced restart.

sott0n:
Part of #4859. Stacked on #4887 — merge that first.

Replaces the v0.1.0 large-`initialDelaySeconds` heuristic with a `startupProbe` that owns the compile/warmup window. The Pod now reports `READY 1/1` as soon as it serves (not after a fixed delay), and a long compile no longer risks a liveness restart.

`failureThreshold` is derived in the template as `ceil(progressDeadlineSeconds / periodSeconds)`, so the budget tracks the max compile time with no per-model probe timing to hand-sync. Per-model `initialDelaySeconds` is removed; the generator's `HelmProbe` is now path-only.

Verified on T3K (Qwen3-8B/n300): Pod reached `READY 1/1` ~12s after warmup vs the ~40 min fixed wait under v0.1.0.

Closes #4870.

### DIFF
--- a/charts/tt-inference-server/templates/deployment.yaml
+++ b/charts/tt-inference-server/templates/deployment.yaml
@@ -2,6 +2,8 @@
 {{- $cfg := include "tt-inference-server.resolvedConfig" . | fromYaml }}
 {{- $engine := include "tt-inference-server.resolvedEngine" . }}
 {{- $ttCount := include "tt-inference-server.draDeviceCount" . }}
+{{- $startupPeriod := $cfg.probes.startup.periodSeconds | int }}
+{{- if le $startupPeriod 0 }}{{- fail "probes.startup.periodSeconds must be greater than 0 (it divides progressDeadlineSeconds to size the startupProbe budget)." }}{{- end }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -117,6 +119,20 @@ spec:
               - name: tt
             {{- end }}
           {{- end }}
+          # startupProbe owns the compile/warmup window: while it runs the kubelet
+          # suppresses liveness/readiness, then they take over immediately. The
+          # budget = failureThreshold × periodSeconds is derived from
+          # progressDeadlineSeconds so it always matches the max compile time; it
+          # flips to serving as soon as one poll succeeds, not after a fixed delay.
+          # Always rendered — disabling it would let liveness (initialDelaySeconds 0)
+          # kill the container mid-compile.
+          startupProbe:
+            httpGet:
+              path: {{ $cfg.probes.readiness.path | default "/health" }}
+              port: http
+            periodSeconds: {{ $startupPeriod }}
+            failureThreshold: {{ ceil (divf $cfg.progressDeadlineSeconds $startupPeriod) | int }}
+            timeoutSeconds: 5
           {{- if $cfg.probes.liveness.enabled }}
           livenessProbe:
             httpGet:
--- a/charts/tt-inference-server/templates/deployment.yaml
+++ b/charts/tt-inference-server/templates/deployment.yaml
@@ -124,8 +124,8 @@ spec:
           # budget = failureThreshold × periodSeconds is derived from
           # progressDeadlineSeconds so it always matches the max compile time; it
           # flips to serving as soon as one poll succeeds, not after a fixed delay.
-          # Always rendered — disabling it would let liveness (initialDelaySeconds 0)
-          # kill the container mid-compile.
+          # Always rendered — disabling it would let liveness (initialDelaySeconds
+          # defaults to 0) kill the container mid-compile.
           startupProbe:
             httpGet:
               path: {{ $cfg.probes.readiness.path | default "/health" }}
@@ -142,7 +142,6 @@ spec:
               path: {{ $cfg.probes.liveness.path }}
               {{- end }}
               port: http
-            initialDelaySeconds: {{ $cfg.probes.liveness.initialDelaySeconds }}
             periodSeconds: 30
             timeoutSeconds: 5
             failureThreshold: 3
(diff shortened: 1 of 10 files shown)


bgoelTT (approved):
Great idea!

sott0n:
@acvejicTT @mdobrosavljevicTT @mjeremicTT @vcankovicTT @vmaksimovicTT could you review it?

charts/tt-inference-server/values.yaml
```diff
@@ -67,14 +67,24 @@ defaults:
...
+    # initialDelaySeconds is 0, restart the container mid-compile.
+    startup:
+      periodSeconds: 15
     liveness:
       enabled: true
       path: /v1/models
-      initialDelaySeconds: 2400
+      # 0: startupProbe gates the compile window, so liveness starts immediately
+      # once startup succeeds.
+      initialDelaySeconds: 0
```
mjeremicTT:
```suggestion
```
This is the default value for initialDelaySeconds if not defined, we can remove it

charts/tt-inference-server/values.yaml
```diff
@@ -67,14 +67,24 @@ defaults:
...
     readiness:
       enabled: true
       path: /health
-      initialDelaySeconds: 2400
+      initialDelaySeconds: 0
```
mjeremicTT:
```suggestion
```

Same here

charts/tt-inference-server/templates/deployment.yaml
mjeremicTT: Since it is 0 now, it could be removed here also

mjeremicTT: Same here

charts/tt-inference-server/values.yaml
sott0n: addressed

mjeremicTT (approved):

PR #4836, opened 2026-08-01
Adding forge DP embedding models on BH galaxy (Qwen3-Embedding-0.6B/-4B, bge-m3)

jazpurTT:
### Summary
Wires three forge-vllm-plugin embedding models to serve data-parallel on BH galaxy. Qwen 0.6B is new, 4B gains a BH-galaxy path, bge-m3 gets a FORGE path alongside MEDIA. 8 files: constants/runner/fabric + dev+prod embedding.yaml + CI config + eval + perf. CI blocker on tt-shield (fixed here, needs green run), bh galaxy runners have not been able to pick up any jobs. We are tracking the issue in https://github.com/tenstorrent/tt-shield/issues/925

### Current status: 
- Wiring done + Tested locally:
    - config resolves for all three
    - JSON/YAML/schema valid
    - runs on a bh-galaxy machine in <1 min.
- CI pending green needs CI blocker to be resolved.

Issue #2892, opened 2026-04-10
BGE-M3 MTEB GPU SCORE


gtobarTT:
Model 1: BAAI/bge-m3
Revision: 5617a9f61b028005a4858fdac845db406aefb181
Default modalities: text
Tasks (1):
  - STS12: main_score=0.7873

PR #4920, opened 2026-08-12
 Onboard 3 forge embedding models (Qwen3/BGE-M3) on QB2 and Nightly CI (p300x2)

kmabeeTT:
## Problem

Adding Qwen3-Embedding-0.6B, Qwen3-Embedding-4B and bge-m3, building on @jazpurTT's integration work (commit 1 unchanged). These are the first forge embedding models to go through `--workflow release`, and that path did not work: the benchmark client posted to `localhost:8000` regardless of `--server-url`, sent a hardcoded `model: "test-model"`, and had no concurrency bound or failure check — one run "passed" with 14 successful and 986 failed requests. Two dependencies were undeclared, MTEB failures surfaced as a bare exit code, and `bge-m3` resolved to the `tt-sdxl-trace` runner (first match on the runner's model set, and `bge-m3` sits in both a forge and an older non-forge spec).

The models use the conventional media layout — one worker per chip, `DEVICE_IDS='(0),(1),(2),(3)'` — so nothing outside the embedding specs changes. Needs **tt-xla `822f6b615` or newer**; before it, pinning a worker to any chip but 0 aborted in device init (https://github.com/tenstorrent/tt-xla/issues/5521). Nightly CI uses latest forge wheel, and seperate PR is being raised to uplift forge wheel for non-Nightly CI runs:

- https://github.com/tenstorrent/tt-inference-server/pull/4921

## What's Changed

**Specs and config**

- `dev/embedding.yaml` + `constants.py` — p300x2 entries for all three: 4 workers x 8 seqs, `max_model_length` 128. Retargeted from `blackhole_galaxy`, which was never validated end to end — its one dispatch was cancelled after a 13h queue.
- **dev specs only.** `prod/` belongs to `promote_dev_spec_to_prod.py`, which promotes `ci.release` models and injects the pins dev omits. These are `nightly`, so not candidates; `charts/values.yaml` generates from prod, so it needs nothing either. CI runs `--dev-mode` and reads these directly.
- `MODEL_RUNNER` pinned per model, as `image.yaml` and `training.yaml` already do — this is the `bge-m3` mis-resolution fix, and is not p300x2-specific.
- `TT_MESH_GRAPH_DESC_PATH` — pinning one chip per worker leaves a single-chip cluster, which tt-metal classifies as `CUSTOM` and refuses to initialise without an explicit mesh graph descriptor (`TT_FATAL ... is_custom_fabric_mesh_graph_desc_path_specified`). Points at the forge wheel's bundled `p150` descriptor (one Blackhole chip per worker), following the absolute-path convention `gemma-4-31b-it` and `Qwen3-32B` already use for p300x2. SPMD would not need this, since the whole board stays visible.
- `BENCHMARK_MAX_CONCURRENCY` — the server's true in-flight capacity (workers x seqs = 32). `max_concurrency` is the worker count *and* drives the `full_board` health gate, so the benchmark cannot reuse it without under-driving the server, and raising it would make the gate demand 32 ready workers from 4. Defaults to `max_concurrency`, so no existing spec changes.
- `models-ci-config.json` — all three registered **nightly** on p300x2 (4B was `weekly`-only). `model_performance_reference.json` — rows were keyed `blackhole_galaxy` while the specs are `p300x2`, so `get_performance_targets` would never have found them.

**Release-path fixes**

- `embedding_benchmark_tests.py` — target the workflow's server, send the spec's model name, take concurrency from the spec, report failed alongside successful counts, and fail on zero successes.
- `embedding_load_test.py` — send the loaded repo id; forge runners reject anything else, so `"test-model"` was a guaranteed 500. This is not exercised on p300x2 (see the `test_suites` note below); it lands the fix for whenever spec tests are enabled there.
- `vllm_forge_qwen_embedding_runner.py` — `enable_const_eval` on (off re-runs weight prep every call, ~153x on the served path); `enable_trace` off (it compiles once const-eval is on, but measures 2-19% slower — no decode loop to amortise). It also now passes `gpu_memory_utilization` explicitly, matching every sibling forge runner. **Reviewer note:** that arg was previously omitted, so vLLM's own default (0.9) applied; it now takes this repo's forge default of 0.1. Qwen3-Embedding-4B's pre-existing n150/n300/t3k/galaxy configs are affected, and 4B runs weekly on N150 — validated on p300x2 only.
- Supporting: declare `tiktoken` and floor `tqdm` (`requirements/*`), provision the vllm benchmark venv for `ModelType.EMBEDDING` (`workflow_dispatch.py`), surface MTEB's stderr (`embedding_eval_tests.py`), register the runners and eval entries (`runner_fabric.py`, `eval_config.py`), ignore `**/generated/fabric/`.

## Scope + Testing

Scoped to these three models: no shared-infrastructure changes and no other model's config touched. `p300x2` is deliberately absent from `test_suites/embedding.json`, so a release there runs evals + benchmarks and no spec tests. The benchmark concurrency fallback was checked against all 19 embedding specs: every pre-existing one keeps its current value, and only the three new p300x2 entries read the new key.

Full release flow on QB2 (`bh-qb-ge` / `p300x2`) per model — full MTEB STS12 (3108 pairs, no downsampling) plus benchmarks, `PASS (0 blocker(s))`, 1000/1000 successful requests and zero failures on each:

| Model | Result | `main_score` | Throughput | CI run |
| --- | --- | --- | --- | --- |
| Qwen3-Embedding-0.6B | ✅ PASS | 0.769775507348377 | 216 req/s | [31544890873](https://github.com/tenstorrent/tt-shield/actions/runs/31544890873) |
| bge-m3 | ✅ PASS | 0.7874187728614176 | 343 req/s | [31544892807](https://github.com/tenstorrent/tt-shield/actions/runs/31544892807) |
| Qwen3-Embedding-4B | ✅ PASS | 0.827020641076769 | 115 req/s | [31544894320](https://github.com/tenstorrent/tt-shield/actions/runs/31544894320) |

  Local runs on the same hardware produce **bit-identical** scores at 202 / 324 / 111 req/s, so CI reproduces local within ~6%. Ready times were 271s / 165s / 375s locally.

- Scores match the published MTEB references (bge-m3 0.7874 vs 0.7873 GPU) because nothing is downsampled. An earlier revision subsetted each split over a "10h timeout" — that was `3108 x 11.5 s`, the const-eval regression fixed here; the full dataset now runs in ~6 min.
- Concurrency matters: driven at the worker count (4) rather than capacity (32), 0.6B measures 78 req/s and bge-m3 146 — 2.6x and 2.2x low. 4B is unchanged at 110, already compute-bound with 4 in flight. The CI runs report `Maximum request concurrency: 32`, confirming the knob reaches the client.
- Batch/seq and trace swept, 15/15 passing, and the shipped config won: batch 64 costs 21-22%, batch 64 / seq 512 costs 39-46% plus 2.5-3.2x compile time, trace on costs 2-19%; scores unchanged throughout. Swept SPMD on an equivalent branch, but these deltas belong to the runner and per-worker batch, not the DP layout.

## Follow-ups

Perf targets are still placeholders (0.6B's are byte-identical to 4B's); there are real numbers now, but targets gate CI, so calibrating belongs in its own change. MTEB evals are sequential (`batch_size: 1`), so the eval leg cannot use the concurrency the benchmark now does.

## Related issues

- Per-chip device pinning: tenstorrent/tt-xla#5521 (fixed in `822f6b615`)
- Original integration: #4836
- bge-m3 GPU reference score: #2892


idjuricTT (approved):

requirements/benchmarks-vllm-forge.txt
```diff
@@ -13,6 +13,8 @@
 
 vllm==0.19.1
 transformers==5.5.1
+# transformers 5.x imports tqdm.contrib.logging.
+tqdm>=4.67.1
```
fivanovicTT: I am okay with leaving it as is, but could be more prudent to pin it to specific version

requirements/evals-embedding.txt
```diff
@@ -9,3 +9,5 @@
 
 mteb>=2.6.6
 openai
+# mteb's OpenAIModel imports tiktoken but does not declare it.
+tiktoken
```
fivanovicTT: Maybe we could pin a version here as well?

fivanovicTT (approved):

requirements/benchmarks-vllm-forge.txt
kmabeeTT: thx done - tqdm==4.70.0

requirements/evals-embedding.txt
kmabeeTT: done, tiktoken==0.13.0

kmabeeTT:
Thx for review going to merge soon.  I pushed a followup commit for review feedback on requirements pinning:

```
commit 2b74be259eadd64b933664dcdb4e555e956a78e5 (HEAD -> kmabee/embedding-models-qb2, origin/kmabee/embedding-models-qb2)
Author: Kyle Mabee <kmabee@tenstorrent.com>
Date:   Wed Aug 12 13:27:54 2026 +0000

    requirements - Pin tqdm==4.70.0 and tiktoken==0.13.0 based on PR Review
```

fivanovicTT (approved):

kmabeeTT:
Re-dispatch on main after merge for good measure and extra testing since release window is near.

- (pass) Qwen3-Embedding-0.6B: https://github.com/tenstorrent/tt-shield/actions/runs/31612251235
- (pass) Qwen3-Embedding-4B : https://github.com/tenstorrent/tt-shield/actions/runs/31612256798
- (pass) bge-m3 : https://github.com/tenstorrent/tt-shield/actions/runs/31612254092

PR #4924, opened 2026-08-12
[Helm v0.2] hugepages-free mode (hugepages.enabled toggle) (#4866)

### LINKED ISSUE #4866: [Helm v0.2] hugepages-free mode (hugepages.enabled toggle)
Part of #4859.

Add a `hugepages.enabled` toggle (default `true` for backward compatibility). When `false` (IOMMU + KMD 1.29.0+, where hugepages are not needed), drop:
- the `hugepages-1Gi` resource request/limit,
- the hostPath `/dev/hugepages-1G` volume/mount,
- the `cleanup-hugepages` initContainer (removes the 2nd privileged container),
- and switch the strategy `Recreate` -> `RollingUpdate`.

Needs on-hardware verification.

sott0n:
Part of #4859. Stacked on #4919 — merge that first.

Adds a `hugepages.enabled` toggle (default `true`). When `false` — for IOMMU + KMD 1.29.0+ clusters where boards don't need hugepages — the chart drops the `hugepages-1Gi` request/limit, the `/dev/hugepages-1G` volume + mounts, and the `cleanup-hugepages` initContainer. Default-on rendering is unchanged.

The `Recreate → RollingUpdate` item from #4859 is already handled by #4887 (`maxSurge=0` is still needed under DRA to free the board before the replacement starts), so it's not repeated here.

Verified on T3K (Qwen3-8B/n300) with `hugepages.enabled=false`: host at zero 1Gi hugepages under IOMMU DMA, device fabric init + inference served, Pod `READY 1/1`.

Closes #4866.

### DIFF
--- a/charts/tt-inference-server/values.yaml
+++ b/charts/tt-inference-server/values.yaml
@@ -29,6 +29,13 @@ cache:
 # (tt-media-server) env vars are set so each server picks up the name it reads.
 hfCacheDir: ""
 
+# Tenstorrent boards need 1Gi hugepages unless the cluster runs IOMMU + KMD
+# 1.29.0+, where they aren't required. When disabled, the chart drops the
+# hugepages-1Gi request/limit, the /dev/hugepages-1G volume + mounts, and the
+# cleanup-hugepages initContainer.
+hugepages:
+  enabled: true
+
 # ---------------------------------------------------------------------------
 # Defaults — merged with (and overridden by) per-model config below.
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


charts/tt-inference-server/values.yaml
```diff
@@ -29,6 +29,13 @@ cache:
 # (tt-media-server) env vars are set so each server picks up the name it reads.
 hfCacheDir: ""
 
+# Tenstorrent boards need 1Gi hugepages unless the cluster runs IOMMU + KMD
+# 1.29.0+, where they aren't required. When disabled, the chart drops the
+# hugepages-1Gi request/limit, the /dev/hugepages-1G volume + mounts, and the
+# cleanup-hugepages initContainer.
+hugepages:
+  enabled: true
```
mjeremicTT, line 37: Since we are defaulting to true here, this means that we will always have hugepages enabled by default.  How will we handle the case where cluster runs IOMMU + KMD 1.29.0+ ?

sott0n: IOMMU + KMD 1.29.0+ clusters set `hugepages.enabled=false`. It's a cluster-level property, same as `hfCacheDir`, so it lives in the cluster's values file (or --set), documented in the values comment and the README values table.

The default stays true to keep existing (v0.1.0) clusters working, and it's the safe direction because the two failure modes aren't symmetric. Left at true on a hugepages-free node, the Pod just stays Pending, verified on our T3K, where the node reports `hugepages-1Gi: 0` and the scheduler says `0/1` nodes are `available: 1` Insufficient hugepages-1Gi. No container starts, no board is held, and it's one `--set` away from fixed. Defaulting to false on a hugepages cluster would instead fail at runtime on device open, which is much harder to diagnose.

mjeremicTT (approved):

Issue #4925, opened 2026-08-12
[Helm v0.3] Per-engine Grafana dashboards (vLLM / media-server / forge)


sott0n:
Follows #4867 (App-metrics PodMonitor). Target: Helm chart v0.3.0.

## Background
#4867 adds a `PodMonitor` so the inference server's `/metrics` reach Prometheus, but the chart ships **no Grafana dashboard**. Metric schemas differ per engine — vLLM exposes `vllm:*`, media-server exposes a different set, and forge may expose none — so a single unified dashboard isn't feasible (non-matching engines would show "No data").

For v0.2 the chart deliberately ships no dashboard; the README points vLLM users to the official vLLM dashboard for import and media/forge users to Grafana Explore.

## Proposal
Ship **per-engine** Grafana dashboards, rendered only for the resolved engine:
- Gate behind `grafanaDashboard.enabled` (default `false`).
- Render a dashboard `ConfigMap` (label `grafana_dashboard=1`) selected by `tt-inference-server.resolvedEngine`, so only the matching engine's dashboard is emitted. Mirrors tt-telemetry's dashboard mechanism.

## Prerequisites / open questions
- Capture the actual metric schemas for **media-server** and **forge** (inspect `/metrics` on hardware; forge may expose none).
- For vLLM: vendor the official vLLM dashboard vs. a slimmed custom one.
- Document the coupling: the ConfigMap+sidecar path assumes a Grafana running the dashboard sidecar watching `grafana_dashboard=1`.

## Out of scope
- device × app correlation (needs a supply-side bridge exporter) — tracked separately.

Refs: #4867, #4859.

sott0n:
Two corrections to the prerequisites in the body.

**media's metric schema needs no hardware.** All 90 families are declared in `tt-media-server/telemetry/{telemetry_client,audio_metrics,image_metrics,video_stage_metrics}.py` — request/stage durations, per-domain STT/TTS/image/video metrics, device warmup, a `tt_canary_*` family — with labels worth building on (`model_type`, `device_id`, `status`, `request_type`, `resolution`, `voice`). Note the multiprocess collector (`telemetry/multiprocess_setup.py`): panels read the aggregated `_bucket`/`_count`/`_sum`, not a single process.

[...]

PR #4927, opened 2026-08-12
[Helm v0.2] add optional app-metrics PodMonitor (#4867)

### LINKED ISSUE #4867: [Helm v0.2] App-metrics PodMonitor (optional)
Part of #4859.

Add a PodMonitor template to scrape the inference server's own `/metrics` (vLLM / media-server already expose it). Gate it behind `podMonitor.enabled` (default `false`), since a PodMonitor requires the Prometheus Operator CRDs (`monitoring.coreos.com`) and would otherwise fail at install time.

Note: this is app-level metrics only; device telemetry is handled by tt-operator's tt-telemetry. device x app correlation (join) is future work.

sott0n:
Part of #4859. Stacked on #4924 — merge that first.

Adds an optional `PodMonitor` (`podMonitor.enabled`, default `false`) scraping the inference server's own `/metrics` on the API port. Default off because a `PodMonitor` requires the Prometheus Operator CRDs (`monitoring.coreos.com`), which would fail to install without them. `podMonitor.labels` lets operators set the label their Prometheus's `podMonitorSelector` matches (e.g. `release: kube-prometheus-stack`).

No Grafana dashboard is bundled — metrics differ per engine (vLLM `vllm:*`; media-server/forge differ). The README points vLLM users to the official vLLM dashboard for import and media/forge to Grafana Explore; per-engine dashboards are tracked for v0.3 in #4925.

Verified e2e on T3K: with kube-prometheus-stack installed, both app metrics (vLLM) and device telemetry (tt-operator's tt-telemetry) are scraped and queryable in Grafana.

Closes #4867.

### DIFF
--- a/charts/tt-inference-server/values.yaml
+++ b/charts/tt-inference-server/values.yaml
@@ -41,9 +41,13 @@ hugepages:
 # CRDs (monitoring.coreos.com), and rendering one without them fails at install.
 podMonitor:
   enabled: false
-  # Set the label your Prometheus's podMonitorSelector matches (e.g.
-  # release: kube-prometheus-stack), or the PodMonitor won't be scraped.
-  labels: {}
+  # Labels added to the PodMonitor so your Prometheus's podMonitorSelector
+  # matches it, or it won't be scraped. The default matches tt-telemetry's chart
+  # and the kube-prometheus-stack convention (its podMonitorSelector defaults to
+  # release: <the stack's own release name>). Override when your Prometheus was
+  # installed under a different release name.
+  labels:
+    release: prometheus
   interval: 30s
   path: /metrics
 
--- a/charts/tt-inference-server/values.yaml
+++ b/charts/tt-inference-server/values.yaml
@@ -36,6 +36,17 @@ hfCacheDir: ""
 hugepages:
   enabled: true
 
+# PodMonitor scraping the inference server's own /metrics (vLLM and media-server
+# expose it). Disabled by default: a PodMonitor needs the Prometheus Operator
+# CRDs (monitoring.coreos.com), and rendering one without them fails at install.
+podMonitor:
+  enabled: false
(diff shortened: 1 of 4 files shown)


charts/tt-inference-server/values.yaml
```diff
@@ -36,6 +36,17 @@ hfCacheDir: ""
...
   enabled: true
 
+# PodMonitor scraping the inference server's own /metrics (vLLM and media-server
+# expose it). Disabled by default: a PodMonitor needs the Prometheus Operator
+# CRDs (monitoring.coreos.com), and rendering one without them fails at install.
+podMonitor:
+  enabled: false
+  # Set the label your Prometheus's podMonitorSelector matches (e.g.
+  # release: kube-prometheus-stack), or the PodMonitor won't be scraped.
+  labels: {}
```
mjeremicTT: Should we have a default value here? Otherwise I see that every time when podMonitor is enabled, those labels needs to be changed as well.

sott0n: Agreed — defaulted it to `release: prometheus` in 9f1cf4605, matching tt-telemetry’s chart (`helm/values.yaml`) so device + app metrics do not need two different label values on one cluster.

`podMonitor.enabled=true` is now enough on a standard kube-prometheus-stack. Override with `--set podMonitor.labels.release=<name>`, or `--set podMonitor.labels=null` when your `podMonitorSelector` matches every PodMonitor.

sott0n:
@mjeremicTT could you confirm updates?

charts/tt-inference-server/values.yaml
mjeremicTT: I will approve this - but we need to be careful with this because this value: `release: prometheus` could change over time and we are basically making dependency to it right now.

mjeremicTT (approved):

PR #4929, opened 2026-08-12
Model support for XTTS v2

### LINKED ISSUE #53926: [XTTS-v2] Add model support in inference server
Add XTTS-v2 model support in inference server.

### LINKED ISSUE #54904: [XTTS-v2] Integrate only existing evals/benchmarks/spec tests for the experimental release
SpeechT5 implemented some evals, benchmarks and spec tests. This will be a good starting point for XTTS v2 for experimental release.

acicovicTT:
# XTTS-v2 model support (tt-media-server)

Adds coqui/XTTS-v2 text-to-speech, served by tt-media-server on N150, P150, and BH QuietBox 2 (p300x2). 

## What's included

- Runner (`xtts_v2_runner.py`): single-chip pipeline, data-parallel on p300x2 (one worker per chip, 4 concurrent requests); voice cloned once at warmup from XTTS_REF_AUDIO.
- 17 languages: optional language request field (default `en`). Sentence chunking is CJK-aware with per-language chunk budgets from coqui's character limits. Romanizer deps (`ja`/`ko`/`zh`) added to `requirements.txt`, lazy-loaded per language. speecht5 modified to reject non-English explicitly.
- CI parity with speecht5: `EvalConfig`, spec-test suite (functionality includes seed-reproducibility and language handling, load, param, integration), and benchmark perf targets for all three devices. `TTSParamTest` now pins `seed: 0` so its identical requests check holds for stochastic TTS (no-op for speecht5).

## Validation (on QuietBox 2)

- Eval: avg WER 3.3% (score 96.7 vs 80 target), 10/10 samples.
- Spec tests: 6/6 PASS. Perf: ~6.1–6.4 ms/token decode, RTF ~6.6 per chip.

## Notes

- Requires a tt-metal build containing `models/experimental/xtts_v2` (commit pin TBD at promotion from dev to prod).
- XTTS-v2 weights are under the Coqui Public Model License (non-commercial) -- flagged in the docs, TBD.


tt-media-server/domain/text_to_speech_request.py
```diff
@@ -39,6 +43,45 @@ def validate_text(cls, text):
...
+    def validate_language(cls, v):
+        if v is None:
+            return DEFAULT_TTS_LANGUAGE
+        base = str(v).strip().lower().split("-")[0]
+        if base not in XTTS_SUPPORTED_LANGUAGES:
+            raise ValueError(
+                f"Unsupported language {v!r}; supported: {sorted(XTTS_SUPPORTED_LANGUAGES)} "
+                "(region variants like 'pt-br' are accepted)"
+            )
+        return base
```
vpetrovicTT, line 81: This is the shared TTS request model, but the validator seems to be XTTS-specific.
I am not sure if this will break downstream for different model like **speecht5**(for example it looks like language: "fr" would pass here but crash in the model runner).

tt-media-server/tt_model_runners/xtts_v2_runner.py
```diff
@@ -0,0 +1,329 @@
...
+        from models.experimental.xtts_v2.frontend import CHAR_LIMITS
+
+        if voice is None:
+            voice = self.default_voice
+        budget = min(CHUNK_CHAR_LIMIT, CHAR_LIMITS.get(language, CHUNK_CHAR_LIMIT))
+        chunks = chunk_text(text, max_chunk_size=budget)
+        pieces: list[np.ndarray] = []
+        gap = np.zeros(
+            int(INTER_CHUNK_SILENCE_SECONDS * XTTS_SAMPLE_RATE), dtype=np.float32
+        )
```
vpetrovicTT, line 248: Should we record some metrics for this model? 
For example, **speecht5** records TtsChunkProgress per chunk and record_tts_input_tokens on success..

tt-media-server/domain/text_to_speech_request.py
acicovicTT: It validates the content of the request to weed-out invalid values. The runner decides whether to accept that parameter or no. SpeechT5 runner won't crash but it'll surface an error. You are right though that this constant should not be named `XTTS_SUPPORTED_LANGUAGES` but something like `VALID_LANGUAGE_CODES`, since Voxtral and potentially other models that we are working on also support multiple languages.

tt-media-server/tt_model_runners/xtts_v2_runner.py
acicovicTT: We plan to work on metrics in a more general PR targeting multiple TTS models. For now, TTFT and RTF are already recorded for this model which is fine with us for experimental release.

vpetrovicTT (approved):

PR #4936, opened 2026-08-13
fix min num inference steps value incosistency for video models

### LINKED ISSUE #4899: Update model config for video models
Model config for video models should be in sync with restrictions -> min number of inference steps should be 4.

vpetrovicTT:


tt-media-server/config/constants.py
```diff
@@ -264,6 +264,11 @@ class ModelServices(Enum):
     }
 )
 
+# Client-facing video API bounds
+MIN_VIDEO_INFERENCE_STEPS = 4
+MAX_VIDEO_INFERENCE_STEPS = 50
+DEFAULT_VIDEO_INFERENCE_STEPS = 20
```
fivanovicTT, line 270: Does this makes more sense to be part of `settings.py` instead?

fivanovicTT: Let's leave it `constants.py` for now

fivanovicTT (approved):

PR #4940, opened 2026-08-13
Fix Mochi pipeline call to match latest tt-metal API (merge this first)

jashansinghTT:
Mochi video generation fails on every current media image because `TTMochi1Runner.run` still uses the pre-refactor `MochiPipeline.__call__` signature. The model loads, runs the full denoise, then dies with `MochiPipeline.__call__() got an unexpected keyword argument 'prompt'` (and after that, postprocessing rejects the PIL output with `'list' object has no attribute 'ndim'`).

Same treatment Wan2.2 got in #4032:

- [x] Pass `prompts`/`negative_prompts` as lists; drop `num_frames`/`height`/`width`/`output_type`, which moved into `MochiPipelineConfig` defaults (168 frames, 480x848)
- [x] Convert the pipeline output (PIL frame lists, or a tensor) into the `(batch, frames, H, W, C)` uint8 array the MP4 exporter expects
- [x] Verified on a Blackhole QuietBox 2 (P300x2, 2x2 mesh) against the `0.18.0-c49bb76` media image with this exact change applied: warmup completes and a 12-step generation returns a valid MP4

Note the prod spec still pins mochi to `0.10.0-555f240`, whose bundled tt-metal predates the Mochi BH mesh fixes and rejects the P300x2 mesh outright — spec bump is a separate small PR.

### DIFF
--- a/tt-media-server/tt_model_runners/dit_runners.py
+++ b/tt-media-server/tt_model_runners/dit_runners.py
@@ -8,6 +8,7 @@ import io
 import os
 from abc import abstractmethod
 
+import numpy as np
 import ttnn
 from config.constants import (
     WAN22_NUM_FRAMES,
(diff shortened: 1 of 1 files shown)


fivanovicTT (approved):
Please leave here in the comments  a successful CI run from tt-shield

PR #4941, opened 2026-08-13
Re-pin Mochi prod spec to the 0.18.0 media image and add DiT cache env var (merge this after #4940)

jashansinghTT:
The prod spec has Mochi frozen at `0.10.0-555f240`, whose bundled tt-metal predates the Mochi blackhole mesh fixes — on P300x2 the pipeline asserts ("Mochi has only been successfully tested on 2x2") before loading anything. Newer tt-metal also requires `TT_DIT_CACHE_DIR` in the container env; the runner no longer sets a default, so the model load fails without it. The dev template already carries the env var; only prod lags (Mochi has no `ci.release` entry in models-ci-config.json, so release automation never re-pins it — flagging that for maintainers rather than changing it here).

Same shape as the Wan2.2 entry below it (pinned `0.17.0/8c48a10` with the same env var).

- [x] Bump Mochi to `0.18.0` / `c49bb76` (newest published media image)
- [x] Add `TT_DIT_CACHE_DIR` at template level, matching the Wan2.2 entry
- [x] Verified on a Blackhole QuietBox 2 (P300x2): with this image + env var the 2x2 mesh initializes and the model loads and warms up

One dependency to be aware of: end-to-end video generation on this image also needs the runner-side call fix in #4940 — once a media image containing that lands, this pin should move to it.

### DIFF
--- a/workflows/model_specs/prod/video.yaml
+++ b/workflows/model_specs/prod/video.yaml
@@ -7,13 +7,15 @@ templates:
 # =============================================================================
 - weights:
     - genmo/mochi-1-preview
-  version: "0.10.0"
-  tt_metal_commit: 555f240
+  version: "0.18.0"
+  tt_metal_commit: "c49bb76"
   impl: tt_transformers
   min_disk_gb: 60
   min_ram_gb: 32
   model_type: VIDEO
   inference_engine: MEDIA
+  env_vars:
+    TT_DIT_CACHE_DIR: /home/container_app_user/cache_root/tt_dit_cache
   device_model_specs:
     - device: T3K
       max_concurrency: 1


vcankovicTT:
Hello Jashan,

Thanks for the PR — I appreciate you flagging this, but we're not able to support directly editing the version and commit in `prod.yaml` for a specific model/device outside of our standard release process.

Reasons:

1. All model/version updates need to go through our Models CI Release Process, which includes evals, benchmarks, and confirmation against acceptance criteria before anything is promoted, for the given code.
2. Scope of the change: Pinning a specific version/commit isn't actually a one-file change. Beyond `prod.yaml`, it typically touches the dev catalogue, prod catalogue, release model specification file, and many more documentation files that need to stay in sync. A direct edit would leave those out of sync and break traceability.

If you'd like to support the model with the latest code, here is the documented procedure:

👉 https://tenstorrent.atlassian.net/wiki/spaces/PSE/pages/2517696767/Models+CI+Release+Process

- Submit a request in #model-requests to have this included in the next release (`v0.21.0` or `v0.22.0`), so it goes through the full validation pipeline.
- In the meantime, if you want to validate/test with the current code, you can kick off an on-dispatch job run from `tt-shield` to confirm everything works as expected or some developer's involvement will be requested.

Let me know if you have questions about the release process.
Thanks!

PR #4951, opened 2026-08-14
Embedding path v1 cleanup: pybind11 runner, config-driven model selection, mock runner

### LINKED ISSUE #2216: Clean up experimental cpp_server embedding flow
Clean up the existing experimental cpp_server embedding implementation before migrating non-Forge embedding models.
## Scope
- Review the current experimental embedding service/runner classes in cpp_server.
- Remove dead or obsolete embedding code.
- Consolidate reusable pieces with the current cpp_server service/runner structure introduced for SDXL.
- Keep only the minimal infrastructure needed for the upcoming non-Forge embedding migration.

jzivanovicTT:
## Summary
- Rewrote `embedding_runner.cpp` from the raw CPython C API to pybind11: no manual
  refcounting, Python tracebacks surfaced as per-request errors, fixed an
  out-of-bounds read when Python returns fewer results than requests.
- Model selection is now configuration: `MODEL_RUNNER_TYPE` picks a row in a table
  in `settings.cpp`; the forked worker derives and exports `MODEL`/`DEVICE`/
  `MODEL_RUNNER` itself and verifies `max_batch_size` against Python at boot.
  The runner class is resolved by Python's `runner_fabric`, not duplicated in C++.
- Batches are capped by the model's own `max_batch_size` instead of the unrelated
  `MAX_IN_FLIGHT_COUNT` (previously caused HTTP 500s on bursts).
- New `IEmbeddingRunner` interface + `MockEmbeddingRunner`: deterministic,
  device-free, Python-free runner so the HTTP/batching/codec path is testable in CI.
- Unit tests for the IPC codec (9 cases), golden-vector comparison script, and a
  serving guide under `cpp_server/docs/`.

## Testing
- BGE-large on n150: all 10 golden prompts bit-exact sequentially; concurrent
  burst of 16 within threshold; PCC warmup gate passed (0.9148).
- Mock path: boots with no hardware/venv, deterministic 1024-dim unit vectors,
  burst of 16 all HTTP 200.
- `ctest`: green (one pre-existing, unrelated failure: `CancellationE2E` requires
  a `--blaze` build).

Closes #2216


tt-media-server/cpp_server/src/config/settings.cpp
```diff
@@ -511,6 +512,93 @@ void readMediaRunnerConfig(MediaRunnerConfigBase& cfg) {
...
+}
+
+std::string joinNames(const std::vector<std::string>& names) {
+  std::string out;
+  for (const auto& n : names) {
+    if (!out.empty()) out += ", ";
+    out += n;
+  }
+  return out;
+}
```
nanicicTT, line 570: If we have fmt, we can do

```cpp
return fmt::format("{}", fmt::join(names, ", "));
```

nanicicTT: If not then I would avoid repeatedly checking `out.empty()` and optionally reserve memory:

```cpp
std::string joinNames(const std::vector<std::string>& names) {
    if (names.empty()) {
        return {};
    }

    std::string out = names.front();

    for (auto it = std::next(names.begin()); it != names.end(); ++it) {
        out += ", ";
        out += *it;
    }

    return out;
}
```

jzivanovicTT: Done, used fmt::join.

nanicicTT (approved):

PR #4954, opened 2026-08-14
Add Qwen3.8-27B model specs (dev + prod pins)

stisiTT:
Adds model specs for **Qwen/Qwen3.8-27B**, released this morning (2026-08-14).

**Why this is a small change:** Qwen3.8-27B is the same architecture as Qwen3.6-27B — same `model_type: qwen3_5`, and its config.json differs from 3.6's by only the transformers version stamp. It runs through the existing `qwen36_blackhole` tt-metal impl with no code changes. The two entries mirror the Qwen3.6-27B specs with the model name swapped.

**Validated on release day, bare metal on P300X2:**
- tt-metal `text_demo` traced_128: passed — TTFT 160 ms, decode 26.53 tok/s (at parity with the registered Qwen3.6-27B targets; the tt-metal pin includes the #51556 bucketing fix)
- Live vLLM OpenAI server: streaming, thinking mode (new default in 3.8), and clean stops with the new eos token pair all verified

**What's in the PR:**
- `dev/llm.yaml`: Qwen3.8-27B entry, P300X2 only (the validated device), mirroring the 3.6 dev entry. No commit pins — dev specs can't carry them.
- `prod/llm.yaml`: matching entry carrying the build pins, since that's where `build_docker_images.py` reads them: `tt_metal_commit: 9a04829`, `vllm_commit: c127c17`, version 0.20.0, status EXPERIMENTAL.

**One thing reviewers should know:** `vllm_commit` here is a **vllm-tt-plugin** SHA, not a tenstorrent/vllm fork SHA like the entries above it — since #4907 the Dockerfile clones vllm-tt-plugin, so this is the first entry pinned under the new scheme. Our bare-metal validation used the fork (`7c99bd3`); the docker image built from these pins uses the plugin path and still needs its own smoke run. Happy to adjust the pin or version if 0.21.0 conventions land differently.

Checks run locally: both catalogs load (299 dev / 226 prod specs), all 179 catalog/spec tests pass, and the prod entry resolves to docker image tag `0.20.0-9a04829-c127c17`.

CI run: https://github.com/tenstorrent/tt-shield/actions/runs/31826395848

### DIFF
--- a/charts/tt-inference-server/values.yaml
+++ b/charts/tt-inference-server/values.yaml
@@ -7921,3 +7921,43 @@ models:
               - name: VLLM__MIN_CONTEXT_LENGTH
                 value: '32'
     defaultEngine: forge
+  Qwen3.8-27B:
+    vllm:
+      p300x2:
+        defaultImpl: qwen36_blackhole
+        impls:
+          qwen36_blackhole:
+            progressDeadlineSeconds: 5400
+            image:
+              repository: ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64
+              tag: 0.20.0-9a04829-c127c17
+            probes:
+              liveness:
+                initialDelaySeconds: 2400
+              readiness:
+                initialDelaySeconds: 2400
+            resources:
+              requests:
+                memory: 67Gi
+            env:
+              - name: ARCH_NAME
+                value: blackhole
+              - name: HF_MODEL
+                value: Qwen/Qwen3.8-27B
+              - name: MESH_DEVICE
+                value: (1, 4)
+              - name: QWEN36_MAX_TOKENS_ALL_USERS
+                value: '525312'
+              - name: TORCHDYNAMO_DISABLE
+                value: '1'
+              - name: TT_MESH_GRAPH_DESC_PATH
+                value: ../../tt-metal/tt_metal/fabric/mesh_graph_descriptors/p300_x2_mesh_graph_descriptor.textproto
+              - name: TT_QWEN35_TEXT_VER
+                value: qwen36_blackhole
+              - name: VLLM_CONFIGURE_LOGGING
+                value: '1'
+              - name: VLLM_RPC_TIMEOUT
+                value: '900000'
+              - name: VLLM_TARGET_DEVICE
+                value: tt
+    defaultEngine: vllm
--- a/reference_config/evals/eval_config.py
+++ b/reference_config/evals/eval_config.py
@@ -1372,6 +1372,76 @@ _eval_config_list = [
         ],
     ),
     EvalConfig(
+        # Qwen3.8-27B reuses the Qwen3.6-27B architecture byte-for-byte (same
+        # model_type qwen3_5), so it serves through the qwen36_blackhole impl
+        # and takes the same eval setup as Qwen3.6-27B above. This entry is
+        # what puts the model into EVAL_CONFIGS; without it the release
+        # workflow fails fast in validate_setup (evals are required).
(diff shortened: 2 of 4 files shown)


pasta-paul:
Confirming this approach on the target hardware: we have `Qwen/Qwen3.8-27B` serving live on a **QuietBox 2 (P300X2 — the device this PR targets)** using the same `qwen36_blackhole` impl and the same vLLM fork pin (`7c99bd3`) — the 3.8 checkpoint is arch-identical to 3.6 (`qwen3_5`, hidden 5120, 64 layers, vocab 248320, max_pos 262144), so a deep-copied 3.6 spec entry serves it with no model-side changes. On-box validation (OpenAI-compatible endpoint, mesh (1,4), bfp8, 262144 ctx):

- Correctness battery 4/4 + long-context needle PASS at 24,738 prompt tokens (10.7s).
- Decode ~17.8 tok/s e2e single-stream; 8-way concurrency all `finish=stop`, ~52.8 tok/s aggregate.
- Thinking honored both ways: `enable_thinking=False` → 4 tok/0.5s vs default 36 tok/2.2s on a trivial probe; on genuinely hard items the `qwen3` reasoning parser **separates** the trace into `message.reasoning` (observed 8,702 chars distinct from a 741-char final answer).
- The new `reasoning_effort` knob (`xhigh`/`medium`/`low` via `chat_template_kwargs`) is accepted without error; the no-kwarg baseline is byte-identical to `xhigh` → **xhigh is the serving default** (worth documenting for users expecting a "thinking off" default).

Two deployment deltas + one caveat, for the entry's accuracy: (1) we run `max_num_seqs=4` (latency-oriented) vs this entry's `max_concurrency: 32` — both work on P300X2, ours is the low-latency point; (2) we're on a custom 0.19.0-line rebuild (`0.19.0-06979e388e-7c99bd3b88`) rather than the 0.20.0/`9a04829`/`c127c17` prod pins here, so this also confirms the impl is robust across both pins; (3) the `supported_modalities: [text]` on a batched entry is correct — at `max_num_seqs>1` the `qwen36` impl asserts text-only (multimodal needs `max_num_seqs=1`), and we've filed that the assert currently hard-crashes the engine instead of returning a clean 4xx: https://github.com/tenstorrent/tt-metal/issues/53187

*Drafted by CQL AI (Canada Quant Labs' AI research agent) and verified by a human who takes responsibility for its accuracy; all numbers from a live QuietBox 2 deployment.*

vcankovicTT:
Hello,

`values.yaml` and production catalogue should never (in any scenario) be changed directly, they must be populated automatically through the regular releasing process as many, many other related files are depending or derived from these files, and all of those files should be in sync, and such files are populated through the scripting, not manually (there is a valid reason behind it).

Additionally we should not release something until we have complete confirmation on evals and benchmarks results, spec tests and acc. criteria, Furthermore, these performance results for the specific model and device must be presented in reports, outputs, and the Models Dashboard application. Users access this dashboard to get a complete overview of the exact values obtained and confirmed as the final output of the release process for a given model and device on a specific version.

My proposal is to cancel this PR and submit a request for this model in the upcoming release (v0.21.0 or v0.22.0) so this can be handled through the regular procedure, that we are trying to establish and follow here in Tenstorrent.

@bgoelTT  please comment if you think differently, that this should go outside the usual way.

Thank you

vcankovicTT (changes requested):
Explained in comment why in my opinion this PR should not be approved, or leave changes solely for dev catalogue and evals-config.

bgoelTT:
@vcankovicTT @stisiTT yes, the prod.yaml should only be touched by the release automation which also updates documentation & publishes perf data.

PR #4956, opened 2026-08-14
Re-pin Motif-Image-6B-Preview to the 0.18.0 media image

jashansinghTT:
## Problem

The prod image spec pins `Motif-Image-6B-Preview` to `tt-media-inference-server:0.9.0-c180ef7`, but that image cannot launch Motif via `run.py` on any device:

1. run.py derives `MODEL=Motif-Image-6B-Preview` from the spec weights, but the 0.9.0 server's `ModelNames` enum still has the old lowercase value `motif-image-6b-preview`, so the server dies at import with `ValueError: 'Motif-Image-6B-Preview' is not a valid ModelNames` before ever binding the port.
2. The 0.9.0 server has no Motif device configs for P150X8 or P300X2, both of which this spec entry lists.

So the documented quickstart (`python3 run.py --model Motif-Image-6B-Preview --device p300x2 --workflow server --docker-server`) fails on a 30s container-start timeout.

## Fix

Re-pin to `0.18.0-c49bb76` (the newest published media image): its enum accepts the exact-case MODEL value and it carries Motif device configs for all four listed devices.

## Verification

Hardware-verified on p300x2 via TT-Studio: with this image, Motif reaches pipeline creation on all devices. Full p300x2 serving additionally needs two one-line fixes — the tt-metal `(2, 2)` mesh preset (PR to tt-metal) and the trace-region raise (#4955) — after which warmup completes and 1024x1024 generations return in ~7s. T3K/GALAXY/P150X8 launch correctly with this re-pin alone.

## Steps to confirm

1. `python3 run.py --model Motif-Image-6B-Preview --device <t3k|p300x2> --workflow server --docker-server` with the old pin → container dies at import (see server log: `ValueError: ... not a valid ModelNames`).
2. Same command with this re-pin → server boots and reports the model on `/v1/models`.

### DIFF
--- a/workflows/model_specs/prod/image.yaml
+++ b/workflows/model_specs/prod/image.yaml
@@ -170,8 +170,8 @@ templates:
 
 - weights:
     - Motif-Technologies/Motif-Image-6B-Preview
-  version: "0.9.0"
-  tt_metal_commit: c180ef7
+  version: "0.18.0"
+  tt_metal_commit: c49bb76
   impl: tt_transformers
   min_disk_gb: 15
   min_ram_gb: 6


vcankovicTT:
Hello Jashan,

production catalogue should never (in any scenario) be changed directly, they must be populated automatically through the regular releasing process as many other related files are depending or derived from these files, and all of those files should be in sync, and such files are populated across the repository automatically through the scripting, not manually (there is a valid reason behind it).

We never change production catalogue directly. We start with dev catalogue, it goes through the testing in release pipelines, and once we have satisfying results and everything confirmed, we publish it in  the production catalogue too, at the end of the release cycle.

For specific model and device on specific version and commit we are publishing evals and benchmarks results, spec tests and model status based on the calculated acc. criteria. Furthermore, these performance results for the specific model and device must be presented in reports, outputs, and the Models Dashboard application. Users access this Dashboard to get a complete Model overview of the exact values obtained and confirmed as the final output of the release process (for that specific referenced Version) for a given model and device.

Please  submit model request in #model-requests channel (through the new Epic) so we can sort this out in the next release, and publish `Motif-Image-6B-Preview` with the correct pinned version along with the measured evals and benchmarks for the given version and commit.

Issue #4869, opened 2026-08-06
[Helm v0.2] Layered CI for the chart: kind install gate + thin real-hardware DRA smoke


sott0n:
Part of #4859.

Establish a layered CI strategy for the chart, where each layer tests only what it owns. Today we have strong **static** coverage (L1); this issue adds a hardware-free **install** gate (L2) and a thin **real-hardware** smoke (L3).

Explicitly *not* re-testing DRA driver behaviour here: that a Pod can claim a device and see `/dev/tenstorrent` is already proven on real silicon by tt-operator's `dra-smoke` (integration-rke2). Our tests assume a working DRAlayer and verify that *our chart* installs and brings up a responding server on top of it.

## Layers

| Layer | Runner | Proves | Frequency |
|---|---|---|---|
| **L1 static** (exists) | `ubuntu-latest` | manifests render + kubeconform; DRA claim structure (`ResourceClaimTemplate`, `deviceClassName: tenstorrent.com`, per-device board count); generator ↔ values.yaml in sync | every PR |
| **L2 install (no hardware)** — new | `ubuntu-latest` + kind | `helm install` succeeds; API server accepts the `ResourceClaim`; pod spec is schedulable (Pending without a device is fine); RBAC, PodMonitor CRD gating, hugepages toggle actually apply | every PR |
| **L3 real-hardware smoke** — new | self-hosted TT (`tt-ubuntu-2204-n150-stable`) | chart-specific runtime assumptions on real silicon | nightly / manual / chart-path + label gate (not per-PR) |

## L2 — install gate (kind, no hardware)

Mirrors tt-operator's kind tier (fake the NFD `tt-present` label, reach as far as image-pull). Catches install-time bugs that render tests pass through: RBAC, immutable-field conflicts, admission webhooks, apiVersion drift, values-schema regressions.

## L3 — real-hardware smoke (thin)

Preconditions (not re-tested): tt-operator already installed. Scope: **one small model on one device shape** (e.g. n150). Assert only what static/L2 cannot:

1. A **real inference image** Pod schedules via the DRA claim and gets `/dev/tenstorrent` (vs `dra-smoke`'s busybox probe).
2. The Pod reaches Ready **within the startupProbe compile window** (#4870) — validates the computed window against real compile time.
3. `/v1/...` returns 200 with a valid response ("the inference server responds").
4. `helm uninstall` releases the claim cleanly — no Pod stuck `Terminating`.

[...]

PR #4957, opened 2026-08-17
[Helm v0.2] CI: hardware-free install gate on kind

sott0n:
Part of #4859. Addresses the **install-gate layer of #4869**, whose layered-CI design this PR implements.

Adds a live-install CI gate for the chart that runs on a `kind` cluster with **no Tenstorrent hardware**. It sits between the existing `helm-chart-validate` render job and real-hardware smoke:

- **`helm-chart-validate` (exists):** proves the manifests *render* (`helm template` + kubeconform).
- **`helm-install-kind` (this PR):** proves a *real API server installs them*.
- **real-hardware smoke (future):** proves the server actually serves on silicon.

The middle layer catches a whole class of bugs that render passes through — CEL compile errors, `apiVersion` drift, admission rejections, CRD-gating regressions — on every PR, on free `ubuntu-latest`, isolated from hardware flakiness.

## What it checks

On a k8s 1.34 kind node (where DRA / `resource.k8s.io/v1` is GA):

| Check | Proves |
|---|---|
| `helm install` succeeds | the release applies against a live API server |
| ResourceClaimTemplate accepted (deviceClassName, count) | the DRA claim — CEL selector and all — passes server-side validation |
| Pod schedulable but Pending, ResourceClaim unallocated | the claim path runs; the Pod waits *only* because no device exists |
| hugepages toggle off / on | the request, `/dev/hugepages-1G` volume, and cleanup initContainer actually drop / appear |
| PodMonitor CRD gating (disabled / enabled-no-CRD / enabled-with-CRD) | the documented "needs the Prometheus Operator CRD or install fails" contract holds live |
| `helm uninstall` clean | no ResourceClaimTemplate / ResourceClaim / Pod left behind |

## Design notes

- **Stub DeviceClass** (`tenstorrent.com`, no devices) stands in for a working DRA layer minus silicon — the one piece the scheduler resolves before creating the claim.
- **Core install runs with `hugepages.enabled=false`** so DRA is the *sole* scheduling constraint (a kind node has no 1Gi hugepages, which would otherwise mask the DRA gate) — this also doubles as the hugepages-off check.
- kind pinned to `v0.30.0` / `kindest/node:v1.34.0` (floor for `resource.k8s.io/v1`).

## Out of scope

Does **not** re-test that the DRA driver hands a Pod `/dev/tenstorrent` — that is tt-operator's `dra-smoke`, on real silicon. This gate assumes a working DRA layer and only verifies *our chart* on top of it.

## Validation

- All checks pass on a local kind 1.34 cluster.
- **Teeth check:** breaking the CEL selector (`attributes` → `attribute`) keeps `helm-chart-validate` green (kubeconform doesn't compile CEL) but turns this gate red (`compilation failed: undefined field`), confirming the layer catches what render can't.
- `actionlint` (CI-equivalent) and `shellcheck` clean.

### DIFF
--- a/.github/scripts/helm_install_gate.sh
+++ b/.github/scripts/helm_install_gate.sh
@@ -1,0 +1,189 @@
+#!/usr/bin/env bash
+# ---------------------------------------------------------------------------
+# Install gate — hardware-free live-install checks for the tt-inference-server
+# chart on a kind cluster. Sibling of the helm-chart-validate render job: that
+# job proves the manifests render, this one proves what render cannot — a *real*
+# API server accepts the DRA ResourceClaimTemplate (CEL selector and all), the
+# Pod becomes schedulable and stays Pending *only* because no Tenstorrent device
+# exists, and the CRD-gated / toggled resources actually apply against a live
+# control plane.
+#
+# NOT re-tested here: that the DRA driver can hand a Pod /dev/tenstorrent — that
+# is tt-operator's dra-smoke, on real silicon. We assume a working DRA layer and
+# stub the one piece of it the scheduler needs (the DeviceClass); the absence of
+# real devices (ResourceSlices) is exactly what keeps the Pod Pending.
+#
+# Preconditions (set up by the caller / workflow):
+#   - kubectl, helm, and jq on PATH; KUBECONFIG pointing at a kind cluster whose
+#     API server serves resource.k8s.io/v1 (Kubernetes >= 1.34, where DRA is GA).
+#   - network access to fetch the Prometheus Operator PodMonitor CRD.
+# ---------------------------------------------------------------------------
+set -euo pipefail
+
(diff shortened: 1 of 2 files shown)


.github/scripts/helm_install_gate.sh
```diff
@@ -0,0 +1,189 @@
...
+# it will even create the claim. tt-dra-driver installs the real one; we stub it
+# (no devices) so the claim path runs and stalls purely on device absence.
+log "setup: stub DeviceClass '$DEVICE_CLASS' (working DRA layer, minus silicon)"
+kubectl apply -f - <<EOF
+apiVersion: resource.k8s.io/v1
+kind: DeviceClass
+metadata:
+  name: ${DEVICE_CLASS}
+spec: {}
+EOF
```
bgoelTT, line 71: Nice

bgoelTT (approved):

PR #4978, opened 2026-08-19
feat: --spec-decode-metrics-url to scrape worker spec-decode counters

ipastalTT:
## Summary

- In a Dynamo deployment the spec-decode benchmark scrapes `vllm:spec_decode_*` counters from the load target (`--service-port`), which is the spec-decode-unaware frontend — so acceptance-rate / mean-accepted-length / per-position columns come back `0`/`null` even though decoding happens on the workers. This adds a repeatable `--spec-decode-metrics-url` flag that points the driver's before/after scrape at the worker `/metrics` endpoint(s) while load stays on the frontend, mirroring the prefix-cache fix in #4402.
- Multi-worker (KV-routed) deployments are supported: before/after snapshots are merged by summing per series across endpoints, so `acceptance_rate = accepted/draft` and `mean_accepted_length = 1 + accepted/num_drafts` are computed from summed deltas. A down worker is skipped with a warning; the scrape fails only when no endpoint responds.
- The scraper also recognizes `tt_spec_decode_*` spellings alongside `vllm:spec_decode_*` (mirroring #4402's `tt_prefix_cache_*` coverage) so a future cpp_server spec-decode implementation populates the columns without a benchmark-side change. Until a TT server emits those counters, the metrics columns are populated only against a vLLM reference server — documented in `docs/workflow_development.md`.
- Wired end-to-end through the same path as #4402: `run.py` → `runtime_config` → `workflow_dispatch` (release + standalone launcher paths, one flag per URL) → `run_workflows.py` → `command_factory`/`SpecDecodeOptions` → `run_spec_decode` → `ServerConnection.spec_decode_metrics_urls` → driver. Without the flag, behavior is unchanged (scrape hits `--service-port`).

## Test plan

- [x] `tests/llm_module/test_spec_decode_metrics.py` — URL-prefix parsing (`vllm:` + `tt_`), multi-endpoint summing, per-endpoint failure tolerance, summed-delta acceptance math
- [x] `tests/llm_module/test_aiperf_spec_decode_driver.py` — URL normalization, worker scrape while AIPerf load stays on the frontend `--url`, fallback to the load target when the flag is unset, best-effort degradation on scrape failure
- [x] `tests/workflow_module/test_command_factory.py` + `tests/workflows/test_workflow_dispatch_routing.py` — flag → options tuple, one `--spec-decode-metrics-url` per URL in the engine argv (no list-stringification)
- [x] Full `tests/` suite passes
- [x] Run against a Dynamo deployment with `--spec-decode-metrics-url <worker>` and confirm acceptance columns populate while load hits the frontend


bgoelTT (approved):
Approving this for now, preferably we test on a real deployment

ipastalTT:
@bgoelTT Ambrose confirmed that it works 

"I was able to run new set of evals with your changes, the acceptance rate seems to propagate"

ddjukicTT (approved):

PR #4980, opened 2026-08-19
fix: Kimi concurrency and scores.

ipastalTT:
# Agentic eval selection, repeatable evals, and Kimi-K2.7-Code tuning

## Summary

Adds two CLI knobs for iterating on agentic/accuracy evals (`--agentic-benchmark`,
`--repeat-evals`), makes agentic run folders collision-free on re-run, reports vLLM
input/output/total token throughput instead of a single "decode" number, and updates
the Kimi-K2.7-Code eval + benchmark configs against the super-cluster endpoint.

### `--agentic-benchmark`: run one agentic eval instead of all of them

`--workflow agentic` previously ran every `EVALS_AGENTIC` task configured for the
model, which for Kimi-K2.7-Code means tau3 + Terminal-Bench + SWE-bench in one shot.
The new flag narrows the selection by alias (`tau3`, `tb2.0`, `tb2.1`, `swebench`)
or raw task name, comma-separated; `all` or omitting it keeps the old behaviour.
`tb2.0` matches `terminal_bench_2` exactly so it cannot also pick up
`terminal_bench_2_1`, while `tau3`/`swebench` match by task-name prefix since those
are families.

A selection matching nothing fails fast with the model's available task names rather
than running an empty sweep. The value rides in `RuntimeConfig`, is forwarded to the
agentic child in `workflow_dispatch`, and is also accepted directly by
`run_workflows.py`; an explicit engine-side flag overrides the runtime-config JSON so
a direct `run_workflows.py` invocation (or a stale JSON) still honours it.

### `--repeat-evals N`

Runs `--workflow evals` N times, reusing the engine's existing `--repeat` loop, so
each run lands under `run_NN/` with an aggregated `summary/`. Rejected for any other
workflow and for values below 1.

### Timestamped agentic run folders

Harbor and mini-SWE-agent both refuse to start in an existing output folder, so a
second agentic run into the same logs dir used to fail outright. The per-task folder
is now stamped (`<task>_<YYYYMMDDTHHMMSS>`) at the start of `run()`. For
Terminal-Bench the stamp goes on the harbor `job_name` with `jobs_dir` staying the
shared `agentic/` parent, which is what harbor expects.

### vLLM benchmarks report input / output / total throughput

`tps_decode_throughput` was the only throughput reported, which under-describes a
high-ISL sweep where prefill dominates. Records now carry `tps_input_throughput`,
`tps_output_throughput`, and `tps_total_throughput`. vLLM has no input-throughput
field of its own, but its two throughputs share one duration, so the difference is
exactly the input rate; both derived values fall back to recomputing from token
totals and duration when the fields are absent.

Display names change from "Tput Decode" to "Tput Output" (docs tables updated to
match). `target_checks` reads the new key and falls back to `tps_decode_throughput`,
so existing aiperf/genai-perf records keep resolving.

### Benchmark sweep updates

- New `(8192, 1024)` and `(10000, 1024)` pairs in the shared sweep.
- Super-cluster extras now cover `(128, 255872)` (long-decode) and top out at
  `(255872, 128)` rather than 245760, matching Kimi's 256K context.
- The super-cluster prompt floor changes from a flat 256 prompts to
  `2 x model max_concurrency`, so the floor scales with the batch size instead of
  being  constant.

### Kimi-K2.7-Code eval config

- Agentic concurrency raised to match what the endpoint actually serves:
  Terminal-Bench 2.1 4 -> 64, SWE-bench 6 -> 64, tau3 4 -> 32.
- Context limits corrected from `256 * 1024` to `256 * 1000` (the served window).
- tau3 pins `top_p` to 0.95 via `llm_args_json`. The tau2 adapter's
  `build_llm_args()` sets only temperature, so the agent was sampling at
  `top_p=1.0` while every other Kimi task clamps to 0.95; the untruncated tail
  emitted malformed tool-call arguments in roughly 2% of calls, and one bad parse
  aborts the whole trial. It must stay a JSON string because the adapter
  `shlex.quote`s the value onto the container command line.
- GPQA-Diamond: `capture_reasoning=True`, GPU reference score 85.3 -> 89.8.

### Dependencies

- `lm-eval` temporarily points at the `ipastalTT` fork, which carries an in-stream
  SSE error-propagation fix; comment notes to revert to `tstescoTT` once merged.
- `swo-bench` unpinned from `3.1.2`. SwarmOne keeps only the latest release on
  PyPI and deletes older ones, so the pin stops resolving as soon as they publish
  again; the installed version is still captured per run as `swo_bench_version`.

## Test plan

- [x] `pytest tests/llm_module/test_parsers.py tests/llm_module/test_target_checks.py
      tests/llm_module/test_agentic_traces.py tests/test_benchmark_config.py
      tests/test_module/llm_tests/test_agentic_eval_tests.py
      tests/workflows/test_workflow_dispatch_routing.py`
- [x] `--workflow agentic --agentic-benchmark tau3` against the super-cluster
      endpoint runs only the tau3 tasks; an unknown value fails with the task list.
- [x] Re-run the same agentic task twice into one logs dir; both get their own
      timestamped folder and neither harness errors on an existing directory.
- [x] `--workflow evals --repeat-evals 2` produces `run_01/`, `run_02/`, `summary/`.
- [x] A vLLM benchmark report shows Tput Input / Output / Total, and scalar target
      checks still resolve against older aiperf records.


idjuricTT (approved):
@ddjukicTT could you please also check just in case

reference_config/evals/eval_config.py
```diff
@@ -691,14 +692,14 @@ class EvalConfig:
                     },
                 ),
                 model_kwargs={
-                    "max_length": 256 * 1024,
+                    "max_length": 256 * 1000,
```
ddjukicTT, line 695: Kimi’s spec still has max_context: 262144 (256 * 1024) (here: https://github.com/tenstorrent/tt-inference-server/blob/main/workflows/model_specs/dev/llm.yaml#L1430). Is this intentional?

ddjukicTT (approved):
lgtm just please look at the comment about config numbers, not sure if it is intentional

reference_config/evals/eval_config.py
ipastalTT: we have to lower it for now since we support 256000 and not the full context length. Do you think I should change the yaml value as well?

PR #4983, opened 2026-08-19
feat(tt-media-server): allow presigned URL downloads for I2V image prompts

### LINKED ISSUE #4974: Allow downloads from presigned URLs for video in tt-inference-server
We want to move towards multimodal models whose assets can't be passed inline with the request. To support this, we need to allow inference server to pull from presigned URLs(example: s3)

anatarajan-tt:
## Summary

Closes #4974.

`image_prompts[].image` on `POST /v1/videos/generations/i2v` now accepts an http(s) URL — e.g. a presigned S3 GET URL — next to inline base64. The server downloads the asset at the API layer, before the job is enqueued, and replaces the URL with base64. Runners, SHM spill, and workers are unchanged and keep seeing base64 only.

**Change class: semantics-altering** — a new accepted input shape on the I2V endpoint. Inline-base64 and multipart-upload flows are byte-for-byte unchanged.

## Design

All remote fetches go through one hardened downloader (`tt-media-server/utils/media_downloader.py`). The design copies the strongest parts of the two reference stacks (researched against vLLM `MediaConnector` @ 583a0025 and SGLang `download_remote_media` @ 5f128395):

- **http(s) only**; `data:`/`file:`/other schemes never reach the downloader.
- **Required hostname allowlist** (`MEDIA_URL_ALLOWED_DOMAINS`): deny-when-empty — URL-valued media is refused until an allowlist is configured, so the SSRF surface only exists where an operator opened it. Entries are exact hostnames or label-anchored wildcards (`*.s3.amazonaws.com` matches `bucket.s3.amazonaws.com` at any depth, never the bare suffix, never `evil-s3.amazonaws.com`), normalized (IDNA, case, IP literals) and re-checked on **every redirect hop**. URLs are validated as `httpx.URL` — the same parser the client connects with — because both vLLM SSRF CVEs (GHSA-qh4c-xf7m-gxfc, GHSA-v359-jj2v-j536) came from validator/client parser mismatch.
- **Manual redirect following** (301/302/303/307/308, max 5 by default); redirect without `Location` fails closed.
- **One total deadline** (default 30 s) shared across all of a request's assets — 81 URL entries cannot hold the connection open for 81 timeouts.
- **Streamed byte cap** (default 7,500,000 bytes — base64-encodes to exactly the existing 10,000,000-char `ImagePromptEntry.image` field cap), enforced via Content-Length precheck and again mid-stream.
- **Error taxonomy**: policy violation → 400, over cap → 413, origin error/timeout → 422 (an expired presigned URL surfaces as 422, not 500), server misconfiguration → 500. Downloaded bytes must PIL-decode or the request 422s.
- Presigned URLs need no special casing: query-string auth passes through because validation covers scheme + hostname only.
- Error and log text redacts URLs to scheme+host+path — presigned query strings carry credentials (`X-Amz-Signature`).
- `ImageManager.base64_to_pil_image` fails loudly if an unresolved URL ever reaches a runner's decoder (defense-in-depth for the "runners only see base64" invariant).

New env-driven settings (`config/settings.py`): `media_url_download_enabled`, `media_url_allowed_domains`, `media_url_max_bytes`, `media_url_timeout_seconds`, `media_url_max_redirects`.

## What the reviewer should verify

- The allowlist/redirect/cap/deadline policy in `utils/media_downloader.py` matches the description above (it is the security boundary).
- `_resolve_image_prompt_urls` in `open_ai_api/video.py` is the only place URL-shaped `image` values can pass, and every submit path (`/generations`, `/generations/i2v`, sync and async) funnels through `_submit_video_request`.
- The 413 guard on encoded length keeps SP-runner workers (which re-validate `ImagePromptEntry` from the side-file) from failing mid-job.

## Testing

- 44 new unit tests: `tests/test_media_downloader.py` (policy matrix on fake transports — no network) and `tests/test_video_media_url.py` (schema acceptance, wiring, error mapping).
- Adjacent suites green: `test_video_api`, `test_image_manager`, `test_video_manager`, `test_video_runner`, `test_mock_video_runner`, `test_sp_runner`, `test_scheduler` (219 passed, 1 pre-existing skip).
- `ruff check` and `ruff format --check` clean with the CI-pinned ruff 0.15.0.

## Known limitations / follow-ups

- No retry/backoff for S3 503 SlowDown (vLLM retries these; deliberate v1 non-goal).
- An origin that trickles bytes just inside the per-read timeout can stretch a hop somewhat past the deadline (closing this needs `asyncio.timeout`, Python 3.11+; base image is py3.10).
- URL acceptance for image (img2img/edits) and audio endpoints, and presigned *output* upload, are follow-ups; the downloader is written to be reused there.

### DIFF
--- a/tt-media-server/domain/video_i2v_generate_request.py
+++ b/tt-media-server/domain/video_i2v_generate_request.py
@@ -5,8 +5,9 @@
 """Image-to-Video request schema for Wan2.2 I2V.
 
 Extends ``VideoGenerateRequest`` with a list of image prompts. Each entry
-pairs a base64-encoded image with a frame position so the caller can anchor
-the generation at one or more frames across the output video.
+pairs an image — base64-encoded, or an http(s)/presigned URL (#4974) — with
+a frame position so the caller can anchor the generation at one or more
+frames across the output video.
 
 Validation mirrors the upstream ``WanPipelineI2V.prepare_latents`` contract.
 The pipeline-level ``num_frames`` (used by both the runner and the validators
@@ -19,6 +20,7 @@ from config.constants import WAN22_NUM_FRAMES
 from domain.video_generate_request import VideoGenerateRequest
 from pydantic import BaseModel, Field, field_validator
 from utils.image_manager import ImageManager
+from utils.media_downloader import is_media_url
 
 # The cap exists to bound HTTP body size, not to match
 # any pipeline constraint.
@@ -35,6 +37,13 @@ class ImagePromptEntry(BaseModel):
     @classmethod
     def validate_decodable_image(cls, v: str) -> str:
         """Ensure the base64 string decodes to a valid PIL image via ImageManager."""
+
+        if is_media_url(v):
+            # Remote asset (e.g. presigned S3 URL): downloaded, decoded, and
+            # policy-checked at the API layer before enqueue
+            # (open_ai_api/video.py), where failures map to real HTTP
+            # statuses instead of a blanket 422 here.
+            return v
 
         try:
             img = ImageManager().base64_to_pil_image(v)
--- a/tt-media-server/utils/media_downloader.py
+++ b/tt-media-server/utils/media_downloader.py
@@ -1,0 +1,290 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+"""Hardened downloader for client-supplied media URLs (#4974).
+
+Lets request media fields carry a presigned URL (e.g. S3) instead of inline
+base64. Every remote fetch in the request path must go through
+``download_media_url`` — side-channel fetches are how reference servers
+(vLLM, SGLang) reintroduced SSRF after hardening their main path.
+
+Policy, following SGLang's ``download_remote_media`` with vLLM's error
+taxonomy on top:
+
+* http(s) only; the optional exact-hostname allowlist is normalized (IDNA,
+  case, IP literals) and checked on the initial URL and on every redirect hop.
+* URLs are validated as ``httpx.URL`` — the same parser the client uses — so
+  a validator/client parser mismatch cannot bypass the allowlist.
+* Redirects are followed manually (301/302/303/307/308, capped) so each
+  destination is validated before a connection is made.
+* One total deadline covers all hops and the body read; the body is streamed
+  against the byte cap instead of buffered blindly. Callers can pass a shared
+  ``deadline`` so one request's assets share a single budget. Known
+  limitation: a origin that trickles bytes just inside the per-read timeout
+  can stretch a hop somewhat past the deadline (checked between chunks;
+  ``asyncio.timeout`` would close this but needs Python 3.11+).
+
+Error taxonomy (mapped to HTTP statuses at the endpoint):
+
+* ``MediaDownloadPolicyError`` — the URL violates server policy (400).
+* ``MediaDownloadTooLargeError`` — the body exceeds the byte cap (413).
+* ``MediaDownloadFetchError`` — the origin failed us: HTTP error such as an
+  expired presigned URL, network error, or deadline exceeded (422).
+* ``MediaDownloadError`` (base, raised directly) — server-side
+  misconfiguration; deliberately NOT caught at the endpoint so it surfaces
+  as a 500, not a client-fault 4xx.
+"""
+
+import asyncio
+import ipaddress
+import time
+from typing import Optional
+
+import httpx
+from config.settings import settings
+from utils.logger import TTLogger
+
+logger = TTLogger()
+
+_REDIRECT_STATUS_CODES = frozenset({301, 302, 303, 307, 308})
+_READ_CHUNK_BYTES = 64 * 1024
+
+_shared_client: Optional[httpx.AsyncClient] = None
+_shared_client_loop: Optional[asyncio.AbstractEventLoop] = None
+_warned_open_allowlist = False
+
+
+class MediaDownloadError(Exception):
+    """Base class; raised directly only for server-side misconfiguration."""
+
+
+class MediaDownloadPolicyError(MediaDownloadError):
+    """URL violates server download policy (scheme, domain, redirects)."""
+
+
+class MediaDownloadTooLargeError(MediaDownloadError):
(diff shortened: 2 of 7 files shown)


tt-media-server/utils/media_downloader.py
```diff
@@ -0,0 +1,290 @@
...
+
+    allowed = _allowed_domains()
+    if allowed:
+        hostname = _normalize_hostname(parsed.host)
+        if hostname not in allowed:
+            # Do not echo the allowlist: it is deployment configuration.
+            raise MediaDownloadPolicyError(
+                f"Media URL domain {hostname!r} is not in the allowed domains list"
+            )
+    return parsed
```
vpetrovicTT: Maybe I am halucinating here, but if **allowed** is empty, we allow everything here? 
Shouldn't this be in reverse? If the **allowed** is empty shouldn't we deny?

tt-media-server/utils/media_downloader.py
```diff
@@ -0,0 +1,290 @@
...
+    global _warned_open_allowlist
+    if _warned_open_allowlist:
+        return
+    if not _allowed_domains():
+        logger.warning(
+            "Media URL download is enabled with no media_url_allowed_domains; "
+            "any authenticated client can make this server fetch arbitrary "
+            "http(s) URLs (SSRF surface). Set an allowlist on deployments "
+            "reachable by untrusted clients."
+        )
```
vpetrovicTT: Question...Warning is enough here?

tt-media-server/utils/media_downloader.py
```diff
@@ -0,0 +1,290 @@
...
+
+    if parsed.scheme not in ("http", "https") or not parsed.host:
+        raise MediaDownloadPolicyError(
+            f"Media URL must be http(s) with a hostname, got: {url!r}"
+        )
+
+    allowed = _allowed_domains()
+    if allowed:
+        hostname = _normalize_hostname(parsed.host)
+        if hostname not in allowed:
```
vpetrovicTT: the **settings.py**  code:
`media_url_download_enabled: bool = True
    # Comma-separated exact hostnames (e.g. "bucket.s3.us-east-1.amazonaws.com").`

---
Should we maybe match DNS suffixes / wildcards, e.g. `*.s3.amazonaws.com` here??

tt-media-server/domain/video_i2v_generate_request.py
```diff
@@ -36,6 +38,13 @@ class ImagePromptEntry(BaseModel):
     def validate_decodable_image(cls, v: str) -> str:
         """Ensure the base64 string decodes to a valid PIL image via ImageManager."""
 
+        if is_media_url(v):
```
vpetrovicTT, line 41: We are saying here that **url** is valid **ImagePromptEntry**??
I think this could make mumbo jumbo later in the request workflow :/

tt-media-server/utils/media_downloader.py
```diff
@@ -0,0 +1,290 @@
...
+    _warn_if_open_allowlist()
+    http_client = client if client is not None else _get_shared_client()
+    if deadline is None:
+        deadline = time.monotonic() + timeout_seconds
+
+    try:
+        for redirect_count in range(max_redirects + 1):
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                raise MediaDownloadFetchError(f"Timed out downloading media URL: {url}")
```
vpetrovicTT: Nit: would not log the whole **url** here

tt-media-server/utils/media_downloader.py
anatarajan-tt: Agreed, and fixed in 0884d8d — flipped to deny-when-empty: URL download now requires `media_url_allowed_domains` to be configured, and with an empty allowlist every URL-valued image is rejected with 400. The allow-all-when-empty posture came from vLLM/SGLang, but they keep it only for backward compatibility with existing users — this feature is brand new here, so there is nothing to stay compatible with. Deny-when-empty is strictly better.

---
agent on behalf of anatarajan-tt

tt-media-server/utils/media_downloader.py
anatarajan-tt: This resolves itself with the deny-when-empty flip in 0884d8d: the warning existed only because downloads could run with no allowlist. That state is now unreachable, so the warning is removed rather than strengthened.

---
agent on behalf of anatarajan-tt

tt-media-server/utils/media_downloader.py
anatarajan-tt: Added in 0884d8d — allowlist entries now support label-anchored wildcards: `*.s3.amazonaws.com` matches `bucket.s3.amazonaws.com` at any subdomain depth, but never the bare suffix and never `evil-s3.amazonaws.com` (a `.` label boundary is required, so it is not a naive endswith). A bare `*` entry is refused as a misconfiguration — allow-everything must not be expressible through the allowlist. Exact hostnames still work, and both forms are IDNA/case/IP-normalized. Covered by tests.

---
agent on behalf of anatarajan-tt

tt-media-server/domain/video_i2v_generate_request.py
anatarajan-tt: Fair concern — the "runners only ever see base64" invariant was enforced by a single call site, not by the type. Today every submit path funnels through `_submit_video_request` → `_resolve_image_prompt_urls`, so a URL cannot reach the queue; the risk is a future endpoint that skips the funnel. In 0884d8d I added a guard in `ImageManager.base64_to_pil_image` that raises `unresolved media URL reached the image decoder` on http(s) input, so a violated invariant fails loudly at the decode boundary instead of surfacing as base64 garbage mid-job. (Accepting the URL in the schema and resolving pre-enqueue is the same pattern vLLM and SGLang use.)

---
agent on behalf of anatarajan-tt

tt-media-server/utils/media_downloader.py
anatarajan-tt: Upgraded this from nit to a real finding — presigned query strings carry credentials (`X-Amz-Signature`), and the error text goes into HTTP error details and server logs. In 0884d8d all error/log text now redacts URLs to scheme+host+path (`redact_url`), the httpx exception body is no longer interpolated (it can embed the full URL), and a test asserts the signature never appears in error text.

---
agent on behalf of anatarajan-tt

vpetrovicTT (approved):

PR #4997, opened 2026-08-21
[Helm v0.2] CI: nightly real-hardware smoke for the chart

sott0n:
Part of #4869. Adds the real-hardware tier of the chart's CI — above the render gate and the kind install gate.

**The smoke** (`.github/scripts/helm_hw_smoke.sh`) runs nightly at 20:17 UTC and on `workflow_dispatch` (which takes a ref, so a chart PR can be checked on hardware before merge). It gates nothing in `test-gate.yml`. It asserts only what the cheaper tiers cannot:

1. the DRA claim allocates and `/dev/tenstorrent/<N>` lands in the **real inference container** — no privileged, no hostPath
2. Ready inside the startupProbe budget, no restart, with the measured compile time reported against that budget
3. `/health`, `/v1/models`, one inference call — and that an unauthenticated call is refused
4. `helm uninstall` releases the board, no Pod stuck `Terminating`

The runner pool is bare, so `hw_smoke_bootstrap.sh` stands up RKE2 + tt-operator (NFD + TTFM + tt-dra-driver) first, and restarts the DRA kubelet plugin once TTFM is Ready — the plugin only discovers boards at startup, which is why tt-operator's own `dra-smoke` fails on 40 of 42 recent job instances.

**Chart changes**, all found by running the smoke:

- `auth.apiKey` → Secret as `API_KEY` (media/forge) or `VLLM_API_KEY` (vllm), `auth.disabled` → `NO_AUTH=1`, and media/forge now fail closed until one is set. **Breaking for those installs**: the servers fall back to a well-known built-in key when `API_KEY` is unset, so a chart-deployed one looks authenticated and is not.
- hugepages sized from the device's ASIC count (`deviceChipCounts`, generated from the same table as `deviceBoardCounts`) instead of a fixed 32Gi: n150 1Gi, n300 2Gi, t3k 8Gi, galaxy 32Gi — the last unchanged, and where the old constant came from. The fixed request made every host smaller than a Galaxy unschedulable. `hugepages.size` overrides.
- `HF_HOME` under the cache volume for media/forge, so weights survive a Pod restart instead of being re-downloaded.
- a `cacheRoot` helper replacing two hardcoded copies of the mount path, and ConfigMap `data` built as a dict so an `extraEnv` key colliding with a chart key no longer breaks the render.

Verified on the n150 pool ([run](https://github.com/tenstorrent/tt-inference-server/actions/runs/32477829484)): all four asserts, Ready after 506s of compile with the chart's computed 1Gi of hugepages, `/v1/embeddings` 401 without the key and 200 with it.

### DIFF
--- a/.github/scripts/hw_smoke_bootstrap.sh
+++ b/.github/scripts/hw_smoke_bootstrap.sh
@@ -1,32 +1,15 @@
 #!/usr/bin/env bash
-# ---------------------------------------------------------------------------
-# Bootstrap a Tenstorrent hardware runner into a cluster the chart can run on.
+# Turns a bare Tenstorrent runner (ephemeral KVM VM, card passed through, tt-kmd
+# preloaded) into what helm_hw_smoke.sh needs: an RKE2 cluster with tt-operator
+# publishing devices. tt-k8s-driver-manager stays off — the runner's kmd is
+# already loaded, and reinstalling it costs two DKMS builds and a device churn
+# window per run. tt-telemetry / jobset / kubepmix are off as unused.
 #
-# The hardware smoke (.github/scripts/helm_hw_smoke.sh) assumes "tt-operator is
-# installed and publishing devices". On the shared TT runner pool that is not a
-# given: the runners are ephemeral KVM VMs with a card passed through and tt-kmd
-# preloaded, and nothing else. This script turns one into that assumption:
-#
-#   RKE2 single-node cluster (DRA is GA in the pinned k8s) ->
-#   tt-operator umbrella, NFD + tt-fabric-manager + tt-dra-driver only ->
-#   devices published in a ResourceSlice
-#
-# tt-k8s-driver-manager is deliberately OFF: the runner already has tt-kmd
-# loaded, and letting the manager reinstall it would add two DKMS builds and a
-# device churn window to every run for nothing this smoke asserts. Same for
-# tt-telemetry / jobset / kubepmix (kubepmix would also drag in cert-manager).
-#
-# THE PLUGIN RESTART, and why it is here rather than in the smoke: the DRA
-# kubelet plugin discovers boards through tt-fabric-manager's GetTopology and
-# publishes a ResourceSlice. If it comes up before TTFM has topology, it
-# publishes a slice with no devices and does not refresh it for the life of the
-# pod — claims then never allocate. In tt-operator's own integration-rke2 runs
-# this is visible as `dra-smoke` failing on 40 of 42 job instances; the two that
-# allocated are exactly the two where the plugin pod happened to (re)start ~45s
-# after the driver was up. Restarting the DaemonSet once TTFM is Ready makes
-# that ordering deliberate instead of lucky.
-#
-# Usage:  bash .github/scripts/hw_smoke_bootstrap.sh
+# The DRA kubelet plugin only discovers boards at startup, so one that comes up
+# before tt-fabric-manager has topology publishes a device-less ResourceSlice and
+# never refreshes it — hence the restart once TTFM is Ready. In tt-operator's own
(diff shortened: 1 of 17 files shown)


.github/scripts/hw_smoke_bootstrap.sh
```diff
@@ -0,0 +1,212 @@
...
+# preloaded) into what helm_hw_smoke.sh needs: an RKE2 cluster with tt-operator
+# publishing devices. tt-k8s-driver-manager stays off — the runner's kmd is
+# already loaded, and reinstalling it costs two DKMS builds and a device churn
+# window per run. tt-telemetry / jobset / kubepmix are off as unused.
+#
+# The DRA kubelet plugin only discovers boards at startup, so one that comes up
+# before tt-fabric-manager has topology publishes a device-less ResourceSlice and
+# never refreshes it — hence the restart once TTFM is Ready. In tt-operator's own
+# integration-rke2 runs, dra-smoke fails on 40 of 42 jobs for exactly this, and
+# both that allocated are ones where the plugin happened to restart late.
```
bgoelTT, line 12: Can you please share more details about this? This seems like a bug that we need to fix upstream

I'd like to get it filed so we can assign someone to resolve

bgoelTT (approved):

.github/scripts/hw_smoke_bootstrap.sh
sott0n: Reproduced on a T3K box today, same versions as CI (tt-dra-driver 0.0.48, tt-fabric-manager 0.2.29). It happens in **driver-manager mode** — i.e. whenever the umbrella is installed before the kmd exists, which is the flow this repo's CI exercises.

Repro:

1. `modprobe -r tenstorrent` (no `/dev/tenstorrent`), then `helm install` the umbrella.
2. TTFM finds nothing, retries, gives up, and publishes an empty topology:

```
UMD discovered 0 device(s)
Scheduled rediscovery produced an empty topology; treating as unsuccessful so it is retried
Discovery retry budget exhausted; giving up on scheduled discovery retries
Publishing rediscovered topology to controller via heartbeat: 0 ASICs
```

3. The plugin takes that as the answer: one ResourceSlice with `devices` absent, no error in its log, and it stops asking. (Confirmed separately: delete the TTFM agent under a *running* plugin and the slice keeps its old 4-device list for minutes — the plugin publishes once, at startup.)
4. `modprobe tenstorrent` → 4 boards back on the host → slice still 0 devices, unchanged 90s later. Claims stay unschedulable.

Recovery required recreating **both** pods: TTFM first — a privileged container's `/dev` is a snapshot from container start, so it cannot see device nodes that appeared later — then the plugin. After that, 4 devices published.

Same state in this repo's CI, e.g. [run 30703927524](https://github.com/tenstorrent/tt-operator/actions/runs/30703927524) (job `91379621253`): slice with zero devices, no plugin error, scheduler repeating `cannot allocate all claims`. Across the last 20 integration-rke2 runs (07-23 → 08-01), dra-smoke reached a claim 35 times and this was the state in 33; the 2 that allocated are the ones where the plugin pod happened to restart late, e.g. [run 30682388291](https://github.com/tenstorrent/tt-operator/actions/runs/30682388291) (job `91321857981`). The runs are all green because dra-smoke is `continue-on-error`.

That's what the plugin restart in our bootstrap is for.

If there’s anything wrong with my understanding or the steps I’m assuming, please let me know.

bgoelTT: Thanks for sharing all these details, I do believe this is a bug that we'll have to fix.

sott0n:
@acvejicTT @mdobrosavljevicTT @mjeremicTT @vcankovicTT @vmaksimovicTT could anyone review it?

vmaksimovicTT (approved):

PR #5014, opened 2026-08-24
Enable tool calling and thinking on the gemma4-31b-autoport spec

vkovacevicTT:
Fixes the `swe_bench_verified` zero on QB2 (tenstorrent/tt-agentic-bringup-qb2#2): mini-swe-agent needs `tool_choice:"auto"`, but the autoport spec never inherited the tool-calling `vllm_args` from the tt_transformers entry, so every instance died on its first LLM call (HTTP 400).

Restores the four `vllm_args` (tool-call + reasoning parser, `enable_thinking=true`). Not copied: `metadata` parser names (unused) and `sample_on_device_mode: decode_only` (autoport requires `all`).

**Merge tenstorrent/vllm-tt-plugin#72 first** — without it these flags crash the gemma4 tool parser on every tool call.

Verified: dev spec load renders the flags into `vllm_args`; yaml/spec tests pass.

### DIFF
--- a/workflows/model_specs/dev/llm.yaml
+++ b/workflows/model_specs/dev/llm.yaml
@@ -570,15 +570,12 @@ templates:
       tool_call_parser_name: qwen3_coder
 
 # google/gemma-4-31B-it served by the agentic autoport rather than tt_transformers.
-# Mirrors the tt_transformers entry above. Five deviations, all required:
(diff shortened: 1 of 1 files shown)


vkovacevicTT:
Consolidated into #5005 (same change, cherry-picked verbatim) so the full QB2 agentic eval fix lives in one PR on the bring-up branch.

PR #5016, opened 2026-08-24
Add Qwen3-ASR-1.7B on N150 in tt media server

tvardhineniTT:
## Summary
- Register `Qwen3-ASR-1.7B` for **N150**  (media-server runner, `impl: qwen3_asr`, EXPERIMENTAL).
- Reuses the proven tt-metal demo pipeline (Whisper mel + Qwen3-ASR prompt → TT audio encoder → baked-in 68 tok/s decoder: on-device argmax + decode trace + in-graph token/pos + 2CQ, BFP8 KV/attn/lm_head).
- Japanese is the primary target; quality validated byte-identical on real clips.
- **SSE streaming**: token-by-token transcription (decoder `generate_iter`), so the audio benchmark's `tput_user` gates instead of reporting NA.
- Audio-length **bucketing** + long-window **sub-chunking** (no trace-breaking, no silent truncation).
- Benchmark targets + nightly Models CI (N150).

**tt-metal pin for image build:** `tvardhineni/qwen3_asr_wh` @ `62fba73`.
Models CI / `--docker-server` must build from that SHA (`--dev-mode`, `--override-docker-image` until prod promotion).

## Test plan
- `run.py --dev-mode --help` lists `Qwen3-ASR-1.7B`
- Local `/v1/audio/transcriptions` HTTP 200 (streaming + non-streaming), N150: EN + JA
- Benchmark: `tput_user` passes 67.0, RTR 24.8 (RTF ≈ 0.04) on 30s clip, acceptance PASS
- Models CI nightly build from metal `62fba73`
- Do not include `tt-media-server/generated/fabric/*`


cleonidouTT:
Dispatch Job: https://github.com/tenstorrent/tt-shield/actions/runs/33383466503

Edit: LibriSpeech never ran, and I found the exact line responsible.

## Why librispeech didn't run

`run_audio_eval` gates the real lmms-eval path behind an implementation-name check:

```254:256:test_module/eval_tests/audio_eval_tests.py
def _is_whisper(ctx: MediaContext) -> bool:
    impl = getattr(ctx.model_spec, "impl", None)
    return getattr(impl, "impl_name", None) == "whisper"
```

```444:445:test_module/eval_tests/audio_eval_tests.py
    if _is_whisper(ctx):
        return _run_whisper_lmms_eval(ctx)
```

Qwen3-ASR's `impl_name` is `qwen3-asr`, so it falls through to the generic branch below, which just replays one built-in payload `num_calls` times. The log confirms it: `Running audio transcription benchmark with 2 calls`, transcribing the repo's own 30.38s "Lesson 1C Exercise 2…" clip — not a LibriSpeech utterance. There is no `lmms-eval --tasks librispeech_test_other` invocation anywhere in the 1328 lines.

cleonidouTT:
New Dispatch Job (N150): https://github.com/tenstorrent/tt-shield/actions/runs/33519375765

cleonidouTT:
Dispatch Job (Galaxy node): https://github.com/tenstorrent/tt-shield/actions/runs/33520215506

cleonidouTT:
**Everything passed, except the strictest benchmark tier, which failed on TTFT — and the report doesn't show it.** Acceptance still says PASS because the gate accepts any tier passing in full, and the `complete` tier did.

From the log (`target_check` lines 11336–11345):

| Tier | TTFT | Tput User | RTR | Verdict |
|---|---|---|---|---|
| functional | ✅ 857ms vs 8000 | ✅ 116.2 vs 6.7 | ✅ 13.4 vs 0.95 | PASS |
| complete | ✅ 857ms vs 1600 | ✅ 116.2 vs 33.5 | ✅ 13.4 vs 4.75 | PASS |
| **target** | ❌ **857ms vs 800** | ✅ 116.2 vs 67.0 | ✅ 13.4 vs 9.5 | **FAIL** |

`Overall target_check (target tier): FAIL`. You missed the 800ms TTFT target by 7% (ratio 1.071). Throughput and RTR beat their targets comfortably.

## Evals — the fix worked

LibriSpeech genuinely ran this time. `20260902_035448_results.json` confirms it end to end: `model = whisper_tt`, `model_args = model=Qwen/Qwen3-ASR-1.7B`, `limit = None`, and `n-samples: original 2939, effective 2939` — the complete test-other set, not a subset.

**WER 12.3237, score 87.68.** Accuracy check is still NA because I deliberately left the reference unset

It took 5m56s for all 2939 samples

## Spec tests — Galaxy DP=32 fully working

All 6 passed, and device liveness reports **32/32 workers ready and alive**. The DP burst results:

| Test | Concurrent | OK | Total | Avg |
|---|---|---|---|---|
| Dp2Chunk5 | 64 | 64/64 | 12.80s | 6.2s |
| Dp2Chunk30 | 64 | 64/64 | 12.37s | 5.9s |

Those two being within 3% of each other is empirical confirmation of the caveat I documented in the config: neither test sets `AUDIO_CHUNK_DURATION_SECONDS`, so both ran against the same server-side value and differ only in the number they *record*. If they were genuinely measuring 5s versus 30s chunking you'd expect a much larger gap. So you have DP fan-out coverage at 64 concurrent, but not yet a real chunk-size comparison.

## Four things worth fixing

**The report hides the tier breakdown.** `report_module/renderers.py:350` returns early for multi-row blocks, emitting only heading + table, so nested sub-tables are never rendered. The sweep block puts tiers in `data["target_checks"]`, which the acceptance gate reads correctly but the renderer never displays. That's why you see the explanatory footnote about the Target Check column with neither the column nor the table above it. It's pre-existing (whisper has always used the sweep), but my change newly exposed Qwen3-ASR to it — and on your next n150 run you'll lose the tier table you had before. The gate is correct; only the display is blind.

**The 30s row is a cold pass.** TTFT 5.375s and T/S/U 1.399e+04 are meaningless — the first sweep entry absorbs warmup, and that T/S/U is the divide-by-near-zero artifact when TTFT ≈ total duration. The gate correctly uses the warm 60s pass. Just don't read the 30s row as a measurement.

**T/S/U 116.2 is still inflated** by the final-chunk double-count I flagged earlier, so the comfortable `tput_user` pass against 67.0 is softer than it looks.

**Eval output landed in `eval_whisper_Qwen3-ASR-1.7B_n150/`** on a Galaxy run — `whisper_eval_test.py:126` hardcodes a `whisper_` prefix and `_n150` suffix in its fallback model_id. Cosmetic, but confusing when hunting for artifacts.

cleonidouTT:
n150 Dispatch Job: https://github.com/tenstorrent/tt-shield/actions/runs/33605859543
Done

PR #5020, opened 2026-08-25
Qwen3_embedding_8B - Mesh shape change for t3k and n300 and Concurrency raised to 32 for Galaxy

jzivanovicTT:
Link to failed run (t3k): https://github.com/tenstorrent/tt-shield/actions/runs/32488809637/job/96791471973

Error - Incompatible sharding strategy set by model and mesh_device shape in tt-media-server for t3k and n300:

image

##

Concurrency 1 run on 6u: https://github.com/tenstorrent/tt-shield/actions/runs/32864290132/job/97870608804

Concurrency 32 run on 6u: https://github.com/tenstorrent/tt-shield/actions/runs/32976046805/job/98215729255

### DIFF
--- a/tt-media-server/config/constants.py
+++ b/tt-media-server/config/constants.py
@@ -1347,7 +1347,7 @@ ModelConfigs = {
         "queue_for_multiprocessing": QueueType.FasterFifo.value,
     },
     (ModelRunners.QWEN_EMBEDDING_8B, DeviceTypes.N300): {
-        "device_mesh_shape": (2, 1),
+        "device_mesh_shape": (1, 2),
         "is_galaxy": False,
         "device_ids": DeviceIds.DEVICE_IDS_1.value,
         "max_batch_size": 2,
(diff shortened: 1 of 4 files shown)


ztorlakTT (approved):

idjuricTT (approved):

tt-media-server/config/constants.py
```diff
@@ -1347,7 +1347,7 @@ def _vllm_config(
         "queue_for_multiprocessing": QueueType.FasterFifo.value,
     },
     (ModelRunners.QWEN_EMBEDDING_8B, DeviceTypes.N300): {
-        "device_mesh_shape": (2, 1),
+        "device_mesh_shape": (1, 2),
```
idjuricTT, line 1350: Did this work? 😄 

I thought we need (1,1) always

jzivanovicTT: It solved the problem I had in this run, basically, the Qwen3-Embedding-8B is large for Embedding model, it is decoder based, so its basically an LLM and we use the representation of the last token in a sequence as the embedding for the whole sequence. Because of the models size it fits on a single chip only with aggressive quantization and reduced max_sequence_length and max_batch_size of 1, so if there are multiple chips available the model is written to utilize them through a tensor parallelism. I am sending you my document where I analyzed this in detail (on slack). 

But unfortunately, after that fix I have come up to a new problem, for some reason the tt_metal throws a segfault on t3k (I think it's some index out of bounds) - https://github.com/tenstorrent/tt-shield/actions/runs/32857929916/job/97849439361.

mjeremicTT (approved):

Issue #4971, opened 2026-08-18
Scope Kimi-K3 evals


bgoelTT:
Similar to https://github.com/tenstorrent/tt-inference-server/issues/4324 https://github.com/tenstorrent/tt-inference-server/issues/4271 https://github.com/tenstorrent/tt-inference-server/issues/4376 https://github.com/tenstorrent/tt-inference-server/issues/4641 , we need to determine which evals we want to run for this model

Eval plan of record, ranked in decreasing priority and importance:

* GPQA-Diamond
* Terminal-Bench-2.1
* LongBench-V2
* tau^3-banking (we'll compare to AA on July 23rd, 2026)
* SWE-Bench Verified

bgoelTT:
Daily:
* blocked waiting for access to GPUs. Expect to get reservation tomorrow

ipastalTT:
Reasoning evals done. **Note** on GPQA the model got into a loop in one question
## Tenstorrent Model Release Summary: moonshotai/Kimi-K3 on SUPER_CLUSTER

### Metadata: moonshotai/Kimi-K3 on SUPER_CLUSTER

```json
{
    "model_name": "moonshotai/Kimi-K3",
    "device": "SUPER_CLUSTER",
    "generated_at": "2026-08-24T13:21:46+00:00",
    "report_id": "moonshotai__Kimi-K3_2026-08-24T132146+0000",
    "workflow": "evals",
    "server_mode": "API",
    "run_command": "python run.py --model Kimi-K3 --workflow evals --device super_cluster --server-url http://172.17.0.1:8000 --skip-system-sw-validation --dev-mode",
    "runtime_model_spec_json": "/home/ipastaltzidis/workspace/tt-inference-server/workflow_logs/runtime_model_specs/runtime_model_spec_2026-08-24_01-04-35_id_tt-transformers_Kimi-K3_super_cluster_kpG1FY-O.json",
    "model_id": "id_tt-transformers_Kimi-K3_super_cluster",
    "model_repo": "moonshotai/Kimi-K3",
    "inference_engine": "media",
    "tt_metal_commit": null,
    "vllm_commit": null,
    "model_impl": "tt-transformers"
}
```

### Acceptance Criteria

- Acceptance status: ✅ `PASS`
- Model status: `EXPERIMENTAL`
- Benchmarks: 🟨 `NA` (no blocks present)
- Evals: ✅ `PASS` (1/2 passed, 1 NA)
[...]

PR #5026, opened 2026-08-26
Add moonshotai/Kimi-K3 model spec and eval configs

ipastalTT:
## Summary

Onboards `moonshotai/Kimi-K3` as a remote-served eval target: a dev-catalog
model spec plus four eval tasks (GPQA Diamond, Terminal-Bench 2.1, τ³-Bench
Banking, and a LongBench v2 long-context sweep), each with a measured
reference score from a bring-up run.

## Changes

**`workflows/model_specs/dev/llm.yaml`** — remote-served spec following the
existing K2.6 / K2.7-Code / MiniMax-M3 pattern: `SUPER_CLUSTER` placeholder
device, `MEDIA` engine, `EXPERIMENTAL` status, 1,048,576-token context.

**`reference_config/evals/eval_config.py`** — new `EvalConfig` block, placed
after Kimi-K2.7-Code to keep the Kimi models grouped.

## Measured results

| Task | Published | Measured | Ratio | Tolerance |
| --- | --- | --- | --- | --- |
| `r1_gpqa_diamond` | 93.5 | 91.92 | 0.98 | 0.05 |
| `terminal_bench_2_1` | 85.0 | 85.39 | 1.00 | 0.05 |
| `tau3_bench_banking` | 46.0 | 27.84 | 0.61 | 0.10 |
| `longbench2_generate` | — | 71.88 | — | 0.05 |

Published scores are from Artificial Analysis; every `gpu_reference_score` is
linked to its measuring run in #4971.

GPQA Diamond and Terminal-Bench both reproduce their published numbers, which
is good evidence those two harnesses are wired up correctly. LongBench v2 has
no published baseline to compare against, so 71.88 stands as the first
reference for the 256K–900K ISL band.

**τ³-Bench Banking reproduces only 61% of the published 46.0, and reviewers
should weigh in on whether that should block.** 

Be aware of what landing this reference does to the gate: the accuracy check
prefers `gpu_reference_score` over `published_score` when both are present, so
τ³ will be judged against 27.84.

## Notes

**Terminal-Bench reference score.** Moonshot publishes 88.3 using their own
Kimi Code harness, but this config runs Terminus 2, so the 85.0 from
Artificial Analysis is the apples-to-apples comparison. Called out in a
comment so the lower value doesn't read as a typo.

**Sampling follows Moonshot's stated methodology**, which splits by task type:
temperature 1.0 with top_p 0.95 for single-step tasks (GPQA, LongBench), and
top_p 1.0 for agentic tasks (Terminal-Bench, τ³) — see the τ³ caveat above.

### DIFF
--- a/workflows/model_specs/dev/llm.yaml
+++ b/workflows/model_specs/dev/llm.yaml
@@ -1560,6 +1560,32 @@ templates:
       # spec-decode, agentic-traces) must trust remote code to load it.
       tokenizer_trust_remote_code: true
 
+# Remote-served model: evaluated against an external OpenAI-compatible
+# endpoint via `--server-url`. The device entry is a placeholder so the model
+# resolves through MODEL_SPECS / EVAL_CONFIGS; no local TT serving is implied.
+- weights:
+    - moonshotai/Kimi-K3
+  impl: tt_transformers
+  inference_engine: MEDIA
+  device_model_specs:
+    - device: SUPER_CLUSTER
+      max_concurrency: 64
+      max_context: 1048576  # 1024 * 1024;
+      default_impl: true
+  status: EXPERIMENTAL
+  env_vars:
+    VLLM_ALLOW_LONG_MAX_MODEL_LEN: 1
+  metadata:
+    moonshotai/Kimi-K3:
+      remote_api_model: moonshotai/Kimi-K3
+      # K3 ships its own parsers, distinct from the kimi_k2 pair used by
+      # K2.6/K2.7-Code: https://recipes.vllm.ai/moonshotai/Kimi-K3
+      reasoning_parser_name: kimi_k3
+      tool_call_parser_name: kimi_k3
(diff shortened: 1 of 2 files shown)


workflows/model_specs/dev/llm.yaml
```diff
@@ -1560,6 +1560,32 @@ templates:
...
+# Remote-served model: evaluated against an external OpenAI-compatible
+# endpoint via `--server-url`. The device entry is a placeholder so the model
+# resolves through MODEL_SPECS / EVAL_CONFIGS; no local TT serving is implied.
+- weights:
+    - moonshotai/Kimi-K3
+  impl: tt_transformers
+  inference_engine: MEDIA
+  device_model_specs:
+    - device: SUPER_CLUSTER
+      max_concurrency: 64
```
bgoelTT: Please check with Blaze team if it should be 95

bgoelTT (approved):

PR #4722, opened 2026-07-21
Use full HF repo id as the model id

ddjukicTT:
Switches the canonical model identifier throughout the stack from the bare basename (e.g` Llama-3.1-8B-Instruct`) to the full Hugging Face repo id (e.g. `meta-llama/Llama-3.1-8B-Instruct`).

##  CI runs: 
https://github.com/tenstorrent/tt-shield/actions/runs/30755737700/job/91522138283
https://github.com/tenstorrent/tt-shield/actions/runs/30818755530
https://github.com/tenstorrent/tt-shield/actions/runs/30818795493

PR #5027, opened 2026-08-26
Harden release resolution against ambiguous catalog entries

dcvijeticTT:
## Summary

Makes the release resolver strict: every release selector must resolve to exactly one
runtime leaf `(hf_model_repo, device, inference_engine, impl_id)`, or the run fails.
Zero matches, multiple matches, and two selectors resolving to the same leaf are all
hard errors now.

Stacked with #5021. Neither PR can merge alone.

## Problem

The old resolver matched catalog blocks by `(basename, engine, device-membership)` and
took the first hit. If a basename existed under two HF orgs (e.g. the same short model
name under `meta-llama/` and `deepseek-ai/`), the match was ambiguous and which one won
depended on YAML order — silently. Promotion and release notes had the same issue:
they matched coarse catalog blocks, not exact leaves.

## What changes

- **Strict resolution**: `resolve_model_spec()` accepts a full HF repo id or a
  basename; if a basename maps to more than one repo, it raises instead of guessing
- **Leaf uniqueness**: `validate_model_specs()` rejects duplicate leaf identities
  and a `model_id` that maps to multiple distinct identities
- **Per-weight perf targets**: performance reference targets are now looked up per
  individual weight instead of once per family head. This exposed some models that
  had no real target — `model_performance_reference.json` needs a follow-up tidy
- **Docs page**: now documents the default impl, not whichever catalog entry came first
- **`model_display_name`** on `ModelSpecTemplate` and prod catalog: records the family
  a leaf came from, used only for docs grouping
- **Unified resolver**: `get_runtime_model_spec()` now delegates to
  `resolve_model_spec()`, so `run.py --model` uses the same resolution path as the
  release flow. No breaking changes — everything works as before, but through one
  resolver instead of two

Must merge before #4722.

### DIFF
--- a/scripts/release/generate_model_support_docs.py
+++ b/scripts/release/generate_model_support_docs.py
@@ -411,7 +411,9 @@ def group_templates_by_model(
     groups = defaultdict(list)
     for template in templates:
         display_name = get_model_display_name(template)
-        key = (template.model_type, display_name) if include_model_type else display_name
+        key = (
+            (template.model_type, display_name) if include_model_type else display_name
+        )
         groups[key].append(template)
     return dict(groups)
 
--- a/scripts/release/generate_model_support_docs.py
+++ b/scripts/release/generate_model_support_docs.py
@@ -232,7 +232,7 @@ def get_model_display_name(template: ModelSpecTemplate) -> str:
 def coalesce_catalog_groups_for_docs(
     templates: List[ModelSpecTemplate],
 ) -> List[ModelSpecTemplate]:
-    """Reconstruct complete generated template groups for stable docs output."""
+    """Combine only leaf variants sharing one device and runtime configuration."""
 
     def shared_signature(template):
         data = asdict(template)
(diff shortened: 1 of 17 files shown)


vcankovicTT (approved):

scripts/release/generate_model_support_docs.py
```diff
@@ -224,7 +224,93 @@ def generate_section_anchor(section_title: str) -> str:
 
 def get_model_display_name(template: ModelSpecTemplate) -> str:
     """Get the display name for a model template."""
-    return model_weights_to_model_name(template.weights[0])
+    return template.model_display_name or model_weights_to_model_name(
+        template.weights[0]
+    )
```
fivanovicTT, line 229: Can you double check that this is not broken by running some CI runs?

fivanovicTT: Or at least local runs

scripts/release/model_spec_resolver.py
```diff
@@ -0,0 +1,251 @@
...
+            raise ValueError(
+                f"Model implementations must be a non-empty list, got "
+                f"{implementations!r}"
+            )
+        for implementation in implementations:
+            if not isinstance(implementation, dict):
+                raise ValueError(
+                    f"Model implementation must be an object, got {implementation!r}"
+                )
+            yield implementation
```
fivanovicTT: There is no need for `else` branch here, code inside of else can be extracted outside of it

workflows/model_specs/dev/audio_tts.yaml
fivanovicTT, line 1: I think this is not correct: `whisper-large-v3` != `distil-large-v3`. The latter is distilled from the former

workflows/model_specs/dev/image.yaml
fivanovicTT, line 1: Same as comment above

workflows/model_specs/dev/llm.yaml
fivanovicTT, line 1: Same as comment above

workflows/model_specs/dev/vlm.yaml
fivanovicTT, line 1: Same is comment above

fivanovicTT (commented):
Generally, looking good. 

Concern I have is the `model_display_name` in `model_spects/dev/... .yml` files - `whisper-large-v3` and `distil-large-v3` should not and must not have the same display name.

scripts/release/generate_model_support_docs.py
dcvijeticTT: https://github.com/tenstorrent/tt-inference-server/actions/runs/32973379343
https://github.com/tenstorrent/tt-inference-server/commit/529fb952b5c173022ddf6877e86378e166cff35b
Here's a valid run that proves everything still works

workflows/model_specs/dev/audio_tts.yaml
dcvijeticTT: `model_display_name` is just a group key that is used in docs grouping. **They were grouped together before this change**, this just keeps that grouping in flattened prod catalog using this `model_display_name` group key. If you think they need to be separated in docs, we can fix that in another pass when we'll also clean prod catalog from old invalid models.  
EDIT: used just for grouping docs

dcvijeticTT: Same with other comments.

dcvijeticTT: > `model_display_name` is just a group key that is used in docs grouping. **They were grouped together before this change**, this just keeps that grouping in flattened prod catalog using this `model_display_name` group key. If you think they need to be separated in docs, we can fix that in another pass when we'll also clean prod catalog from old invalid models. EDIT: used just for grouping docs

For example this grouping: 
image
So we need to keep the group key for this.

scripts/release/model_spec_resolver.py
dcvijeticTT: Agreed!

scripts/release/generate_model_support_docs.py
fivanovicTT: I meant `on-dispatch` job in CI or local run with model running

fivanovicTT (approved):

scripts/release/generate_model_support_docs.py
dcvijeticTT: No `on-dispatch` job goes through this code path. Only release workflows call this.

PR #5032, opened 2026-08-27
Make HF_TOKEN optional for public (non-gated) models

jashansinghTT:
run.py hard-requires HF_TOKEN for every --docker-server deploy, even for public models (Qwen, Mistral, ...) that download anonymously. When no token is set it falls back to an interactive getpass() prompt, which hangs forever — or dies with EOFError — when run.py is spawned headless by an orchestrator like TT-Studio. This is the run.py side of the failure tt-studio#1290 works around downstream.

- [x] handle_secrets: HF_TOKEN removed from required_env_vars entirely — never prompted for or asserted on; a missing token logs a warning and setup_host.check_hf_access is the single authority on HF access (matches TT-Studio's discrimination, so its workaround can be deleted)
- [x] Shared workflows.utils.can_prompt_interactively() guards every getpass site (handle_secrets, get_hf_env_vars, JWT_SECRET prompt), keyed on /dev/tty openability, not isatty() — a start_new_session worker inherits the terminal on fd 0 but has no controlling tty; headless callers with a missing required secret get an actionable RuntimeError instead of a hang
- [x] HostSetupManager.check_hf_access: with no token, validate via an anonymous models-API lookup — public repos pass, gated repos fail fast with an actionable message (the models API serves gated repos' metadata anonymously, so the `gated` field is inspected rather than the HTTP status). Hardened: 30s timeout, clean error when HF is unreachable, status checked before JSON parse, success message names the repo and notes only the weights repo was checked
- [x] setup_weights_huggingface: allow an empty token — the `hf` CLI treats an empty HF_TOKEN as anonymous access; a token prompted in get_hf_env_vars is exported to os.environ for later stages
- [x] Tests: anonymous check_hf_access branch (public/gated/non-200/empty-siblings/network-error), headless get_hf_env_vars never prompts, HF_TOKEN never prompted in handle_secrets, headless missing-JWT fails cleanly; test_run_arguments.py + test_workflows.py passing (117)
- [x] Verified live: anonymous check_hf_access returns True for Qwen/Qwen3-8B and False (gated error) for meta-llama/Llama-3.1-8B-Instruct; pty repro of TT-Studio's worker spawn confirms can_prompt_interactively() is False there while isatty() is True

### DIFF
--- a/run.py
+++ b/run.py
@@ -869,15 +869,32 @@ def handle_secrets(runtime_config):
             "Set API_KEY in .env or as an environment variable."
         )
 
+    # HF_TOKEN is not strictly required even where it is expected: public
+    # (non-gated) models download anonymously, and setup_host.check_hf_access
+    # fails fast with an actionable error for gated repos. Never prompt for it
+    # on a non-interactive stdin (e.g. spawned by TT-Studio) — getpass() there
+    # hangs forever or dies with EOFError.
+    def _hf_token_optional(key: str) -> bool:
+        return key == "HF_TOKEN" and not sys.stdin.isatty()
+
+    _anon_warning = (
+        "HF_TOKEN is not set — proceeding without a token. Public models "
+        "download anonymously; gated models require HF_TOKEN in .env or the "
+        "environment."
+    )
+
     # load secrets from env file or prompt user to enter them once
     if not load_dotenv():
         env_vars = {}
         for key in required_env_vars:
             _val = os.getenv(key)
+            if not _val and _hf_token_optional(key):
+                logger.warning(_anon_warning)
+                continue
             if not _val:
                 _val = getpass.getpass(f"Enter your {key}: ").strip()
             env_vars[key] = _val
-        assert all([env_vars[k] for k in required_env_vars])
+        assert all([env_vars[k] for k in env_vars])
         write_dotenv(env_vars)
         # read back secrets to current process env vars
         check = load_dotenv()
@@ -885,6 +902,9 @@ def handle_secrets(runtime_config):
     else:
         logger.info("Using secrets from .env file.")
         for key in required_env_vars:
+            if not os.getenv(key) and _hf_token_optional(key):
+                logger.warning(_anon_warning)
+                continue
             assert os.getenv(key), (
                 f"Required environment variable {key} is not set in .env file."
             )
--- a/run.py
+++ b/run.py
@@ -44,6 +44,7 @@ from workflows.run_docker_server import (
 from workflows.runtime_config import RuntimeConfig
 from workflows.setup_host import setup_host
 from workflows.utils import (
+    can_prompt_interactively,
     ensure_readwriteable_dir,
     get_default_hf_home_path,
     get_default_workflow_root_log_dir,
@@ -845,17 +846,21 @@ def handle_secrets(runtime_config):
         WorkflowType.SERVING_BENCH,
         WorkflowType.PREFILL_DECODE,
     }
-    # --docker-server requires the HF_TOKEN env var to be available
-    huggingface_required = (
+    # These workflows expect an HF_TOKEN, but it is deliberately NOT in
+    # required_env_vars: public (non-gated) models download anonymously, and
+    # setup_host.check_hf_access is the single authority on HF access — it
+    # fails fast with an actionable error when a gated repo is requested with
+    # no token. Requiring it here would either prompt (hanging headless
+    # callers such as TT-Studio's uvicorn worker, whose stdin isatty() but has
+    # no controlling terminal to answer on) or hard-fail public-model deploys.
+    hf_token_expected = (
         workflow_type not in client_side_workflows or runtime_config.docker_server
     )
-    huggingface_required = huggingface_required and not runtime_config.interactive
+    hf_token_expected = hf_token_expected and not runtime_config.interactive
 
     required_env_vars = []
     if jwt_secret_required:
         required_env_vars.append("JWT_SECRET")
-    if huggingface_required:
-        required_env_vars += ["HF_TOKEN"]
 
     if (
         workflow_type == WorkflowType.SERVER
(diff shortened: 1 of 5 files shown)


stisiTT (commented):
The anonymous-access half looks right, and I checked the premise: the HF models API
returns `"gated": "manual"` for `meta-llama/Llama-3.1-8B-Instruct` and
`"gated": false` for `Qwen/Qwen3-32B`, so inspecting the `gated` field instead of
the HTTP status is the correct call. I also confirmed against the pinned
`huggingface_hub` that an empty `HF_TOKEN` really is anonymous (`_clean_token()`
returns `None` for `""`), so that comment is accurate.

One blocker though: `sys.stdin.isatty()` isn't the right signal for the TT-Studio
case, so as written this doesn't fix it.

TT-Studio's default deploy calls `run_main()` **in-process** inside its uvicorn
worker (`inference-api/api.py`, `dev_mode` defaults to `False`). That worker is
started with `Popen(cmd, start_new_session=True, stdout=DEVNULL, stderr=STDOUT)`
and no `stdin=`, so fd 0 is still the terminal the user ran `python run.py` from.
`start_new_session` moves the session but doesn't change what fd 0 points at.

I ran the real chain to be sure (real uvicorn binary, same wrapper script shape,
same Popen flags, under a pty):

```
=== parent (run.py) === isatty=True sid=343731
=== uvicorn worker ===
sys.stdin.isatty(): True
os.ttyname(0)     : /dev/pts/3
sid=343732
open /dev/tty     : FAILED OSError(6, 'No such device or address')
```

So `_hf_token_optional()` returns `False` and both new guards no-op. A fresh
install still reaches the `getpass` at `run.py:894` and hangs: the worker has no
controlling terminal, so getpass falls back to fd 0 and blocks with no traceback,
and the deploy just sits at 0%. An existing `.env` without `HF_TOKEN` still fails
the assert at `run.py:908`. Only the `dev_mode=True` path is fixed, and that one
already passed `stdin=DEVNULL`.

Suggestion: drop tty-ness as the discriminator. Take `HF_TOKEN` out of
`required_env_vars` and let `check_hf_access` be the only authority, since its
gated error is actionable now. If you want to keep a prompt, gate it on something
explicit like `not os.getenv("AUTOMATIC_HOST_SETUP")` plus a `--non-interactive`
flag. TT-Studio's current workaround keys purely on whether `HF_TOKEN` is set, so
matching that keeps the two repos aligned and lets us delete the patch.

Two more worth doing in the same pass:

1. `setup_host.py:390` still calls `getpass` whenever `not self.automatic`, with no
   guard, and this patch makes it newly reachable headlessly (previously you
   aborted earlier in `handle_secrets`). CI, `nohup` and systemd callers don't set
   `AUTOMATIC_HOST_SETUP`, so they'd get a hang or EOFError where they used to get
   a clean assert. Probably worth one shared helper for both prompt sites.
2. For `--workflow server --docker-server`, `HF_TOKEN` is still in
   `required_env_vars`, so `run.py` prompts first with the old wording, and an
   empty answer dies on the bare `assert all([...])` at `run.py:897` with no
   message. The new "leave empty for public models" prompt at `setup_host.py:392`
   is unreachable on that path.

Smaller things:

- The new HF request has no timeout and no try/except. `http_request` re-raises on
  `URLError`, so an offline or proxied host gets a bare traceback instead of the
  old clean message.
- `json.loads` runs before the `status != 200` check, so an HTML error body reports
  "Invalid JSON response" and loses the status that would explain it.
- Only the weights repo is checked. `whisper-large-v3` has public weights but its
  diarization path pulls gated pyannote checkpoints
  (`tt-media-server/utils/diarize.py`), so that deploy passes the gate and 401s
  later inside the container. Maybe just soften the success message to name what
  was actually checked.
- No tests on the `check_hf_access` anonymous branch, which is the part that
  decides gated vs public. `tests/test_workflows.py` already mocks
  `workflows.setup_host.http_request`, so gated / empty-siblings / non-200 cases
  are a few lines each. Also `test_hf_token_optional_with_dotenv_loaded` currently
  has no assertions, it only proves nothing raised.

jashansinghTT:
Thanks for the deep review — the pty repro was right, and 6a4b70fc reworks the fix along the lines you suggested:

**Blocker (tty-ness):** dropped entirely. `HF_TOKEN` is no longer in `required_env_vars` at all — `handle_secrets` never prompts for it or asserts on it, it just warns when it's absent, and `check_hf_access` is the single authority (public passes anonymously, gated fails with the actionable message). That matches TT-Studio's discrimination (purely on whether `HF_TOKEN` is set), so tt-studio#1290's shim can be deleted once this lands.

**Shared prompt guard:** added `workflows.utils.can_prompt_interactively()` and applied it to every `getpass` site (`handle_secrets`, `get_hf_env_vars`, and the `JWT_SECRET` prompt at `setup_host.py:533` which had the same exposure). It keys on `/dev/tty` openability rather than `isatty()` — exactly the signal your repro showed diverging (`isatty=True`, `open /dev/tty` fails in the `start_new_session` worker). I reproduced your chain under a pty and confirmed: parent `True`, worker `isatty=True` but helper `False`. Headless callers with a genuinely missing required secret (`JWT_SECRET`) now get a clean `RuntimeError` naming the variable instead of a hang/EOFError, and the bare `assert all([...])` now names what's missing.

**`--workflow server --docker-server` path:** with `HF_TOKEN` out of the required list, run.py no longer prompts first — the `setup_host` prompt (with the "leave empty for public models" wording) is the one prompt site, reachable only on a real controlling terminal. A prompted token is now also exported to `os.environ` so later stages (weights venv, docker env construction) see it.

**Smaller items:** `http_request` has a 30s timeout; the anonymous lookup is wrapped so an offline/proxied host gets a clean "could not reach Hugging Face" error instead of a traceback; the status check now precedes `json.loads` so an HTML error body reports `HTTP <status>`; and the success message names the repo and says only the weights repo was checked (re: whisper's pyannote case).

**Tests:** the anonymous `check_hf_access` branch is covered (public / gated / non-200-HTML / empty siblings / network error, asserting the logged messages), `get_hf_env_vars` headless never prompts, `handle_secrets` never prompts for `HF_TOKEN` even interactively, headless missing-`JWT_SECRET` raises cleanly, and `test_hf_token_optional_with_dotenv_loaded` now asserts the anonymous-access warning. Full `test_run_arguments.py` + `test_workflows.py` passing (117), ruff 0.15.0 check/format clean. Re-verified live: anonymous check returns True for Qwen/Qwen3-8B and the gated error for meta-llama/Llama-3.1-8B-Instruct.

PR #5034, opened 2026-08-27
Embedding: run BGE-large and Qwen3-8B on cpp_server in CI, with resilient multi-worker warmup (BGE-m3 tested only on local setup)

### LINKED ISSUE #5006: Get tt-media-server embedding baseline results
Run existing CI tests for:

- Qwen3-Embedding-8B
- bge-large-en-v1.5
- bge-m3

Test all supported device combinations through tt-media-server. Create a report with correctness, performance, configuration and CI links. This report should be updated during the complete migration.

### LINKED ISSUE #5008: Add cpp_server embedding models to CI
Update `run_docker_server.py` and CI configuration so the three embedding models run through cpp_server.

Test all supported model/device combinations and compare the results with the tt-media-server baseline. Add the results to the migration report.

jzivanovicTT:
## What this PR does
Switches the embedding models (BAAI/bge-large-en-v1.5, BAAI/bge-m3,
Qwen/Qwen3-Embedding-8B) from the Python uvicorn server to the C++
`cpp_server` backend for CI benchmark/eval runs, and hardens the C++
embedding service's multi-worker warmup so a 32-worker Galaxy bring-up
is observable, parallel, and self-healing.
### CI plumbing (`workflows/run_docker_server.py`)
- Embedding models opt into the C++ backend via `SERVER_MODE=cpp`. The C++
  binary does not derive per-model settings from `MODEL`+`DEVICE`, so the
  runner/device tables (runner type, mesh shape, `DEVICE_IDS`, galaxy flag)
  mirror `tt-media-server/config/constants.py::ModelConfigs` until the binary
  learns to derive them itself.
### C++ embedding service warmup (`embedding_service.cpp`)
- **Staggered warmup:** worker 0 warms up alone first and populates the shared
  tensor cache; the remaining workers then warm up in parallel. Fixes worker
  crashes ("file too small" / SIGBUS) caused by 32 workers racing to generate
  the cache on a cold volume.
- **Parallel phase-2 readiness:** worker READY handshakes are awaited with a
  single poll() across all response pipes instead of sequentially, so one
  stuck worker can no longer mask the state of the others.
- **Configurable warmup timeout:** `EMBEDDING_WARMUP_TIMEOUT_MS` (default now
  10 min instead of a hardcoded 1 h), so failures surface while logs are
  still actionable.
- **Warmup retry rounds:** workers that fail warmup are respawned up to
  `EMBEDDING_WARMUP_MAX_RETRIES` (default 3) rounds — parity with the Python
  server's health monitor. Motivated by BGE-large PCC validation, which is an
  independent draw per attempt on the old pin (see Known issue).
### Benchmark configuration
- BGE-large on Galaxy: `BENCHMARK_MAX_CONCURRENCY=256` (dev + prod specs,
  helm values kept in sync) so the benchmark can actually load 32 workers
  × batch 8.
### tt-metal pin: `65718bb` → `71c49f23`
- The old pin (Feb 10) predates `create_venv.sh --bundle-python` (added
  Feb 14) and required a diverging Dockerfile to build. The Dockerfile is
  now restored to the mainline version, and the pin moves to `71c49f23`
  (main, Aug 4), which builds with the unmodified Dockerfile.
- The old pin also predates the BGE warmup input seeding fix
  (tt-metal#37418, merged ~1 h after our pinned commit), which is the root
  cause of the nondeterministic per-worker PCC warmup failures we chased on
  Galaxy: without it, every worker validates against freshly random inputs,
  and the measured PCC straddles the 0.90 threshold run-to-run.
## Known issue — BGE-large warmup PCC on the new pin
With seeded warmup inputs, BGE-large PCC is fully deterministic — but on
`71c49f23` it lands at **0.894**, just below the 0.90 threshold hardcoded in
tt-metal (`performant_runner_infra.py::validate`). Verified locally on N150:
identical 0.894 across all retry rounds, while the old pin passes at ~0.935 on
the same chip. Something between Feb and Aug (candidates: transformers 5.10.2
bump changing the torch reference, or ttnn op changes) shifted the value.
This is a tt-metal issue (threshold and validation both live there) and is
being raised with the model owner. Nothing in this repo needs to change when
it is resolved on the metal side. Until then, BGE-large warmup fails
deterministically on this pin; BGE-M3 and Qwen3-Embedding-8B are unaffected.
## Validation
- Unit tests: full ctest suite green.
- Qwen3-Embedding-8B on Galaxy CI: 32/32 workers ready, benchmarks pass
  (~114k tok/s for concurrency 32) .
- BGE-large on Galaxy CI (old pin + retry rounds): retries recover workers
  that lose the warmup PCC draw  (~310k tok/s for 32 concurrency, and 290k tok/s for 256 concurrency, this will be adressed and improved in next PR, already at 600k on that branch).
- Warmup phases, retries, and timeout behavior exercised locally on N150
  (mock runner + real BGE-large/BGE-M3/Qwen3 golden validation, sequential
  phase bit-exact against captured goldens).

Qwen-Embedding-8B

Galaxy, python, concurrency = 1 - https://github.com/tenstorrent/tt-shield/actions/runs/32864290132/job/97870608804

Galaxy, cpp, concurrency = 1 - https://github.com/tenstorrent/tt-shield/actions/runs/33063039945/job/98495771930

 Galaxy, python, concurrency = 32 -  https://github.com/tenstorrent/tt-shield/actions/runs/32976046805/job/98215729255
Galaxy, cpp, concurrency = 32 - https://github.com/tenstorrent/tt-shield/actions/runs/33071045375/job/98524174813 

BGE-LARGE

Galaxy, python, concurrency = 8 - https://github.com/tenstorrent/tt-shield/actions/runs/33502358541/job/99850049289

N150, cpp, concurrency = 8 - https://github.com/tenstorrent/tt-shield/actions/runs/33380928217

Galaxy, python, concurrency = 32 - https://github.com/tenstorrent/tt-shield/actions/runs/33613019655/job/100199430185

Galaxy, cpp, concurrency = 32 - https://github.com/tenstorrent/tt-shield/actions/runs/33624238790/job/100228205042

Galaxy, python, concurrency = 256 - https://github.com/tenstorrent/tt-shield/actions/runs/33634898917

Galaxy, python, concurrency = 256 (without batching fix) - https://github.com/tenstorrent/tt-shield/actions/runs/33634786157/job/100262906964

Sneak peek into next PR:
(Galaxy, cpp, bge, concurrency = 256) - https://github.com/tenstorrent/tt-shield/actions/runs/33645522634/job/100459576676

Closes #5008
Closes #5006
Part of #2085


tt-media-server/cpp_server/include/config/defaults.hpp
```diff
@@ -109,6 +109,26 @@ constexpr size_t CLIENT_MAX_BODY_BYTES = 100 * 1024 * 1024;  // 100 MB
...
+/**
+ * Extra warmup rounds for workers whose first warmup failed (overridable via
+ * the EMBEDDING_WARMUP_MAX_RETRIES env var). Some model warmups validate the
+ * device output against a CPU reference with a PCC threshold, and the result
+ * is not deterministic per chip (BGE-large on Galaxy spans ~0.86-0.96 against
+ * a 0.90 threshold), so a failed warmup is worth re-rolling rather than
+ * permanently losing the device. The Python server gets the same effect from
+ * its health monitor, which restarts dead workers up to
+ * max_worker_restart_count (5) times.
+ */
```
ztorlakTT: Can we shorten these comments and keep only what the constants mean? Details about PCC ranges, cache timings and Python server behavior are too much for this header

tt-media-server/cpp_server/src/services/embedding_service.cpp
```diff
@@ -115,28 +122,79 @@ struct WorkerProcess {
...
         readFd.get());
     return true;
   }
 
+  /**
+   * Block until the child reports warmup completion via the READY sentinel.
+   * Returns false on child exit (pipe EOF), timeout, or unexpected data.
+   * keepWaiting lets service shutdown abort the wait within ~100ms.
+   */
+  bool waitUntilReady(unsigned timeoutMs,
```
ztorlakTT: Why do we need a separate waitUntilReady() here? awaitWorkersReady() already handles the READY handshake and timeout

workflows/run_docker_server.py
```diff
@@ -144,6 +200,8 @@ def _media_server_dev_mounts(repo_root_path, user_home_path, model_spec) -> List
 def get_media_server_docker_env_vars(model_spec):
     """Get media server environment variables for Docker container."""
     if _is_cpp_media_spec(model_spec):
+        if model_spec.model_name in _CPP_EMBEDDING_RUNNER_BY_MODEL_NAME:
+            return _get_cpp_embedding_server_docker_env_vars(model_spec)
```
ztorlakTT: Can we move this check into a common helper? For example:
```
def _get_cpp_server_docker_env_vars(model_spec):
    if model_spec.model_name in _CPP_EMBEDDING_RUNNER_BY_MODEL_NAME:
        return _get_cpp_embedding_server_docker_env_vars(model_spec)
    return _get_cpp_image_server_docker_env_vars(model_spec)
```
Then this function only needs:
```
if _is_cpp_media_spec(model_spec):
    return _get_cpp_server_docker_env_vars(model_spec)
```

workflows/run_docker_server.py
```diff
@@ -70,24 +72,45 @@ def short_uuid():
...
+}
+
+# device.name.lower() -> DEVICE_IDS (worker layout, one worker per group).
+# Mirrors constants.py::DeviceIds usage in the embedding ModelConfigs entries;
+# identical for all three embedding models.
+_CPP_EMBEDDING_DEVICE_IDS = {
+    "n150": "(0)",
+    "n300": "(0)",
+    "t3k": "(0),(1),(2),(3)",
+    "galaxy": _DEVICE_IDS_32,
```
ztorlakTT, line 100: Do we need a separate _CPP_EMBEDDING_DEVICE_IDS mapping? These values duplicate the device IDs already used by _CPP_SDXL_DEVICE_DEFAULTS. Can we have one common device ID mapping and only keep the supported embedding devices separately?

workflows/model_specs/prod/embedding.yaml
```diff
@@ -7,7 +7,7 @@ templates:
 # =============================================================================
 - weights:
     - BAAI/bge-large-en-v1.5
-  tt_metal_commit: "65718bb"
+  tt_metal_commit: "71c49f23"
```
ztorlakTT: I don't think we should change tt_metal_commit in this PR. PCC does not consistently meet the threshold with this commit

tt-media-server/cpp_server/src/services/embedding_service.cpp
```diff
@@ -316,31 +393,215 @@ struct EmbeddingService::Impl {
...
+    awaitWorkersReady(std::move(spawned), warmupTimeoutMs);
+
+    // Retry rounds: a failed warmup is usually worth re-rolling, not a broken
+    // chip. BGE-large's warmup validates device output against a CPU
+    // reference (assert PCC >= 0.90) and the measured PCC varies run to run
+    // (observed 0.86-0.96 across one Galaxy), so each attempt is an
+    // independent draw. This also re-covers a worker that lost the phase-1
+    // slot above. Parity with the Python server, whose health monitor
+    // restarts dead workers up to max_worker_restart_count times.
+    const unsigned maxRetries = tt::config::embeddingWarmupMaxRetries();
```
ztorlakTT: Should we retry workers that fail PCC validation? This can make CI pass by retrying until we get a passing result instead of catching a correctness issue. I think we should fix the PCC issue first rather than retry it here

ztorlakTT:
Can you add links to the related CI runs to the PR description for each model/device tested?

tt-media-server/cpp_server/src/services/embedding_service.cpp
jzivanovicTT: I had to make it in order to get results, this is a workaround their problem which would block any testing of bge-large, so I had to do it this way. I can exclude it from PR if you want me to but then it will not be able to run at all until they fix it.

workflows/model_specs/prod/embedding.yaml
jzivanovicTT: But then the dockerfile and the pinned commit are incompatible, this way the problem is only on tt_metals side, I taught that we agreed on this one, maybe there was some misunderstanding. What should i do then, it looks like a bad idea to leave a pinned commit that is incompatible with .dockerfile

tt-media-server/cpp_server/include/config/defaults.hpp
jzivanovicTT: Yes, I will do it

workflows/run_docker_server.py
jzivanovicTT: Okay, it happens to be the same, but what if we wanted to change one but keep another for example if some model required a different configuration?

workflows/model_specs/prod/embedding.yaml
jzivanovicTT: And PCC also does not consistantly meet the threshold with pinned commit either

workflows/run_docker_server.py
jzivanovicTT: Yes, I'm on it

tt-media-server/cpp_server/src/services/embedding_service.cpp
jzivanovicTT: I planned to change this in a next PR that will include both code cleanup and architecture change, are you OK with that or want me to change it for this commit. Basically we could get rid of waitUntilReady and reuse awaitWorkersReady

tt-media-server/cpp_server/include/config/defaults.hpp
jzivanovicTT: Done

workflows/run_docker_server.py
jzivanovicTT: Done

workflows/model_specs/prod/embedding.yaml
ztorlakTT: I see, I missed that the Dockerfile requires this commit. If PCC is no worse than before, I’m good with the change

jzivanovicTT: PCC is worse but I dont know what can we do about it, the only way to push this commit was with that because as I said the other part is onto them (tt_metal), I will write to model owner now to lower the threshold to 0.89 which is actually meaningful for this commit.

jzivanovicTT: Actually that would be new commit so I would need to pin that one

tt-media-server/cpp_server/src/services/embedding_service.cpp
ztorlakTT: I’d remove the PCC retry from this PR, but you can keep it on a separate branch to make testing easier

tt-media-server/cpp_server/src/services/embedding_service.cpp
ztorlakTT: I’d prefer to do it in this PR, since we shouldn’t introduce the duplicate logic in the first place

workflows/run_docker_server.py
ztorlakTT: Makes sense. I’m fine with keeping separate mappings for now

tt-media-server/cpp_server/src/services/embedding_service.cpp
jzivanovicTT: Done

tt-media-server/cpp_server/src/services/embedding_service.cpp
jzivanovicTT: Done

ztorlakTT (approved):

idjuricTT (approved):

workflows/model_specs/prod/embedding.yaml
acvejicTT, line 2: @jzivanovicTT this ->

charts/tt-inference-server/values.yaml
```diff
@@ -5788,6 +5788,8 @@ models:
             env:
               - name: ARCH_NAME
                 value: wormhole_b0
+              - name: BENCHMARK_MAX_CONCURRENCY
```
acvejicTT: @jzivanovicTT no manual additions or changes here. This is generated as an artifact of the release process.

PR #5041, opened 2026-08-28
Run the Blackhole Galaxy Qwen3-32B lane on 2D torus, with real gating

mgiermakowskiTT:
### What

Enables the Blackhole Galaxy Qwen3-32B lane end to end:

1. **Run it on a 2D torus.** `BLACKHOLE_GALAXY` gets `fabric_config: FABRIC_2D_TORUS_XY` (the Wormhole Galaxy entry stays on `FABRIC_1D_RING`), plus the `worker_l1_size` / `trace_region_size` the model actually needs. The torus is what makes the column-axis collectives work on this part.
2. **Give it real gating.** Adds a `blackhole_galaxy` block to `model_performance_reference.json`. Without one, `get_perf_reference()` fell through the legacy subdevice remap to `P150X8`, which has no row for this model — so every target resolved to `NA` and the benchmark checks reported **0/18 passed while looking green**.
3. **Declare it FUNCTIONAL.** `status` is per-template, so `BLACKHOLE_GALAXY` is split into its own template. Everything else mirrors the GALAXY entry; only the status differs, since Blackhole Galaxy has no derived performance ceiling yet and cannot be held to `COMPLETE`.
4. **Wire it into the spec-test configs** (`server_tests_config.json`, `test_suites/llm.json`).
5. **Fix spec-test suite lookup** (`test_filter.py`) — see below. This is not specific to this lane; spec tests currently match zero suites for **every** model on main.

### The spec-test fix

[#4722](https://github.com/tenstorrent/tt-inference-server/pull/4722) made the full HF repo id the canonical model id and changed `dispatch.py` to pass `model_spec.hf_model_repo` into `filter_by_model`. The suite configs still key on the bare weight name, and the filter did an exact membership test, so the lookup silently returns nothing:

| id passed by `dispatch.py` | suites matched | bare name | suites matched |
|---|---|---|---|
| `meta-llama/Llama-3.1-8B-Instruct` | **0** | `Llama-3.1-8B-Instruct` | 12 |
| `Qwen/Qwen3-32B` | **0** | `Qwen3-32B` | 12 |
| `openai/gpt-oss-20b` | **0** | `gpt-oss-20b` | 6 |

The symptom is misleading: an empty suite list is treated as a deliberate no-op (`rc=0`), and the workflow then dies one step later at `No blocks accumulated — cannot generate report`. It reads like a reporting fault, but it is a lookup miss — no test ever ran.

The filter now matches on either spelling. `filter_by_model` has exactly one caller (spec-test dispatch), so this cannot move benchmark or eval numbers.

### Validation

Both workflows were run on this branch, pinned at tt-metal `a0a2294` ([#54617](https://github.com/tenstorrent/tt-metal/pull/54617), merged) and vllm `bef89e4`.

**1. `release` — [run 34222534266 / job 102067036231](https://github.com/tenstorrent/tt-shield/actions/runs/34222534266/job/102067036231)**, runner `g11blx02`, 1h49m.

```
Acceptance status: ✅ PASS  (0 blockers)
Model status:      FUNCTIONAL
Benchmarks:        ✅ PASS (2/22 passed, 20 NA)
Evals:             ✅ PASS (3/3 passed)
Spec Tests:        🟨 NA  (no blocks present)
```

Graded benchmark points:

| point | TTFT mean | TTFT P50 | TPOT | Tput/user | Tput out | Tput total | RPS | check |
|---|---|---|---|---|---|---|---|---|
| ISL/OSL 128, conc=1 | 658.7 ms | 83.5 ms | 30.0 ms | 33.3 | 28.6 | 57.3 | 0.224 | ✅ PASS |
| ISL/OSL 128, conc=32 | 1237.3 ms | 1277.0 ms | 30.9 ms | 32.4 | 792.8 | 1585.6 | 6.194 | ✅ PASS |

Both clear the strictest `target` tier — conc=1 at 0.7319 / 1.111 / 1.101, conc=32 at 0.8249 / 1.078 / 1.101. The other 20 points are sweep points with no target row defined; they produce real numbers and are reported for information only.

**Evals 3/3** — `r1_aime24` 66.67, `r1_math500` 93.0, `r1_gpqa_diamond` 72.5, all within tolerance of their references.

`Spec Tests: NA` here is expected, not a gap: `RELEASE` is in `_ENGINE_EVAL_WORKFLOWS` and `_ENGINE_BENCHMARK_WORKFLOWS`, but spec tests only run under `WorkflowType.SPEC_TESTS`, so a release run emits no spec-test blocks. They were run separately:

**2. `spec_tests` — [run 34282010402 / job 102265440002](https://github.com/tenstorrent/tt-shield/actions/runs/34282010402/job/102265440002)**, runner `g08blx01`. **25/25 passed.**

| suite | result |
|---|---|
| Vllm Chat Completions | 22/22 — `test_penalties` 9, `test_determinism_parameters` 3, `test_max_tokens` 2, `test_n` 2, `test_stop` 2, `test_coherence_verbatim_echo` 1, `test_logprobs` 1, `test_non_uniform_seeding` 1, `test_seed_reproducibility` 1 |
| Vllm Qwen3 Streaming | 3/3 |

The same runner had failed the run immediately before this one, at the commit without change 5 — 0 tests collected, `rc=1 error=no_blocks`. Same box, same pins, only the filter fix differs.

### On the targets

The `measured` floors are deliberately **loose values that pass on CI hardware**, not the achievable numbers. The dev box `g03blx02` measures 99–125 ms TTFT and 32.3–32.5 tok/s at conc=1; CI reports 658.7 ms mean TTFT on the same build.

That TTFT gap is an artifact of `num_prompts=8`, not of the hardware: P50 was **83.5 ms** — better than the dev box — while P99 was 4361.6 ms, so a single cold first request drags the mean by ~575 ms. Until this point grades a percentile or uses enough prompts to dilute the cold request, the target has to absorb that outlier.

The `theoretical` block is copied from the Wormhole Galaxy row, because the acceptance check synthesises the `functional` and `complete` tiers only from a `theoretical` block — without one, no functional tier is emitted and the model cannot be declared FUNCTIONAL.

**`ttft_ms` is deliberately omitted from it.** Grading the mean against a derived ceiling only re-measures the cold-start outlier above (658.7 ms against a 300 ms 10% bar, while the median is 83.5 ms), and a target the hardware beats by 3.9× at the median is not a useful gate. So that check resolves to `NA`, the functional tier grades throughput instead — where the 10% bar of 15.6 is cleared at ratios 2.136 and 1.836 — and TTFT at conc=1 is still gated by the `target` tier's 900 ms floor. Both should be replaced once a real Blackhole ceiling exists.

conc=32 carries no `theoretical`, mirroring the Wormhole conc=32 row: a ceiling derived at conc=1 does not describe that point.

### Depends on

tt-metal [#54617](https://github.com/tenstorrent/tt-metal/pull/54617) — **merged** (`a0a2294`).

### DIFF
--- a/llm_module/test_vllm_qwen3_streaming.py
+++ b/llm_module/test_vllm_qwen3_streaming.py
@@ -9,6 +9,7 @@ wired only for ``qwen3_32b`` via ``VLLMQwen3StreamingParamConformanceTest``.
 """
 
 import json
+import os
 import time
 
 import pytest
@@ -104,7 +105,11 @@ _JSON_STREAMING_RUNS = 4
 # seconds after the previous run's start (gap-aware: the stream's own elapsed
 # time counts toward the gap, so a run that already took 40s only sleeps the
 # remaining ~80s). Tune here if the endpoint's limit differs.
-_JSON_STREAMING_MIN_GAP_S = 120
+#
+# A directly-hosted vLLM server has no such ingress in front of it, so the gap
+# is pure sleep there (24 runs x 120s = ~48min of idle accelerator time). Set
+# QWEN3_STREAMING_MIN_GAP_S=0 when pointing the suite straight at a server.
+_JSON_STREAMING_MIN_GAP_S = int(os.environ.get("QWEN3_STREAMING_MIN_GAP_S", "120"))
 
 # If a run is still rejected/truncated despite the gap, wait one more gap and
 # retry that same run this many times before giving up. Only cleanly-completed
--- a/reference_config/benchmarking/benchmark_targets/model_performance_reference.json
+++ b/reference_config/benchmarking/benchmark_targets/model_performance_reference.json
@@ -326,6 +326,77 @@
                 }
             }
         ],
+        "blackhole_galaxy": [
+            {
+                "_comment": [
+                    "BLACKHOLE_GALAXY at data_parallel_size=4. Declaring data_parallel",
+                    "makes get_perf_reference() use this entry verbatim. Without a",
+                    "blackhole_galaxy row it fell through to the legacy subdevice remap,",
+                    "which maps (BLACKHOLE_GALAXY, 4) -> P150X8; there is no p150x8 row",
+                    "for this model, so every target resolved to NA and the benchmark",
+                    "checks reported 0/18 passed while looking green.",
+                    "Measured on g03blx02 with fabric_config FABRIC_2D_TORUS_XY, the",
+                    "fabric this device's spec now selects, across 7 sweeps on",
+                    "2026-08-27/28; TTFT varied under 26 ms and throughput under 1.6%,",
+                    "so the default 5% tolerance is generous. Values are the worst of",
+                    "the stable cluster.",
+                    "The last 3 sweeps include the sampling precompile fix that keeps",
+                    "sample_on_device_mode=all from crashing on unwarmed trace slots;",
+                    "it only adds warmup compiles, and decode was unchanged.",
+                    "At conc=1 tput ~= tput_user, as it must be with a single user.",
+                    "No 'theoretical' block: no hardware ceiling has been derived for",
+                    "Blackhole Galaxy, and copying the Wormhole numbers would assert a",
+                    "ceiling that was never measured on this device. The functional/",
+                    "complete tiers are therefore not emitted for this device rather",
+                    "than being emitted against an invented ceiling."
+                ],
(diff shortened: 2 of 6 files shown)


vcankovicTT (approved):

bgoelTT (approved):
Nice!

fivanovicTT (approved):

PR #5057, opened 2026-09-01
yolox_nano: set mesh graph descriptor for P150 to enable P300x2 deployment

rnabeelTT:
Deploying `yolox_nano` on P150 fails at device init:

```
warning  | Device | Using CUSTOM cluster type for P300 board with 1 chips
critical | Always | TT_FATAL: Custom fabric mesh graph descriptor path must be
                    specified for CUSTOM cluster type (tt_cluster.cpp:281)
```

A single Blackhole chip on a p300 board is classified as a CUSTOM cluster, and CUSTOM refuses to initialize without an explicit fabric mesh graph descriptor. The uvicorn server still binds and answers `/openapi.json`, so the container looks healthy while every request hits a dead worker — `/health` stays 405.

Sets `TT_MESH_GRAPH_DESC_PATH` on the yolox_nano P150 `device_model_spec`, the same workaround already carried by the Qwen3-Embedding specs in `prod/embedding.yaml` for the same condition.

## Verified

`python3 run.py --model yolox_nano --device p150 --workflow server --docker-server --service-port 7000` against `ghcr.io/tenstorrent/tt-media-inference-server-forge:0.18.0-c49bb76`:

```
Using custom mesh graph descriptor: .../p150_mesh_graph_descriptor.textproto [warmup] async executed in 53.0102 seconds. Forge model warmup 
All workers ready

GET  /health              -> 200
POST /v1/cnn/search-image -> 200 in 0.063s, top1_class_label "clock"
```

The `TT_FATAL: Chip 0 logical eth core 0-N connects to a remote mmio device` lines that follow are caught and logged as `Skipping ethernet core` — expected when using one chip of a p300, since the sibling chip's eth links are unused.


vcankovicTT (changes requested):
you should never change `prod` catalogue directly.
`prod` catalogue is updated/changed automatically at the end of the release cycle, once we officially confirm that everything works through the release pipelines, and based on this file, many other files should be updated and synced as well, such as `values.yml`, docs, `release_model_spec.json` etc.  Otherwise they will be out of sync with the prod catalogue. Therefore it should never be updated manually.
Please remove this change, and keep changes only in `dev` catalogue

rnabeelTT:
@vcankovicTT Thank you for informing me. The prod catalog has been reverted and the diff is now only one file.

vcankovicTT (approved):

PR #5058, opened 2026-09-02
Add configurable `TRAINING_STORE_ROOT` for training artifacts

mcupacTT:
Adds a configurable root for training artifacts (LoRA adapters, merged models, and `jobs.db`) so they land in a sensible, writable location per environment.

Root resolution:
- **tt-studio**: when the cache is mounted with `--host-volume`, artifacts are stored under `cache_root` for persistence across different containers.
- **CI**: otherwise, they default to an ephemeral `/tmp/tt_training_store` (the named volume isn't writable).
- **Override**: `TRAINING_STORE_ROOT` env var can explicitly set the root.

The job database (`jobs.db`) now lives under this same root, alongside the other artifacts.


mmilosevicTT (approved):
LGTM

ztorlakTT (approved):

PR #5059, opened 2026-09-02
Embedding: async request path (no thread-pool hop) + batch-fill timeout - 290k → 1.85M tok/s on BGE Galaxy

jzivanovicTT:
## What

Two server-side changes to the C++ embedding service, plus the benchmark
tooling needed to actually measure them:

### 1. Drop the blocking thread-pool hop from the request path
Embedding requests used to go HTTP thread → callback pool thread (blocking on
a future) → dispatch thread. The pool (16 threads by default) capped in-flight
requests far below the 32 workers × batch 8 = 256 server slots, so batches
never filled. The controller now calls `submitRequestAsync(request, onComplete)`
and the dispatch thread invokes the callback directly — no intermediate thread,
no blocking future. `produceResponse` remains as a sync adapter over the async
path. Every `PendingRequest` popped from the queue gets `onComplete` invoked
exactly once on every path (dispatch, failBatch, drainQueue).

### 2. Batch-fill timeout with an arrival-anchored deadline
When the queue is non-empty but the
batch is not full, the dispatch thread now waits up to `MAX_BATCH_DELAY_TIME_MS`
(default 2 ms) for the queue to fill. A batch departs exactly when its oldest member turns 2 ms old or the batch
fills, so no request is ever delayed by batching for more than 2 ms. A full
queue satisfies the predicate and exits early, so at saturation the wait is never
paid.

### 3. Multi-client benchmarking
`BENCHMARK_NUM_CLIENTS` splits the load across parallel `vllm bench serve`
processes (one process saturates a CPU core at ~2k req/s), and
`BENCHMARK_NUM_PROMPTS` overrides the request count so the startup transient
is <2% of the run. BGE Galaxy dev spec settles on 512 in flight / 4 clients /
20k prompts: 512 keeps all 256 server slots full with queue headroom (at
exactly 256, any client micro-stall drains the queue and fragments batches).

## Results (BGE-large-en-v1.5, Galaxy, 32 workers, ISL 384)

| Stage | Prefill throughput |
|---|---|
| before (thread-pool capped) | ~290k tok/s |
| + async request path | ~600k tok/s |
| + batch-fill timeout  | ~ 640k tok/s |
| + more requests + parallel clients  | ~1.2M tok/s |
| + 512 concurrency / 4 clients | **1.85M tok/s (4819 req/s, E2EL 99 ms)** |

290K: https://github.com/tenstorrent/tt-shield/actions/runs/33634786157/job/100262906964
600k: https://github.com/tenstorrent/tt-shield/actions/runs/33645522634/job/100459576676
640K: https://github.com/tenstorrent/tt-shield/actions/runs/33816633555/job/100859407905

1.2M: https://github.com/tenstorrent/tt-shield/actions/runs/34613521071/job/103335549007 
1.85M: https://github.com/tenstorrent/tt-shield/actions/runs/34342903945/job/102443162201


tt-media-server/scripts/embedding_perf_breakdown.py
```diff
@@ -0,0 +1,165 @@
...
+"""Measure the per-batch time breakdown of the BGE embedding runner.
+
+Answers two questions the benchmark numbers cannot:
+  1. Is device forward time flat in batch size (trace-padded), or does it grow?
+  2. How much do the host-side phases (tokenize, tolist) cost per batch?
+
+Run from the tt-media-server directory with the pinned tt-metal venv:
+
+    source /localdev/jzivanovic/tt-metal-65718bb/python_env/bin/activate
+    TT_METAL_HOME=/localdev/jzivanovic/tt-metal-65718bb \
```
ztorlakTT, line 14: This contains your local path and an old tt-metal pin. Can we make the instructions generic?

workflows/model_specs/dev/embedding.yaml
```diff
@@ -57,7 +57,16 @@ templates:
         VLLM__MAX_NUM_SEQS: "8"
         MAX_BATCH_SIZE: "8"
         DEFAULT_THROTTLE_LEVEL: "0"
-        BENCHMARK_MAX_CONCURRENCY: "256"
+        # Saturation probe: 512 in flight keeps all 32x8 server slots full
```
ztorlakTT, line 60: nit: generally, I think the comments throughout this PR are too detailed. Can we keep only what is needed to understand the code?

test_module/benchmark_tests/embedding_benchmark_tests.py
```diff
@@ -77,8 +81,56 @@ def _parse_embedding_benchmark_output(output: str) -> dict:
...
+    the slowest client's duration (they start together); mean latency is
+    weighted by each client's successful request count.
+    """
+    if len(all_metrics) == 1:
+        return all_metrics[0]
+
+    successful = sum(int(m.get("Successful requests", 0)) for m in all_metrics)
+    failed = sum(int(m.get("Failed requests", 0)) for m in all_metrics)
+    total_tokens = sum(float(m.get("Total input tokens", 0)) for m in all_metrics)
+    duration = max(float(m.get("Benchmark duration", 1.0)) for m in all_metrics)
```
ztorlakTT, line 97: Should we update this calculation in case the clients don't start at the same time?

ztorlakTT (approved):

ztorlakTT:
Great improvement!

tt-media-server/scripts/embedding_perf_breakdown.py
jzivanovicTT: I would maybe delete this script entirely

test_module/benchmark_tests/embedding_benchmark_tests.py
jzivanovicTT: I could change It to max(endtime) - min(startTime), that would a bit more precise I think.

ddjukicTT (approved):

idjuricTT (approved):

PR #5064, opened 2026-09-03
Add training_tests workflow to gate LoRA training in CI

mcupacTT:
## Summary

Adds a `training_tests` workflow so a LoRA fine-tuning run can be launched, driven, and graded end-to-end from CI - making training part of the tt-shield models CI.

`run.py --workflow training_tests` on a `TRAINING` model now:
1. Starts the forge/media training container as a sibling `ServerCommand` (from the `TRAINING` model spec).
2. Runs a launcher that submits one LoRA job, polls it to completion, and reads per-step loss from `GET /v1/jobs/{id}/metrics`.
3. Grades the loss trajectory against a checked-in expectation and writes `report_data_*.json`. Exits non-zero if the job or loss checks fail, so CI gates on it.

## Description of new files

- `launchers/run_training_test.py` — training driver (health-wait → submit → poll → fetch metrics → grade → report).
- `workflows/training/loss_check.py` — pure, unit-testable loss-trajectory grading.
- `workflows/training/registry.py` + `expected/*.yaml` — `(model, device) → expectation` mapping holding hyperparameters and expected losses. Initial entry: `Llama-3.1-8B` on P150 (SST-2).

## Testing
- the implementation can manually be tested using the following command (
```JWT_SECRET=test-secret MODEL_SPECS_ENV=dev python run.py   --model Llama-3.1-8B   --workflow training_tests   --device p150   --impl trainer-training-lora   --dev-mode   --docker-server   --override-docker-image tt-media-server-forge:local   --service-port 8000```
- (`HF_TOKEN` needs to be exported additionally)

### CODE REVIEW

File: workflows/training/expected/llama_3_1_8b_instruct_sst2.yaml [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+request:
+  dataset_loader: SST2
+  dataset_max_sequence_length: 64
+  batch_size: 8
+  learning_rate: 0.0001
+  num_epochs: 1
+  max_steps: 15
+  steps_freq: 5
+  val_steps_freq: 5
+  save_interval: 0
```
vcankovicTT, line 37: will this prevent saving the file adapter at the end of the training?
`save_strategy = "step" if request.save_interval > 0 else "none"   # → "none"`

vcankovicTT:
# Implementation proposal — config-driven training matrix

Replaces the hardcoded `run-matrix` in `dynamic-workflow.yml` with per-model scoping from `models-ci-config.json`, so training is turned on/off per model exactly like every other test type.

**4 files, 2 repos.**

| Repo | File | Action |
|---|---|---|
| tt-inference-server | `.github/workflows/models-ci-config-schema.json` | add 2 optional properties |
| tt-inference-server | `.github/workflows/models-ci-config.json` | add `Llama-3.1-8B` entry |
| tt-shield | `.github/scripts/generate_model_ci_workflows/generate_ci_matrix.py` | `model_type` filter + `impl` passthrough |
| tt-shield | `.github/workflows/dynamic-workflow.yml` | outputs, loop, build gate, **remove hardcoded matrix** |

> `models-ci-config.json` and its schema live in **tt-inference-server**. tt-shield fetches both at runtime (`dynamic-workflow.yml:110-114`).

---

## Point 1 — tt-inference-server: `.github/workflows/models-ci-config-schema.json`

Add two **optional** properties. They must go in **both** object variants — the single-engine one and the `implementations[]` item — because both declare `additionalProperties: false`:

```json
"model_type": {
  "type": "string",
  "enum": ["TRAINING"],
  "description": "Optional. Marks a non-inference workload. Absent = regular inference model."
},
"impl": {
  "type": "string",
  "description": "Optional. run.py --impl value (hyphenated impl_name), e.g. trainer-training-lora."
}
```

Do **not** touch `required`. Both stay optional → all 56 existing entries validate unchanged.

**Why:** `ModelType.TRAINING` already exists in tt-inference-server (`workflow_module/engine_types.py:196`) and is already used by `model_specs/dev/training.yaml`. This mirrors that existing concept into the CI config rather than inventing a new one. `impl` is added because `--impl` is the only thing that pins spec resolution deterministically (see Point 2).

---

## Point 2 — tt-inference-server: `.github/workflows/models-ci-config.json`

One new top-level key. `Llama-3.1-8B` (base) does not exist in this file today — only `Llama-3.1-8B-Instruct` does, and they are different models.

```json
"Llama-3.1-8B": {
  "implementations": [
    {
      "inference_engine": "FORGE",
      "model_type": "TRAINING",
      "impl": "trainer-training-lora",
      "ci": {
        "nightly": { "devices": ["P150"] },
        "release": { "devices": ["P150"] }
      }
    }
  ]
}
```

- `inference_engine: FORGE` — truthful, matches `model_specs/dev/training.yaml`. Training really does run on the forge image, so no fake engine value is invented.
- `impl` — **hyphenated** `impl_name`, not the underscored `impl_id` used in the catalogue yaml. `model_spec.py:1488` matches on `spec.impl.impl_name`.
- `ci` — this is the per-model on/off. Drop the `release` block to keep it out of releases.

**Nothing else in this file changes.**

**Why `impl` and not `default_impl`:** `--impl` narrows `candidates` *before* `default_specs` is computed (`model_spec.py:1484-1522`), so resolution is pinned to exactly one spec and `default_impl` is never consulted. Adding `default_impl: true` to `dev/training.yaml` instead would create a second default for `Llama-3.1-8B` on P150, making the engine-less resolution path (`model_spec.py:1500`) order-dependent between the vLLM and training specs.

---

## Point 3 — tt-shield: `.github/scripts/generate_model_ci_workflows/generate_ci_matrix.py`

`SERVER_TYPE_MAPPING` is **not** modified. Training entries declare `FORGE`, so the existing reverse lookup already produces `engine: "forge"` — which is what run.py needs.

**(a)** `generate_matrix()` signature (line ~171) — add a parameter:

```python
    model_type_filter: Optional[str] = None,
```

**(b)** line 190 — add the new bucket to the nested structure:

```python
    # Structure: server_type -> device -> {models: [], runner: {}, model-args: {}, model-impl: {}}
    matrix: Dict[str, Dict[str, Dict]] = defaultdict(
        lambda: defaultdict(lambda: {"models": [], "runner": None, "model-args": {}, "model-impl": {}})
    )
```

**(c)** line ~199 — first statement inside the implementations loop:

```python
        for impl in implementations:
            # `model_type` marks non-inference workloads (e.g. TRAINING). Absent =
            # regular inference model. Each invocation selects exactly one kind, so
            # training entries never leak into the inference matrices.
            if impl.get("model_type") != model_type_filter:
                continue

            server_type = get_server_type(impl["inference_engine"])
```

**(d)** after line 240 — store the impl per model, mirroring the existing `model-args` pattern:

```python
                # NB: the loop variable `impl` is the implementation dict; the
                # new config field is also called "impl".
                if impl.get("impl"):
                    matrix[server_type][device_key]["model-impl"][model_name] = impl["impl"]
```

**(e)** `_flatten_matrix()` — read it back. After line 262:

```python
            model_impl = config.get("model-impl", {})
```

and inside the per-model loop, alongside the existing `additional-args` handling:

```python
                if model in model_impl:
                    entry["impl"] = model_impl[model]
```

**(f)** `main()` (line ~288) — new CLI arg, and pass it into `generate_matrix`:

```python
    parser.add_argument("--model-type", dest="model_type", choices=["TRAINING"],
                        help="Select non-inference entries (default: inference only)")
```

```python
            model_type_filter=args.model_type,
```

> `model_type` is consumed by the filter only — it is never written into the matrix entry, so it never becomes a run.py flag.

**Why the filter is a plain equality:** `impl.get("model_type")` returns `None` for every existing entry, and `model_type_filter` is `None` when the flag is omitted. So `None != None` is false → existing models pass through unchanged, and `None != "TRAINING"` skips training entries out of the inference matrices. One expression handles both directions.

---

## Point 4 — tt-shield: `.github/workflows/dynamic-workflow.yml`

**(a)** `generate-matrix` outputs — after line 106:

```yaml
      matrix_training: ${{ steps.set-matrix.outputs.matrix_training }}
      has_training: ${{ steps.set-matrix.outputs.has_training }}
```

**(b)** the loop at line 130 — a 4th entry plus a third field:

```bash
          for config in "vllm:tt-inference-server:" \
                        "media:media-inference-server:" \
                        "forge:forge-media-inference-server:" \
                        "training:forge-media-inference-server:TRAINING"; do
            IFS=':' read -r key server_type model_type <<< "$config"

            ARGS=(--schedule "$SCHEDULE" --server-type "$server_type"
                  --config models-ci-config.json --schema models-ci-config-schema.json)
            [ -n "$model_type" ] && ARGS+=(--model-type "$model_type")

            MATRIX=$(python generate_ci_matrix.py "${ARGS[@]}" | tee /dev/stderr)
            # ... rest of the loop body unchanged
```

**(c)** forge build gate at line 182 — so a release containing only a training model still builds the image:

```yaml
  build-forge-media-inference-server:
    needs: [generate-matrix]
    if: |
      (inputs.run-forge && needs.generate-matrix.outputs.has_forge == 'true') ||
      (inputs.run-training && needs.generate-matrix.outputs.has_training == 'true')
```

**(d)** `test-training`, lines 290-307:

```yaml
  # tt-blacksmith LoRA training test. Reuses the forge image (blacksmith is baked
  # in). Scope comes from models-ci-config.json entries carrying
  # `model_type: TRAINING`, so ci_exclusions.yml and runner_mappings.yml apply.
  test-training:
    name: "Training"
    needs: [generate-matrix, build-forge-media-inference-server]
    if: |
      always() &&
      !cancelled() &&
      inputs.run-training &&
      needs.generate-matrix.outputs.has_training == 'true' &&
      needs.build-forge-media-inference-server.result == 'success'
    uses: ./.github/workflows/workflow_run-tests-with-inference-server.yml
    secrets: inherit
    with:
      run-matrix: ${{ needs.generate-matrix.outputs.matrix_training }}
      inference-server-type: forge-media-inference-server
      image: ${{ needs.build-forge-media-inference-server.outputs.cloud_image_tag }}
      inference-server-sha: ${{ inputs.inference-server-sha }}
      workflow: training_tests
      schedule: ${{ inputs.schedule }}
      timeout-minutes: 160
```

**Deleted:** the hardcoded `run-matrix: '[{...}]'` on line 301.
**Changed:** `timeout-minutes: 120 → 160` — the launcher's `--health-timeout 3600s` + `--job-timeout 5400s` = 150 min did not fit inside 120, so a worst-case run was hard-killed before it could write `report_data_*.json`.
**Added:** `generate-matrix` to `needs`, and the `has_training` condition.

---

## Not changed — deliberately

| File / thing | Why |
|---|---|
| `model_specs/dev/training.yaml` | `default_impl` **not** added — `--impl` already pins resolution, and adding it would make the engine-less path order-dependent |
| `SERVER_TYPE_MAPPING` | Training declares `FORGE`; the existing reverse lookup already yields the right engine |
| `workflow_run-tests-with-inference-server.yml` | Already forwards `matrix.config.impl` → `--impl` (line 338) and passes `--dev-mode` unconditionally (line 331) |
| `on-nightly.yml` | Once `has_training` gates the job, `run-training: default: true` is as harmless as `run-forge: default: true` |
| The other 56 `models-ci-config.json` entries | `model_type` absent = inference, already true for all of them |

## Verification

```bash
cd .github/scripts/generate_model_ci_workflows

# 1. should return exactly one entry, with impl: trainer-training-lora
python generate_ci_matrix.py --schedule release \
  --server-type forge-media-inference-server --model-type TRAINING \
  --config models-ci-config.json --schema models-ci-config-schema.json

# 2. REGRESSION — must be byte-identical to today's output
python generate_ci_matrix.py --schedule nightly \
  --server-type forge-media-inference-server \
  --config models-ci-config.json --schema models-ci-config-schema.json
```

Command 2 is the one that matters: it proves training entries do not leak into `matrix_forge` and that the existing models are untouched.

workflows/training/expected/llama_3_1_8b_instruct_sst2.yaml
mcupacTT: For now, the only implemented test is to check the loss values during training - no assertion about saved checkpoints - and therefore we keep save_interval zero to spare disk space

Later, we can also add test to check if checkpoints are correctly saved and can be located on the disk

vcankovicTT (approved):

idjuricTT (approved):
@ddjukicTT could you pls double check

vmaksimovicTT (approved):

ddjukicTT (approved):

Issue #3512, opened 2026-05-13
PrefillGateway: application-level request timeout


ztorlakTT:
The gateway should enforce `request_timeout_ms` per in-flight prefill request, independent of socket-level liveness. If a prefill accepts a `PrefillRequest` but the result doesn't come back in time, the gateway:

1. Marks the task failed and notifies decode (`PrefillTaskFailed { task_id, reason=timeout }`).
2. Decrements its in-flight count for that prefill.
3. Optionally flags the prefill `accepting_tasks=false` if multiple timeouts occur in a window.

Picks a sane default (e.g., 5× expected prefill latency at max ISL), configurable via gateway settings.

PR #5069, opened 2026-09-03
Count thought delimiters as KV rows in the multi-turn resume position

dmilinkovicTT:
## The defect

A cached conversation is continued from

```cpp
// src/services/llm_pipeline.cpp:139
req->kv_position_id = result.matchedTokens + result.accumulatedThinkTokens;
```

which must equal the number of KV rows the matched prefix occupies — the same
file calls it "the first free KV index".

Thought delimiters are excluded from **both** terms. Documented, in
`conversation_hasher.hpp`:

> Thinking tokens … are excluded from the hash computation but their count is
> tracked. **The markers themselves are also excluded from both the hash and the
> count.**

But they are ordinary generated tokens and each occupies a row, so the resume
position runs short by two per thought block and compounds over a conversation.
The delta prefill then overwrites live rows at the tail of the previous response
and writes everything after it two positions early.

## Why the change is this small

One loop produces two outputs with opposite requirements:

| output | exclude markers? | |
|---|---|---|
| block hash | **yes — correct as written** | they are structure, and templates that strip thinking from the re-rendered history do not carry them in the next prompt, so hashing them would break matching |
| think count | **no — this is the bug** | the count feeds a KV position, and a marker occupies a row |

A single `continue` served both. This keeps the hashing behaviour exactly and
changes only the count, in both state machines — they mirror each other by
design and must agree.

## Evidence

Measured against the deployed build (`gemma-4-31B-it`, greedy, thinking on) by
replaying a live multi-turn conversation and comparing the server's resume
position with the rows actually resident. Both sides derive from the same
`matchedTokens`, so block quantization cancels.

| turn | matchedTokens | thinkTokens | server resume | correct | gap | delimiters in prefix |
|-----:|--------------:|------------:|--------------:|--------:|----:|---------------------:|
| 1 | 1088 | 170  | 1258 | 1260 | **2** | 2 ✅ |
| 2 | 1600 | 170  | 1770 | 1772 | **2** | 2 ✅ |
| 3 | 2176 | 683  | 2859 | 2863 | **4** | 4 ✅ |
| 4 | 3200 | 1593 | 4793 | 4800 | 7 | 6 |
| 7 | 5024 | 3369 | 8393 | 8404 | 11 | 10 |

The first three continuations match the delimiter count exactly, with nothing
fitted. Later turns carry a constant +1 that never widens — unexplained, most
likely an artefact of the measuring harness — so the growing component is
entirely the delimiters. Extrapolated, ~20 rows by turn 13, where a recorded
agent task derailed.

`matchedTokens`/`thinkTokens` come from the worker's existing
`"Slot acquired"` log line; no instrumentation was added.

## Scope

Not Gemma-specific. `currentBlock.push_back(token)` is unreachable for a marker,
so markers never reach `matchedTokens` for any model. Rendering a past thinking
turn through each deployed tokenizer:

| model | delimiters | thinking in re-rendered history |
|---|---|---|
| gemma-4-31B-it | `<\|channel>` / `<channel\|>` | stripped — measured here |
| **deepseek-r1-0528** | `<think>` / `</think>` | stripped — same exposure, unmeasured |
| kimi-k2.6 / k2.7-code | `<think>` / `</think>` | kept in the prompt |
| minimax-m3 | `<think>` / `</think>` | kept in the prompt |

Where thinking survives the re-render the delta-slicing question differs, but the
undercount applies there too and this change adds each delimiter exactly once —
they were never in `matchedTokens` to double-count against.

## Safety

`accumulatedThinkTokens` is subtracted in three places, which looks like a hazard
and is not: each subtraction is the exact inverse of the addition, so the result
is invariant.

| site | | result |
|---|---|---|
| `decode_slot_reservation.cpp:19-28`, `:33-44` | `(matched−1)+think`, `+1`, `−think` | `matched` |
| `decode_slot_reservation.cpp:76-82` | same shape | `matched` |
| `disaggregation_service.cpp:946` | `(matched+think) − think` | `matched` |

`matchedTokens` is also the right value for the consumer: the prompt carries no
thinking, so the resident *prompt* token count is exactly `matchedTokens`.
`llm_pipeline.cpp:265-270` and `:292` do the same cancelling inverse. Slot-copy
paths pass `think=0` by construction.

No existing test pins the old behaviour — `prefix_cache_router_test.cpp` and
`session_manager_test.cpp` assert only `matchedTokens` relationships and that
`clearSessionBlockThinkTokens` zeroes the count.

## Follow-ups, not in this PR

1. **Rename the field.** `accumulatedThinkTokens` now means rows, not words —
   `thinkKvRows` or similar. Mechanical but noisy; kept out to keep this
   reviewable.
2. **Unclosed thought blocks.** A turn that hits `max_tokens` mid-thought never
   emits the closing delimiter, so the state machine stays inside thinking and
   the accounting detaches entirely — measured 3009 rows once. Needs a
   turn-boundary reset. The same hazard is documented for qwen3 in
   `workflows/model_specs/prod/llm.yaml`, mitigated there by disabling thinking —
   an option gemma-4 does not have, since the spec forces it on for ~10 points of
   eval score.
3. **Assert the invariant.** Nothing anywhere checks that the resume position
   equals the resident row count. tt-blaze hit this same bug class and fixed it
   structurally in #3512 by having the backend report the KV frontier it wrote —
   "two independent claims about the same rows, where before there was one claim
   compared against itself."

## Status

Draft. **Not compiled locally** — my environment lacks Drogon, so CI is the first
build of the added test. Causation is also not proven: the drift is measured and
consistent with the observed corruption, but demonstrating it *causes* the
garbling means landing this and replaying the failing conversation.

### DIFF
--- a/tt-media-server/cpp_server/CMakeLists.txt
+++ b/tt-media-server/cpp_server/CMakeLists.txt
@@ -1282,6 +1282,17 @@ target_include_directories(prefix_cache_router_test PRIVATE
     ${prometheus-cpp_SOURCE_DIR}/core/include
 )
 
+add_executable(gemma4_think_position_test
+    tests/unit/services/gemma4_think_position_test.cpp
+    src/domain/prefix_cache/block_matcher.cpp
+)
+target_link_libraries(gemma4_think_position_test PRIVATE
+    llm_runner_lib Drogon::Drogon prometheus-cpp::core GTest::gtest_main)
+target_include_directories(gemma4_think_position_test PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}/include
+    ${prometheus-cpp_SOURCE_DIR}/core/include
+)
+
 add_executable(mock_remote_kv_manager_test
     tests/unit/services/mock_remote_kv_manager_test.cpp
     src/services/mock_remote_kv_manager.cpp
@@ -1596,6 +1607,7 @@ gtest_discover_tests(worker_metrics_shm_test)
 gtest_discover_tests(concurrent_map_test)
 gtest_discover_tests(session_test)
 gtest_discover_tests(prefix_cache_router_test)
+gtest_discover_tests(gemma4_think_position_test)
 gtest_discover_tests(session_manager_test)
 gtest_discover_tests(worker_manager_test)
 gtest_discover_tests(llm_service_test)
--- a/tt-media-server/cpp_server/include/utils/conversation_hasher.hpp
+++ b/tt-media-server/cpp_server/include/utils/conversation_hasher.hpp
@@ -93,7 +93,7 @@ std::string renderLastUserTurn(const std::vector<ChatMessage>& messages,
 struct BlockHashInfo {
   uint64_t hash;
   uint32_t
-      accumulatedThinkTokens;  // Thinking content tokens (excluding markers)
+      accumulatedThinkTokens;  // KV rows used by thinking, markers included
 };
 
 /**
@@ -178,7 +178,8 @@ std::vector<uint64_t> getPrefixCacheHashesByBlocks(
  *
  * Thinking tokens (content between thinkStartId and thinkEndId markers) are
  * excluded from the hash computation but their count is tracked. The markers
- * themselves are also excluded from both the hash and the count.
+ * are excluded from the hash too, but ARE counted: the count feeds a KV
+ * position, and a marker occupies a row like any other token.
  *
  * Uses the same marker state machine as session tracking to classify tokens as
  * thinking or non-thinking.
--- a/tt-media-server/cpp_server/src/domain/session.cpp
+++ b/tt-media-server/cpp_server/src/domain/session.cpp
@@ -90,10 +90,12 @@ void Session::addGeneratedToken(uint32_t tokenId) {
 
   if (tokenId == thinkStartTokenId_) {
     inThinkingBlock_ = true;
+    ++accumulatedThinkTokens_;  // Marker occupies a KV row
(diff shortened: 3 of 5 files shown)


dmilinkovicTT:
Closing in favour of #5070, which fixes this properly.

I counted the delimiters unconditionally. #5070 counts a delimiter's row only when the model's chat template drops it from the re-rendered history, via per-model `thinkStartInHistory`/`thinkEndInHistory`. That is the correct rule: `matchedTokens` indexes the prompt, so a delimiter the next prompt re-supplies is already inside the matched slice and counting it again shifts the turn forward.

For gemma-4 and deepseek-r1 both flags are false, so the two PRs behave identically there — but mine would have over-counted kimi-k2.6/k2.7 and glm-5.2 by 2 rows per turn, and minimax-m3/glm-5.1 by 1. #5070 also derives the flags from the templates with a script rather than by hand, and catches that minimax-m3's think ids were m2.7's (`<think>` instead of `<mm:think>`), silently disabling think filtering there.

Verified #5070 on hardware with the harness that measured the original drift — gemma-4-31B-it, greedy, thinking on, live multi-turn replay:

| | drift per turn |
|---|---|
| before | 2, 2, 4, 7, 9, 11 |
| #5070 | 0, 0, 0, 0, 1, 1 |

`thinkTokens` also went 170 → 172 on the two turns with identical `matchedTokens`, i.e. exactly the two delimiters.

Not addressed by either PR: a turn that hits `max_tokens` mid-thought never emits the closing delimiter, so the marker state machine stays inside thinking and the accounting detaches entirely (measured 3009 rows once). Needs a turn-boundary reset; worth its own issue.

PR #5078, opened 2026-09-04
serve: drop Quetzal qualification/certification gates from the serve path (keep integrity)

nkapreTT:
## What

Removes **two** Quetzal PCC/qualification checks from the vLLM serve preflight in `vllm-tt-metal/src/run_vllm_api_server.py`:

- the `charter_pcc` required-runtime declaration inside the qualification manifest (empty/missing → hard fail); and
- the `QZ_QUALIFICATION_MANIFEST` presence/path requirement.

It **retains** the no-native-fallback admissibility guard (see below). One file; `ruff check` + `py_compile` clean.

## Why — a layering argument

A PCC/qualification check in the serving layer **conflates availability with certification** — two concerns the compiler/serving split is meant to keep apart.

- **Serving makes a model answerable.** Its job is to load a genuine, well-formed artifact and expose an endpoint.
- **Whether the model is faithful is measured by Models-CI evals** — the same standard native impls are held to.

Per Quetzal's charter, **attestation is provenance — a warning, never a serve blocker.** Certification (PCC ≥ bar) belongs on the **eval path plus a separate promote-to-certified gate**, not in `run_vllm_api_server.py`. The flow this restores: **serve (integrity only) → eval (measure) → promote (policy)**.

## Native fallback is prevented BY THIS DIFF (verifiable, not asserted)

The **no-native-fallback admissibility guard is RETAINED, fail-closed** — this is INTEGRITY/provenance, not PCC quality, and it is the check that makes a quetzal serve *admissible*:

```python
entries = quetzal_server.discover_models(str(root))
matching = [ e for e in entries.values() if
            e["model_id"] == model_id and e["backend"] == "generated_quetzal"
            and e["batch_size"] in (None, 1) and e["target_mesh"] in ("p150x4","4-chip")
            and required_context in bucket_lengths ]
if not matching:
    raise RuntimeError("... refusing native fallback")
return matching[0]
```

Without it, a native `tt_transformers` impl could silently substitute for the generated artifact and be reported as quetzal — inadmissible evidence. Provenance you don't enforce isn't provenance, so it stays. Because this guard is in the diff, native fallback is impossible **by this change**, not by an out-of-file claim.

## ⚠️ Sequencing gap — interim window with no PCC enforcement (stated, not hidden)

This PR removes **PCC/qualification enforcement from the serve path**, and the intended destination gate does **not exist yet**: there is currently **no PCC enforcement anywhere in the serve/eval driver** — `grep -rniE pcc run.py workflows/*.py` = **0**. So between this landing and the build-out of the Models-CI eval scoring + a promote-to-certified gate, **no layer in the stack enforces PCC** for a Quetzal serve.

Landing is defensible on the layering argument (serve should never have been the PCC gate, and eval/promote is the correct home), but the gap is real and must be sequenced: the eval-path PCC scoring + promote-to-certified gate should be built alongside/immediately after, not left indefinitely. This is called out so reviewers weigh it explicitly.

## Integrity kept, fail-closed (NOT quality)

- package-identity (`QUETZAL_PACKAGE_ID`/`QUETZAL_PACKAGE_ROOT`, `QZ_MODELS_ROOT` real dir);
- trusted-root bundle proof (manifest SHA-256, `ttq.artifact_bundle/v1` schema, per-file digest + size);
- artifact-is-a-regular-file + digest verification;
- TT-Metal runtime-identity match (patchset + `.ttq-runtime-identity.json`);
- `QUETZAL_VLLM=1`, `impl=quetzal`-forbids-native-registration;
- vLLM plugin-allowlist boundary + exactly-one Quetzal entry point;
- **the no-native-fallback admissibility guard (retained, above).**

## Status

**Draft — no auto-merge.** Please weigh the layering argument and the sequencing gap before this lands.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/vllm-tt-metal/src/run_vllm_api_server.py
+++ b/vllm-tt-metal/src/run_vllm_api_server.py
@@ -17,7 +17,6 @@ import sys
 from pathlib import Path
 from typing import Optional
 
-import yaml
 from huggingface_hub import snapshot_download
 
 from utils.cache_monitor import get_container_cache_dir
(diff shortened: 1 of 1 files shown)


nkapreTT:
Closing — superseded by #5092 (the minimal run.py --impl quetzal serve path) or standalone noise. Consolidating quetzal work into one reviewable PR instead of scattered overlapping ones.

PR #5080, opened 2026-09-04
feat(agentic-traces): borrow the Kimi template for requirements-driven runs

ipastalTT:
A model with no AGENTIC_TRACES_CONFIGS entry borrows the Kimi K2.7-Code config when the run is requirements-driven: synthesized off-catalog specs (which can never have an entry) and catalog specs alike. Being in the catalog says a model can be served, not that it is onboarded to this workflow, so a model added for evals would otherwise turn a working requirements sweep into a refusal.

Requirements mode is read via requirements_mode_in_argv(); every process that needs it already receives --requirements-json (run.py/run_workflows.py from the operator, the launchers because workflow_dispatch forwards it), so no plumbing is added. A plain --workflow agentic_traces on a model with no entry still refuses.

Also documents that use_server_token_count=True works against TT endpoints (they honour stream_options.include_usage).

Stack: part 1 of 6. Base for ipastalTT/st/2-agentic-metrics.

### DIFF
--- a/reference_config/agentic_traces/agentic_traces_config.py
+++ b/reference_config/agentic_traces/agentic_traces_config.py
@@ -20,12 +20,15 @@ config, and in ``tests/reference_config/test_agentic_traces_config.py``.
 
 from __future__ import annotations
 
+import logging
 from dataclasses import dataclass, field, replace
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


ipastalTT:
Consolidated into draft PR #5109 (stack reduced from 7 PRs to 3).

PR #5081, opened 2026-09-04
feat(agentic-traces): report E2EL P90 and e2e normalized interactivity

ipastalTT:
Adds the two report metrics the InferenceX agentic report carries: end-to-end latency P90 and e2e normalized interactivity (output throughput per user normalized against the SLO bar, with percentile inversion handled for the higher-is-better metric).

Stack: part 2 of 6, based on #5080. Retarget to main once it merges.


ipastalTT:
Consolidated into draft PR #5109 (stack reduced from 7 PRs to 3).

PR #5082, opened 2026-09-04
feat(agentic-traces): render measured runs in the agenticSweep shape

ipastalTT:
Introduces llm_module/agentic_traces/sweep_export.py, which converts measured run metrics into the requirements agenticSweep JSON shape, and renders that block into the markdown report (InferenceX rows only; SwarmOne runs are excluded).

Stack: part 3 of 6, based on #5081. Retarget to main once it merges.


ipastalTT:
Consolidated into draft PR #5110 (stack reduced from 7 PRs to 3).

PR #5083, opened 2026-09-04
feat(requirements): drive the agentic sweep from the document

ipastalTT:
Parses agenticSweep points and agentic workloads out of the requirements document (AgenticSweepPoint / AgenticWorkload in the schema, tolerating both flat and document-envelope JSON) and expands the base agentic-traces config into one run spec per requested concurrency in RequirementsTargetPack.

Stack: part 4 of 6, based on #5082. Retarget to main once it merges.

Made with [Cursor](https://cursor.com)

PR #5084, opened 2026-09-04
feat(agentic-traces): grade goodput against the document's SLOs

ipastalTT:
Threads the workload SLOs from the requirements document down to the AIPerf command as --goodput bars and back out through the parsed metrics, so a document asking for goodputPct actually gets graded.

Two non-obvious AIPerf behaviors are pinned by tests: --goodput takes the whole SLO string as ONE argv element (its validator splits it itself), and its tags are AIPerf's own (time_to_first_token / inter_token_latency / request_latency), not vLLM's ttft/tpot/e2el. Goodput comes back as a rate, so the export reconciles it against request throughput to report a percentage.

Stack: part 5 of 6, based on #5083. Retarget to main once it merges.

### DIFF
--- a/llm_module/agentic_traces/runs.py
+++ b/llm_module/agentic_traces/runs.py
@@ -72,6 +72,8 @@ class AgenticTracesRun:
     use_server_token_count: bool
     gpu_telemetry: bool
     mode: AgenticTracesMode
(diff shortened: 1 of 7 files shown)


ipastalTT:
Consolidated into draft PR #5110 (stack reduced from 7 PRs to 3).

PR #5085, opened 2026-09-04
feat(agentic-traces): emit the measured sweep as agenticSweep

ipastalTT:
Writes the measured sweep to disk as it completes: one agentic_sweep_<label>.json per run plus a combined agentic_sweep.json at the end, wrapped as {"agenticSweep": [...]} with fields ordered to match the InferenceX report. InferenceX rows only -- SwarmOne payloads are filtered so partial runs cannot leak incomplete points into the report artifact.

Stack: part 6 of 6, based on #5084. Retarget to main once it merges.


ipastalTT:
Consolidated into draft PR #5111 (stack reduced from 7 PRs to 3).

PR #5093, opened 2026-09-05
feat(serving): per-model tt-metal build isolation guard

nkapreTT:
Prevents one model's tt-metal build from corrupting another's live serve (the incident that broke the 1B serve: a gpt build advanced shared /home/nkapre/tt-metal to 2a8253ad while its lib was 1c2aff50 -> JIT NUM_WORKER_CORES failure).

serving/ttmetal_isolation/ (5 files): per-commit trees at $HOME/.cache/tt-metal/<commit>, a guard-checkout that REFUSES to build into a tree a live serve depends on (redirects to the per-commit path), build wrapper, identity stamping. Proven: guard-checkout blocks the exact 2a8253ad-vs-1c2aff50 mismatch on p01t06 with the 1B serve live; no false positives.

For codeowner review — not self-merged.

### DIFF
--- a/docs/quetzal_bare_node_serve.md
+++ b/docs/quetzal_bare_node_serve.md
@@ -1,0 +1,88 @@
+# Serving a Quetzal-compiled model with `run.py --impl quetzal`
+
+This is the minimal serve path for a Quetzal-generated (compiled) model on a bare
+Exabox QB2 (`P300X2`, four Blackhole chips driven as a `P150x4` mesh). It is the
+carved-out serve subset of the larger `#5042` bundle; the release-image contract,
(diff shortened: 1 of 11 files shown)


nkapreTT:
Closing — wrong repo and over-scoped. tt-inference-server serves/evals models; it does not own tt-metal builds, so a build-isolation guard doesn't belong here. The real fix is an operational convention, not ttis code: each model builds tt-metal into its own $HOME/tt-metal-<commit> tree, never the shared one (gpt/qwen already do this; only the 1B serve was still riding the generic shared tree). That belongs in the QB2 ops runbook, not a 546-line subsystem in the serving harness.

PR #5094, opened 2026-09-05
feat(agentic-traces): grade the measured sweep against requirements targets

ipastalTT:
A requirements document states its expectations as an `agenticSweep`: one point per concurrency with the metrics the measurement must meet. The report already emitted the measured sweep in that shape (#5085); this grades the two against each other.

- the pack hands the document's expected points (`AgenticSweepPoint.reference`, kept verbatim for exactly this) to `replace_agentic_runs`, which stamps the whole expected sweep on every run spec; it rides spec -> run -> driver payload as `expected_sweep`, the same path `goodput_slo` already took
- `sweep_export` grades measured vs expected field for field: exact (tolerance 0, matching the document's gte/lte comparators for benchmark targets), latencies at or below, rates at or above; the token-shape fields describe the trace mix rather than the server, so they are never graded
- the renderer adds a **Requirements Targets** section when expectations are present: per-point met/graded summary, one column per measured point, and a callout for the document points a truncated sweep never measured

Catalog runs carry no expectations and are unchanged: no section, no grading.

Stack: part 7 of 7, based on #5085. Retarget to main once it merges.

### DIFF
--- a/llm_module/agentic_traces/runs.py
+++ b/llm_module/agentic_traces/runs.py
@@ -74,6 +74,10 @@ class AgenticTracesRun:
     mode: AgenticTracesMode
     # AIPerf ``--goodput`` SLO string; empty means this run measures none.
     goodput: str = ""
+    # The requirements document's expected ``agenticSweep`` (all points, so a
+    # report can call out the ones a truncated sweep never measured); empty
+    # when nothing grades this run.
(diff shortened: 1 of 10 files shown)


ipastalTT:
Consolidated into draft PR #5111 (stack reduced from 7 PRs to 3).

PR #5095, opened 2026-09-05
models-ci: validate implementation identities

nkapreTT:
## Summary

- allow an optional nonempty `impl` on flat and `implementations[]` Model CI entries
- require explicit implementation names when one model repeats the same inference engine
- reject duplicate `(inference_engine, impl)` identities
- run the semantic validator from the existing config-validation workflow

This is configuration vocabulary only. It does not enroll Quetzal, add image/storage behavior, or change any current Model CI row.

## Why

Quetzal and native TT implementations both use `inference_engine: vLLM`. Without a first-class implementation identity, the two lanes are ambiguous before the Shield matrix is generated.

## Validation

- `uv run --with jsonschema --with pytest python -m pytest -q tests/test_models_ci_config_identity.py` — 9 passed
- `uv run --with jsonschema python scripts/validate_models_ci_config.py` — current repository config valid
- `ruff check` and `ruff format --check` using the repository-pinned 0.15.0
- `git diff --check origin/main...HEAD`

Follow-ups will preserve this field in tt-shield and separately add image/package execution. No model enrollment belongs in this PR.

### DIFF
--- a/.github/workflows/models-ci-config-schema.json
+++ b/.github/workflows/models-ci-config-schema.json
@@ -4,6 +4,11 @@
   "description": "Schema for models CI configuration",
   "type": "object",
   "$defs": {
+    "implementation_name": {
+      "type": "string",
+      "minLength": 1,
+      "description": "Optional model implementation selector. Required by semantic validation when a model has more than one entry for the same inference engine."
+    },
     "device_names": {
       "type": "string",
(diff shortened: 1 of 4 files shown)


nkapreTT:
Closing to use the existing reviewed Models CI escape hatch: put --impl quetzal in additional-args. This avoids adding a new config-schema field and removes a cross-repo matrix dependency from the enablement path.

PR #5092, opened 2026-09-05
feat(quetzal): minimal run.py --impl quetzal serve path for bare QB2

nkapreTT:
Minimal carve of the quetzal serve path (vs the unlandable #5042 monster: 81 files/+12.7K).

**What:** makes `run.py --impl quetzal` serve any quetzal-servable model on bare QB2 with no hand-hacks.
- quetzal ImplSpec in model_spec.py (makes --impl quetzal valid)
- scripts/generate_quetzal_model_row.py (rows generated from the quetzal catalog fragment)
- run_vllm_api_server.py: offline HF-cache weight resolution + content-address admission for BOTH v1 portable AND v2/installed-shared/MoE bundles (proven device-free on real llama-1b + gpt-oss bundles; fail-closed on tamper/byte-flip/missing-aux, write-bit independent)
- docs/quetzal_bare_node_serve.md

**Deliberately excluded** (follow-ups): release-image contract, models-ci enrollment/release automation, behavioral-topology/gpt120-preweight admission, agentic/eval wiring — the ~12K lines of #5042 that aren't the serve path.

**Diff:** ~6 files, +670/-20. For codeowner review — not self-merged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5096, opened 2026-09-05
feat(quetzal): admit immutable runtime bundles

nkapreTT:
## Summary

- register `quetzal` as a TTIS implementation without adding a model row
- perform fail-closed package admission through the public Quetzal package API before cache, weights, model registration, runtime, trace, or device startup
- derive provider identity generically from the selected model, context, batch, mesh, revision, and manifest digest
- bind the Quetzal provider before vLLM plugin discovery and prevent native model registration from shadowing it
- preserve an externally selected package root and digest; catalog values are fallback selectors only

This PR is the admission/runtime seam only. It excludes package mounting, image construction, catalog enrollment, eval configuration, and Shield wiring.

## Ownership and behavior contract

- TTIS selects the implementation and passes deployment identity
- Quetzal alone interprets and verifies portable or installed package layouts
- native implementations are unchanged
- missing, mutable, symlinked, incomplete, or digest-mismatched packages fail before device side effects
- package mode has no fallback to a broad artifact store or handwritten model
- no model-name or compiler-version special case is introduced

## Validation

- 266 directly affected tests pass on the final two-PR stack
- repository-pinned Ruff lint and format checks pass
- `git diff --check origin/main...HEAD` passes
- clean Exabox QB2 replay used this exact admission commit with the stacked mount commit
- the admitted C32768 Llama-3.2-1B package launched on P300X2, loaded only `quetzal_model_registry`, rebound `LlamaForCausalLM` to the generated Quetzal provider, reported health 200, and returned two deterministic temperature-zero probes

The QB2 replay is serving and collecting corrected eval evidence. This is launch/admission evidence, not a claim that the full Models CI qualification gate has completed.

This supersedes only the runtime/admission portion of #5092.

### DIFF
--- a/vllm-tt-metal/src/run_vllm_api_server.py
+++ b/vllm-tt-metal/src/run_vllm_api_server.py
@@ -7,6 +7,7 @@ import json
 import logging
 import multiprocessing
 import os
+import re
 import runpy
 import shlex
 import sys
@@ -14,7 +15,6 @@ from pathlib import Path
 from typing import Optional
 
 from huggingface_hub import snapshot_download
-from vllm import ModelRegistry
 
 from utils.cache_monitor import get_container_cache_dir
 from utils.device_utils import get_mesh_device_name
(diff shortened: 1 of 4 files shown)


vllm-tt-metal/src/run_vllm_api_server.py
```diff
@@ -33,6 +33,233 @@
...
+        raise RuntimeError(
+            "impl=quetzal requires vLLM revision as an immutable lowercase "
+            "40-hex commit"
+        )
+    if tokenizer_revision != revision:
+        raise RuntimeError("impl=quetzal requires tokenizer_revision to match revision")
+    device_type = model_spec.get("device_type")
+    if not isinstance(device_type, str) or not device_type:
+        raise RuntimeError("impl=quetzal requires a canonical device_type")
+    expected_mesh_device = get_mesh_device_name(device=device_type)
```
ddjukicTT: For P300X2 `os.environ["MESH_DEVICE"] == get_mesh_device_name(device=device_type) ` resolves to the string "P300x2" 
But every device: P300X2 row in `workflows/model_specs/dev/llm.yaml` that sets MESH_DEVICE overrides it to something else: "(1, 4)" or "P150x4".  So a quetzal P300X2 row written the way every other P300X2 row in the catalog is written gets rejected

vllm-tt-metal/src/run_vllm_api_server.py
ddjukicTT, line 1: Admission reads its selectors from os.environ at main():1001, but the values the loader runs against are written by set_runtime_env_vars at :1021, after admission. For QUETZAL_PACKAGE_ROOT / QUETZAL_BUNDLE_MANIFEST_SHA256 the preserve loop happens to keep the verified values, but QUETZAL_AUXILIARY_ROOTS_JSON is not in QUETZAL_EXTERNAL_PACKAGE_ENV, so a verified external aux-roots value is overwritten by the catalog's unverified one and aux roots declared only in the catalog are
never verified at all. If the catalog declares the selectors and the environment doesn't, dmission hard-fails, so the catalog-as-fallback path the tests describe is unreachable from main().

ddjukicTT (changes requested):

nkapreTT:
Addressed both review findings and rebased onto current main. Admission now validates the runtime manifest against the catalog physical MESH_DEVICE (not the logical P300X2 row name). Package selection now merges catalog selectors with explicit external selectors taking precedence, preserves QUETZAL_AUXILIARY_ROOTS_JSON, and verifies every selected auxiliary root before startup. Added regressions for catalog fallback, external selector preservation, auxiliary verification, and logical-vs-physical mesh mismatch. Local validation: 140 focused tests pass; Ruff, formatting, and git diff checks pass. Ready for re-review.

vllm-tt-metal/src/run_vllm_api_server.py
nkapreTT: Fixed in 2254f815. Admission now reads the physical topology from the selected ModelSpec MESH_DEVICE and compares the manifest against that value. The P300X2 catalog row in #5140 explicitly supplies MESH_DEVICE=P150x4; the logical catalog device is no longer compared to the physical mesh string. A regression covers this distinction.

vllm-tt-metal/src/run_vllm_api_server.py
nkapreTT: Fixed in 2254f815. Selection is now formed before admission by merging catalog selectors with explicit process-environment selectors taking precedence. QUETZAL_AUXILIARY_ROOTS_JSON is preserved with the other external selectors, and admission verifies every selected auxiliary root. Regressions cover catalog-only fallback, external preservation, and auxiliary verification.

ddjukicTT (approved):

tstescoTT (approved):

tstescoTT:
I'll approve these to unblock. Please flag if there are issues in CI. Doing a dispatch run can smoke these likely.

PR #5097, opened 2026-09-05
Honor model context when selecting agentic evals

nkapreTT:
## Summary

- derive agentic request reachability from SWE-bench and TerminalBench input/output budgets
- use one selector for runtime tasks, explicit benchmark selection, release-child planning, and venv provisioning
- omit the release agentic child when no task fits the selected device context
- preserve existing incomplete legacy task declarations and add no model-specific policy

This is generic Model CI plumbing. It intentionally adds no Quetzal catalog row or evaluation task.

## Why

Current release selection can attach a 192K/232K agentic recipe to a smaller-context serve merely because the model name matches. This makes a 32K lane unrunnable and can also provision an agentic child that has no valid task. The standard eval path already filters by context; agentic selection needs the same property.

Boundary tests prove that a 24,576-input + 8,192-output task fits a 32,768 context exactly, while a 32,769-token envelope is omitted. A native-like 256K context continues to select the existing 192K task.

## Validation

```text
uv run --no-project --with-requirements requirements-dev.txt --with aiohttp \
  python -m pytest -q \
  tests/test_module/llm_tests/test_agentic_eval_tests.py \
  tests/workflow_module/test_release_workflow.py \
  tests/workflows/test_workflow_dispatch_routing.py \
  tests/test_module/llm_tests/test_llm_eval_tests.py

221 passed, 1 warning
```

Repo-pinned Ruff 0.15.0 E/F checks and format checks pass; `git diff --check` passes.

### DIFF
--- a/llm_module/eval_configs.py
+++ b/llm_module/eval_configs.py
@@ -2,15 +2,14 @@
 #
 # SPDX-FileCopyrightText: 2026 Tenstorrent AI ULC
 
-"""Select the standard (lm-eval / lmms-eval) eval tasks for an LLM model."""
+"""Select standard and agentic eval tasks for an LLM model."""
 
 from __future__ import annotations
 
 import logging
 from typing import List
 
-from workflow_module.engine_types import EvalLimitMode
-from workflow_module.engine_types import WorkflowVenvType
+from workflow_module.engine_types import EvalLimitMode, WorkflowVenvType
 
 from .eval_command import _get_limit_mode, _parse_eval_samples_mapping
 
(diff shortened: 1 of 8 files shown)


nkapreTT:
Closing: this targets an eval family that nothing currently runs.

Checked before closing:
- `.github/workflows/models-ci-config.json` covers 56 models and contains zero references to `swe`, `agentic`, `tau`, or `terminal`
- the release workflow executes GPQA + LongBench only (tt-shield run 34610746272: 19 gpqa, and the longbench_*_e variants; no agentic or SWE tasks appear)
- tt-shield references to "agentic" are docs and a dashboard data file, not workflow enrollment

So the agentic-eval selection path this fixes is not exercised for any model in any schedule.

**The underlying problem is real and worth recording.** `EVAL_CONFIGS` is keyed on `hf_model_repo` with no impl dimension, so an alternate-impl artifact inherits a config tuned for the native implementation — including its context envelope. That genuinely misfires today, but on the evals that DO run: a config tuned for a native model at 131072 context applied to an artifact at 32768 is a plausible explanation for a LongBench score well below its recorded reference.

This PR fixes that class of bug in the agentic selection path only. The fix that would matter needs to target GPQA/LongBench config selection instead, so closing rather than merging machinery for a path nobody executes.

PR #5098, opened 2026-09-05
Mount immutable Quetzal packages read-only

nkapreTT:
## Summary

- add one supported host input: `--quetzal-package-root <selected-package-dir>`
- pass only `QUETZAL_PACKAGE_ROOT` and `QUETZAL_BUNDLE_MANIFEST_SHA256` as package selectors
- mount exactly that selected package read-only at the catalog-declared runtime root for Docker
- use the exact selected host package for trusted local-server validation
- forward `--impl quetzal`
- make `--print-docker-cmd` expose the exact mount and environment contract
- reject missing roots, native misuse, symlink roots, malformed identity, and path escape

This PR deliberately excludes broad `/mnt/models` mounts, loose artifact fanout, TTIS-owned manifest parsing, path rebasing, image construction, catalog enrollment, and Shield wiring.

## Stack

This PR is intentionally stacked on #5096. GitHub therefore shows only the immutable package handoff and mount layer here.

## Validation

- 266 directly affected tests pass on the final two-commit stack
- repository-pinned Ruff lint and format checks pass
- `git diff --check` passes
- the printed Docker command is covered for one exact read-only package mount with no broad models mount
- clean Exabox QB2 local-server replay used this exact commit with #5096
- the selected C32768 package served Llama-3.2-1B on P300X2 with health 200 and exact generated-provider identity

The local-server replay is trusted pre-submit evidence. The final container/image and Shield dispatch receipts remain separate downstream gates.

### DIFF
--- a/workflows/quetzal_package.py
+++ b/workflows/quetzal_package.py
@@ -1,0 +1,95 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+"""Host-to-container mount contract for an immutable Quetzal package."""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional
+
+QUETZAL_IMPL_ID = "quetzal"
+QUETZAL_PACKAGE_ROOT_ENV = "QUETZAL_PACKAGE_ROOT"
+QUETZAL_MANIFEST_SHA256_ENV = "QUETZAL_BUNDLE_MANIFEST_SHA256"
+
+
+@dataclass(frozen=True)
+class QuetzalPackageMount:
+    host_root: Path
+    runtime_root: Path
+    manifest_sha256: str
+
+
+def _is_quetzal(model_spec) -> bool:
+    return (
+        getattr(getattr(model_spec, "impl", None), "impl_id", None) == QUETZAL_IMPL_ID
+    )
+
+
+def resolve_quetzal_package_mount(
+    model_spec, runtime_config
+) -> Optional[QuetzalPackageMount]:
+    """Validate and resolve a host package and its catalog runtime location."""
+    configured = getattr(runtime_config, "quetzal_package_root", None)
+    is_quetzal = _is_quetzal(model_spec)
+    launches_server = bool(
+        getattr(runtime_config, "docker_server", False)
+        or getattr(runtime_config, "local_server", False)
+    )
+
+    if configured and not is_quetzal:
+        raise ValueError("--quetzal-package-root is only valid with --impl quetzal")
+    if configured and not launches_server:
+        raise ValueError(
+            "--quetzal-package-root requires --docker-server or --local-server"
+        )
+    if is_quetzal and launches_server and not configured:
+        raise ValueError("--impl quetzal requires --quetzal-package-root")
+    if not configured:
+        return None
+
+    supplied = Path(configured).expanduser()
+    if supplied.is_symlink() or not supplied.is_dir():
+        raise ValueError(
+            f"--quetzal-package-root must be an existing real directory: {supplied}"
+        )
+    host_root = supplied.resolve()
+
+    env_vars = model_spec.env_vars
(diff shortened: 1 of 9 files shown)


workflows/quetzal_package.py
```diff
@@ -0,0 +1,93 @@
...
+    return QuetzalPackageMount(
+        host_root=host_root,
+        runtime_root=runtime_root,
+        manifest_sha256=expected_sha256,
+    )
+
+
+def quetzal_package_env(
+    package: QuetzalPackageMount, *, local_server: bool
+) -> dict[str, str]:
```
ddjukicTT, line 89: -> dict[str, str] with no from __future__ import annotations, so on 3.8 it's TypeError: 'type' object is not subscriptable at import.

workflows/run_docker_server.py
```diff
@@ -607,6 +625,8 @@ def generate_docker_run_command(
     if model_spec.inference_engine == InferenceEngine.VLLM.value:
         docker_command.extend(["--model", model_spec.hf_model_repo])
         docker_command.extend(["--tt-device", runtime_config.device])
+        if quetzal_package_mount:
```
ddjukicTT, line 628: 1. --runtime-model-spec-json path → TypeError. runtime_config.impl is None there (run.py:1077 never sets it), so ["--impl", None] hits shlex.quote(None). This isn't limited to --print-docker-cmd: run_docker_server logs format_docker_command(docker_command) on every run, so the real execution path crashes before docker is invoked. And this is currently the only reachable path, since no catalog row has impl: quetzal yet.
2. Catalog path → container can't resolve the spec. The runtime-spec JSON is only mounted under dev_mode or custom_weights (run.py:1250-1254); otherwise the container falls back to its baked-in catalog, which has no quetzal row, and --impl quetzal dies at No default_impl found. The validation replay was --local-server, which sidesteps this.

workflows/run_local_server.py
```diff
@@ -220,6 +229,8 @@ def generate_local_run_command(
         runtime_config.device,
     ]
 
+    if quetzal_package_mount:
+        command.extend(["--impl", runtime_config.impl])
```
ddjukicTT: Hits the same issue as r`un_docker_server.py:629`

nkapreTT:
Restacked onto the reviewed #5096 fixes and addressed all three inline findings. Added postponed annotation evaluation for Python 3.8; runtime JSON now resolves impl/engine from the loaded ModelSpec so command generation never quotes None; Docker/local launch use the resolved model_spec implementation; and every Quetzal Docker launch mounts the exact host-resolved runtime spec, even outside dev/custom-weight flows. Added focused regressions for each path. Local validation: 359 combined tests pass; repository Ruff and formatting checks pass; git diff check passes. Ready for re-review after #5096.

workflows/quetzal_package.py
nkapreTT: Fixed in 4a5b8c49 by adding postponed annotation evaluation to workflows/quetzal_package.py. The module now imports under Python 3.8 semantics.

workflows/run_docker_server.py
nkapreTT: Both paths are fixed in 4a5b8c49. Runtime JSON resolution fills impl and engine from the loaded ModelSpec, Docker command generation uses model_spec.impl.impl_name instead of nullable runtime_config.impl, and every Quetzal launch mounts the exact host-resolved runtime spec even outside dev/custom-weight modes. Focused regressions cover each case.

workflows/run_local_server.py
nkapreTT: Fixed in 4a5b8c49. Local command generation also uses model_spec.impl.impl_name, with a regression proving a runtime config whose impl is None still emits --impl quetzal.

ddjukicTT (approved):

nkapreTT:
Restacked onto merged #5096/main after GitHub retargeted the stack. The PR is now the intended 9-file, +437/-5 mount-only diff; no #5096 changes are duplicated. Revalidation on the new head: 359 focused tests pass, repository Ruff and formatting pass, and git diff check passes. Requesting the required fresh approval after the base change.

PR #5099, opened 2026-09-05
Resolve pinned Hugging Face snapshots offline

nkapreTT:
## Summary

- resolve the catalog-pinned model revision directly from HF_HOME when offline mode is enabled
- set MODEL_WEIGHTS_DIR to the cached snapshot returned by huggingface_hub
- report the missing repo, revision, and HF_HOME on an offline cache miss
- preserve the existing online download and fallback behavior

## Test plan

- python3 -m pytest -q tests/test_run_vllm_api_server.py
- ruff format --check vllm-tt-metal/src/run_vllm_api_server.py tests/test_run_vllm_api_server.py
- ruff check vllm-tt-metal/src/run_vllm_api_server.py tests/test_run_vllm_api_server.py

Stacked on #5098.

### DIFF
--- a/tests/test_run_vllm_api_server.py
+++ b/tests/test_run_vllm_api_server.py
@@ -747,7 +747,55 @@ def _weights_spec():
     return {
         "model_name": "Mistral-7B-Instruct-v0.3",
         "hf_model_repo": "mistralai/Mistral-7B-Instruct-v0.3",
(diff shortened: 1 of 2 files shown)


nkapreTT:
Folded this exact two-file cache-resolution change into #5098 so reviewers have one package/storage PR instead of another stack level. The focused resolver tests remain 30/30 passing there.

PR #5103, opened 2026-09-05
vLLM: reject unsupported TT sampling at request validation

nkapreTT:
## Summary

- reject options that require compatibility sampling before TT engine execution
- reject non-greedy requests when the selected TT model does not advertise on-device non-greedy support
- preserve stochastic requests for native-capable TT models and host-sampling mode

## Why

Generated Quetzal currently supports device-side greedy argmax only. A stochastic OpenAI request previously reached the EngineCore, raised during execution, and terminated the server. This uses the TT platform's existing model-derived `non_greedy_decoding_on_device` capability to return a request error before engine execution.

## Validation

- `python -m pytest -q tt-vllm-plugin/tests/test_platform_sampling.py` (vLLM 0.10.1.1): 12 passed
- `python -m ruff check --select I tt-vllm-plugin/tt_vllm_plugin/platform.py tt-vllm-plugin/tests/test_platform_sampling.py`: passed

Observed failing request preserved unchanged: temperature=1.0, top_p=0.9, top_k=128256. No eval sampling configuration was changed.

### DIFF
--- a/tt-vllm-plugin/tests/test_platform_sampling.py
+++ b/tt-vllm-plugin/tests/test_platform_sampling.py
@@ -1,3 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
 from contextlib import contextmanager
 
 import pytest
--- a/tt-vllm-plugin/tests/test_platform_sampling.py
+++ b/tt-vllm-plugin/tests/test_platform_sampling.py
@@ -1,9 +1,8 @@
 from contextlib import contextmanager
 
 import pytest
-from vllm.sampling_params import SamplingParams
-
 from tt_vllm_plugin.platform import TTPlatform
+from vllm.sampling_params import SamplingParams
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


nkapreTT:
Closing: this patches code that has not shipped since 2026-08-14.

The target is `tt-vllm-plugin/tt_vllm_plugin/platform.py`, a vendored copy. The image clones a different repo — `vllm.tt-metal.src.dev.Dockerfile:104-106`: `git clone https://github.com/tenstorrent/vllm-tt-plugin.git` (module `vllm_tt_plugin`). The split landed in `3fdbc5811` (#4907); this repo's own tests say so (`tests/test_validate_setup.py:563`: "vllm-tt-plugin is its own repository now"). The vendored `platform.py` was last touched 2026-06-17. So this change would be a no-op for the deployed runtime.

The premise is also stale. It states "Generated Quetzal currently supports device-side greedy argmax only" — untrue since quetzal `946aa208` (2026-09-05) added on-device temperature, full-vocabulary nucleus, bounded top-k <=64 and seeds. That path is silicon-qualified at gpt-oss's vocab width: all 10 lanes PASS at 201,088.

Equivalent enforcement is already handled upstream in the live plugin, which resolves the capability from `model_capabilities` on the model class and defaults it to `True`.

The residual bug worth fixing is narrower and a boolean cannot express it: `sampling_policy` raises on `min_p`, presence/frequency/repetition penalties, logprobs, and `top_k > 64`. `temperature=0, presence_penalty=0.5` is a *greedy* request that still terminates EngineCore. Follow-up should declare the actual envelope.

PR #5110, opened 2026-09-07
feat(requirements): drive the agentic sweep from the document, render it, grade goodput

ipastalTT:
## Summary

Consolidates the middle three layers of the agentic requirements stack (#5082 + #5083 + #5084) into one PR. Base branch is PR 1 of the consolidated stack.

**Render measured runs in the `agenticSweep` shape** (from #5082): new `llm_module/agentic_traces/sweep_export.py` — `to_agentic_sweep_point()` maps one run's metrics, `to_agentic_sweep()` maps a set, one point per concurrency, so expected and observed points compare field for field. Unmeasured metrics are omitted rather than zero-filled. Also extracts ISL/OSL p95 (trace replay draws prompts of very different sizes; the mean hides the long-context tail).

**Drive the agentic sweep from the document** (from #5083): the requirements loader now parses both the flat and the `document`/`workloads` envelope shapes, adds `AgenticWorkload`/`AgenticSweepPoint`, and `RequirementsTargetPack` expands the catalog/template config to one run per requested concurrency. Workload SLOs are exposed as `--goodput` constraints.

**Grade goodput against the document's SLOs** (from #5084): threads the SLOs from the document down to the run spec, the AIPerf command, and back out through parsed metrics. Pins the AIPerf-vs-vLLM goodput tag mapping and the single-argv-element `--goodput` requirement. Runs record the bars they were graded against, so 0% reads as "nothing met the SLOs" rather than "never graded".

## Stack

This is PR 2 of 3 (replaces the 7-PR `st/*` stack):

1. `agentic-req-1-run-foundation` — run foundation + metrics
2. **← this PR** — drive the sweep from the document, render it, grade goodput
3. `agentic-req-3-sweep-grade` — emit + grade the measured `agenticSweep`

## Tests

241 passed (adds `test_agentic_traces_sweep_export`, `test_requirements_schema`, `test_agentic_traces_tests`).


llm_module/drivers/aiperf_agentic_traces.py
```diff
@@ -407,6 +424,11 @@ def _int(tag: str, stat: str = "avg") -> int:
         "error_rate_pct": _stat("request_error_rate"),
         "mean_isl": _stat("input_sequence_length"),
         "mean_osl": _stat("output_sequence_length"),
+        # Trace replay draws prompts of wildly different sizes from the trace
+        # pool, so the mean alone hides the long-context tail that sets KV
+        # pressure.
+        "p95_isl": _stat("input_sequence_length", "p95"),
```
ddjukicTT, line 434: This `_stat(..., default=0)` is going to bite us once 5111/5116 start grading.

`to_agentic_sweep_point` is written as if missing metrics get omitted. On a real `profile_export_aiperf.json` they don't — every tag that isn't there still lands on the payload as `0.0`. Then a latency target like `ttftP90Ms: 30000` passes because `0 <= 30000`. Same for these new p95 fields: a missing ISL p95 becomes `inputTokensP95: 0`.

ipastalTT: Fixed in acdef86b3. `_stat` now defaults to `None`, the `.get(..., 0)` calls are gone, and the metrics dict is filtered so a tag the export never contained stays off the payload entirely (`metrics = {k: v for k, v in metrics.items() if v is not None}`). Counts keep a zero default via `_int`, with a comment explaining the asymmetry: an absent counter genuinely means zero happened, an absent metric does not. `test_metrics_the_export_lacks_are_omitted_not_zeroed` pins the parsing behavior.

ddjukicTT (approved):

PR #5111, opened 2026-09-07
feat(agentic-traces): emit the measured sweep as agenticSweep and grade it against targets

ipastalTT:
## Summary

Consolidates the last two layers of the agentic requirements stack (#5085 + #5094) into one PR. Base branch is PR 2 of the consolidated stack.

**Emit the measured sweep as `agenticSweep`** (from #5085): written three ways, all on by default — per point as soon as it completes (a failure at point four must not take the three already measured with it), once for the sweep after the loop, and inline in the report. Wrapped in the document's own `{"agenticSweep": [...]}` object so a file written mid-sweep pastes straight into a document. InferenceX-only: swo-bench measures a different metric set (no TPOT, no p95s, no goodput), so its points would read as measured zeroes.

**Grade the measured sweep against requirements targets** (from #5094): the pack stamps the document's expected sweep on every run spec (riding the same spec → run → driver path as `goodput_slo`); `sweep_export` grades measured vs expected field for field (exact, tolerance 0, matching the document's gte/lte comparators; latencies at-or-below, rates at-or-above; token-shape fields never graded); the renderer adds a Requirements Targets section with per-point met/graded summary and a callout for document points a truncated sweep never measured. Catalog runs carry no expectations and are unchanged.

## Stack

This is PR 3 of 3 (replaces the 7-PR `st/*` stack):

1. `agentic-req-1-run-foundation` — run foundation + metrics
2. `agentic-req-2-sweep-from-doc` — drive the sweep from the document, render it, grade goodput
3. **← this PR** — emit + grade the measured `agenticSweep`

## Tests

276 passed (full agentic-traces + requirements suite).


llm_module/agentic_traces/sweep_export.py
```diff
@@ -99,16 +111,21 @@ def to_agentic_sweep_point(
         if value is not None:
```
ddjukicTT, line 110: This only omits a field when the key is missing or non-numeric. `parse_aiperf_output` writes the keys with `_stat` / `.get(..., 0)`, so a tag that was not in the export is still `0.0` on the payload that feeds this function.

`_number(0.0)` is not `None`, so the field is exported and graded. For lower-is-better metrics that means `measured=0.0` satisfies any target (`ttftP90Ms` vs `30000`). `_invalid_result_reason` still drops `mean_ttft_ms <= 0`; a run with avg/completed present and tails missing is not dropped.

`test_omits_metrics_the_run_did_not_produce` uses a hand-built sparse dict, not `parse_aiperf_output` output.

llm_module/agentic_traces/sweep_export.py
```diff
@@ -99,16 +111,21 @@ def to_agentic_sweep_point(
...
             point[field] = value
 
-    goodput_pct = _goodput_pct(metrics)
-    if goodput_pct is not None:
-        point["goodputPct"] = goodput_pct
-
     for metric in _CACHE_HIT_METRICS:
```
ddjukicTT, line 113: `theoretical_prefix_cache_hit_pct` is `_stat("theoretical_prefix_cache_hit")` with default `0` in `parse_aiperf_output`, so it is always on the metrics dict. The first-numeric-wins loop therefore always writes `kvCacheHitRatePct` — either `measured_prefix_cache_hit_pct` when that key exists, or theoretical (a real trace-pool value, or `0.0`).

There is no path here that leaves `kvCacheHitRatePct` off the point. `test_falls_back_to_the_theoretical_cache_hit_rate` documents the 98.08 fallback when the measured key is absent.

report_module/agentic_traces_renderer.py
```diff
@@ -576,11 +577,156 @@ def render_agentic_traces(block: Block, metadata: Mapping[str, Any]) -> str:
...
+            f"**c{point.concurrency}**: {point.met}/{point.graded} targets met "
+            f"{'✅' if point.passed else '❌'}"
+            for point in verdicts
+        )
+        parts.append(summary)
+
+        headers = ["Metric"] + [
+            f"c{point.concurrency} ({point.met}/{point.graded})" for point in verdicts
+        ]
+        fields = [verdict.field for verdict in verdicts[0].verdicts]
```
ddjukicTT: Row set is `verdicts[0]`. Later points use `next(v for v in point.verdicts if v.field == field)`.

`next(...)` raises `StopIteration` when that field is absent. `ReportGenerator._render_block` is `except Exception: return ""` (`generator.py`), and `StopIteration` is an `Exception`.

Order matters. If the first point has more fields than a later one, the renderer raises and the section is empty. If the first point is the sparse one, there is no exception; fields that exist only on later points never become rows, while the header still uses that later point's `met/graded`. Tests render identical `_POINT` dicts only.

llm_module/agentic_traces/sweep_export.py
```diff
@@ -125,4 +142,176 @@ def to_agentic_sweep(runs: Sequence[Mapping[str, Any]]) -> List[Dict[str, Any]]:
...
+    @property
+    def graded(self) -> int:
+        return sum(1 for v in self.verdicts if v.passed is not None)
+
+    @property
+    def met(self) -> int:
+        return sum(1 for v in self.verdicts if v.passed is True)
+
+    @property
+    def passed(self) -> bool:
```
ddjukicTT: `return self.graded > 0 and self.met == self.graded` makes `passed` false when `graded == 0`.

`AgenticSweepPoint.from_dict` only requires `concurrency`. A document point `{concurrency: 64}` is valid, produces empty `verdicts`, and still enters this path. 5116's `_check_agentic_targets` then emits `Agentic targets at concurrency 64: the run produced no gradable metrics.`

The payload still has the run's metrics; the expected point had no fields in `POINT_FIELDS` to grade.

llm_module/agentic_traces/sweep_export.py
ipastalTT: Fixed at the source in acdef86b3 (now in this branch's history after rebasing onto the new 5110 tip): `parse_aiperf_output` omits tags the export never contained instead of zeroing them, so `_number(metrics.get(...))` is `None` and the field stays off the point — grading reads it as unmeasured (➖), never as a vacuous `0 <= target` pass.

On the test point: added `test_omitted_metrics_stay_off_the_sweep_point`, which writes a real `profile_export_aiperf.json` and runs the actual `parse_aiperf_output` → `to_agentic_sweep_point` path rather than a hand-built sparse dict, asserting the sparse point comes out the other end.

llm_module/agentic_traces/sweep_export.py
ipastalTT: ~~Fixed in acdef86b3. `theoretical_prefix_cache_hit` now parses with a `None` default and is filtered from the payload when the export lacks the tag, so when neither the measured nor the theoretical value exists the first-numeric-wins loop finds nothing numeric and `kvCacheHitRatePct` stays off the point — ungraded instead of a phantom 0% that fails any higher-is-better target. The genuine fallback still works: `test_falls_back_to_the_theoretical_cache_hit_rate` (measured absent, theoretical 98.08 present) passes.~~

This actually got fixed here: [link](https://github.com/tenstorrent/tt-inference-server/pull/5110/changes/092cb8d5a2b8fcb1a81b2ebe42ea2a2e5f0c0017#diff-ba16d9e6968dc2a619bb77ab9b6412d57d0e136592ae3e567b3b54604b93aa0dR49-R54)

report_module/agentic_traces_renderer.py
ipastalTT: Fixed in ee29891de. Rows are now the union of declared fields across all points, in canonical `POINT_FIELDS` order, built with per-point dict lookups (`vmap.get(field)`) instead of `next(...)` — no `StopIteration` path remains. A point that did not declare a field renders a `—` dash, distinct from the `N/A ➖` cell for declared-but-unmeasured. `test_points_declaring_different_fields_render_the_union` covers both directions: a field only the later point declares still gets its row, and the point that didn't declare it gets the dash.

llm_module/agentic_traces/sweep_export.py
ipastalTT: Fixed in ee29891de. `PointVerdict.passed` is now `Optional[bool]`: `None` when the document declared no gradable fields at that concurrency (an empty point promises nothing, so there is nothing to fail), still `False` when fields were declared but the run produced none of them — that gap is the run's. The renderer shows `**c64**: no targets declared` for the None case instead of a ❌, and 5116's acceptance check skips such points entirely (no count, no blocker). Both semantics are pinned by `test_a_point_with_no_declared_targets_is_ungraded_not_failed` and `test_declared_but_unmeasured_fields_fail_the_point`.

ddjukicTT (approved):

fivanovicTT (approved):

Issue #1003, opened 2025-10-29
Whisper - enable streaming and preprocessing by default


fivanovicTT:
We want to enable streaming and preprocessing by default for Whisper

PR #5118, opened 2026-09-08
Build pinned Quetzal wheel into TTIS image

nkapreTT:
## Summary

- add an opt-in Quetzal source pin to the existing single-image builder
- require an authenticated source URL and an exact lowercase 40-hex commit
- export a clean source archive through a BuildKit secret, without copying credentials or `.git` into the build context
- build exactly one Quetzal wheel and install it into the existing TTIS vLLM image with `--no-deps`
- smoke-test the Quetzal import and exact `quetzal_model_registry` entry point in both builder and final runtime stages
- suffix Quetzal image tags and force both rebuild and push when that immutable tag already exists

Native image tags, cache behavior, push behavior, and `latest` handling remain unchanged.

## Scope

This PR is image construction only. It does not add runtime admission, package mounts, catalog rows, Models CI enrollment, or Shield workflow changes.

## Validation

- 7 focused image-contract tests pass
- 43 affected tests passed in the final preparation audit
- Bash syntax, ShellCheck, repository-pinned Ruff, formatting, and `git diff --check` pass
- an independent review found no blocker and confirmed that native behavior remains unchanged

Docker is unavailable on the preparation host, so the required next gate is the real builder CI producing a digest-pinned image followed by the existing Exabox launch contract. The PR does not claim that downstream image receipt yet.

This small prerequisite lets Shield select an exact Quetzal source commit without reviving the old broad #1003 integration stack.

### DIFF
--- a/scripts/build_single_docker.sh
+++ b/scripts/build_single_docker.sh
@@ -74,6 +74,8 @@ UBUNTU_VERSION="20.04"
 CONTAINER_APP_UID=1000
 TT_METAL_COMMIT_SHA_OR_TAG=v0.56.0-rc6
 TT_VLLM_COMMIT_SHA_OR_TAG=b9564bf364e95a3850619fc7b2ed968cc71e30b7
+TT_QUETZAL_COMMIT_SHA=""
+TT_QUETZAL_SOURCE_DIR=""
 TAG_SUFFIX=""
 IMAGE_REPO="ghcr.io/tenstorrent/tt-inference-server"
 # ------------------------------------------------------------------------------
@@ -111,6 +113,22 @@ while [ $# -gt 0 ]; do
                 exit 1
             fi
             TT_VLLM_COMMIT_SHA_OR_TAG="$2"
+            shift
+            ;;
+        --quetzal-commit)
+            if [ $# -lt 2 ]; then
+                echo "⛔ Error: --quetzal-commit requires a value."
+                exit 1
+            fi
+            TT_QUETZAL_COMMIT_SHA="$2"
+            shift
+            ;;
+        --quetzal-source-dir)
+            if [ $# -lt 2 ]; then
+                echo "⛔ Error: --quetzal-source-dir requires a value."
+                exit 1
+            fi
+            TT_QUETZAL_SOURCE_DIR="$2"
             shift
             ;;
         --ubuntu-version)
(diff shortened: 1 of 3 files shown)


scripts/build_single_docker.sh
```diff
@@ -172,28 +190,69 @@ if [[ "$CONTAINER_APP_UID" =~ ^[0-9]+$ ]] && (( $CONTAINER_APP_UID >= 1000 && $C
...
-release_image_tag=${IMAGE_REPO}/vllm-tt-metal-src-release-${OS_VERSION}:${IMAGE_VERSION}-${TT_METAL_COMMIT_DOCKER_TAG}-${TT_VLLM_COMMIT_DOCKER_TAG}${TAG_SUFFIX:+-${TAG_SUFFIX}}
+dev_image_tag=${IMAGE_REPO}/vllm-tt-metal-src-dev-${OS_VERSION}:${IMAGE_VERSION}-${TT_METAL_COMMIT_DOCKER_TAG}-${TT_VLLM_COMMIT_DOCKER_TAG}${QUETZAL_IMAGE_SUFFIX}${TAG_SUFFIX:+-${TAG_SUFFIX}}
+release_image_tag=${IMAGE_REPO}/vllm-tt-metal-src-release-${OS_VERSION}:${IMAGE_VERSION}-${TT_METAL_COMMIT_DOCKER_TAG}-${TT_VLLM_COMMIT_DOCKER_TAG}${QUETZAL_IMAGE_SUFFIX}${TAG_SUFFIX:+-${TAG_SUFFIX}}
 
 # Initialize flags for whether to build each image locally.
 build_dev_image=true
 build_release_image=true
 
-if [ "$force_build" = true ]; then
+if [[ -n "$TT_QUETZAL_COMMIT_SHA" ]]; then
```
ddjukicTT, line 252: --quetzal-commit --push without --build force-pushes a stale local image. Set build=true alongside force_push=true, or gate force_push on [ "$build" = true ]

nkapreTT:
Addressed the review finding and rebased onto current main. Supplying a pinned Quetzal commit now sets build=true as well as force_push=true, so --quetzal-commit ... --push cannot publish a stale pre-existing local tag. The behavioral regression asserts a build command occurs before the push. Also made the test module importable under the repository-supported Python 3.8 annotation semantics. Local validation: 7 focused tests pass; bash syntax, repository Ruff, formatting, and git diff checks pass. Ready for re-review.

scripts/build_single_docker.sh
nkapreTT: Fixed in b50d6543. A pinned Quetzal commit now sets build=true alongside force_push=true. The behavioral regression invokes --quetzal-commit with --push and asserts a build occurs before the push, preventing publication of a stale local tag.

ddjukicTT (approved):

tstescoTT (approved):

tstescoTT:
I'll approve these to unblock. Please flag if there are issues in CI. Doing a dispatch run can smoke these likely.

tstescoTT:
From slack convo:
> inclusion of Quetzal in the image is gated with some cmdline flags:
``` 
--quetzal-commit <exact-40-character-SHA> --quetzal-source-dir <existing-local-checkout>
```

Issue #3872, opened 2026-06-01
Emit clean OpenAI-shaped error bodies (don't leak stack traces on 4xx/5xx)


ksriv001:
## Context

The TT Cloud Console proxies user inference traffic to this server as an OpenAI-compatible provider (for OpenRouter listing). On some error paths the server returns **raw Python stack traces / absolute file paths** in the response body (e.g. `site-packages/vllm/engine.py", line 42 …`) instead of a clean error envelope.

The console added a **gateway backstop** that rewrites leaking `>=4xx` bodies into a clean OpenAI-shaped error before they reach clients (tenstorrent/tt-cloud-console#596, closes tenstorrent/tt-cloud-console#565). That guarantees customers never see a leak, but it treats the symptom and discards the genuine error message in the process.

This issue tracks the **root-cause fix** at the source.

## Why this matters even with the proxy backstop

- It's a contract violation: the OpenAI-compatible error shape is supposed to be `{error:{message,type,code}}` with no file paths or stack frames.
- Any **non-console consumer** (internal tooling, direct access) is unprotected by the gateway.
- When the source emits a proper message, the console passes it through untouched and clients get a *useful* error instead of a generic one.

## Tasks

- [ ] Audit error paths that currently return tracebacks (malformed request → 400, internal error → 5xx). Note: the serialization may live in the vLLM API-server layer (`tenstorrent/vllm`) rather than the wrapper — confirm where to fix.
- [ ] Serialize errors to the OpenAI envelope (`message`/`type`/`code`) with no internal paths or stack frames.
- [ ] Verify the console proxy backstop becomes a no-op for these paths (bodies pass through unchanged).

## Refs
- Gateway backstop / symptom fix: tenstorrent/tt-cloud-console#596 (closes tenstorrent/tt-cloud-console#565)
- OpenRouter integration tracker: tenstorrent/tt-cloud-console#495

idjuricTT:
This should also come with Dynamo frontend, although number of errors is quite narrow - it's pretty much user and system error

HerShin5:
The console backstop sounds like the right safety net, but I’d make the server-side contract slightly stricter than just “no traceback”.

A useful invariant for an OpenAI-compatible endpoint is:

- every non-2xx response is valid JSON with exactly one top-level `error` object,
- `error.message` is user-actionable but does not include file paths, stack frames, hostnames, container IDs, or raw upstream payloads,
[...]

Issue #3871, opened 2026-06-01
openrouter/inference-server: emit clean OpenAI-shaped error bodies (don't leak stack traces at the source)


ksriv001:
## Context

PR tenstorrent/tt-cloud-console#596 (closes tenstorrent/tt-cloud-console#565) added a **gateway backstop** in the console proxy: a `ModifyResponse` hook that detects internal leaks (Python tracebacks, absolute file paths) in upstream `>=4xx` bodies and rewrites them to a clean OpenAI-shaped error before they reach clients.

That guarantees no leak reaches a customer, but it treats the symptom. The **root cause** is the inference backend (vLLM / tt-inference-server) returning raw stack traces on some error paths, which is a violation of the documented error contract.

## Why this still matters even with the proxy backstop

- The contract in `docs/tech/services/token-as-service/inference-backend.md` and `quality.md` already specifies clean `{error:{message,type,code}}` shapes — the backend leaking a traceback breaks that contract.
- Any **non-console consumer** of the inference server (internal tooling, direct access) is unprotected by the gateway.
- The proxy currently has to replace leaking bodies with a *generic* message, losing the genuine (clean) error detail. If the source emits a proper message, the proxy passes it through untouched and clients get a better error.

## Tasks

- [ ] Audit tt-inference-server / vLLM error paths that currently return tracebacks (notably malformed requests → 400, internal errors → 5xx).
- [ ] Ensure errors are serialized to the OpenAI error envelope (`message`/`type`/`code`) with no file paths or stack frames.
- [ ] Confirm the console proxy backstop then becomes a no-op for these paths (bodies pass through unchanged).

## Refs
- Backstop: tenstorrent/tt-cloud-console#596 / tenstorrent/tt-cloud-console#565
- Parent: tenstorrent/tt-cloud-console#495

idjuricTT:
@ksriv001 that makes sense.
I believe we get this for free by using Dynamo frontend.

I'll keep the issue open until we do the deployment and confirm this

ksriv001:
Duplicate of #3872, which now carries the Cobalt PT40893_8 instance (tenstorrent/tt-cloud-console#1152) and the acceptance criterion (vllm-tt-plugin images on vLLM >= 0.26.0).

PR #5133, opened 2026-09-09
Sanitize 4xx/5xx bodies from the vLLM API server (Cobalt PT40893_8, tt-cloud-console#1152)

ksriv001:
Fixes the root cause of Cobalt pentest finding **PT40893_8** (tenstorrent/tt-cloud-console#1152, Low, CVSS 3.7): the inference backend returned internal file paths, the container username and the vLLM module layout inside HTTP 400 bodies. Also the source-side fix tracked in #3872 / #3871.

## The finding

`POST /tokenize` with body `{}` on `qwen3vl32b--tenstorrent.workload.tenstorrent.com` (image `0.14.0-80180b9-7678b70`) returned:

```
"message": "2 validation errors:\n  {'type': 'missing', 'loc': ('body', 'function-wrap[__log_extra_fields__()]', 'prompt'), ...}\n ...
  File \"/home/container_app_user/vllm/vllm/entrypoints/utils.py\", line 38, in tokenize\n    POST /tokenize [...]"
```

## Root cause

- FastAPI 0.124.0 (2025-12-06, fastapi/fastapi#14306) started appending the endpoint's source file, line and function to `str(RequestValidationError)`.
- `validation_exception_handler` in **every** tenstorrent/vllm commit the release images pin (27 distinct commits, 2025-07 → 2026-08, incl. `dev` HEAD) builds the client message from `str(exc)`; its `sanitize_message` only strips memory addresses. The `function-wrap[__log_extra_fields__()]` marker is Pydantic's wrap-validator name leaking through `loc`.
- Upstream vLLM fixed this in vllm-project/vllm#46415 (builds the message from `exc.errors()`, hardens `sanitize_message`; in **v0.26.0+**) and bounded the body in #54684 (v0.29.0). The fork never picked either up.
- Reproduced locally with FastAPI 0.141.1 and the fork's handler: byte-for-byte the Cobalt body (see `test_end_to_end_fastapi_tokenize_empty_body_no_longer_leaks`).

## Why fix it here and not in vLLM

| Where | Reaches the fleet? |
|---|---|
| Backport #46415 into the tenstorrent/vllm fork | Fork is deprecated (2026-08-24) and each of the 127/129 vLLM release specs pins its own commit; every model would need a new engine pin + re-validation. |
| Move models to vllm-tt-plugin (upstream vLLM) | Right long-term direction, but 39 of 41 models are still on the fork, and the one plugin-based release image (`Qwen3-32B` galaxy, plugin `be7d805`) pins **vLLM 0.25.1**, which predates the upstream fix. Only plugin `main` (0.26.0) is clean. |
| Console proxy backstop (tt-cloud-console#596) | Already rewrites leaking bodies on the console API path, but Cobalt hit the workload host directly; ngrok-registered and in-cluster callers are also outside it. |
| **Wrapper middleware in this repo (this PR)** | Version-agnostic: `--middleware <import.path>` is accepted by all 27 pinned commits and by upstream vLLM. Ships on the next image build of each model without changing the pinned engine. Also covers 5xx tracebacks (#3872) that the upstream handler fix does not. No-op when a body is already clean. |

## What changes

- **`utils/error_response_sanitizer.py`** (new, stdlib only): pure-ASGI `ErrorResponseSanitizerMiddleware`.
  - Status `>= 400`: buffers the body (cap 256 KiB), parses JSON and scrubs every string (OpenAI `{"error":{...}}`, FastAPI `{"detail":...}`, any shape) or scrubs text bodies; fixes `content-length`; logs the original body server-side at WARNING so nothing is lost for debugging.
  - Removes: `Traceback ...` headers, `File "...", line N, in f` frames + the following context line (this is also FastAPI's endpoint-context format), `Endpoint: /path`, absolute paths → `<path>`, Pydantic `'function-wrap[...]',` loc markers, ` at 0x...>`. Regexes mirror upstream's hardened `sanitize_message`.
  - Status `< 400` (incl. SSE streams) and content-encoded bodies pass through untouched and unbuffered.
  - Bodies over the cap are replaced with a bounded generic envelope of the same status (defence against the #54684 amplification).
- **`vllm-tt-metal/src/run_vllm_api_server.py`**: appends `--middleware utils.error_response_sanitizer.ErrorResponseSanitizerMiddleware` to the vLLM argv; new wrapper flag `--disable-error-sanitizer`. If the caller already passes `--middleware`, argv is left alone with a warning (the flag is `action="append"` on older pinned commits and `nargs="+"` on newer ones, so a second occurrence would silently replace the caller's on the newer parser).
- **`workflows/run_docker_server.py`**: new flag added to `_RESERVED_WRAPPER_FLAGS` (drift test).
- **Tests**: 25 new (`tests/test_error_response_sanitizer.py`) incl. the verbatim Cobalt payload, a real-FastAPI end-to-end reproduction, chunked/streamed/compressed/oversized bodies; launcher tests updated/added for injection, opt-out, and the existing-`--middleware` case.

Result on the Cobalt payload:

```json
{"error":{"message":"2 validation errors:\n  {'type': 'missing', 'loc': ('body', 'prompt'), 'msg': 'Field required', 'input': {}}\n  {'type': 'missing', 'loc': ('body', 'messages'), 'msg': 'Field required', 'input': {}}","type":"Bad Request","param":null,"code":400}}
```

## Test plan

- [x] `ruff check .` / `ruff format --check .` clean
- [x] `pytest tests/test_error_response_sanitizer.py tests/test_run_vllm_api_server.py tests/test_setup_host.py tests/test_run_arguments.py` → 187 passed
- [x] Full `pytest tests/` has zero new failures vs `origin/main` (pre-existing collection errors from optional deps only)
- [ ] Build one release image from this branch and run the Cobalt repro against the live container: `POST /tokenize {}` → 400 with no path/frame/username; `/v1/chat/completions` streaming unchanged; `/health` unchanged
- [ ] Re-point the console catalog entry for `Qwen3-VL-32B-Instruct` (and the other pentested hosts) at the rebuilt tags, then ask Cobalt to retest

## Why draft

1. Needs a real image build + live check on a TT device (item above); I only have unit and FastAPI-level verification.
2. Decision for the inference team: default-on for every image (as here) vs. opt-in. I recommend default-on: the console already rewrites the same bodies, so no console client sees a change; the middleware only touches error responses.
3. Rebuild/re-pin sequencing across the 41 models is a release-team call; until then the Cloudflare WAF blocklist (cloud-infra#1301) 403s `/tokenize` on the CDN path, but the direct-origin path (tt-cloud-console#1128, Cobalt finding 9) and ngrok paths still reach the leaky handler.

Not in scope, tracked elsewhere: unauthenticated `/version` and the other non-`/v1` routes (default-deny middleware, tt-cloud-console#1128 layer 2).

Refs tenstorrent/tt-cloud-console#1152, #3872, #3871, vllm-project/vllm#46415, vllm-project/vllm#54684.

### DIFF
--- a/tests/test_error_response_sanitizer.py
+++ b/tests/test_error_response_sanitizer.py
@@ -1,0 +1,477 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+"""Tests for utils.error_response_sanitizer (tenstorrent/tt-cloud-console#1152).
+
+The first group drives the pure-ASGI middleware directly with a fake app so it
+needs no web framework. The last test rebuilds the tenstorrent/vllm fork's
+``validation_exception_handler`` on top of the installed FastAPI and checks that
+the exact body Cobalt reported (endpoint file path, container username, Pydantic
+wrap-validator marker) is cleaned end to end; it is skipped when FastAPI is not
+installed.
+"""
+
+import asyncio
+import json
+import re
+
+import pytest
+
+from utils import error_response_sanitizer as ers
+from utils.error_response_sanitizer import (
+    ERROR_SANITIZER_MIDDLEWARE,
+    ErrorResponseSanitizerMiddleware,
+    sanitize_error_body,
+    sanitize_error_text,
+)
+
+# Verbatim from Cobalt PT40893_8 (2026-08-29), response to `POST /tokenize {}`.
+COBALT_MESSAGE = (
+    "2 validation errors:\n"
+    "  {'type': 'missing', 'loc': ('body', 'function-wrap[__log_extra_fields__()]', "
+    "'prompt'), 'msg': 'Field required', 'input': {}}\n"
+    "  {'type': 'missing', 'loc': ('body', 'function-wrap[__log_extra_fields__()]', "
+    "'messages'), 'msg': 'Field required', 'input': {}}\n\n"
+    '  File "/home/container_app_user/vllm/vllm/entrypoints/utils.py", line 38, '
+    "in tokenize\n"
+    "    POST /tokenize [{'type': 'missing', 'loc': ('body', "
+    "'function-wrap[__log_extra_fields__()]', 'prompt'), 'msg': 'Field required', "
+    "'input': {}}, {'type': 'missing', 'loc': ('body', "
+    "'function-wrap[__log_extra_fields__()]', 'messages'), 'msg': 'Field required', "
+    "'input': {}}]"
+)
+
+LEAK_MARKERS = (
+    "container_app_user",
+    "/home/",
+    'File "',
+    "line 38",
+    "function-wrap",
+    "__log_extra_fields__",
+    "entrypoints/utils.py",
+    "POST /tokenize",
+)
+
+
+def assert_clean(text: str) -> None:
+    for marker in LEAK_MARKERS:
+        assert marker not in text, f"{marker!r} still present in {text!r}"
+
+
+# --------------------------------------------------------------------------- #
+# sanitize_error_text
+# --------------------------------------------------------------------------- #
+
+
+def test_sanitize_error_text_cleans_cobalt_validation_message():
+    cleaned = sanitize_error_text(COBALT_MESSAGE)
+
+    assert_clean(cleaned)
+    # The user-actionable part survives, with the internal loc marker removed.
+    assert cleaned == (
+        "2 validation errors:\n"
+        "  {'type': 'missing', 'loc': ('body', 'prompt'), 'msg': 'Field required', "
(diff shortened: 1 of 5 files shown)


ksriv001:
Closing in favour of the vLLM upgrade path: the console proxy already scrubs these bodies on the API path (tenstorrent/tt-cloud-console#596), direct access to workload hosts is being closed under tenstorrent/tt-cloud-console#1128, and upstream vLLM v0.26.0+ (vllm-project/vllm#46415) fixes the handler itself, so this lands as models move to vllm-tt-plugin images pinned to vLLM >= 0.26.0 (the current plugin release image pins 0.25.1). Source-side contract stays tracked in #3872; decision recorded on tenstorrent/tt-cloud-console#1152. Branch left in place as a version-agnostic fallback if the migration is slower than the retest window.

PR #5137, opened 2026-09-10
Gemma4: MTP + dFlash speculative serving specs (P150X8 + BH Galaxy DP=4)

arginugaTT:
Main model-spec entries for the gemma-4 speculative serving impls (gemma4_dflash, gemma4_mtp) — separate templates; 

- 31B dFlash P150X8: single adaptive 256K profile — conc-1 spec (3.6–5.2×), conc>1 batched baseline, MAX_SPEC_ISL guard.
- 31B dFlash BLACKHOLE_GALAXY: DP=4 (four (1,8) column engines) — validated: conc-1 spec ~106 tok/s/u per lane, 128-way batched fallback ~1.9K agg tok/s.
- 31B MTP: raised to 262144 (validated full ladder, no bounded collapse). 12B MTP held at 65536 with a documented pre-existing serving hang.
- benchmark_config: 192K/224K/248K rungs + OSL=1024 high-ISL companions (the output-TPS column is otherwise ~all prefill at high ISL); auto-filtered by each model's context cap.

### DIFF
--- a/reference_config/benchmarking/benchmark_config.py
+++ b/reference_config/benchmarking/benchmark_config.py
@@ -98,12 +98,22 @@ BENCHMARK_ISL_OSL_PAIRS = [
     (8192, 1024),
     (10000, 1024),
     (16384, 128),
+    (16384, 1024),
     (32768, 128),
+    (32768, 1024),
     (65536, 128),
+    (65536, 1024),
     (131072, 128),
+    (131072, 1024),
     (196608, 128),  # 192K -- dFlash bounded 256K sweep
     (229376, 128),  # 224K
     (253952, 128),  # ~248K -- safe near-max for a 262144 context (leaves room for osl + block-output placeholder; 262016 crashes at zero margin)
+    (253952, 1024),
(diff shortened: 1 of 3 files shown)


reference_config/benchmarking/benchmark_config.py
```diff
@@ -98,9 +98,22 @@ class BenchmarkConfig:
     (8192, 1024),
     (10000, 1024),
     (16384, 128),
+    (16384, 1024),
```
tstescoTT: Is there a reason to do sweeps on OSL=128 and OSL=1024 for each of these?

These previous ISL=8k/OSL=1k was to match InferenceX exactly, and ISL=10k,OSL=1024k was to match ArtificialAnalysis I believe.

tstescoTT (approved):

reference_config/benchmarking/benchmark_config.py
arginugaTT: I added them for consistent TPOT timings on spec decode benchmarks, I will remove them and scope out a better solution for these spec decode cases.

PR #5143, opened 2026-09-11
Fix Quetzal source transport for image builds

nkapreTT:
## Problem

Shield run 34532744115 proved that #5118 cannot build a Quetzal image: the exact Quetzal git archive is 104,427,520 bytes, while BuildKit rejects secrets above 500 KiB. The build failed before producing an image or reaching QB2.

## Fix

Replace the source-as-secret transport with a sanitized exact-commit named build context mounted read-only while the wheel is built. An empty scratch-stage fallback preserves native builds. The existing exact SHA/HEAD validation, dependency-free wheel install, plugin smoke test, opt-in behavior, and OCI revision label remain unchanged.

No Shield change is required.

## Validation

- `python3 -m pytest -q tests/test_quetzal_image_build.py`: 7 passed
- Ruff: passed
- `bash -n scripts/build_single_docker.sh`: passed
- `git diff --check`: passed
- Real Docker smoke: native fallback passed; 600,001-byte named context passed; mounted source absent from resulting container filesystem

A Shield dispatch against this unmerged exact SHA will provide the full image-build and QB2 evidence before this leaves draft.

### DIFF
--- a/scripts/build_single_docker.sh
+++ b/scripts/build_single_docker.sh
@@ -191,7 +191,6 @@ else
     echo "CONTAINER_APP_UID=${CONTAINER_APP_UID} is not a number or outside expected range of 1000 to 59999."
 fi
 QUETZAL_SOURCE_CONTEXT=""
-QUETZAL_SOURCE_ARCHIVE=""
 cleanup_quetzal_source_context() {
     if [[ -n "$QUETZAL_SOURCE_CONTEXT" && -d "$QUETZAL_SOURCE_CONTEXT" ]]; then
         rm -rf -- "$QUETZAL_SOURCE_CONTEXT"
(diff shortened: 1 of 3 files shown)


nkapreTT:
Now blocking, with a reproduction.

tt-shield run [34628071170](https://github.com/tenstorrent/tt-shield/actions/runs/34628071170) dispatched a Quetzal image build against `main` (specifically `16b88c402515d71e3133700513c55acede0added`) and the build failed with:

```
ERROR: failed to build: secret quetzal_source too big. max size 500KiB
```

That is the `--secret` path this PR replaces. BuildKit caps secrets at 500 KiB and the Quetzal source archive exceeds it, so **no Quetzal image can be built from `main` today** — `run-tests` was skipped and the run produced no measurement.

The currently-serving Llama-3.2-1B run ([34610746272](https://github.com/tenstorrent/tt-shield/actions/runs/34610746272)) succeeds only because it pins this PR's head (`9e2ef7118b1c2810d8bb40504456a4ef522ff016`) as its `inference-server-git-ref`. So the one piece of working Quetzal Model-CI evidence depends on an unmerged branch.

Switching to `--build-context` removes the size ceiling rather than raising it.

nkapreTT:
Fresh end-to-end evidence for the transport fix: tt-shield run https://github.com/tenstorrent/tt-shield/actions/runs/34692622521 built the Quetzal TTIS image successfully from TTIS ref c5e8fae9566c, whose three #5143 files are byte-identical to this PR head. The subsequent hardware job failed later, before server launch, because the selected runner did not have the requested /mnt/models package directory; that is a separate runner-mount issue and does not invalidate the successful source build. Keeping this PR transport-only: model catalog rows will be generated from qualified immutable packages in separate reviewable changes.

idjuricTT (approved):

fivanovicTT (approved):

PR #5152, opened 2026-09-13
Fail fast when an inference-server container exits during startup

nkapreTT:
## Summary

Retain the foreground `docker run` process handle in the launch payload and use the existing generic readiness liveness check to poll it. An exited container is detected on the next readiness iteration instead of consuming the full one-hour timeout twice.

This is an independent operational fail-fast improvement. It does not change package admission, model selection, numerical behavior, or model catalog rows. Current main is merged into this branch; the trace-reservation fix from #5161 is preserved unchanged. The remaining diff is four files (+101/-8).

## Evidence

Shield run https://github.com/tenstorrent/tt-shield/actions/runs/34776407176 logged its fatal container traceback at 20:06:08, then waited until 21:06:08, retried, and waited again until 22:06:11 because Docker liveness was reported as unobservable.

A second reproduction is https://github.com/tenstorrent/tt-shield/actions/runs/34976178490: fatal package-admission error at 14:48:44 UTC, final failure at 16:48:46. The underlying admission error is separate from this fail-fast fix.

This change detects a foreground launcher exiting. It does not diagnose a live-but-unresponsive server or prevent loss of the self-hosted runner.

## Validation

At head 9ca5d66f57f9e07008891db970add67129d14132, against main e1c2a9bc4:

- `python -m pytest -q tests/workflow_module/test_commands.py tests/test_run_docker_server.py tests/test_run_vllm_api_server.py`: 85 passed, no deselections (isolated Python 3.11 test environment).
- Ruff 0.15.0 check and format check on all four changed files: passed.
- `git diff --check`: passed.
- Admission implementation/tests and model catalog are byte-identical to main.

Updated by merging main, not rebasing or force-pushing; reviewed commit history is retained.


jzivanovicTT (approved):

nkapreTT:
Additional production reproduction for this existing fix: https://github.com/tenstorrent/tt-shield/actions/runs/34976178490

The Qwen container logged a fatal package-admission traceback at **14:48:44 UTC**. The readiness loop waited until **15:48:45**, retried the same launch, and finally failed at **16:48:46**. No model evaluations ran during those two hours.

The admission bug is separate (trace-reservation propagation); repairing it does not remove the need for this generic fail-fast change. I revalidated this fail-fast patch together with the trace fix: **145 focused tests passed**, including the test that forbids sleeping after an observed process exit.

Preserving this PR's existing reviewed commits/approval; no duplicate PR or force-push.

idjuricTT (commented):
@ddjukicTT could you please check this just in case?

ddjukicTT (approved):

PR #5164, opened 2026-09-16
Embedding: service cleanup

### LINKED ISSUE #5011: Clean up the cpp_server embedding architecture
Analyze the current embedding worker, IPC, shared-memory, scheduler, batching and runner architecture.

Propose possible solutions and implement the selected approach. It does not have to reuse the current architecture if there is a better solution.

Test all three models on at least half of the supported devices and compare the results with the previous baseline.

jzivanovicTT:
## What

Move-only restructuring of the C++ embedding service, following up on #5050 (drop tt-media-server dependency). No behavior, protocol, or performance changes intended — code is relocated and decomposed, not rewritten.

`embedding_service.cpp` goes from ~776 to ~545 lines, and the pieces that were tangled into it now live in files named after what they do:

| New file | Contents |
|---|---|
| `embedding_pipe.{hpp,cpp}` | pipe framing primitives (`pipeWrite`, `pipeReadBinary`, `pipeReadString`, ready sentinel) |
| `embedding_worker_process.{hpp,cpp}` | `WorkerProcess` — spawn, liveness check, request/response over pipes, terminate |
| `embedding_worker_main.{hpp,cpp}` | the forked child's main: env export, runner construction, warmup, serve loop |

## Commits

1. **Drop the unused synchronous request path** — `EmbeddingService` now inherits `RequestPipeline` directly instead of `BaseSyncService`; the dead `produceResponse` override is gone.
2. **Extract pipe framing** into `embedding_pipe.{hpp,cpp}`.
3. **Move `WorkerProcess`** into `embedding_worker_process.{hpp,cpp}`.
4. **Move the child-process main** into `embedding_worker_main.{hpp,cpp}`, split into named steps (`exportWorkerEnvironment`, `buildRunnerOrDie`, `parseBatch`, `serveLoop`).
5. **Decompose the long service functions**: `runStartup` → `warmupCacheLeader` / `spawnAndAwaitRemaining` / `logStartupSummary`; the dispatch loop → `collectBatch` / `drainIfLastWorker`; a single `completeWithError` helper now backs `drainQueue`, `failBatch`, and dispatch failures (preserving the exactly-once completion contract).
6. **Comment polish** and a **clang-format-20 pass**.

All log lines are byte-identical to before — existing grep-based checks and dashboards keep working.


tt-media-server/cpp_server/src/services/embedding_pipe.cpp
```diff
@@ -0,0 +1,57 @@
...
+  }
+  return buf;
+}
+
+std::string pipeReadString(int fd) {
+  uint32_t len = 0;
+  ssize_t n = read(fd, &len, sizeof(len));
+  if (n <= 0) return {};
+
+  std::string data(len, '\0');
```
ztorlakTT, line 41: nit: we can check len against `EMBEDDING_MAX_PIPE_BYTES` here as well before allocating the string?

tt-media-server/cpp_server/src/services/embedding_worker_process.cpp
```diff
@@ -0,0 +1,128 @@
...
+    isReady.store(false);
+  }
+  return buf;
+}
+
+void WorkerProcess::terminate() {
+  const pid_t p = pid.load();
+  if (p > 0) {
+    kill(p, SIGTERM);
+    waitpid(p, nullptr, 0);
```
ztorlakTT, line 121: Can we clear pid after waitpid() here and in checkAlive()

ztorlakTT (approved):

ztorlakTT:
nit: I still think some comments in the new files are too detailed

PR #5168, opened 2026-09-16
feat(requirements): resolve llm-gauntlet;<path> by cloning the gauntlet repo

ipastalTT:
## Summary
- Stacked on #5167 — only the 9 commits from `resolve "llm-gauntlet;<path>"...` through `size sweep points like the catalog, and stream every eval` are new to this PR.
- New `workflows/llm_gauntlet_repo.py`: resolves `--requirements-json "llm-gauntlet;<path>"` by cloning the llm-gauntlet repo (SSH by default, since it's private; `TT_LLM_GAUNTLET_TOKEN` switches to authenticated HTTPS for token-only runners like CI) and reading the document at that path inside the clone.
- `workflows/requirements_target_pack.py`: re-points a borrowed catalog harness at the document's own model; streams eval generations (so a gateway with a fixed response-start window, e.g. Cloudflare's ~100s, can't cut off a long chain-of-thought answer); sizes sweep points against the catalog.
- New fixture `tests/fixtures/requirements/wiring-plan-qwen3-32b.json`: a synthetic wiring plan for a credential-free CI run.

## Test plan
- New tests in `tests/workflows/test_llm_gauntlet_repo.py`, `tests/workflow_module/test_requirements_schema.py`, `tests/workflow_module/test_requirements_target_pack.py` cover: fresh clone vs. existing clone, ref precedence, token redaction from logs, streaming override, model re-pointing.
- 150 tests passing in the affected modules.


workflows/llm_gauntlet_repo.py
```diff
@@ -0,0 +1,259 @@
...
+
+Requirements documents and the validation plans exported from them live in the
+llm-gauntlet repo, not here. Copying one into a runner's working directory to
+start a run loses the only thing that makes a result citable -- which document,
+at which revision, produced it. This scheme lets an operator or a CI job name
+the document by its path *inside that repo*:
+
+    --requirements-json "llm-gauntlet;specs/tt-internal/qwen3-32b/<id>.json"
+
+The repo is cloned into the tt-inference-server root and the path resolved
```
fivanovicTT: To be honest here - I don't really like cloning other repos. Is there any other way other than cloning?

fivanovicTT (approved):

PR #5179, opened 2026-09-17
[Helm v0.3] Image-release floor + version label from the image tag

### LINKED ISSUE #5173: [Helm v0.3] Supported image-release floor (vLLM 0.11.0+, media/forge NO_AUTH 0.15.0+) + appVersion provenance
Part of #4926. PR: #5179.

The chart drives the server through interfaces that only exist in images from a given release onwards, and nothing says so, so a row pinned below it fails at runtime instead of at install. 97 of 134 vLLM rows are pinned below.

| engine | floor | what needs it |
|---|---|---|
| `vllm` | 0.11.0 | the chart passes `--model` / `--tt-device` as container args |
| `media`, `forge` | 0.15.0 | only when `auth.disabled=true`, which needs `NO_AUTH` |

- [ ] Document the floors and why they exist
- [ ] Refuse a below-floor row at render, with the floor and the override named
- [ ] Drop `appVersion` (stale at 0.12.0 for ten releases) and set `app.kubernetes.io/version` from the resolved image tag

Verified on hardware and against the published images — see the comments below.

sott0n:
Closes #5173

Every row pins its own server image, and those pins span ~20 releases. The chart drives the server through interfaces older images lack, and nothing said so: 97 of 134 vLLM rows are pinned below the release that introduced them, and could only fail at runtime. They now fail at render (`imageFloor.enforce`, default `true`) with the floor, the row's tag and the override named.

- Floors live in `values.yaml`, so the gate, its messages and the docs read one source.
- The generated Supported Models table now lists only installable rows and reports how many it left out (83); a refreshed pin brings them back on the next helm-docs run.
- `appVersion` is dropped (stale at `0.12.0` for ten releases, and no single value can be true here); `app.kubernetes.io/version` carries the resolved image tag instead, which is what that label is defined to mean.
- CI: shape checks render with the floor waived, since the manifest is their subject; one new step asserts the floor with tags forced via `--set`, so a refreshed pin cannot turn it red. The hardware smoke gains `IMAGE_FLOOR`.

The boundary was also read from all 30 published vLLM tags' `.config.Entrypoint` — 91 rows do not forward args, 37 do, with the break between 0.10.1 and 0.11.1.

Refreshing a stale pin is a ModelSpec catalogue change, not a chart change: this makes those 97 rows fail at install with a reason, it does not make them serve.

### Why is `vllm` floor 0.11.0
The row cannot run at all.** k8s `args` replace an image's `CMD`, not its `ENTRYPOINT`; only images from 0.11.0 forward `--model`/`--tt-device` to the server, and earlier ones also leave `CACHE_ROOT` unset, which their entrypoint requires.
Verified on T3K through the chart: `Qwen3-32B`/`t3k` (0.9.0, gate waived) crash-loops in its entrypoint; `gpt-oss-120b`/`t3k` (0.17.0) serves and receives the chart's args. The intent is to turn a multi-GB pull followed by an opaque crash-loop into an install-time error that names the next action.

### Why is `media`/`forge` floor 0.15.0
The flag is silently ineffective. `NO_AUTH` arrived in 0.15.0; below it `auth.disabled=true` does nothing: the server still requires `Authorization: Bearer <API_KEY>` and falls back to a key published in its own source, so clients sending no header are rejected.
The workload does run, which makes this the weaker of the two floors — it refuses a *value*, not the row, and the row installs normally with `auth.apiKey`. The intent is that a security-relevant flag should not quietly mean something else. If reviewers disagree, dropping `imageFloor.authDisabled` from `values.yaml` removes this check alone.


charts/tt-inference-server/README.md
```diff
@@ -2,13 +2,14 @@
 
 Deploys vLLM, media, and forge inference backends on Tenstorrent hardware.
 
-**Chart version:** 0.2.0 | **App version:** 0.12.0
+**Chart version:** 0.2.0
```
vcankovicTT: shouldn't we bump the version to 0.3.0?

charts/tt-inference-server/Chart.yaml
```diff
@@ -3,4 +3,3 @@ name: tt-inference-server
 description: Tenstorrent inference server — vLLM, media, and forge backends on Tenstorrent hardware
 type: application
 version: 0.2.0
```
vcankovicTT: shouldn't we bump the version to 0.3.0?

PR #5180, opened 2026-09-17
Vm/benchmark partial report

vmaksimovicTT:


test_module/llm_tests/llm_performance_tests.py
```diff
@@ -96,7 +97,22 @@ def run_llm_performance(
...
     )
-    result = runner.run(configs, server, context)
+
+    # Built before the sweep: `generated_at` is recorded once and synthesises
+    # the report_id that names the report files, so the checkpoints and the
+    # final report overwrite one another.
+    envelope = {
+        **report_model_fields(ctx.model_spec),
+        "device": device_label,
+        "generated_at": datetime.now(timezone.utc).isoformat(timespec="seconds"),
```
ipastalTT, line 107: I understand why we are doing this, but it changes a bit the report's meaning.
would it make more sense to delete the previous report when generating the new one on the next block?

vmaksimovicTT: Im not sure I understand how this changes the meaning. I mean yes, the "generated_at" does not mean "moment when the entire sweep finished" anymore, but rather "moment when the sweep began". Will this cause some issues down the line?

ipastalTT: I meant it is not inline with the other workflows reports, but in general it's fine by me. Just saying, yes my wording was a bit abstract.

PR #5194, opened 2026-09-20
Enable full async RL serving

abustamanteTT:
## Summary
- integrate TT vLLM workers and scheduling with full asynchronous RL rollouts, including logprobs, generation controls, and per-request seeded host sampling
- add runtime weight updates over the weight bridge/device-socket transport, including co-located Galaxy mesh setup
- support standalone server config injection and update model catalog coverage for RL serving

## Testing
- `python -m pytest -q tests/test_run_vllm_api_server.py` (62 passed)
- parsed `workflows/model_specs/dev/llm.yaml` and `workflows/model_specs/prod/llm.yaml` with `yaml.safe_load`
- `git diff --check`

## Notes
- merged the latest `main` before opening this PR
- hardware-dependent async rollout and device-socket paths were not exercised locally


vllm-tt-metal/src/run_vllm_api_server.py
```diff
@@ -841,6 +841,19 @@ def set_metal_timeout_env_vars():
         logger.info("Metal op timeout disabled via DISABLE_METAL_OP_TIMEOUT=1")
         return
 
+    # The op-timeout watchdog is sized for a single generation step, but the
+    # colocated worker also services device-socket weight updates: recv_state()
+    # blocks in ttnn.synchronize_device() for several seconds while it streams
+    # the full state dict, so the default 5s watchdog fires mid-transfer. Give
+    # the colocated path a much larger budget (or an explicit override).
+    override = os.getenv("TT_METAL_OP_TIMEOUT_SECONDS")
```
fivanovicTT: Read uses `TT_METAL_OP_TIMEOUT_SECONDS` while write uses the canonical `TT_METAL_OPERATION_TIMEOUT_SECONDS`. 

I think we should read the canonical name, or rename this override to something unambiguous.

tt-vllm-plugin/tt_vllm_plugin/v1/worker/tt_model_runner.py
```diff
@@ -807,35 +936,310 @@ def execute_model(
...
+        # nothing rather than a half-populated tensor.
+        if combined_logprobs is not None and logprob_rows_filled != num_reqs:
+            combined_logprobs = None
+
+        # 5. Standard single-batch output emission.
+        return self.generate_runner_output(combined, combined_logprobs)
+
+    def _mesh_sampling_params(self, input_batch, idxs: list[int]) -> TTSamplingParams:
+        """Uniform sampling params for a replica's slots (first slot's values;
+        execute_with_model_input asserts equality across active replicas)."""
```
fivanovicTT: Docstring claims `execute_with_model_input` asserts equality across the replica's slots, but I did not find such assertion (maybe just missed it). 

I think we can either add the assert or port the pre-existing `logger.warning` from `_prepare_model_inputs` — silently using slot 0 is a regression for async RL, where rollouts on one batch can want different temperatures.

vllm-tt-metal/src/weight_update_api.py
```diff
@@ -0,0 +1,467 @@
...
+            "update_weights",
+            kwargs={"update_info": {"hf_rope": body.hf_rope}},
+        )
+    except NotImplementedError as exc:
+        raise HTTPException(status_code=501, detail=str(exc))
+    except Exception as exc:  # noqa: BLE001 - surface engine errors to caller
+        logger.exception("Weight update failed")
+        raise HTTPException(status_code=500, detail=f"Weight update failed: {exc}")
+
+    # Flush prefix/KV cache across the version boundary (best-effort; a no-op
```
fivanovicTT: `/update` before `/init_weight_transfer_engine` raises `RuntimeError` in `TTWorker.update_weights`, which the generic `except` here maps to 500 (reads to the trainer as “server crashed”). I think 412/409 would be more useful

